PR #5, opened 2016-02-17
Permalink is displayed after deployment (along with the alias URL).

raphaelyancey:
Useful when working on several features that must be tested separately. It cuts the need to go to the *Build* page on the site to get the unique URL.


biilmann:
Thanks! Awesome

PR #8, opened 2016-08-01
instead of using CWD use '.' for deploying the root dir

rybit:


calavera:
why? This is not portable and it won't work on Windows.

rybit:
Otherwise you record the current directory ~ which means that if you move that dir it is going to break.

biilmann:
What would be the equivalent on windows? Actually hoped the `.` might work, we might need to make the empty string work otherwise

calavera:
> which means that if you move that dir it is going to break.

Sorry, I'm not familiar with this implementation. Where are we recording that path? is the path absolute?

rybit:
yeah it used to record the cwd in the .netlify file

biilmann:
We're storing the path in the `.netlify` config file. The issue is when there's no path specified and we fall back to `process.cwd()` and that ends up being an absolute path (where obviously `.netlify` ought to store a relative path.

calavera:
Store a relative path to the `.netlify` file and expand when it's necessary. Don't store anything if `result.path` is empty.

rybit:
that means that each time you deploy you'll be asked what dir you want to deploy.

PR #18, opened 2016-12-30
Add "env" command

mraerino:
*Improvement of developer experience*

This command takes the environment variables from netlify and writes them either to stdout or to a file.
It should be especially useful for people using dotenv locally, because the output uses dotenv compatible syntax.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -13,6 +13,7 @@ var program    = require("commander"),
     list       = require("../lib/commands/list_sites"),
     updateSite = require("../lib/commands/update_site"),
     openSite   = require("../lib/commands/open"),
+    env        = require("../lib/commands/env"),
     updateNotifier = require('update-notifier'),
     pkg = require('../package.json');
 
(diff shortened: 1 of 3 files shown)


bin/cli.js
```diff
@@ -78,6 +79,13 @@ program
   .action(config.wrap(program, init.cmd));
 
 program
+  .command("env")
+  .description("Output configured env variables")
+  .option("-s --site-id [id]", "Fetch from site with <id>")
+  .option("-f --file [filename]", "Save to file called <filename>")
```
rybit: I think cutting this is better. It simplifies the code. You can just use a redirect b/c this is a cli

mraerino: Unfortunately the site selection ui uses stdOut, so if I redirect the output without passing a site id I don't get the site selection ui in my terminal.
This should happen on stdErr, but since this would require large amounts of re-structuring, I solved it this way.

mraerino:
@rybit have you considered my last comment?

rybit:
Sorry for the delay - yeah it totally sounds good.

PR #38, opened 2018-01-24
deprecatin'

fool:
time to start work on deprecating this.  This is a soft announcement since there's no new doc on our site yet which is kinda required before we can fully deprecate.


calavera:
LGTM

PR #39, opened 2018-01-31
Prevent global and local config paths to clash

### LINKED ISSUE #32: Calling netlify sites inside ~ causes error
When i run `netlify sites` inside my home ~ directory, I get this error
```
diwu@burnt ~ » netlify sites
fs.js:732
  var r = binding.read(fd, buffer, offset, length, position);
                  ^

Error: EISDIR: illegal operation on a directory, read
    at Error (native)
    at Object.fs.readSync (fs.js:732:19)
    at tryReadSync (fs.js:487:20)
    at Object.fs.readFileSync (fs.js:535:19)
    at readLocalConfig (/usr/local/lib/node_modules/netlify-cli/lib/settings/config.js:18:26)
    at newConfig (/usr/local/lib/node_modules/netlify-cli/lib/settings/config.js:129:15)
    at Command.<anonymous> (/usr/local/lib/node_modules/netlify-cli/lib/settings/config.js:150:5)
    at Command.listener (/usr/local/lib/node_modules/netlify-cli/node_modules/commander/index.js:301:8)
    at emitTwo (events.js:106:13)
    at Command.emit (events.js:191:7)
```

Inside any other directory, it seems to work just fine.

loicteixeira:
Fix #32.

Currently, calling `netlify` commands from within the home directory will crash because `netlify` first saves the *global* config at `~/.netlify/config` (so `~/.netlify/` is a directory) and then try to save the *local* config at `~/.netlify` (as a file).

With this PR, if the CLI detects that the command is ran from the home directory, it will save the local config as `.netlify.local` instead of `.netlify` to prevent the names from clashing. In any other cases, it keeps the existing `.netlify` to allow backward compatibility.

Another approach would be to allow the local config file name to be overwritten via environment variable (e.g. `LOCAL_CONFIG_FILENAME`), the same way it is already possible some other options. Note that it would mean that the user has to first encounter the error, learn about such setting and then use it. On the other hand, with the change from this PR, the CLI will work automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,5 @@ Netlify creates different sites with each their own URL for each of your environ
 - netlify-cli is known to hang with an "ECONNRESET" error (parsed as JSON, it will look odd) when used from many CI environments.  This is a known issue that is only fixed in our alternate and current CLI:  https://github.com/netlify/netlifyctl
 
 - netlify-cli is known to hang when used with Node.js version 8.1.0.  Version 8.1.2 works well
+
+- the local config file is named `.netlify.local` instead of `.netlify` when you run commands from within your home directory.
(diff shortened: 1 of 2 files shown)


loicteixeira:
@biilmann @calavera: Is there anything I should do to get this merged? Maybe you'd prefer another approach?

rybit:
Hi - sorry to inform you (and that we didn't earlier) but this project is no longer supported (noted in the readme). 

Please use: https://github.com/netlify/netlifyctl

loicteixeira:
Oh, I can see now that the deprecation warning got merged on the same day I created the PR so I missed it. Anyway, thanks for the answer @rybit.

Issue #79, opened 2018-08-29
Opening browsers isn't working on windows


bcomnes:
The open command is having issues opening on windows and linux.  Look into improving this.

verythorough:
Actually, it appears to be working on Linux, at least on Ubuntu with a full (default) GUI.

To clarify on the Windows issue, the browser _will_ open for the `netlify login` command. However, it will not open for the `netlify open` commands.

This is the output for `netlify open:site`:
```
Opening "cli-function-test" site url:
> https://cli-function-test.netlify.com
```

DavidWells:
`this.exit`
https://github.com/netlify/cli/blob/master/src/commands/open/site.js#L35
might be causing the process to die early.

Can you comment out that line (if you have it running locally) and see if
that fixes the `netlify sites:open` command?

PR #89, opened 2018-08-31
Use an alternate open browser lib

bcomnes:
Closes https://github.com/netlify/cli/issues/79


verythorough:
Tested `open:site` and `open:admin` locally on Windows and both worked! 🎉

PR #100, opened 2018-09-08
install base modules

DavidWells:
Install base modules

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,6 +1,6 @@
 [build]
   # This will be run the site build
-  command = "cd site && npm install && npm run build"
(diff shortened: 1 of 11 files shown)


netlify.toml
```diff
@@ -1,6 +1,6 @@
 [build]
   # This will be run the site build
-  command = "cd site && npm install && npm run build"
+  command = "npm install && cd site && npm install && npm run build"
```
bcomnes: Does the build bot not install top level automatically when you cd around and npm i?

bcomnes: TIL

PR #109, opened 2018-09-14
Use netlify/js-client

bcomnes:
Extract the open-api js-client into its own module and consume it in the cli.


src/commands/sites/list.js
```diff
@@ -48,6 +49,9 @@ this.log(`
         if (s.repo_url) {
           console.log(`  ${chalk.whiteBright.bold('repo:')} ${chalk.white(s.repo_url)}`)
         }
+        if (s.account_name) {
+          console.log(`  ${chalk.whiteBright.bold('account:')} ${chalk.white(s.account_name)}`)
```
bcomnes: Added which account the site lives under

src/commands/sites/list.js
```diff
@@ -9,14 +9,15 @@ class SitesListCommand extends Command {
     const { api } = this.netlify
     await this.authenticate()
 
-    const sites = await api.listSites()
+    const sites = await api.listSites({ filter: 'all' })
```
bcomnes: List all sites you have access too

bcomnes: We should discuss how to support pagination in the api client.

src/commands/deploy.js
```diff
@@ -61,7 +58,7 @@ ${chalk.cyanBright.bold('netlify deploy --dir your-build-directory')}
 # deploy to live URL with the --prod flag
 ${chalk.cyanBright.bold('netlify deploy --dir your-build-directory --prod')}
 `)
-    this.exit()
+      this.exit()
```
bcomnes: All of these formatting changes were done by prettier.  Sorry about that.

src/commands/addons/create.js
```diff
@@ -1,5 +1,5 @@
 const Command = require('../../base')
-const { getAddons, createAddon } = require('../../utils/api/addons')
+const { getAddons, createAddon } = require('netlify/src/addons')
```
bcomnes: I can start porting this over to the open-api repos.

PR #114, opened 2018-09-18
Expose fetch error details

bcomnes:
Expose more details when fetch request fail.


bcomnes:
The issue that required this was resolved upstream.  We can add something else if errors are leaking from js-client.

Issue #117, opened 2018-09-19
Add plugins command


DavidWells:
Add ability for users to install additional commands

```
# list installed plugins
netlify plugins
# install plugin
netlify plugins:install netlify-name-of-thing-plugin
# remove plugin
netlify plugins:uninstall netlify-name-of-thing-plugin
# updates installed plugins to latest versions
netlify plugins:update
```

Plugins should follow a naming convention like: `netlify-name-of-thing-plugin`.  

(`netlify-{name-of-thing}-plugin`)

Future:

```
# Search plugin repo for available plugins 
netlify plugins:search query
```

- [ ] Add commands
- [ ] document how to create plugins

sw-yx:
ideally also document how plugins are different from addons (different access levels etc)

bcomnes:
Oh yeah, good point.  Might be a point of confusion.

PR #132, opened 2018-10-05
Add a plugin system

bcomnes:
We want to add a plugins system so you can add additional tools that need to interact with your netlify account from the CLI.  This PR will implement and experiment with a plugins system.

Closes https://github.com/netlify/cli/issues/117


bcomnes:
Doing it now.

PR #140, opened 2018-10-09
Simplify command reference site

verythorough:
**- Summary**
- Removes content duplicated in Netlify docs. 
- Simplifies site structure to a single level.
- Comments out "Edit this page" links that point to auto-generated content. We could add these links back if they point to the src files, but they're not as straighforward as editing your typical md docs files.
- Fixes docs-generating scripts for running on Windows

**- Test plan**
Check the deploy preview and README for broken links, or any other textual errors.

Goats in a tree


README.md
```diff
@@ -35,130 +31,70 @@ Welcome to the Netlify CLI! The new 2.0 version (now in beta) was rebuilt from t
 </details>
 <!-- AUTO-GENERATED-CONTENT:END -->
 
-## Installation
```
bcomnes: We can't get rid of installation and a basic usage example from the README.  Its slightly duplicate content, but provides expected context if you arrive to the CLI on the GitHub repo or npm page.

README.md
```diff
@@ -35,130 +31,70 @@ Welcome to the Netlify CLI! The new 2.0 version (now in beta) was rebuilt from t
...
-```
-
-### CI and Environment Variables.
-
-The following environment variables can be used to override configuration file lookups and prompts:
-
-- `NETLIFY_AUTH_TOKEN` - an access token to use when authenticating commands. **KEEP THIS VALUE PRIVATE**
-- `NETLIFY_SITE_ID` - override any linked site in the current working directory.
-
-## Getting Started + Docs
```
bcomnes: We can remove this part though, since I already moved these links to the very tippy top of the README

README.md
```diff
@@ -35,130 +31,70 @@ Welcome to the Netlify CLI! The new 2.0 version (now in beta) was rebuilt from t
...
 
-### [watch](/docs/commands/watch.md)
+### [watch](/docs/watch.md)
 
 Watch for site deploy to finish
 
 
 <!-- AUTO-GENERATED-CONTENT:END -->
 
-### telemetry
```
bcomnes: Whats the reason for removing this?

site/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "start": "npm run sync && run-p start:*",
     "build": "npm run clean && run-s build:*",
-    "build:docs": "DOCS_GEN=TRUE node ../scripts/docs.js",
```
bcomnes: What does this ENV var do again?

src/utils/header.js
```diff
@@ -2,8 +2,8 @@ const chalk = require('chalk')
 const pkg = require('../../package.json')
 
 module.exports = function header(context) {
-  const title = `${chalk.bgBlack.cyan('⬥ Netlify CLI')} ${chalk.red('(BETA)')}`
-  const docsMsg = `${chalk.greenBright('Read the docs:')} https://cli.netlify.com`
+  const title = `${chalk.bgBlack.cyan('⬥ Netlify CLI')}`
+  const docsMsg = `${chalk.greenBright('Read the docs:')} https://www.netlify.com/docs/cli`
```
bcomnes: Good catch

README.md
DavidWells: I agree with bret. We cannot remove this

README.md
verythorough: I moved it into the main doc: https://deploy-preview-1786--www.netlify.com/docs/cli/#usage-data-collection

bcomnes: Ahh yeah, seeing that now.  Fine with me.

site/package.json
verythorough: Removes the special characters made by chalk. But we don't need to set it before calling docs.js because it's set within the script itself: https://github.com/netlify/cli/blob/master/scripts/docs.js#L7

bcomnes: re slack discussion: we set this in the script file.  This way we can avoid cross-env https://github.com/netlify/cli/pull/140/files#diff-032384f2d383805f5dabad18c6e37353L7

README.md
verythorough: 👍 Updating now

Issue #177, opened 2018-11-08
Allow additional configuration of manual deploys


sam3d:
I apologise if this isn't the right place to post this, but I'd like to make a feature request.

---

When running the `netlify init --manual` process, certain options that would be useful to configure are not available. For example:

1. The CLI will attempt to parse the `origin` URL (which currently suffers from `https` parsing errors as per #78), and no other remote can be specified. Ideally, there would be an additional prompt that contains the default parsed `ssh` url, and _allows for a custom SSH URL to be specified_ should the user wish (this should also stipulate that the URL be of type SSH, and probably validate against a non-ssh URL).

2. The "Deploys from manual" in the web ui has a few things about it that aren't ideal. For example, that text that says "Manual" should be either be entirely user configurable (so they can, say, put `Custom Ops Repo` or `GitLab Personal Env`), or so that it uses all or part of the SSH git URL that's provided (such as `organisation/repo`, `personal-gitlab-instance.com:organisation/repo`, or `git@gl-instance.com:org/repo`, for example).

<img width="723" alt="screenshot 2018-11-08 at 01 28 16" src="https://user-images.githubusercontent.com/8385528/48171878-a4875e80-e2f5-11e8-9b6c-3550fbc183c0.png">

3. Also, the URL for clicking on the "manual" appears to take a weird form:

```
https://app.netlify.com/sites/{site_name}/{git_origin_url}
```

Thus making any visitation of that URL, of course, resolve to a `404`:

<img width="1115" alt="screenshot 2018-11-08 at 01 31 21" src="https://user-images.githubusercontent.com/8385528/48172002-0f389a00-e2f6-11e8-8644-e327e0176e2c.png">

This one I'm not entirely sure what the best way to resolve would be. Could be an option to set repository HTTPS URL, and if it isn't set then don't allow the hyperlink at all?

---

Also, for the sake of this issue, I've been focusing on the CLI. Is there anyway that any of these configuration options could make their way into the `netlify.toml` file? I personally believe "infrastructure as code" is an excellent practice, and so the more configuration that can take place in these files the better. Though, I totally understand that might not be the domain of this repo.

bcomnes:
1. Fair, I see your PR ty!
2. Yes, agreed its not ideal.  Ill post back any details I find along the way.
3. Same ^^
[...]

PR #178, opened 2018-11-09
Allow CI setup without a local clone from sites:create command

sam3d:
**- Summary**

This PR attempts to offer some solutions to #177. Therefore it does two things:

1. When performing manual configuration, there is now an additional step that offers the option to change the remote repository (while defaulting to the one that it could read locally) and also validates against non-SSH URL's

2. When using `sites:create`, it now gives the option to set up Continuous Integration using all of the available methods in `configManual`, including the new option to set the remote git URL. My favourite new feature here is that this command can now be used to entirely configure manual CI without needing to have the repository local at all

**- Test plan**

The following is the output when using the `sites:create` with my default team and a random name:

screenshot 2018-11-09 at 02 12 46

This did not require me to have any local git repository, and all of the details provided were used correctly. I've been able to verify this as the site deployment says it "deploys from 'manual'" and when hovering over that link, I am greeted with this hyperlink:

```
https://app.netlify.com/sites/clever-curran-7704a4/git@code.samholmes.net:sam/test.git
```

Thus showing that the remote git URL has been set correctly on the `repo` object. Having set up my personal GitLab instance with the webhook and deploy key, I can confirm that `git push` deploys are working as expected without needing to have a local repo instance on hand.

---

Now, when using the `netlify init --manual`, you can see that I have the option pre-filled in my local repo for the git deployment information:

screenshot 2018-11-09 at 02 08 50

If this does not conform to the following regex (to verify that it is an ssh URL):

```regex
/(ssh:\/\/|[a-zA-Z]*@|[a-zA-Z.].*:(?!\/\/))/
```

([testing of this regex](https://regexr.com/42pg4))

Then the following error will occur:

screenshot 2018-11-09 at 02 09 03

Thus preventing the user from continuing unless a correct URL is provided. This has the added benefit of safe-guarding against any potential incorrect automatic parsing of the URL from HTTPS to SSH, and ensuring that `repo` information cannot be updated unless the remote repository is _definitely_ using SSH (which, all going according to plan, should mean that the user can simply just press return to use the default without a hitch).

Add the ability to set up Continuous Integration from `sites:create`, and the option to manually configure the remote repo that CI uses

image

He's such a good boy.

### DIFF
--- a/src/commands/sites/create.js
+++ b/src/commands/sites/create.js
@@ -5,7 +5,9 @@ const chalk = require('chalk')
 const Command = require('../../base')
 const renderShortDesc = require('../../utils/renderShortDescription')
 const { track } = require('../../utils/telemetry')
-const configManual = require("../../utils/init/config-manual");
+const configManual = require('../../utils/init/config-manual')
+const parseGitRemote = require('parse-github-url')
+const configGithub = require('../../utils/init/config-github')
 
 class SitesCreateCommand extends Command {
   async run() {
@@ -13,11 +15,12 @@ class SitesCreateCommand extends Command {
     const { api } = this.netlify
 
     await this.authenticate()
-    let accountSlug = flags['account-slug']
-    let name = flags.name
+
     const accounts = await api.listAccountsForUser()
     const personal = accounts.find(account => account.type === 'PERSONAL')
-    if (!name || !accountSlug) {
+
+    let name = flags.name
+    if (!name) {
       console.log('Choose a site name or leave blank for a random name. You can update later.')
       const results = await inquirer.prompt([
         {
@@ -25,7 +28,14 @@ class SitesCreateCommand extends Command {
           name: 'name',
           message: 'Site name (optional):',
           filter: val => (val === '' ? undefined : val)
-        },
+        }
+      ])
+      name = results.name
+    }
+
+    let accountSlug = flags['account-slug']
+    if (!accountSlug) {
+      const results = await inquirer.prompt([
         {
           type: 'list',
           name: 'accountSlug',
@@ -38,7 +48,6 @@ class SitesCreateCommand extends Command {
         }
       ])
       accountSlug = results.accountSlug
-      name = results.name
     }
 
     let site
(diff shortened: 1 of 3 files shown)


bcomnes:
Very cool @sam3d! I'll try to take a look today or early next week.

mpartipilo:
Will this work with ssh running under a different port?

sam3d:
Ooooh I didn’t clock that edge case, thank you for the new Regex!

I’ll update it later this evening 😊

mpartipilo:
You're welcome :smiley: Thanks to you, actually. I was stuck trying to deploy something to Netlify. I was struggling with the cli tool, going over the docs, then I found this life saver PR.

I did notice that with your branch `netlify init --manual` doesn't ask for a site name anymore. It just picks a random one. Not a big deal though.

sam3d:
Oops thank you for the the site name - that's not desired behaviour so I'll take a look at that. And you're welcome! Solved an exact issue I had where the web UI doesn't offer the ability to set up Continuous Integration unless it's:

1. An existing remote git repo (which is fine)
2. On a hosted code provider (which is less useful)

---

Also, I didn't realise this when I responded from my phone yesterday, but I'm pretty sure the regex I wrote includes support for different ports (with the top line being the testing example):

screenshot 2018-11-23 at 13 11 31

(https://regexr.com/42pg4)

Did you find this wasn't working?

sam3d:
Ah thank you for explaining! The only issue is that a path can appear directly after the first semicolon in regex, and that appears to un-match those edge-cases where:

1. the `ssh://` isn't provided, and
2. the standard port is used, so a path is the first thing after a colon

screenshot 2018-11-23 at 13 25 33

https://regexr.com/43lot

I was erring on the side of less strict with my regex, with the hope that the additional option to configure the URL and also users seeing the "SSH URL" signifier being enough to prompt anything the regex didn't catch.

mpartipilo:
I tried a few different attempts at a prettier regex and I failed. I understand your approach better now, the port number is not required. A more relaxed rule set is indeed preferable.

sam3d:
I also tried and failed, not to worry 😄

mckelvaney:
Any news on this being reviewed & merged @bcomnes ?
It would be great to if this made it in as this exact issue cost me quite some time yesterday! TIA.

bcomnes:
Very sorry guys, I'll try to put some eyes on this tomorrow.

bcomnes:
I spoke to soon, I wont have focus for this till moday.

sam3d:
That's okay, totally understand you must be busy. Thank you for keeping us in the loop, and let me know if there's anything I can do to make this easier on you :-)

bcomnes:
okay! Reviewing :)

bcomnes:
Cool! I like the premise.  Digging in

src/commands/sites/create.js
```diff
@@ -95,6 +108,12 @@ SitesCreateCommand.flags = {
   'account-slug': flags.string({
     char: 'a',
     description: 'account slug to create the site under'
+  }),
+  'manual': flags.boolean({
```
bcomnes: First pass here, I'm thinking we rename this to `--ci` or something, and that lets you set up CI for the new site, independent of the filesystem / current project you are in.  That would also mean we would need to acomidate the github configurator.  I can take it away from here if you want.

bcomnes:
I also noticed some issues with init today as well, say if you manual deploy then go try to init CI, there are blockers and it requires some roundabout method tp fix.

bcomnes:
Okay added this in a way I can live with.  Its not perfect, but we need to rework a lot of this code someone soon in order to update to gh-apps and move a lot of the logic into our API.

bcomnes:
Thanks for your input everybody!  I'll go ahead and merge this and release something tomorrow.  Please open up any new issues for anything else left on the table.

PR #186, opened 2018-12-04
Improve init flow in various manual depoy states

bcomnes:
This makes init work when you made a deploy to a site manually and want to upgrade to using CI.  It also adds a force flag so you can re-init a site that is already using CI.


bcomnes:
Incorporate after https://github.com/netlify/cli/pull/178

PR #189, opened 2018-12-05
Add simple JSON parsing to parse raw for addons commands

bcomnes:
This adds a naive json parsing algorithm to the addons command.


bcomnes:
I need this sooner than later today, so merging.  We can revise easily as this is an unreleased feature.

PR #253, opened 2019-02-21
Centralize CI in Azure DevOps.

calavera:
So we don't have to use two different services.
Travis' future is pretty unknown at the moment, I rather move away.


bcomnes:
Fine with the change and all on board with centralizing on a single sevoce. 

Do you have wind Travis is going away though?  Curious for my own sake.

bcomnes:
Nvmd just saw the news

PR #254, opened 2019-02-21
[WIP] Add integration tests

DavidWells:
**- Summary**

Start of integration tests for CLI commands

- [x] Addons flow
- [x] Add site creation/teardown to before/after tests

Add integration tests


bcomnes:
Still looking to land this @DavidWells ?

DavidWells:
Yeah we need a NETLIFY api key added to the CI process for it to work.

Do you have access to that @bret?

Basically the integration tests setups up a Netlify site, execs a given command, checks the command did the right thing, then tears down the site

bcomnes:
Good idea.  I have access to travis and appveyor, not azure pipelines.

src/commands/addons/delete.js
```diff
@@ -1,12 +1,11 @@
 const Command = require('@netlify/cli-utils')
 const { getAddons, deleteAddon } = require('netlify/src/addons')
-const { parseRawFlags } = require('../../utils/parse-raw-flags')
 const { flags } = require('@oclif/command')
 
 class AddonsDeleteCommand extends Command {
   async run() {
     const accessToken = await this.authenticate()
-    const { args, raw } = this.parse(AddonsDeleteCommand)
```
DavidWells: raw flags are needed to parse all the variable options for add-on commands

I think its ok to remove for the delete command here but just a heads up.

We might need to add back in, if use case arises where add-ons accept args on deletions.

raeesbhatti: I had to change this because the `delete` command was not respecting the `--force` flag for some reason. But i tried to keep the change to minimal.

PR #289, opened 2019-04-09
Use cli-utils

bcomnes:
Use CLI-utils and incorporate a bit of refactoring and long needed cleanup.

Also fixes tests.  Closes https://github.com/netlify/cli/issues/295


bcomnes:
Rebased

bcomnes:
I need to add at least one test back to the repo since the ones we had have been externalized.

Issue #232, opened 2019-01-30
Better CLI Docs


phae:
Converted from David's note:
Autogenerating animated SVGs of commands would be l33t.

mischkl:
At the moment docs are completely missing any information on the "dev" command. Meanwhile I thought this was a headline feature. IMHO this should be added ASAP! 😉

PR #303, opened 2019-04-24
add dev + functions command to ref site

DavidWells:
Add dev + functions to ref site

https://github.com/netlify/cli/issues/232#issuecomment-486131073


site/package.json
```diff
@@ -11,7 +11,8 @@
   "scripts": {
     "start": "npm run sync && run-p start:*",
     "build": "npm run clean && run-s build:*",
-    "build:docs": "node ../scripts/docs.js",
+    "postbuild": "cp src/_redirects dist/_redirects",
```
bcomnes: Does this need to happen after the build? Or could it be a `build:redirects` step?  If so, lets just make it a build step  so that insinuation isn't there.  If it needs to be post build, then fine.

PR #324, opened 2019-06-02
Branded Install Message

sdras:
**- Summary**
Currently, any output from dependencies is the only thing the user sees when installing this package:
Screen Shot 2019-06-02 at 8 40 30 AM

It would be better if the most bold thing users see is some sort of success and branded install of the CLI. Working on another PR in hopes of suppressing that output as well.

Open to suggestions for what it looks like, just mostly getting something in first.

**- Test plan**
Screen Shot 2019-06-02 at 11 44 28 AM

Create branded install message with ascii art

sudo


sw-yx:
do you mind taking out the `npm-shrinkwrap.json` from your PR? let me handle that.

sdras:
We should talk more about the branding here, it goes against the branding doc I saw internally. I got ahead of myself. :)

PR #327, opened 2019-06-03
fix: Don't show broken url if a site is undeployed

zastavnitskiy:
**- Summary**
`netlify sites:list` shows the wrong URL if a site is undeployed — the URL doesn't contain the hash of the deployment ↓

Screenshot 2019-06-03 at 22 17 17

Todo:

- [x] Show broken status for sites with no successful deployments at all
- [ ] Show status for sites with broken recent deployments(with successful ones in the past)

**- Description for the changelog** ↑

This MR only fixes the client. Instead of using site.ssl_url we will try to use last published deploy to check and display the correct URL.

Otherwise, display error status.

**- Test plan**

1. Create a site that fails to deploy — for example by specifying a wrong build script.
2. Push, wait for the failed build.
3. Run `netlify sites:list`

### DIFF
--- a/src/commands/sites/list.js
+++ b/src/commands/sites/list.js
@@ -19,7 +19,9 @@ class SitesListCommand extends Command {
           id: site.id,
           name: site.name,
           ssl_url: site.ssl_url,
-          account_name: site.account_name
+          account_name: site.account_name,
+          published_deploy: site.published_deploy,
+          admin_url: site.admin_url
         }
 
         if (site.build_settings && site.build_settings.repo_url) {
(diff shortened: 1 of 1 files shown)


src/commands/sites/list.js
```diff
@@ -18,8 +18,9 @@ class SitesListCommand extends Command {
         const siteInfo = {
           id: site.id,
           name: site.name,
-          ssl_url: site.ssl_url,
-          account_name: site.account_name
```
sw-yx: i need to doublecheck this as i lack context on this specific command but whats up with the rename of these props?

zastavnitskiy: `site.ssl_url` is always defined in the API, even if the site is not deployed.

I needed to check if site is deployed, so I used `site.published_deploy`; because published_deploy also contained ssl_url, I decided to use that instead of `site.ssl_url`

I'm not sure if they are the same though. An alternative approach would be to keep using `site.ssl_url` and only use `published_deploy` as a flag.

sw-yx: oh man ok. thanks very much. i dont have bandwidth to look at this right now but appreciate the input and will need to come back to this to confirm that these are the right values.

zastavnitskiy: Hey, just fyi. There is one more case where cli output is broken — the site is deployed, but the recent build fails. It seems, that the API data doesn't contain any information about the recent broken build.

A possible solution might be to list the deployments for each site, and calculate status from there, but probably it's better to add the status of most recent build to the API.

In any case, this MR doesn't cover all use cases, so feel free to close.

sw-yx: thank you. still very tied up here but will get to it eventually. hugely appreciate the PR though!

PR #334, opened 2019-06-18
updates messaging for creating sites and error messages

jschatz1:
**- Summary**
Based on feedback from our CLI call. This PR improves the messaging when you create a site through the CLI and when that command fails. 

**- Test plan**

Screen Shot 2019-06-18 at 11 25 19 AM

Modifies the messaging to include the idea of a global site creation failure rather than a local one. Users will be confused when the site creation fails based on the name if the name is not outright determined to be a global conflict. 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Welcome to the Netlify CLI! The new 2.0 version was rebuilt from the ground up t
   * [dev](#dev)
   * [functions](#functions)
 - [Contributing](#contributing)
+- [Development](#development)
 - [License](#license)
 
 </details>
@@ -138,6 +139,16 @@ Manage netlify functions
 ## Contributing
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for more info on how to make contributions to this project.
+
+## Development
+
+You'll need to follow these steps to run Netlify CLI locally:
+
+    uninstall any globally installed versions of netlify-cli
+    clone and install deps for https://github.com/netlify/cli
+    npm link from inside the cli folder
+
+Now you're both ready to start testing and to contribute to the project!
 
 ## License
 
--- a/src/commands/sites/create.js
+++ b/src/commands/sites/create.js
@@ -2,6 +2,7 @@ const { flags } = require('@oclif/command')
 const inquirer = require('inquirer')
 const prettyjson = require('prettyjson')
 const chalk = require('chalk')
+const sample = require('lodash.sample')
 const Command = require('@netlify/cli-utils')
 const { track } = require('../../utils/telemetry')
 const configManual = require('../../utils/init/config-manual')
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -139,6 +140,16 @@ Manage netlify functions
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for more info on how to make contributions to this project.
 
+## Development
```
sdras: Nice, thanks for adding this.

src/commands/sites/create.js
```diff
@@ -19,7 +20,20 @@ class SitesCreateCommand extends Command {
...
-      console.log('Choose a site name or leave blank for a random name. You can update later.')
+      const userName = await api.getCurrentUser()
+      const suggestions = [
+        `super-cool-site-by-${userName.slug}`,
+        `the-awesome-${userName.slug}-site`,
+        `${userName.slug}-makes-great-sites`,
+        `netlify-thinks-${userName.slug}-is-great`,
+        `the-great-${userName.slug}-site`,
+        `isnt-${userName.slug}-awesome`,
+        `${userName.slug}-blog`
```
sdras: I love these! We don't know that they're making a blog though so this one should be altered or removed.

src/commands/sites/create.js
```diff
@@ -19,7 +20,20 @@ class SitesCreateCommand extends Command {
...
+      ]
+      const siteSuggestion = sample(suggestions)
```
sdras: Probably not a big thing, but not sure we need lodash.sample for this, it's just a one-liner. I guess I could check how big that is though.

src/commands/sites/create.js
```diff
@@ -59,9 +73,10 @@ class SitesCreateCommand extends Command {
         body
       })
     } catch (error) {
-      console.log(`Error ${error.status}: ${error.message} from createSiteInTeam call`)
       if (error.status === 422) {
-        this.error(`A site with name ${name} already exists. Please try a different slug`)
+        this.error(`${name}.netlify.com already exists. Please try a different slug.`)
```
sdras: Much better.

jschatz1:
I didn't want to include another library in `package.json` BUT I didn't want to create a new util file because that didn't exist yet, and since importing `lodash.sample` only adds that one method.

src/commands/sites/create.js
jschatz1: You read my mind. https://github.com/netlify/cli/pull/334#issuecomment-503194907

jschatz1: So I had it as 
```javascript
const siteSuggestion = suggestions[Math.floor(Math.random()*suggestions.length)]
```
But then I thought someone would ask me to put it in a util. Then I looked and saw we didn't have any utils for the CLI like this so I thought rather than rewriting a random function myself  I could inlude lodash's method which has _almost_ the same exact implementation. https://github.com/lodash/lodash/blob/master/sample.js. It should only import this one file and it has a slightly better implementation.

jschatz1: @sdras does that work for you? Or would you rather me put it into a util?

sdras:
Makes sense to me!

sdras:
🐑

PR #335, opened 2019-06-18
use a list of sites to chose from rather than entering the name manually

jschatz1:
**- Summary**
`netlify link` makes you manually enter the name of your site, which may result in human error and a tedious process. I wanted to give users a list of sites so there's no human error involved.

**- Test plan**
Screen Shot 2019-06-18 at 1 56 06 PM

Screen Shot 2019-06-18 at 1 58 47 PM

Screen Shot 2019-06-18 at 1 59 38 PM

Better DX for linking sites in `netlify link`

image

### DIFF
--- a/src/utils/link/link-by-prompt.js
+++ b/src/utils/link/link-by-prompt.js
@@ -8,10 +8,10 @@ const { track } = require('../telemetry')
 module.exports = async function linkPrompts(context) {
   const { api, state } = context.netlify
 
-  const SITE_NAME_PROMPT = 'Site Name'
-  const SITE_ID_PROMPT = 'Site ID'
+  const SITE_NAME_PROMPT = 'Choose from a list of your sites'
+  const SITE_ID_PROMPT = `Use a site ID`
 
-  let GIT_REMOTE_PROMPT = `Use current git remote URL`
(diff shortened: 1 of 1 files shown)


src/utils/link/link-by-prompt.js
```diff
@@ -132,25 +123,19 @@ Run ${chalk.cyanBright('`git remote -v`')} to see a list of your git remotes.`)
       }
 
       if (sites.length === 0) {
-        context.error(`No sites found named ${siteName}`)
+        context.error(`You don't have any sites. Use netlify 'sites:create' to create a site.`)
```
biilmann: At some point we could even consider letting the user pick a prompt to create a new site directly here.

biilmann:
LGTM - have we tested on an account with lots of sites?

src/utils/link/link-by-prompt.js
jschatz1: That's a great idea.

jschatz1:
@biilmann should be good with many sites: tada

verythorough:
I don't think I'd count ~15 sites as "many". Some people have hundreds!

sw-yx:
best to use autocomplete/filter to handle many sites

PR #338, opened 2019-06-20
Revert "Remove shrink-wrap pruning during pack step"

imorente:
**Summary**
This reverts commit c28bd8189b5a9e602a249e76c857c2d6af858ce4.

Dev dependencies listed in `npm-shrinkwrap.json` are getting installed when the CLI is installed globally.

This PR brings back the fix we had in place so only production dependencies are included before publishing.

**Test plan**
1. Run `npm pack`
2. Verify that `npm-shrinkwrap.json` in generated tarball doesn't include dev dependencies, such as ava.

**Description for the changelog**
Prune npm-shrink-wrap during pack step

**A picture of a cute animal (not mandatory but encouraged)**
corgi jumping on the beach


raeesbhatti:
This breaks the package and prevents it from being installed  when doing `npm i -g netlify-cli` for some reason.

raeesbhatti:
Actually, never-mind. The previous behavior was fixed by upgrading `npm`, clean cache and re-publishing.

PR #342, opened 2019-07-01
Allow user to choose another site name if current doesn't exist

raeesbhatti:
**- Summary**

When doing `netlify init` right now, if you choose a `Site name` that is already taken, the process fails and exits. This change allows the user to reenter a new `Site name` for their project without exiting.

**- Test plan**

First attempt with HEAD state:
- Run `netlify init` in a new project
- Choose `Yes, create manually deploy site`
- Enter `test` as `Site name`.
- Choose your team. The command will error out and exit with `Error: test.netlify.com already exists. Please try a different slug.`

Second attempt with this branch:
- Run `netlify init` in a new project
- Choose `Yes, create manually deploy site`
- Choose your team
- Enter `test` as `Site name`. The command will show a warning `Warning: test.netlify.com already exists. Please try a different slug.`. But will prompt you for `Site name` again.
- Enter a new unique name this time.
- The command will succeed

* Moved team selection before site name selection
* Move site name selection into an inline function `inputSiteName `
* Call `inputSiteName` again if specified site name already exists

golden-retriever-dog-breed-info

### DIFF
--- a/src/commands/sites/create.js
+++ b/src/commands/sites/create.js
@@ -18,33 +18,6 @@ class SitesCreateCommand extends Command {
 
     const accounts = await api.listAccountsForUser()
 
-    let name = flags.name
-    if (!name) {
-      const userName = await api.getCurrentUser()
-      const suggestions = [
-        `super-cool-site-by-${userName.slug}`,
-        `the-awesome-${userName.slug}-site`,
-        `${userName.slug}-makes-great-sites`,
-        `netlify-thinks-${userName.slug}-is-great`,
-        `the-great-${userName.slug}-site`,
-        `isnt-${userName.slug}-awesome`
-      ]
-      const siteSuggestion = sample(suggestions)
-
-
-      console.log(`Choose a unique site name (e.g. ${siteSuggestion}.netlify.com) or leave it blank for a random name. You can update the site name later.`)
-      const results = await inquirer.prompt([
-        {
-          type: 'input',
-          name: 'name',
-          message: 'Site name (optional):',
-          filter: val => (val === '' ? undefined : val),
-          validate: input => /^[a-zA-Z0-9-]+$/.test(input) || 'Only alphanumeric characters and hyphens are allowed'
-        }
-      ])
-      name = results.name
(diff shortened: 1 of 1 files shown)


src/commands/sites/create.js
```diff
@@ -61,23 +34,56 @@ class SitesCreateCommand extends Command {
...
+      } catch (error) {
+        if (error.status === 422) {
+          this.warn(`${name}.netlify.com already exists. Please try a different slug.`)
+          await inputSiteName()
+        } else {
+          this.error(`createSiteInTeam error: ${error.status}: ${error.message}`)
+        }
       }
     }
+    await inputSiteName(name)
```
DavidWells: How does this work in CI context?

IE if I'm running this in travis/circle/whateves the `netlify sites:create` (or netlify init that calls it) should error out or default to a randomly generated site name if the site name is taken.

It looks like this will hang in CI forever (until timeout) waiting for the user to input a new name. The user is a machine and Elon musk is still years away from cracking AI 😃

DavidWells: There is a native node lib `tty` that is supposed to know whether or not an interactive prompt (aka user) is present.

https://nodejs.org/api/tty.html

Perhaps we can use this to wrap the interactive bits of the CLI

The other "work around" we have been using is if all the required flags are supplied, never use a prompt

raeesbhatti: I'm not sure about assigning a randomly generated name if required one doesn't exist. I think the kind of flow we have right now, for example, we ask if the person wants to create a new site or link to existing, then we ask for their team and then the site name. This whole flow doesn't work without an interactive shell and without proper flags.
The changes I've made in regards to the site name, still respect the flags.
I can make the change to allow these options to fail if shell is not interactive, but I think we should discuss.
/cc @jschatz1 @sdras

raeesbhatti:
Hey @jschatz1, can you please take a look at this!
Thanks

raeesbhatti:
Can we merge this?

DavidWells:
lets do it

Issue #282, opened 2019-04-04
[easy] add .netlify to .gitignore if exists, on `netlify init`


sw-yx:
**- Do you want to request a *feature* or report a *bug*?**
[easy] add .netlify to .gitignore if exists, on `netlify init`

**- What is the current behavior?**

no action

biilmann:
I don't think we want to ignore the .netlify folder - it's useful that everyone in the same repo get the same site id if doing manual deploys

DavidWells:
`.netlify` will likely contain more things in the future (caches, plugin data, etc). 

If people want to manually set the ID (which limits the reuse of the repo), they can set that in the `netlify.toml` file 

I’d wager this property is not used very often in Netlify.toml. 

I think we should ignore the `.netlify` folder by default

Issue #219, opened 2019-01-23
Automatically ignore local .netlify folder


DavidWells:
After running`netlify link` automatically add `.netlify` folder to `.gitignore` file.

This will help keep local files out of git

DavidWells:
Also for `netlify init`

bcomnes:
Its kind of side-effecty.  Another alternative would be to prompt for this or print a message.

raeesbhatti:
I think this was fixed, please re-open if not.

sionleroux:
The current implementation is a bit annoying because it keeps creating an ignore file that I don't need. I have `/.netlify/` ignored at a higher level than the individual project, so I don't need it in each repo. The link `link dev` command makes it anyway. I found an option to avoid this by running `netlify dev --skip-gitignore` but this is annoying to type so now I have made a bash alias that includes this flag.  When I forget to use it I keep having to clean up unwanted gitignore files every time I run `netlify dev`.

I had a look at the code that implements this issue in the PR mentioned above and it seems to be parsing the gitignore file by hand (instead of using git) and assuming that the file is in the directory the command is being run from. However if I add an ignore rule 1 directory up, my current directory has none, and this implementation will miss it and re-add it even though it is already in effect. Consider the output of the command `git check-ignore .netlify` ([docs](https://git-scm.com/docs/git-check-ignore)) which returns a positive result ".netlify" even when I have no .gitignore in the project because it also reads rules from:

- .git/info/exclude
- the global excludes file [configured with core.excludesfile](https://git-scm.com/docs/gitignore#_configuration)

and returns empty with a return code of 1, if I comment it out e.g. from the global excludes file.

The current implementation naïvely parses only the .gitignore in the current directory, adding an ignore rule when it is already ignored, i.e. git status doesn't show a .netlify folder. The assumption is usually reasonable, but right now it's not the case that the whole team is using the netlify tool and needs this ignore, it's just me using it.

Can we either improve the "ignore detection" or provide an option to configure the CLI to use "skip-gitignore" as the default? If it's a bad practice to shell out to the git-cli maybe there is already a git package in npm that supports something like git check-ignore with a native javascript implementation?

_P.S. Sorry for the essay, I think I got a bit overenthusiastic_

PR #358, opened 2019-07-18
Automatically Ignore .netlify folder on init/link

DavidWells:
Fixes https://github.com/netlify/cli/issues/282 & https://github.com/netlify/cli/issues/219

This will automatically add `.netlify` folder to gitignore folder of folder.

It will ignore `.gitignores` that already contain `.netlify`

Runs whenever `netlify init` or `netlify link` is ran

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ vendor
 # osx
 .DS_STORE
 .vscode
-.netlify
 coverage
 test-site
 site/dist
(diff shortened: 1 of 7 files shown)


DavidWells:
This only runs when `netlify link` and `netlify init` are run. 

Those should only happen once in a given directory

If the user wants to not ignore the `.netlify` folder they can ignore the change.

DavidWells:
If people complain we can add another check to not add to `.gitignore` if the `.netlify` folder is already in a project.

The standard dev flow would be:

```
git clone repo
netlify link (or init)
```

Issue #385, opened 2019-07-31
bug: netlify sites:create --json output is blank


moltar:
**- Do you want to request a *feature* or report a *bug*?**

bug

**- What is the current behavior?**

```
netlify sites:create --name foo -a bar --json
```

Produces an empty output.

**- If the current behavior is a bug, please provide the steps to reproduce.**

```
netlify sites:create --name foo -a bar --json
```

**- What is the expected behavior?**

Output is a valid JSON with information about newly created site.

**- Please mention your node.js, and operating system version.**

❯ node -v
v10.15.3

DavidWells:
Thanks for the report. We need to add `this.logJson(site)` here https://github.com/netlify/cli/blob/master/src/commands/sites/create.js#L154 to support the json output of the site data

PR #391, opened 2019-08-09
sites:create Fix JSON log

raeesbhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/385
Logs site data as JSON when `--json` flag is passed to `sites:create`

**- Test plan**
`netlify sites:create --json`

* Import `lodash.pick`
* Pick compact and meaningful properties from `site` object and display them

### DIFF
--- a/src/commands/sites/create.js
+++ b/src/commands/sites/create.js
@@ -3,6 +3,7 @@ const inquirer = require('inquirer')
 const prettyjson = require('prettyjson')
 const chalk = require('chalk')
 const sample = require('lodash.sample')
+const pick = require('lodash.pick')
 const Command = require('@netlify/cli-utils')
 const { track } = require('../../utils/telemetry')
 const configManual = require('../../utils/init/config-manual')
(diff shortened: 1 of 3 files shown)


src/commands/sites/create.js
```diff
@@ -152,6 +153,10 @@ class SitesCreateCommand extends Command {
       }
     }
 
+    if (flags.json) {
+      this.logJson(pick(site, ["id","state","plan","name","custom_domain","domain_aliases","url","ssl_url","admin_url","screenshot_url","created_at","updated_at","user_id","ssl","force_ssl","managed_dns","deploy_url","account_name","account_slug","git_provider","deploy_hook","capabilities","id_domain"]))
```
DavidWells: Does this log out env vars of the site?

Can't remember where that was in the object but we want to redact those

raeesbhatti: @DavidWells ENV vars are part of `build_settings` which is excluded from the output.

raeesbhatti:
@DavidWells Going to merge this. Can revert if needed.

Issue #105, opened 2018-09-10
Allow account switching


bcomnes:
Allow users to switch which account they are logged in.  This is easily added due to the enhanced config structure we added recently.

PR #395, opened 2019-08-12
Add command to switch active Netlify account

raeesbhatti:
**- Summary**

Adds a new command (`netlify switch`) to allow you to switch between different Netlify accounts in CLI.
Fixes: https://github.com/netlify/cli/issues/105

**- Test plan**

* `netlify switch`
* Select "I would like to login to a new account". You'll be logged into the new account and your current account in CLI will be updated to reflect the latest.
* `netlify switch`, you can now see both accounts and switch between them

* Adds a new flag to `netlify login` (`--new`), that allows you to login to a new Netlify account even if you're already logged in
* Adds a new command `netlify switch` that allows you to switch between already logged in accounts and login to a new one

why-do-i-want-to-crush-cute-animals-1432782372

### DIFF
--- a/src/commands/login.js
+++ b/src/commands/login.js
@@ -1,10 +1,12 @@
+const { flags } = require('@oclif/command')
 const Command = require('@netlify/cli-utils')
 const chalk = require('chalk')
 
 class LoginCommand extends Command {
   async run() {
+    const { flags } = this.parse(LoginCommand)
     const [ accessToken, location ] = this.getConfigToken()
-    if (accessToken) {
+    if (accessToken && !flags.new) {
       this.log(`Already logged in ${msg(location)}`)
       this.log()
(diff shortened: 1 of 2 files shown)


sw-yx:
i never knew we were working on this. imo its fine to run two commands `logout` and `login` again rather than support this

raeesbhatti:
I personally think this will be useful since doing `logout` and `login` requires following the whole login flow each time. While this is simple and quick.
Looking for more opinions.

sw-yx:
david clarified that its useful for "Work vs personal accounts. Also agencies
And support" :) up to you really i was just adding random not very helpful comments and i will stop now

raeesbhatti:
Haha no problem. Your comments are always welcome 🙌🏽

raeesbhatti:
Going to merge this since there's a general positive signal. Can revert if there's an objection.

PR #489, opened 2019-08-28
Deploy: Add a timeout flag

raeesbhatti:
**- Summary**
Add a new flag (`-p`) to `netlify deploy` to allow users to configure timeout for deployments.

**- Test plan**

* Run `netlify deploy -p 5`
* Expect to error out if deployment takes more than five seconds

* Add new flag (`--timeout`) to `deploy` command

🐶


DavidWells:
noiceeeeee!

sw-yx:
offer to persist this preference?

raeesbhatti:
@sw-yx Since this flag will be heavily used in automated environments, it think we should keep the whole flow non-interactive.

sw-yx:
k

PR #504, opened 2019-09-02
Update ISSUE_TEMPLATE.md

aisflat439:
Today I had to file a possible issue. I noticed that the issue template requests 
`--info`
which isn't a valid flag.
I switched it to `--verbose` which potentially is enough.

**- Summary**

**- Test plan**


DavidWells:
Thanks!

Issue #507, opened 2019-09-04
rename prompt from "deploy path" to "publish directory"


sw-yx:
`netlify deploy` currently prompts for "deploy path" 

https://github.com/netlify/cli/blob/b373eecdeab44ee913148666e25cf693a25730d7/src/commands/deploy.js#L89

this is non standard terminology. as i've agreed with @verythorough, we should use the same terminology everywhere to limit confusion.

sw-yx:
also if we read from netlify.toml we should say so but thats another issue..

PR #508, opened 2019-09-04
rename deploy path 

sw-yx:
closes https://github.com/netlify/cli/issues/507

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -80,13 +80,13 @@ class DeployCommand extends Command {
     }
 
     if (!deployFolder) {
-      this.log('Please provide a deploy path relative to:')
+      this.log('Please provide a publish directory (e.g. "public" or "dist" or "."):')
       this.log(process.cwd())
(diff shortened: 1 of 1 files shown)


DavidWells:
nice

src/commands/deploy.js
```diff
@@ -80,13 +80,13 @@ class DeployCommand extends Command {
...
     if (!deployFolder) {
-      this.log('Please provide a deploy path relative to:')
+      this.log('Please provide a publish directory (e.g. "public" or "dist" or "."):')
       this.log(process.cwd())
       const { promptPath } = await inquirer.prompt([
         {
           type: 'input',
           name: 'promptPath',
-          message: 'deploy path',
+          message: 'Publish Directory',
```
verythorough:
```suggestion
          message: 'Publish directory',
```

Minor change, and I'm sure it's not consistent throughout the CLI yet, but we're standardizing on sentence case across the board, so might as well make this particular prompt match before we do a broader update to all prompts.

Issue #443, opened 2019-08-12
blank project with one line SPA _redirects file causes very weird redirect error


sw-yx:
**- Do you want to request a *feature* or report a *bug*?**
bug

**- What is the current behavior?**

Shawn Makinson [1:32 PM]
@swyx I started a fresh quasar project and tried a _redirects file. Was fine when empty, I added the `/*      /index.html       200` and I see the errors occur.

I figured that was a good one to start with
Looks like I see the term spit out ‘Rewrote URL to  /index.html’ about 4 or 5 times from one page load.

https://netlify.slack.com/archives/C7G3WHB7U/p1556040743038600

**- What is the expected behavior?**
it should just work

BrunoQuaresma:
Updates? I'm still having issues with _redirects file and redirects rule in .toml file.

vladimirdjurdjevic:
I am experiencing this issue also. Using netlify-cli 2.20.0. I am using --dir flag for simple static server, and it's still broken.

sw-yx:
this specific issue was resolved. appreciate both of you making new issues with repros as i suspect you have different issues.

Issue #428, opened 2019-08-12
tracker issue - Netlify Dev Redirects and the SPA rule doesnt work well with webpackdevserver


sw-yx:
**- Do you want to request a *feature* or report a *bug*?**

this is a known bug many (incl myself) run into. we should address this asap.

**- What is the current behavior?**

https://github.com/netlify/netlify-redirect-parser/issues/1

**- What is the expected behavior?**

it should work like in prod, which is harder because there is no devserver there

Issue #418, opened 2019-08-12
redirects should not redirect requests for existing files


sw-yx:
**- Do you want to request a *feature* or report a *bug*?**
bug

**- What is the current behavior?**
a simple single page app redirect rule:

```
/*    /index.html   200
```

ironically breaks most SPA's because the requests for js bundles ALSO redirects to index.html.

![image](https://user-images.githubusercontent.com/6764957/60351213-4ea21c80-9993-11e9-9a8e-8cb48d1b2483.png)

this is not how redirects work in production. we check first if there are existing files.

**- If the current behavior is a bug, please provide the steps to reproduce.**

the offending code is either https://github.com/netlify/netlify-dev-plugin/blob/master/src/commands/dev/index.js#L109-L129

or https://github.com/netlify/netlify-rules-proxy/blob/master/index.js#L163

depending where we want to fix this.

sw-yx:
cc @raeesbhatti

raeesbhatti:
I'm looking at the redirect issues right now.

raeesbhatti:
@sw-yx I'm unable to reproduce this issue here: https://github.com/raeesbhatti/netlify-redirect-test. Can you please provide a more elaborate example!

sw-yx:
please use CRA

PR #527, opened 2019-09-12
RulesProxy: ignore rediect rule for base

raeesbhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/418
Fixes: https://github.com/netlify/cli/issues/428
Fixes: https://github.com/netlify/cli/issues/443
Adding the following rule to redirects config breaks the any site that is running its own dev server through `netlify dev`:
```
/* /index.html 200
```
This rule is required for things to work correctly in production because the site will be compiled into html files.
But it causes `netlify dev` server to redirect every non-existing resource to `/index.html` which is usually means all of the scripts, stylesheets and images because those are not files yet but dynamic paths handled by the app's dev server.

The solution is to ignore this rule in `netlify dev` because it already redirects any non-existing resource requests to the application server.

**- Test plan**
* Create a `create-react-app` project
* Create `_redirects` file with `/* /index.html 200` as contents
* `netlify dev`
* See the app in browser

* Parse and take into effect rules defined in `_redirects` file in project root (was not done before)
* Ignore a rules that matches `/* /index.html 200` format

🦊

### DIFF
--- a/src/utils/rules-proxy.js
+++ b/src/utils/rules-proxy.js
@@ -31,34 +31,49 @@ function parseFile(parser, name, data) {
   return result.success
 }
 
-function parseRules(redirectFilePath, distTomlPath, baseTomlPath) {
+function parseRules(projectDir, publicDir) {
   let rules = []
-  if (fs.existsSync(redirectFilePath)) {
+
+  const generatedRedirectsPath = path.resolve(publicDir, '_redirects')
+  if (fs.existsSync(generatedRedirectsPath)) {
     rules = rules.concat(
       parseFile(
         redirectParser.parseRedirectsFormat,
         '_redirects',
-        fs.readFileSync(redirectFilePath, 'utf-8')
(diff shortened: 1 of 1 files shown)


sw-yx:
i think this is a good start, but I wonder how we should handle other forms of splat rules. for example, here are some test cases for netlify dev vs production behavior:

```
/en/* /en/404.html 404
/de/* /de/404.html 404
/:slug  /blog/:slug 200
/path/* param1=:value1 param2=:value2 /otherpath/:value1/:value2/:splat 301
/* /index.html 201
/* /index.html 301
/* /index.html 302
/* /index.html 307
/* /index.html 308
/* / 200
/api/* /.netlify/functions/:splat 200
/app/*  /app/index.html  200!
/api/*  https://api.example.com/:splat  200
/app/* /app 200
```

for the app/* stuff, you can test with https://github.com/sw-yx/jamstack-hackathon-starter which implements an SPA only at the /app/* url.

feel free to split this out to a separate issue if you wish

Issue #517, opened 2019-09-10
Redirect from root fails in dev, works on deploy


fineline:
**- Do you want to request a *feature* or report a *bug*?**

Bug

**- What is the current behavior?**

The following config works when deployed to netlify, but fails in netlify dev:

```toml
[[redirects]]
  from = "/"
  to = "/home"
  status = 200
```

I have tested other redirects such as:

```toml
[[redirects]]
  from = "/test"
  to = "/home"
  status = 200
```

and that works fine, so it appears to be just the root URL redirect (rewrite).

**- If the current behavior is a bug, please provide the steps to reproduce.**

* Add code above to netlify.toml
* Ensure a file exists in `/static/home/index.html`
* run `netlify dev`
* browse to http://localhost:8888
* Observe the message "Forbidden"

**- What is the expected behavior?**

When deployed to netlify, the page at `/home/index.html` is displayed for the raw domain URL. I would expect this same behaviour in netlify dev.

**- Local Environment Information**

────────────────────┐
 Netlify Site Info  │
────────────────────┘
 ›   Warning: Did you run `netlify link` yet?
 ›   Error: You don't appear to be in a folder that is linked to a site

Issue #432, opened 2019-05-09
Wildcard/splat _redirects file processing error


jgarber623:
**- Do you want to request a *feature* or report a *bug*?**

Bug, likely related to netlify/netlify-dev-plugin#147 and netlify/netlify-dev-plugin#155.

I did some digging and found that netlify-dev-plugin depends on the [@netlify/rules-proxy](https://www.npmjs.com/package/@netlify/rules-proxy) npm package which _is not_ open source.

I was able to open up the source for @netlify/rules-proxy on my computer and found it depends on (among other things) [netlify-redirect-parser](https://www.npmjs.com/package/netlify-redirect-parser) and [netlify-redirector](https://www.npmjs.com/package/netlify-redirector) (neither of which is open source). From there, debugging got more difficult, so I didn't pursue further.

**- What is the current behavior?**

Generate a site containing a folder structure something like:

```
|- _redirects
|- posts/
|  |- index.html
|- users/
|  |- index.html
```

In `_redirects`:

```
/posts/*    /posts/:splat.html    200!
/users/*    /users/:splat.html    200!
```

Using `netlify dev` and navigating to `http://localhost:<port>/posts` is erroneously rewritten to `http://localhost:<port>/posts/.html` which does not exist. This behavior is different than Netlify's production service which correctly rewrites correctly. Navigating directly to `http://localhost:<port>/posts/index.html` resolves correctly, but is not expected.

You can observe this by navigating to `https://whoistheorchid.com/releases` (which is backed by a file at `./releases/index.html`.

**- If the current behavior is a bug, please provide the steps to reproduce.**

Steps to reproduce are noted above.

**- What is the expected behavior?**

I'd expect `netlify dev` to process `_redirects` in the same way Netlify's production service does.

**- Please mention your node.js, and operating system version.**

macOS: 10.14.4
Node: v12.1.0
npm: 6.9.0

PR #533, opened 2019-09-14
Dev: Try static files before deciding on redirect

raeesbhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/517
Fixes: https://github.com/netlify/cli/issues/432
Fix the logic to check if a static file for given URL exists.

**- Test plan**
`netlify.toml`:
```
[[redirects]]
  from = "/"
  to = "/home"
  status = 200
```
* Ensure a file exists in /static/home/index.html
* run `netlify dev`
* browse to http://localhost:8888

* Move `alternativePathsFor` function from `dev` command `index.js` to `rules-proxy.js`
* Use `alternativePathsFor` to check if a path has static file before calling `next()`

🐥

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -18,6 +18,7 @@ const {
 } = require('netlify-cli-logo')
 const boxen = require('boxen')
 const { createTunnel, connectTunnel } = require('../../utils/live-tunnel')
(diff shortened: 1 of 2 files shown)


sw-yx:
yes! we've needed this for forever!

PR #535, opened 2019-09-15
Fix: Fixing the relative path in test

polemius:
**- Summary**

This PR fix the failed test. The test trying read file by relative path, and it is failed. I've add global variable `__dirname`__dirname` to path.

**- Test plan**

Please run the command:
```
$ yarn test:ava
```
and make sure that the test `src/utils/redirect-parser/line-parser.test.js` not failed.

Fix the test, use absolute path to path in `fs.readFileSync`


raeesbhatti:
Thanks for this PR @polemius   😊

PR #544, opened 2019-09-17
Dev: Add git pre-commit hook to prettifing only changed files

polemius:
**- Summary**

As far in the repository, we have packages `husky` and `lint-staged` I've added the git-hook that run prettier.

**- Test plan**

Change file with non-prettier config (for example add many new lines in some part of a file) and add some another changes. Make commit with these changes and check with `git diff HEAD^` what was added. This many new lines should disappear.

Add git-hook to run prettier automatically

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,6 +160,17 @@
     "prettier": "^1.18.2",
     "strip-ansi": "^5.2.0"
   },
+  "husky": {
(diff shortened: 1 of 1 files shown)


erezrokah:
Closing as stale

Issue #529, opened 2019-09-13
Problems with netlify dev and redirectors with jwt secret


richtera:
Bug:

It's not possible to assign the jwtSecret and in theory it defaults to "secret" but it it doesn't work.

**- Problem:**
passing JWT secrets is not possible.
sending nf_jwt with "secret" as the secret doesn't work.

**- What is the expected behavior?**
It should work like deployment

**- Local Environment Information**
Paste the results of `netlify status --verbose` here
Netlify is not linked for this, it's only a local dev site.

The code currently doesn't pass jwtSecret.
```javascript
  const rewriter = rulesProxy({ publicFolder: settings.dist })
```

```sh
$ node_modules/.bin/netlify status --verbose
──────────────────────┐
 Current Netlify User │
──────────────────────┘
Name:  Andreas Richter
Email: ****
Teams: 
  SelfStudy:              Owner Collaborator Controller
  Andreas Richter's team: Collaborator
 ›   Warning: Did you run `netlify link` yet?
 ›   Error: You don't appear to be in a folder that is linked to a site
```

calavera:
I think we should generate a shared secret that any client can use to generate those tokens. If I remember right, the default path to check roles is `app_metadata.authorization.roles` which means that we expected a payload like this in the JWT:

```
  {
    "app_metadata": {
      "authorization": {
        "roles": ["admin"]
      }
    },
    "exp": 2149017956
  }
```

The proxy in the dev subsystem should read the `nf_jwt` cookie, verify the token with the secret, and then check that the role is in the payload.

richtera:
Right the webassembly is loaded with the default secret "secret" but does not correctly decode and redirect the rules according to roles.

PR #545, opened 2019-09-18
Dev: Add support for JWT redirects

raeesbhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/529

**- Test plan**
Create a static files project
Have some content in file `/admin/something.html`
Have a valid file at `/login.html`
Have a file with following contents at `/_redirects`:
```
/admin/*			200!	Role=admin
/admin/*	/login	401!
```
`netlify dev`
Make an http request for `http://localhost:8888/admin/something` without any Cookie headers and expect it to fail with a 401 and contents of `/login.html` 
Make another http request for `http://localhost:8888/admin/something` with a Cookie header that has `nf_jwt=YOUR_JWT_HERE`. Expect this request to be successful. Your JWT contents similar to following:
```
{
    "app_metadata": {
      "authorization": {
        "roles": ["admin"]
      }
    },
    "exp": 2149017956
  }
```
We currently don't verify the secret for JWT. The object path for roles property (`app_metadata.authorization.roles` in the example above) can be configured by having a `jwtRolePath` value in your `netlify.toml -> [dev]`. Like so:
```
[dev]
  jwtRolePath = "app_metadata.authorization.roles"
```

Stop `netlify dev`.
Remove the second line from `/_redirects`.
Rerun `netilfy dev`
Make an http request for `http://localhost:8888/admin/something` without any Cookie headers and expect it to fail with a 404 page. 

Decode `nf_jwt` Cookie and use it to match roles defined in `redirects` directives.

🦁


raeesbhatti:
@calavera is `jwtRolePath` configurable in production? If not, then I can remove that option here.

DavidWells:
> @calavera is jwtRolePath configurable in production? If not, then I can remove that option here.

I think `jwtRolePath` is hardcoded to `app_metadata.authorization.roles`. Is this correct? @calavera or @futuregerald ?

futuregerald:
Right now we modify it for people when they request it. It should be available via the UI and at some point it will be. I wouldn't remove it. Anyone using OIDC compliant JWT's that use domain namespacing will need to set a custom path, this particularly affects people using Auth0.

raeesbhatti:
Can we get a thumbs up on this please! @DavidWells @calavera

raeesbhatti:
Going to merge this, let me know if you have any objections and we can revert.

PR #556, opened 2019-09-24
Fix dependencies check

ehmicky:
**- Summary**

We are checking our dependencies two different ways:
  - with `dependency-check`, run by `npm run test:deps`
  - with `eslint-plugin-node`

In my opinion `eslint-plugin-node` does a much better job than `dependency-check` at doing this, so `dependency-check` is redundant and can be removed. Namely:
  - it is integrated with ESLint: integrated with IDE, some rules can be autofixed, etc.
  - it has more settings for custom resolution, nested `package.json`, etc.
  - it has [more rules](https://github.com/mysticatea/eslint-plugin-node#possible-errors) and does more in terms of the check itself

Also note that `npm run test:deps` was currently failing, which was showing up in our CI status.

Fix dependencies check


DavidWells:
removed via https://github.com/netlify/cli/pull/571

ehmicky:
I think you forgot to remove `dependency-check` from the `devDependencies`.

Issue #566, opened 2019-10-09
Serve functions programmatically


pimterry:
It'd be really useful to be able to serve functions programmatically. In my case, I'd like to do this for automated testing: I want to run all my functions with a real server locally, and write tests that make requests to them and check the results.

In theory you could do so by spawning the CLI as a subprocess and automating it through there, but that's messy and requires a separate node process, so you can't use things like [rewire](https://www.npmjs.com/package/rewire) to stub out dependencies.

All the code is here for this already, this would just require a little reorganization. Right now, the best I can see is:

```javascript
import { serveFunctions } from 'netlify-cli/src/utils/serve-functions';

describe("My functions", () => {
  it("should handle requests", () => {
    serveFunctions({
        functionsDir: path.join(__dirname, '..', 'functions'),
        port: 9090
    });
    
    const response = await fetch('http://localhost:9090/netlify/.functions/paddle-webhook');
    expect(response.status).to.equal(200);
  });
});
```

This has a couple of problems though:

* You can't cleanly shut the server down.
* There's no easy way to tell when the server has started: you need to poll `/` until you get a 404 response.
* You're importing an internal undocumented API, which could presumably change any time

Any chance you'd be open to fixing these issues?

sw-yx:
what’s the api you want? 

all this code is internals, yes, but theres nothing really stopping you copying out the code and writing your own tool if u wanna do that. we can even link to you if you think this would help others

pimterry:
> what’s the api you want?

Something like this:

```javascript
import { Server } from 'netlify-cli';

describe("My functions", () => {
  
  const server = new Server();

  beforeEach(async () => {
    await server.start();
  });

  afterEach(async () => {
    await server.stop();
  });

  it("should handle requests", () => {    
    const response = await fetch(`http://localhost:${server.port}/netlify/.functions/paddle-webhook`'`);
    expect(response.status).to.equal(200);
  });
});
```

I.e. take `netlify dev` and separate the CLI part (args parsing etc) from actually running the server, export the server, and let me use it directly. In this perfect world this doesn't need the `functionsDir` arg of the previous example either, as it just reads from my `netlify.toml`, and does exactly the same thing the production environment would do.

> all this code is internals, yes, but theres nothing really stopping you copying out the code and writing your own tool

True, but it'd be nice to have this built-in & supported by Netlify, since you've already written the code for 99% of it, and duplicating it myself means maintaining it as that changes.

[...]

PR #568, opened 2019-10-10
Make it easier to use serve-functions programmatically

pimterry:
**- Summary**

See #566.

This change takes away a few of the immediate pain points for programmatic usage, e.g. building automated tests using netlify CLI to serve your functions.

More specifically, this change:
- Means that the serveFunctions promise doesn't resolve until the server has fully started up.
- Returns the newly started server, so it can be cleanly shutdown later on.
- Adds a 'quiet' option to serveFunctions to disable express-logging

There are other more dramatic ways to achieve the same goals (like a first-class `require('netlify-cli').startServer()` kind of API), but this is a very small easy step that is pretty much just as good imo.

**- Test plan**

I don't have a clear test plan for this, but I've got working automated tests locally using this to integration test my netlify functions for https://github.com/httptoolkit/accounts, where it's working very nicely.

* Make it easier to use `serve-functions` programmatically

Cute hedgehog

### DIFF
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -55,7 +55,7 @@ function buildClientContext(headers) {
   }
 }
 
-function createHandler(dir) {
+function createHandler(dir, watch) {
   const functions = getFunctions(dir)
 
   const clearCache = action => path => {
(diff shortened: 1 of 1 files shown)


sw-yx:
nice! up to @raeesiqbal to consider

pimterry:
One more tweak that I missed: there's now a `watch` option you can pass to serveFunctions, defaulting to true. This lets you disable chokidar watching the test files, because otherwise the test suite won't exit when it's fully completed.

pimterry:
Any thoughts on this @raeesbhatti? I've got all my tests working nicely now, but only locally, because I can't run them in CI until this is published.

pimterry:
Ok I've just rebased this, and for now I've forked netlify-cli entirely and republished it to npm as [@httptoolkit/netlify-cli](https://www.npmjs.com/package/@httptoolkit/netlify-cli).

That unblocks my current work, but it'd be great to get this merged in upstream too!

sw-yx:
agreed.

erezrokah:
Closing as stale.
We could re-visit if needed, but one could also use the current approach the CLI uses for its own tests.
See https://github.com/netlify/cli/blob/6410f72e5e29f7b2bd8d58e8fef002d60557abd3/tests/dev.test.js#L47

Issue #569, opened 2019-10-11
[netlify dev] Redirect not working for functions


screendriver:
**- Do you want to request a *feature* or report a *bug*?**
Bug

**- What is the current behavior?**

I have following section in my `netlify.toml`

```toml
[build]
  publish = "public"
  environment = { AWS_LAMBDA_JS_RUNTIME = "nodejs10.x" }
  functions = "src/functions/"

[[redirects]]
  from = "/api/*"
  to = "/.netlify/functions/:splat"
  status = 200
```

When I start `netlify dev` and open http://localhost:34567/api/graphql for example it shows me a `Function not found...`. But when I deploy this into production on Netlify it works. And it still works when I open http://localhost:34567/.netlify/functions/graphql So the redirect (rewrite) is not working locally.

**- If the current behavior is a bug, please provide the steps to reproduce.**

Just create a `netlify.toml` as described above and a function called `src/functions/graphql.js`

**- What is the expected behavior?**

`netlify dev` should behave as in production.

**- Local Environment Information**
` ›   Error: You don't appear to be in a folder that is linked to a site`
Because I deploy my site from GitHub actions. I am not logged in locally.

coopsmoss:
I am experiencing the same issue but I am using the _redirects method. Lost the better half of a day to this issue. Is there a way to downgrade the CLI in the mean time?

robbishop:
I have had a similar issue. When netlify dev server redirects a function call the last query string parameter has `/index.html` appended to it. If you use the `.netlify/functions` endpoint it works fine, as does both variants when deployed to Netlify 

### /.netlify/functions/test?foo=bar
`event: { queryStringParameters: {"foo":"bar"} }`
### /api/test?foo=bar
`event: { queryStringParameters: {"foo":"bar/index.htm"} }`

jlengstorf:
I'm also hitting this. Redirects don't appear to get picked up from `netlify.toml` or `_redirects`. Repro here: https://github.com/jlengstorf/jamstack-intro/tree/bug/dev

jlengstorf:
More testing shows that redirects _do_ work for e.g. SPA setups. 

```toml
[[redirects]]
  from = '/api/*'
  to = '/.netlify/functions/:splat'
  status = 200

[[redirects]]
  from = '/test/*'
  to = '/test'
  status = 200
```

[...]

PR #572, opened 2019-10-16
fix: remove naming conflict

jlengstorf:
There was a const being declared with the same name as a function in the outer scope. I _think_ this could be responsible for #569, but in any case it’s probably worth renaming to avoid confusion/naming collisions.

**- Summary**

**- Test plan**


raeesbhatti:
@jlengstorf Thanks for the PR. I've found the issue for your project, it's a different one but this was also a bug. Thanks 🎉

PR #576, opened 2019-10-18
[WIP] Refactor all JS function templates to use async/await syntax

### LINKED ISSUE #522: refactor all js function templates to use async/await syntax
bug

some of our [function templates](https://github.com/netlify/cli/tree/master/src/functions-templates/js) use the legacy callback syntax (`(event, context, callback)`) for AWS Lambda.  ([example](https://github.com/netlify/cli/blob/a0160b1b98352914b3818af891c5f8ae15cf4db6/src/functions-templates/js/slack-rate-limit/slack-rate-limit.js))

this isn't something we should encourage as it leads to more complicated code. We should refactor to more modern async/await syntax (`async (event, context)`). ([example](https://github.com/netlify/cli/blob/a0160b1b98352914b3818af891c5f8ae15cf4db6/src/functions-templates/js/stripe-subscription/stripe-subscription.js))

arpitbatra123:
Files covered:
1. `src/functions-templates/js/slack-rate-limit/slack-rate-limit.js`
Will fix #522.

### DIFF
--- a/src/functions-templates/js/slack-rate-limit/slack-rate-limit.js
+++ b/src/functions-templates/js/slack-rate-limit/slack-rate-limit.js
@@ -38,7 +38,7 @@ class IdentityAPI {
 
       if (!response.ok) {
         return response.text().then(data => {
-          return Promise.reject({ stauts: response.status, data })
+          return Promise.reject({ status: response.status, data })
         })
       }
       return response.text().then(data => {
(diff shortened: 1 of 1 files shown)


sw-yx:
nice! still WIP?

arpitbatra123:
I'm thinking of keeping this a WIP until all functions are refactored to the new syntax.

PR #578, opened 2019-10-19
Typo in error message

spences10:
image


DavidWells:
Thanks for the fix!

Speeling is hard! 😅

PR #582, opened 2019-10-22
Allow overriding the api target host

mraerino:
**- Summary**

When working on the Netlify API I want to be able to work against a local or staging environment using the CLI.
This PR will introduce the ability to set an environment variable called `NETLIFY_API_URL` to set the API target host.
If it has no path appended (not even `/`) the standard suffix will be appended.

**- Test plan**

Tested on my local environment with login and status commands.

n/a (internal change)


raeesbhatti:
Just published this to NPM as well.

PR #587, opened 2019-10-24
Only build once on startup and add a small debounce

n4bb12:
**- Summary**

I encountered two problems with the functions builder:

- The functions build runs multiple times on startup.
- When multiple files are saved at once, each file triggers an individual build.

The result is that startup and rebuilds take unnessessarily long and can heavily spike the CPU.

Why this happens:

- The `chokidar` library by default emits an `add` event on each matching file when initializing the watcher. See the [ignoreInitial](https://github.com/paulmillr/chokidar#path-filtering) option. Thus, on startup the build runs `1 + numberOfSourceFiles` times.
- The build is not debounced, so multiple changes in quick succession cause the builds to stack up.

Solution:

- Set `ignoreInitial` to true when initializing the chokidar watcher.
- Add a small debounce time of 100ms between change events and triggering builds. It seems like a good time to me, but in case someone has different needs, I added and documented an option.

**- Test plan**

I `npm link`-ed the cli into one of my projects and verified the startup and change event build is triggered as intended.

There seems to be no tests for the functions builder right now, so adding/updating one is a bit of a major task.

I ran the existing tests and fixed a path issue for an existing test on Windows.

- On startup, the functions builder now only runs once. Enjoy the speed! 🌬
- There is a 100ms debounce time between functions builds on file changes. This can be configured via `functionsBuildDebounceMs`.

🦝


n4bb12:
Perhaps the animal was not cute enough. Let me try again.

erezrokah:
Not sure how we missed this and had to implement a fix recently https://github.com/netlify/cli/pull/974.
Very sorry @n4bb12!

PR #588, opened 2019-10-25
Bump `@netlify/build` version

ehmicky:
This fixes some issues with `zip-it-and-ship-it`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -439,11 +439,11 @@
       }
     },
     "@netlify/build": {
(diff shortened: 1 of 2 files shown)


ehmicky:
Yes you're right, sorry!

Basically there were some bugs in `zip-it-and-ship-it` that made it crash :slightly_smiling_face:
Also `@netlify/build` was not using the right version of `@netlify/config`, which also made it crash.

PR #592, opened 2019-10-29
Fix semver issue with `@netlify/build`

ehmicky:
**- Summary**

We use use the `^` semver range on `@netlify/build`. However currently `@netlify/build` version starts with `0.0.*`. In such cases, every version increment is [considered a breaking change](https://github.com/npm/node-semver#caret-ranges-123-025-004) according to semver, which makes `^` not pick any new versions.

I have released `@netlify/build` `0.1.0` which since it starts with `0.*.*` instead of `0.0.*` should fix this issue.

Upgrade `@netlify/build`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -439,9 +439,9 @@
       }
     },
     "@netlify/build": {
-      "version": "0.0.34",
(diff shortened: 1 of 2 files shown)


ehmicky:
Closed in favor of #593.

PR #594, opened 2019-10-29
Fix missing dependency `@oclif/dev-cli`

ehmicky:
**- Summary**

The dependency `@oclif/dev-cli` is missing. It was installed as part of `@netlify/cli-utils` which was removed by #593.

This PR fixes this. I've checked other dependencies of `@netlify/cli-utils` and this is the only missing one.

Fix missing dependency `@oclif/dev-cli`


ehmicky:
Never mind, it was installed as a `devDependency` :) All is good.

PR #602, opened 2019-11-06
show useful error if targetPort is unknown

erquhart:
**- Summary**

Was getting this error from Netlify Dev, even though **port** was set in my config:

```
ValidationError: 'port' is required.
    at validateParameters (~/.config/yarn/global/node_modules/wait-port/lib/validate-parameters.js:13:20)
    at Promise (~/.config/yarn/global/node_modules/wait-port/lib/wait-port.js:183:9)
    at new Promise (<anonymous>)
    at waitPort (~/.config/yarn/global/node_modules/wait-port/lib/wait-port.js:173:10)
    at startProxy (~/.config/yarn/global/node_modules/netlify-cli/src/commands/dev/index.js:126:9)
    at DevCommand.run (~/.config/yarn/global/node_modules/netlify-cli/src/commands/dev/index.js:406:31)
    at <anonymous>
```

This fix shows this instead (and still exits):

```
◈ Netlify Dev doesn't know what port your site is running on.
◈ Please set --targetPort.
```

It's actually referring to `targetPort` - I'm not using an SSG, so the port couldn't be inferred. In this case, `targetPort` becomes a required config property. Documentation should be updated, but that seemed non-trivial to do with existing verbiage stating all config properties are optional and the API doc being auto generated. This at least informs the developer directly when the error occurs.

**- Test plan**

Tested locally, displays the message and exits only when **targetPort** is unknown.

show useful error if targetPort is unknown


raeesbhatti:
Thanks for the PR @erquhart

PR #603, opened 2019-11-07
Mock identity response with properly-shaped data

tctrautman:
**- Summary**

This PR adds mocked response data that resembles the shape of `body.user` for identity functions in production, which I'd find helpful during development.

**- Test plan**

There are no tests since it's just a few lines.


erquhart:
Thanks for this @tctrautman!

PR #616, opened 2019-11-18
Update Config parsing to work with netlify build

DavidWells:
Update Config parsing to work with netlify build.


DavidWells:
There are some minor differences in config shape. Mainly `command` is gone. I've searched for references to this and came up empty.

The other difference is `build.publish` is a fully resolved path. This appears to work everywhere it's referenced

Before:

```
{
  build: {
    command: 'npm install && cd site && npm install && npm run build',
    publish: 'site/dist'
  },
  dev: {
    command: 'yarn start'
  }
}
```

After

```
{
  build: {
    publish: '/Users/davidwells/Netlify/repos/cli/site/dist',
    lifecycle: {
      build: [Array]
    }
  },
  plugins: [],
  dev: {
    command: 'yarn start'
  },
}
```

src/utils/command.js
```diff
@@ -25,16 +25,28 @@ class BaseCommand extends Command {
...
+    let config = {}
+    try {
+      configPath = await getConfigPath(argv.config, cwd)
+      config = await resolveConfig(configPath, {
+        cwd: cwd,
+        context: argv.context
+      })
+    } catch (err) {
+      // Suppress config not found error for CLI. @TODO Revisit
+      if (err.message.indexOf('No netlify configuration file was found') === -1) {
```
raeesbhatti: Why suppress this error?

DavidWells: Not every site needs netlify config file. They might have build settings set in the UI

The @netlify/build config package was throwing hard errors when config was missing.

DavidWells: It might be updated in build? @ehmicky did you update how config handles this?

raeesbhatti: I see.

DavidWells: ref: https://github.com/netlify/build/issues/497#issuecomment-557100286

ehmicky: Yes this `try`/`catch` should not be needed anymore. Missing configuration files do not throw anymore.

PR #630, opened 2019-12-05
remove "did you mean" feature

DavidWells:
Dev rel team said this doesn't work


XhmikosR:
Are you sure you didn't miss updating the lock file? BTW that's why CI should use `npm ci`

erquhart:
This is working - if there's a bug or proposal to remove, let's open an issue first.

Issue #615, opened 2019-11-17
Make automatic open of the browser optional


janbaer:
Could it be possible to make it possible that the **netlify dev** command opens the browser automatically? My browser is constantly opened and I don't need this feature. It's also some annoying for me because the browser that I use is not the default browser. It's a special nightly build of the Chromium browser. So whenever I run the **netlify dev** command a second instance of my browser will be startet. Could you please make this *feature* optional? Thanks in advance!

frednomoon:
massively in favour of this! very annoying at the moment

therobinkim:
One thing I've noticed is that if I close the browser tab/window too soon after it opens up, I immediately get a `socket hang up` error and `netlify dev` crashes.

raeesbhatti:
I've [created a PR](https://github.com/netlify/cli/pull/631) to allow the users to configure this behavior. I'll try to get this merged and released today. But in the mean time you can do a local `npm link` of `raees/autoLaunchConfig` branch to use the changes.

raeesbhatti:
A new config parameter was added in the latest release of 2.23.0 where you can specify `autoLaunch` in your `netlify.toml` to disable the auto-launch of browser when `netlify dev` starts. e.g.
```
[dev]
  autoLaunch = false
```

PR #631, opened 2019-12-05
Allow configuration of auto launch feature

raeesbhatti:
**- Summary**

Some users have a requested a way to disable the feature in Netlify Dev that automatically launches the local dev server in browser. This change implements a way to disable that feature from `netlify.toml`. Fixes https://github.com/netlify/cli/issues/615

**- Test plan**

Add the following to your `netlify.toml`
```
[dev]
  autoLaunch = false
```
Run `netlify dev`

Check for `config.dev.autoLaunch` to determine if we should launch the browser automatically or not.

🐥

### DIFF
--- a/docs/netlify-dev.md
+++ b/docs/netlify-dev.md
@@ -117,6 +117,7 @@ Netlify Dev is meant to work with zero config for the majority of users, by usin
   targetPort = 3000 # Port of target app server
   publish = "dist" # If you use a _redirect file, provide the path to your static content folder
   jwtRolePath = "app_metadata.authorization.roles" # Object path we should look for role values for JWT based redirects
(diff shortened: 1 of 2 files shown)


ianmartorell:
With this change, the default behaviour changes to **not** auto launching the browser. This is an unexpected change and might break some development environments.

For instance, I just opened #672 to auto detect Expo projects, but I need the browser to be auto launched because the Expo cli clears the terminal when running, making the `netlify-dev` URL disappear, and we can't rely on the user to add the `autoLaunch = true` flag to their `netlify.toml`. Ideally, auto-detected servers should work out of the box.

So the way I see it, one of these should be done before the next release:

1. Revert default behaviour to `autoLaunch = true`
2. Check `autoLaunch` in the auto-detected server settings too

Ideally it would be the first one, to avoid changing default behaviour. I'm happy to open the pull request.

PR #669, opened 2019-12-29
Remove duplicate port allocation - fixes #658

### LINKED ISSUE #658: lambda listening on wrong port
**- Do you want to request a _feature_ or report a _bug_?**
report a bug

run netlify dev, notice the port it mentions is being listened to for lambda.
At the end it says "Waiting for localhost:34567" and hangs there. 

add to netlify.toml
[dev]
  autoLaunch = false
  functionsPort = 34567

* run netlify dev, notice the port it mentions is being listened to for lambda.
* I see something like: Lambda server is listening on 51976, but that port seems to change
* let the dev server (quasar in my case) finish building
* note the message: "Waiting for localhost:34567"

lambda and the last localhost to use the same port and use the one specified for functionsPort if it is specified.

Couple notes:
* this is on a clean install of OSX Catalina.
* netlify-cli is installed with yarn global. yarn global list shows: "netlify-cli@2.25.0"
* node v12.14.0 installed via asdf-vm

I had another setup on High Sierra that did not have this issue. Unsure if its related, but that setup used npm -g for the install of netlify-cli.

130n:
**- Summary**

Based on #658
When running `netlify dev` neither default functionsPort nor port passed in configuration is used.

**- Test plan**

Add to netlify.toml
```
[dev]
  functionsPort = 30303
```
and verify the port is used

Remove duplicate call to getPort for the same port (default: 34567)

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -408,7 +408,7 @@ class DevCommand extends Command {
           functionWatcher.on('change', functionBuilder.build)
           functionWatcher.on('unlink', functionBuilder.build)
         }
-        const functionsPort = await getPort({ port: settings.functionsPort || 34567 })
+        const functionsPort = settings.functionsPort || 34567
 
         // returns a value but we dont use it
         await serveFunctions({
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -408,7 +408,7 @@ class DevCommand extends Command {
           functionWatcher.on('change', functionBuilder.build)
           functionWatcher.on('unlink', functionBuilder.build)
         }
-        const functionsPort = settings.functionsPort || 34567
+        const functionsPort = settings.functionsPort
 
         // returns a value but we dont use it
         await serveFunctions({


src/commands/dev/index.js
```diff
@@ -408,7 +408,7 @@ class DevCommand extends Command {
           functionWatcher.on('change', functionBuilder.build)
           functionWatcher.on('unlink', functionBuilder.build)
         }
-        const functionsPort = await getPort({ port: settings.functionsPort || 34567 })
+        const functionsPort = settings.functionsPort || 34567
```
ianmartorell:
```suggestion
        const functionsPort = settings.functionsPort
```
Probably shouldn't hardcode the default here since it's already in `serve-functions.js`:
https://github.com/netlify/cli/blob/6fe0d2aa49197148804b5292e01fed09621c55f9/src/utils/serve-functions.js#L15
https://github.com/netlify/cli/blob/6fe0d2aa49197148804b5292e01fed09621c55f9/src/utils/serve-functions.js#L194
Otherwise `defaultPort` would never be used.

ianmartorell:
Any idea of when we could get this merged? Without it `netlify dev` literally does not work at all in most cases.

130n:
@RaeesBhatti @ianmartorell It should be just a click away!

RaeesBhatti:
@130n I'm really sorry. We're just getting back from holidays.

ianmartorell:
Figured that was the case after commenting 😳 no worries and thank you!

PR #672, opened 2020-01-07
Add detector for Expo

### LINKED ISSUE #333: Support for Expo (React Native for web) apps

# Feature Request

Now with first-class web support coming to [React Native](https://blog.expo.io/expo-cli-and-sdk-web-support-beta-d0c588221375) it would be great to have feature detection built into the deployment command. 

* Run `expo build:web` to generate a static site
* `netlify deploy` starts the flow
* choose a name
* ? **deploy path** ( . ) - `/web-build`

This last step doesn't seem to persist my choice. It'd be cool if it checked to see if the `app.json` (React Native config file) exists in the root directory, then possibly checks to see if `expo.web.build.output` is defined (custom output path). 

So if `/web-build` exists and `app.json` exists you would get the following:
* ? **deploy path** (web-build)

and if `expo.web.build.output` is defined you would get that value as the auto suggested path.

# How

I'm happy to work on this PR if you guys are interested.

ianmartorell:
**- Summary**
This PR adds a detector for `netlify-cli` to work automagically with Expo projects.

Prior to this, every project needed to be configured using `netlify.toml` before `netlify dev` could be used.

This also fixes #333 since `netlify publish` now uses the detector config to find the publish directory.

**- Test plan**
I created a new blank Expo project, ran the `dev` command on that folder and it started up successfully. Then I ran the `publish` command and it didn't ask for the build directory.

*Note*: I also needed to use #669 or the port allocation of the functions server becomes random and `netlify dev` can't connect.

Add detector for Expo

7241gucrk6941

### DIFF
--- a/src/detectors/expo.js
+++ b/src/detectors/expo.js
@@ -8,16 +8,17 @@ module.exports = function() {
   /** everything below now assumes that we are within expo */
 
   const possibleArgsArrs = scanScripts({
-    // prefer the `start:web` script instead of expo's default `web` since that runs
-    // `expo start --web`, which auto launches the (non-proxied) web browser.
-    preferredScriptsArr: ['start:web'],
-    // `expo start:web` does not auto launch the web browser
-    preferredCommand: 'expo start:web'
+    // This script will run `expo start --web` in a new Expo project.
+    // Note: Expo will automatically launch the browser with your app's
+    // Webpack server listening on port 19006, but the instance proxied
+    // by `netlify dev` will be running on port 8888.
+    preferredScriptsArr: ['web'],
+    preferredCommand: 'expo start --web'
   })
 
   if (possibleArgsArrs.length === 0) {
     // ofer to run it when the user doesnt have any scripts setup! 🤯
-    possibleArgsArrs.push(['expo', 'start:web'])
+    possibleArgsArrs.push(['expo', 'start', '--web'])
   }
   return {
     type: 'expo',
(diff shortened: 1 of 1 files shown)


ianmartorell:
Ideally, there should be a way to inform the user of the URL the proxied server is listening on, because currently the Expo cli clears the terminal and the URL output by `netlify-dev` disappears.

RaeesBhatti:
I tried this locally but the `proxyPort` is `19006` for some reason but the expo server is running on `19002`. I'm not familiar with Expo, can you please explain the different ports and which one is correct.

ianmartorell:
Sure thing! Expo actually starts up a bunch of things on different ports:

- Expo Development Server runs on port `19000` - it's basically the native app's entrypoint
- The React Native packager runs on port `19001` - serves JS and assets
- Expo DevTools runs on port `19002` - I guess this is what you saw?
- The Webpack dev server runs on port `19006` - this is the one we care about

We want to proxy the Webpack server so that's port `19006`. How did you go about trying it locally to see the port `19002`?

---

Something I just realized though, is that when you start an Expo project with `expo start:web` it will only fire up the DevTools and the Webpack server. As opposed to starting it with `expo start --web` which will fire up Expo Dev server and the Packager (only needed if you want to run the app on Android or iOS) and then start the Webpack server and open the project in the browser. 

So the reason I'm inclined to use `expo start:web` is that `expo start --web` automatically opens the Webpack dev server in the browser, which we don't want, and there's no way to configure that on the Expo side as far as I know. However, it seems using `expo start:web` locks you into not being able to start the React Native side of things.

So I guess the best course of action is to just use Expo defaults and add a way to disable opening the browser automatically in `expo-cli`.

RaeesBhatti:
Thanks for the PR @ianmartorell.

sw-yx:
this is great! thanks ian!

PR #673, opened 2020-01-07
Revert default autoLaunch behavior to true

ianmartorell:
**- Summary**
This PR reverts the default `autoLaunch` to `true`, the way it works in the current release of `netlify-cli`. Motivation is explained in https://github.com/netlify/cli/pull/631#issuecomment-571555799.

**- Test plan**
The change is really basic so not much testing is needed, but I tested it with a new basic project with and without a `netlify.toml`.

Probably shouldn't be included in the changelog since it reverts a change present in master but not yet released.

ezgif com-webp-to-png (1)

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -438,7 +438,11 @@ class DevCommand extends Command {
         }
       })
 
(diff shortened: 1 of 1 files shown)


RaeesBhatti:
Thanks for the PR @ianmartorell

Issue #659, opened 2019-12-20
localhost:8888 not showing up after update


hyliancoder:
Ever since updating to 2.25.0 I cant get a server to run.

it no longer shows the 

`   ┌──────────────────────────────────────────────────────────────┐`
  ` │                                                              │`
   `│   [Netlify Dev] Server now ready on http://localhost:8888    │`
   `│                                                              │`
  ` └──────────────────────────────────────────────────────────────┘`

box and shows endless dots and never serves the files on a new port where the functions are working.

`Netlify Dev ◈
◈ Starting Netlify Dev with create-react-app
◈ Function builder netlify-lambda detected: Running npm script build:lambda
 ›   Warning: ◈ This is a beta feature, please give us feedback on how to improve at https://github.com/netlify/cli/
yarn run v1.21.1
$ react-scripts start
Starting the development server...

Waiting for localhost:3000.
◈ Lambda server is listening on 57397

Connected!
Waiting for localhost:34567...◈ lambda/open-graph-preview.js modified, reloading...
..................................................................................................Compiling...
.Compiled with warnings...................................................................................................` 

those warnings being linter warnings that show up when running the normal `yarn start` as well so nothing blocking it from serving. 

was working fine on 2.21.0. but now even when i roll back to 2.21.0 it still no longer works.

reesecode:
after looking at the other issue here: https://github.com/netlify/cli/issues/657

thankfully rolling back to `2.11.13` got it working again, but for sure a pain that I cant update.

michalhudecek:
Same issue here

narrative1st:
Same issue. Rolled back to 2.11.13, endless dots remain...

abettermap:
I think this is the same problem as #657.

narrative1st:
Yes, and I was able to get it back working with a combination of rewriting some of Nuxt's default build scripts, and rolling back to `2.12.0` for `netlify-cli`.

reesecode:
> I think this is the same problem as #657.

for sure, just unfortunately outside of being an issue with gatsby it seems. as my project frontend is just a CRA with apollo

lazno:
same problem here with CRA. tried out versions 2.11.13, 2.12.0 and current 2.25.0 - to no avail

SteveALee:
Arrg - I also hit this during my first experience of netlify-cli and functions using latest gatsby. I spent all the time I had to explore this try to debug / fix

[...]

Issue #634, opened 2019-12-06
Netlify Dev fails using v2.22.0, v2.23.0


talves:
There is an issue running [this repository](https://github.com/ChristopherBiscardi/party-corgi-discord-automations) with `netlify dev` on version > v2.21.0 (2.21.0 is working)

![image](https://user-images.githubusercontent.com/784848/70350531-c1a1a000-181b-11ea-8c92-1436b859151c.png)

**- What is the current behavior?**
If using `v2.21.0` of `netlify-cli`, the server will start without issue.

![image](https://user-images.githubusercontent.com/784848/70352241-8903c580-181f-11ea-9241-01d5c22621b0.png)

When running `v2.22.0` or greater, the server errors out and does not wait.

**- If the current behavior is a bug, please provide the steps to reproduce.**

Clone the repository above at [this head](https://github.com/ChristopherBiscardi/party-corgi-discord-automations/tree/43d449440175c332fe3a708fb3a904fe8a6552e2)
Use `v2.22.0` or `v2.23.0` of `netlify-cli` to run `netlify dev`

**- What is the expected behavior?**

When running `v2.22.0` or greater, the server should wait.

**- Local Environment Information**

This was confirmed on windows and mac.

raeesbhatti:
Thanks for reporting this @talves . Please try the latest release and let us know if the issue persists.

ryancharris:
I am also experiencing this issue 👍

raeesbhatti:
@ryancharris can you please upgrade to the latest version and report if you're still having this problem.

talves:
Yes, it still has that error. If you checkout the repository I linked above, you can check it also.

futuregerald:
Forgive me if I re-opened this in error but it seems like it's not fixed for @talves . Are you still having the same problem?

talves:
Thanks @futuregerald , yes, this is still an issue and having the same issue. Just tested with v2.25.0 (failed).

RaeesBhatti:
@talves Is this still an issue with latest version?
cc @ryancharris

talves:
Just updated to version 2.40.0 and ran on Windows using Powershell and then using WSL (Ubuntu) and all is working as expected calling the functions. 🎉😃

I will close this and @ryancharris can request a re-open if it does not work on Mac.

Thanks for the hard work on this everyone!

talves:
[...]

PR #674, opened 2020-01-08
Fix functions server port assignment

apuyou:
**- Summary**

This PR assigns a port for the functions server in the `dev` command, sets it in the config and then uses it when starting the functions server. Previously, duplicate logic could result in the functions server starting on a specific port and then the CLI waiting for it on another port. This would prevent startup of the whole dev environment (_endless dots_).

The port is elected as follow:
- `functionsPort` option in `netlify.toml` if provided
- `34567` otherwise

If the elected port is currently in use, a random port is assigned.

This fixes the following issues: #659 #634 

**- Test plan**

The test must be run in a context where the functions server is started. This requires a `.functions` directory with appropriate contents.

The code was tested with the following configuration:
- no port specified in `netlify.toml`
- no port specified in `netlify.toml` but port `34567` is busy
- port `11111` specified in `netlify.toml`
- port `11111` specified in `netlify.toml` but port `11111` is busy

Fix function servers port assignment issue preventing `netlify dev` startup

Puppy

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -409,15 +409,13 @@ class DevCommand extends Command {
           functionWatcher.on('change', functionBuilder.build)
           functionWatcher.on('unlink', functionBuilder.build)
         }
-        const functionsPort = settings.functionsPort
+        const functionsPort = await getPort({ port: settings.functionsPort || 34567 })
+        settings.functionsPort = functionsPort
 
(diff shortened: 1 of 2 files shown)


RaeesBhatti:
Thanks for the PR @apuyou

mrlubos:
Nice one @apuyou, thanks!

PR #679, opened 2020-01-09
Handle zombie processes differently

RaeesBhatti:
**- Summary**
The new changes use a different approach where we don't try to kill the processes and instead log the error and let the user terminate the `netlify dev` process which cleanly kills all of the child processes. The previous approach did not fix all of the issues and there were still some edge cases. This can serve as a good patch-fix until we can dedicate some time to solve this problem properly.

**- Test plan**

* `netlify dev` in your existing project

Dont try to kill the child processes, log the error and let the use exit.

# 🐶

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6862,26 +6862,6 @@
         }
       }
     },
-    "find-process": {
-      "version": "1.4.3",
-      "resolved": "https://registry.npmjs.org/find-process/-/find-process-1.4.3.tgz",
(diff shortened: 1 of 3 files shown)


ehmicky:
I like that it removes lots of lines of code :)

RaeesBhatti:
@ehmicky Here is the latest change in the zombie process series: https://github.com/netlify/cli/commit/64d6e67a4eb0c39f3dc49e0061a2f3cedb012f32

PR #681, opened 2020-01-12
Update path to detectors folder

locks:
**- Summary**

I was reading the documentation for `netlify dev` and noticed that the detectors link goes to the archived netlify-dev-plugin.

**- Test plan**

Does not apply.

Documentation now links to current folder of detectors.

image


RaeesBhatti:
Thanks for the PR @locks

PR #690, opened 2020-01-25
addition of two new flags --userId and email to identify the account …

cizza:
**- Summary**

The addition of these two flags will allow the user to switch between netlify accounts based on either an email address or userId. The reason for these flags to exists is to allow users who do not or cannot interact with terminal directly to be able to switch users programatically. 

UserId flag is direct comparison with userId contained in the availableUsersChoices object. 
However the email address is not a direct comparison, but rather an "includes" check as the email property in availableUsersChoices is not always just email address but can contain additional data (e.g. name).

**- Test plan**
Functionality tested manually on my netlify accounts, including testing of existing "switch" functionality.

Two new available flags on switch command - switch by userId or email.

### DIFF
--- a/src/commands/switch.js
+++ b/src/commands/switch.js
@@ -1,15 +1,20 @@
 const Command = require('../utils/command')
+const { flags } = require('@oclif/command')
 const chalk = require('chalk')
 const inquirer = require('inquirer')
 const LoginCommand = require('./login')
 
 class SwitchCommand extends Command {
   async run() {
+
+    let selectedAccount;
+    const { flags } = this.parse(SwitchCommand)
+
     const LOGIN_NEW = 'I would like to login to a new account'
     const availableUsersChoices = Object.values(this.netlify.globalConfig.get('users')).reduce(
-      (prev, current) =>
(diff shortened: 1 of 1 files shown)


src/commands/switch.js
```diff
@@ -19,6 +24,21 @@ class SwitchCommand extends Command {
       }
     })
 
+    if(Object.keys(flags).length) {
```
erquhart: This logic should be activated only if the `userId` or `email` flags are provided, unrecognized flags should be ignored, allowing prompts to continue.

src/commands/switch.js
```diff
@@ -19,6 +24,21 @@ class SwitchCommand extends Command {
...
+
+      selectedAccount = Object.entries(availableUsersChoices).find(([k, v]) => Object.keys(flags)[0] === 'userId' ?
+          k === flags[Object.keys(flags)[0]] : v.includes(flags[Object.keys(flags)[0]]))
```
erquhart: This logic should be broken down so the imperative check for 'email' is explicit.

erezrokah:
Is this still relevant/needed? We can rebase and follow up on the comments if needed

erezrokah:
Closing as stale. Please comment if still relevant

PR #693, opened 2020-02-03
asking for a bit more info and clarifying how to search the issues list

fool:
Is a change to the issue filing template, so no changes to functionality of code/tests should apply


RaeesBhatti:
Thanks for the PR @fool

PR #699, opened 2020-02-07
[dev] Add support for injecting shared environment variables

### LINKED ISSUE #431: Can't access Shared environment variables from `netlify dev`
Report a bug. 

The `netlify dev` command correctly pulls in env vars for the linked project but doesn't pull in team wide env vars from the Shared Environment Variables section of Team Settings. 

Link a project and call `netlify dev`

See the same env vars when calling `netlify dev` as when building remotely in Netlify.

Eddie-CooRo:
Fixes #431

### DIFF
--- a/src/utils/dev.js
+++ b/src/utils/dev.js
@@ -50,27 +50,31 @@ async function addEnvVariables(api, site, accessToken) {
   }
 
   /** from web UI */
-  const apiSite = await api.getSite({ site_id: site.id }).catch(error => {
-    console.error(error)
-    switch (error.status) {
-      case 401:
-        console.error(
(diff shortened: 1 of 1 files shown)


erquhart:
Thanks for opening this @Eddie-CooRo!

@RaeesBhatti the approach here seems reasonable - I'll defer you on review in case there's a non-obvious issue.

Eddie-CooRo:
Hey @erquhart thanks for your reply
@RaeesBhatti have you seen any issue in this PR? We upgraded our account to pro just because of shared env variables, and netlify dev not supporting it is a little troublesome for us. because of this, I have to use a custom build of this package for our team, and using the awesome features you guys are working on, needs some rebasing and dirty works.
It would be great if we could have this feature in the official package.
thanks in forward.

PR #702, opened 2020-02-07
update codeowners

erquhart:
Switch to new netlify-dev team.


erquhart:
@verythorough good catch!

Issue #474, opened 2019-03-29
[DX] Read .env files in the `functions` directory


kentcdodds:
I configure env variables in the netlify UI, but locally I use `.gitignore`d `.env` files and https://www.npmjs.com/package/dotenv-cli to run netlify-lambda with the right env variables.

It'd be cool if netlify-dev-plugin could solve this problem somehow :)

uptownhr:
Not sure if this is the right thread for this but I also ran into env variable issues when trying out netlify dev. I would actually wish that .toml file environmental variables be used in netlify dev and made available not just to the site but also to the functions. This would provide a single place to handle environmental variables for both dev and production

BrunoQuaresma:
Is there any workaround to get `.env` files working on `netlify dev`?

bsgreenb:
Shouldn't Netlify Dev environment just be its own deploy context?

Also am personally finding that Netlify kind of clashes with my dotenv setup, so would definitely appreciate a dotenv friendly approach.

dferber90:
> I'm finding that I cannot override any `process.env` variables when using `dotenv` in combination with `netlify dev`.
> 
> And I can confirm that `dotenv` is called (via the log):
> 
> ```
> Request from ::ffff:127.0.0.1: GET /.netlify/functions/auth/twitter
> isProduction false
> Loading dotenv config...
> ```

@owenconti  This happens because `netlify dev` will set the environment variables before `dotenv` gets a chance to run. `dotenv` doesn't overwrite existing environment variables (see https://www.npmjs.com/package/dotenv#what-happens-to-environment-variables-that-were-already-set), so the value defined in Netlify wins.

---

@BrunoQuaresma Here's a snippet which enables overwriting env vars defined in Netlify with env vars from `.env`. This only works for Netlify functions though, not for your client-side code. Note that this snippet would have to be pasted at the top of every function you want to use.

```js
const dotenv = require("dotenv")
const envConfig = dotenv.config();
// "parsed" will be undefined when no .env file was present,
// in which case we need to skip the call to Object.entries
if (envConfig.parsed) {
  Object.entries(envConfig.parsed).forEach(([key, value]) => {
    process.env[key] = value;
  });
}
```

<details>
<summary>Or as a short version</summary>

```js
const envConfig = require("dotenv").config();
Object.entries(envConfig.parsed || {}).forEach(
  ([key, value]) => (process.env[key] = value)
);
```

</details>

---

@sw-yx 

[...]

Issue #457, opened 2019-04-17
support local environment variables and dev-specific settings in netlify.toml, with optional context flag


sw-yx:
@davidwells already has a prototype of this. 

related: 

- https://github.com/netlify/netlify-dev-plugin/issues/114
- https://github.com/netlify/netlify-dev-plugin/issues/60
- also raised in community
- also raised in slack (internal channel) for an enterprise customer:  https://netlify.slack.com/archives/C8VD55Q23/p1562004405020500

cavatron:
As a workaround, set your variable in the UI to the value you want it to be while developing, which Netlify Dev will use. Then use `netlify.toml` to set up other contexts, which will override the ones in the UI when needed.
```
[context.production.environment]
  MY_ENV = "something for prod"
```

Issue #444, opened 2019-04-10
feature request: support for environment settings specific to [dev]


fool:
**Request:**

folks would like to be able to set dev-specific environment variables via toml file.  These should take precedence over the ones set in our Build & Deploy settings admin UI.  TBD as to whether they should override locally-set-in-the-shell env vars, I leave that detail up to the implementor.  I could see either way making sense!

**Requestors:**

Netlify's Support team is tracking customers who have requested it.  So far just one request in slack (only visible to Netlify employees:  https://netlify.slack.com/archives/CC07WN1GC/p1554911472067300)

bsgreenb:
Hey there, want to second that I'm afraid to use `netlify dev` until it can have its own environmental variables.

Also, feel like we should be allowed to build netlify sites with dotenv entirely, hope that's coming soon as well!

sami616:
Would be great to override env variables specifically when running netlify dev. My use case is connecting to my development database inside a lambda.

Having the option of using a .env file or specific context inside the .toml feels like s nice solution

fjeddy:
So the only real alternative here is to have duplicate env variables, one for production and one for dev, and then write redundant code to switch between the two depending on what env you're on? :S

It seems odd that this wasn't accounted for when it was made in the first place; I assume very few people use the same variables when developing as live.

peterjohansson92:
I agree with @fjeddy 

I think that all the env-variables that you add should be treated as a secret. Even though you can add different env-variables to the netlify.toml, it still needs to be located in the repository.

You should be able to add different env-setups in the UI (like you can do in the netlify.toml) or somehow be able to inject it from a 3:rd party through a pre-script.

https://daniel-krzyczkowski.github.io/Access-Azure-Key-Vault-secrets-in-the-Azure-DevOps-Release-Pipelines/

https://daniel-krzyczkowski.github.io/How-to-inject-Azure-Key-Vault-secrets-in-the-Azure-DevOps-CICD-pipelines/

trevorblades:
@sami616 I posted a comment in a different issue (https://github.com/netlify/cli/issues/405#issuecomment-563105485) about my solution for this problem. In short, I set `_DEV` and `_PROD` environment variables in the Netlify interface, and then conditionally use one over the other based on `process.env.NETLIFY_DEV`, which is `true` when you're running `netlify dev` and `false` when deploying for production/staging. Hope this helps!

fjeddy:
[...]

Issue #405, opened 2019-07-28
What's the proper way of setting environment variable in Netlify functions?


iamskok:
What's the proper way of setting environment variables in netlify? I would like to be able to set different values for the variables depending on the environment.

Pseudo code:

```
let host;
if (process.env.GATSBY_CUSTOM_CONTEXT === 'production') {
	host = process.env.PRODUCTION_HOST
} else if (process.env.GATSBY_CUSTOM_CONTEXT === 'development') {
	host = process.env.DEVELOPMENT_HOST
}
```

I have tried passing env variable thru CLI, like `GATSBY_CUSTOM_CONTEXT=production gatsby build` and I also tried using same command with `cross-env`. 

My other attempt used `netlify.toml`:

```
[build]
  base = "/"
  publish = "public"
  command = "yarn build"
  functions = "src/functions"

[context.production]
  [context.production.environment]
    GATSBY_CUSTOM_CONTEXT = "production"
```

All of these options worked with `netlify dev` locally, but in production `GATSBY_CUSTOM_CONTEXT` is always `undefined`.

P.S. 

Not sure if I used the right repo to post this question. Feel free to tell me if this issue has to be moved some other place. I have also asked this question in https://github.com/netlify/ask-netlify/issues/59 bc I would like to see a more in-depth tutorial.

trevorblades:
Is there no way to differentiate between running `netlify dev` locally and my function running in production on Netlify? If I set an environment variable in the Netlify control panel, I can't override it on my local machine (or I haven't figured out how). In my case, I'm trying to use a different GitHub auth client id/secret in development and production.

iamskok:
@trevorblades thanks for updating your answer! This should solve my problem.

PR #714, opened 2020-02-19
Netlify Dev: Add support for .env files

RaeesBhatti:
**- Summary**
Allows users to use `.env` files for their site generators and Netlify functions written in JS, when using `netlify dev`.
Fixes: https://github.com/netlify/cli/issues/457
Fixes: https://github.com/netlify/cli/issues/444
Fixes: https://github.com/netlify/cli/issues/474
Fixes: https://github.com/netlify/cli/issues/405

**- Test plan**
* Add `.env` file to your function directory with `SOMETHING=true` and see if it is accessible in function using `process.env` while running through `netlify dev`
* Add a similar `.env` file to the root of your project and see if it is available to your site framework (etc. React etc.) when running through `netlify dev`

* Use `dotenv` package to parse `.env` file if exists and set the env variables to newly spawned site server process.
* Do the same for Netlify functions builder when running through `netlify dev`
* Use `lambda-local` to serve JS functions and pass path to function local `.env` file as parameter when executing

🙈


sw-yx:
looks good but needs docs and tests. got to start having a habit of adding these alongside features or it just never gets done

unclejustin:
So this seems to work for new variables in .env, but if I have an existing variable in netlify.com the "remote" variable overrides my local .env. This seems backwards to me. The whole point of a .env file is to use "dev" api key for local development and a production key in actual production. Am I missing something?

RaeesBhatti:
@unclejustin I can confirm this behavior for Netlify Function. The cause for this is that `dotenv` doesn't [override env variables that are already set](https://github.com/motdotla/dotenv/issues/199). Please hold on while we figure this out.

RaeesBhatti:
Being discussed here: https://github.com/netlify/cli/issues/734
cc @unclejustin @jrunestone

oles:
@RaeesBhatti 

No matter what I do - I cannot access any variables from my .env file in my functions. 
It doesn't matter if the .env file is in my root folder, inside the functions folder, inside a function folder.

`netlify dev` do report: `Overriding the following env variables with .env file: BISNODE_API_KEY,BISNODE_API_SECRET`

Though `process.env` do not contain it.

On a side note, this is the third bothersome quirk with `netlify dev` I've experienced with your functions offering. I'd really like to use your product, and it saddens that I'm having such a hard time with these seemingly basic things

RaeesBhatti:
@oles Can you please share an example project with this problem!

oles:
@RaeesBhatti Sure - I'll get to it tomorrow 👍

oles:
@RaeesBhatti Here you go: https://github.com/oles/netlify-functions-env-example

Just run `netlify dev` in it, verify that it reports `Overriding the following env variables with .env file: API_KEY,API_SECRET`.

Then open `http://localhost:8888/.netlify/functions/example`, and see that it is blank, and that it logs out two `undefined`s in the console.

They should be the values from the .env file, as accessed in the function with `const {API_KEY, API_SECRET} = process.env`

RaeesBhatti:
Thanks for reporting this @oles. This was caused by a silly mistake, we've added tests to make sure that doesn't happen again: https://github.com/netlify/cli/pull/748

unclejustin:
🍻 Cheers for the transparency @RaeesBhatti !

oles:
@RaeesBhatti Perfect! Thank you!

RaeesBhatti:
A fix for this has been released in `netlify-cli@2.39.0`

fjeddy:
This only listens for for variables in `.env` and ignores alternatives like `.env.production` and `.env.development`

wesbos:
Yeah this is confusing as Gatsby uses the above env files, not just`.env`

JLarky:
> Yeah this is confusing as Gatsby uses the above env files, not just`.env`

yes, I was also confused by that so I created https://github.com/netlify/cli/pull/924 hope that helps

PR #721, opened 2020-02-20
Adds angular detector

tzmanics:
**- Summary**
Creates a detector that looks for an Angular workspace configuration file to identify it as an Angular project.

**- Test plan**

ran `dev`, `init`, and a few other cli commands to make sure nothing broke

Added a detector to distinguish Angular projects.

baby jaguar

### DIFF
--- a/src/detectors/angular.js
+++ b/src/detectors/angular.js
@@ -1,0 +1,32 @@
+const { hasRequiredDeps, hasRequiredFiles, getYarnOrNPMCommand, scanScripts } = require('./utils/jsdetect')
+
+module.exports = function() {
+  // REQUIRED FILES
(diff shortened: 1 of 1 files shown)


src/detectors/angular.js
```diff
@@ -0,0 +1,31 @@
...
+  })
+
+  if (possibleArgsArrs.length === 0) {
+    // offer to run it when the user does not have any scripts setup! 🤯
+    possibleArgsArrs.push(['ng', 'build', '--prod'])
+  }
+
+  return {
+    type: '@angular/cli',
+    port: 8888,
```
RaeesBhatti: The `command` property is actually required for everything to work correctly. Please revert your last commit. Otherwise everything looks great! 🎉

Issue #716, opened 2020-02-20
Console.log no longer working in netlify-cli/2.35.0 


michalhudecek:
Console.log used to work in previous versions. After updating to 2.35.0, console.log statements no longer show.

RaeesBhatti:
Hey @michalhudecek, can you give an example of this! Do you mean the console.logs in Netlify Cli codebase or your site generator or Netlify Functions?

michalhudecek:
Sorry, I mean in Netlify functions. When I run `netlify dev`  I used to get console.log output from netlify functions when they were called. Now I do not get anything. I tried various functions I have it is the same everywhere.

michalhudecek:
Here is a screenshot of the console. The function is successfully called but no console.logs show anymore.

![image](https://user-images.githubusercontent.com/1215304/74934788-40ec4c00-53e7-11ea-913a-87587bd0fe77.png)

Philip-Owen:
Having the same issue here. Console logs from within exports.handler are not showing up in the terminal. I'm using netlify-cli/2.35.0 as well.

michalhudecek:
If it helps, I am on Windows 10, node.js v12.16.0

RaeesBhatti:
Thanks for reporting this. We have a PR in the works for this one. Please hold on

RaeesBhatti:
If you can, please try out the changes from the PR (https://github.com/netlify/cli/pull/722) and provide feedback on if it is working as intended.

itsravenous:
👋  This appears to have regressed again. I'm on cli version 2.57.0 and can't see console logs called from my handler function.

PR #722, opened 2020-02-20
Dev: Fix Netlify Functions logging

RaeesBhatti:
**- Summary**
Fixes `console.log`s not working in latest version when running Netlify Functions with `netlify dev`.
Fixes: https://github.com/netlify/cli/issues/716 

**- Test plan**

* Add `console.log` to your Netlify Function and run it through `netlify dev`

Use `winston` to provide custom logger for `lambda-local`

🐥

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
     "update-notifier": "^2.5.0",
     "uuid": "^3.3.3",
     "wait-port": "^0.2.2",
+    "winston": "^3.2.1",
     "wrap-ansi": "^6.0.0",
     "write-file-atomic": "^3.0.0"
   },
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -7,6 +7,7 @@ const queryString = require('querystring')
 const chokidar = require('chokidar')
 const jwtDecode = require('jwt-decode')
 const lambdaLocal = require('lambda-local')
+const winston = require('winston')
 const {
   NETLIFYDEVLOG,
   // NETLIFYDEVWARN,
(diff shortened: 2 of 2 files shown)


RaeesBhatti:
1. Yes, we switched from using our custom logic to invoke Netlify Function on `netlify dev` to using `lambda-local`, so that, we can support `.env` files and because `lambda-local` is more featureful. But the default logger in `lambda-local` logs input and output each Netlify Function event, which clutters the terminal. The only way to disable that was to set `verboseLevel` parameter to `0`. But setting `verboseLevel` to a number less than `3` makes `console.log` statements from Netlify Functions not show up in console because of how the default logger logic is implemented.
2. The only way to fix that without changing `lambda-local` codebase is to provide our own logger. Using `getLogger` is not useful in this context because that logger already has some custom logic that is causing the problem. Which is why we provide our own.

erquhart:
Thanks, looks good to me! I’m cool with merging without tests, but that PR should come next - in this case just a test to ensure logs from a function work.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

unclejustin:
@RaeesBhatti hello! Thanks for your effort on this, but it doesn't actually fix the problem for me. I'm on 2.36 which includes this change, but console.log still does nothing.

RaeesBhatti:
@unclejustin Can you please post a sample function to reproduce this issue!

unclejustin:
Certainly. Also note that I can put the `console.log`s anywhere and none of them ever work.

```js
const { retry, respondSuccess, respondError } = require('./utils')
const stripe = require('stripe')(process.env.STRIPE_SK)

const baseUrl = process.env.BASE_URL

exports.handler = (event, context, callback) => {
  console.log('Please work bby')
  
  switch (event.httpMethod) {
    // Code...
  }
}
```

RaeesBhatti:
The tried the following code and it's working as expected in the latest version:
```
const baseUrl = process.env.BASE_URL

exports.handler = (event, context, callback) => {
  console.log('Please work bby')
  
  switch (event.httpMethod) {
    // Code...
  }
}
```
Can you please make sure that you do a `netlify version` in the same directory to make sure you're running the latest version. Also, maybe try global uninstall and then install again. It could also be that if you're running with `yarn` or `npx` they're using different locally installed version than the latest one.
Screen Shot 2020-02-27 at 9 22 52 PM

unclejustin:
It was yarn! @RaeesBhatti 🙇 

Ran `netlify version` as you said and confirmed 2.35 ran which and found that I had installed initially with npm global. But "upgraded" with yarn 🤦‍♂ 

I'm all good, thanks for straightening me out 💪

RaeesBhatti:
No problem. Glad we could figure this out! 🎉

Issue #717, opened 2020-02-20
MaxListenersExceededWarning shows every time I call a function


michalhudecek:
```(node:12448) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 resize listeners added to [WriteStream]. Use emitter.setMaxListeners() to increase limit```

![image](https://user-images.githubusercontent.com/1215304/74907015-d4575a00-53b2-11ea-825e-48cca8879d87.png)

I am using netlify-cli/2.35.0 . It was showing in previous versions as well though.

**Comment by RaeesBhatti** [comment]
Hey @michalhudecek Can you fill out the following:

**- What is the current behavior?**

**- If the current behavior is a bug, please provide the steps to reproduce.**

**- What is the expected behavior?**

**- Local Environment Information**
Paste the results of `netlify status --verbose` and `netlify version` here

**Comment by mthit** [comment]
@RaeesBhatti I was able to get this error with the following setup:

```
netlify-cli/2.36.0 win32-x64 node-v12.13.1
```

I narrowed it down to having this as the only code inside netlify.toml:

```
[[redirects]]
  from = "/*"
  to = "/index.html"
  status = 200
  force = false
```

When I got rid of the toml file and instead used a _redirects file with this content, the error was gone:
```
/* /index.html 200
```

Also, the stack log of the error was as follows:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 keypress listeners added to [ReadStream]. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:268:17)
    at ReadStream.addListener (events.js:284:10)
    at ReadStream.Readable.on (_stream_readable.js:868:35)
    at new Interface (readline.js:230:11)
    at Object.createInterface (readline.js:78:10)
    at handleSignalEvents (C:\Users\HP\AppData\Roaming\nvm\v12.13.1\node_modules\netlify-cli\node_modules\configorama\lib\utils\handleSignalEvents.js:18:25)
    at new Configorama (C:\Users\HP\AppData\Roaming\nvm\v12.13.1\node_modules\netlify-cli\node_modules\configorama\lib\main.js:65:7)
    at module.exports (C:\Users\HP\AppData\Roaming\nvm\v12.13.1\node_modules\netlify-cli\node_modules\configorama\lib\index.js:19:20)
[...]

PR #728, opened 2020-02-26
Dev: Fix a memory leak in redirects parsing

RaeesBhatti:
**- Summary**
The redirects parser was trying to parse on each request when there was only one redirect rule (for homepage). This change fixes the memory leak.
Fixes: https://github.com/netlify/cli/issues/717

**- Test plan**
1. Create a blank project with only index.html and the following contents in your `netlify.toml`:
```toml
[[redirects]]
  from = "/*"
  to = "/index.html"
  status = 200
  force = false
```
2. Run `netlify dev`, open the page in your browser and keep hitting refresh

Move the rules parsing to file change events of redirect files, instead of keeping it in `getMatcher` function.

🦊


RaeesBhatti:
**Context of the problem**:

We parsed the redirect rules in `getMatcher` function and then propagated them to a higher context. And then check if there were any rules defined we would create a matcher for them and propagate it to higher context, otherwise just returned a noop function. And then at the start of `getMatcher` we would check if the matcher is already defined so that we wouldn't have to run the parsing the creating matcher logic again. We also filtered those rules to exclude the homepage redirect because this is done automatically in dev server.

**Problem**:

If there was only one rule and it was for homepage redirect (the one we filter out), it would mean that no rule logic will run and thus the whole parsing and getting matcher would run again. Each time we ran rules parsing, it created some event listeners which run on each request because we were parsing the rules if there were no rules before. This would result in the warning about maximum listeners in the console.

**Solution**:

Move the rules parsing logic out of `getMatcher` and into `onChanges` function which executes on first run and then only when redirect files change. Which means that listeners are not added on each request, fixing our problem.

There is also a possible bug in the `@netlify/config` -> `configurama` package, which I'm going to discuss with David because he's the author. Basically, `configurama` creates those event listeners for async calls only and not on the `.sync` method, and `async` calls are being used in `@netlify/config` which we're using to parse the config files. I think it either shouldn't add those event listeners on async calls or `@netlify/config` should use the `.sync` method.

RaeesBhatti:
I'll create some tests for this today.

Issue #608, opened 2019-11-08
Automatically open browser to netlify dev port on `netlify dev`


DavidWells:
Many users internal & external have reported issues with running `netlify dev` where the logs saying where the netlify dev server is running disappears.

This leaves the user in a state where they cannot proceed and open/dev their app.

This has been seen effecting create react app & gatsby projects.

**- What is the current behavior?**

There is a log message that gets erased by certain tools (CRA/Gatsby)

This leaves the users wondering where the netlify dev app is actually running

**- If the current behavior is a bug, please provide the steps to reproduce.**

**- What is the expected behavior?**

Automatically open the default browser (and re-use tab if one already is open). This is the behavior of create react app and its a nice DX.

So:

1. Print the message "Netlify dev is now running on port:whatever"
2. Then automatically open the dev app in the default browser

rlfrahm:
This is the top result on Google when looking to disable this behavior so posting solution to disable here. I ended up holding my version for a while until I finally found that you could just add `autoLaunch = false` to `[dev]` in `netlify.toml`.

Like so:

```toml
[dev]
  autoLaunch = false
```

Reference: https://github.com/netlify/cli/blob/master/docs/netlify-dev.md#netlifytoml-dev-block

PR #729, opened 2020-02-26
Dev: Change autoLaunch default behavior

RaeesBhatti:
**- Summary**
Disable auto launch of browser on `netlify dev` by default. Requires `autoLaunch = true` in config to enable. I'm introducing this change because in my experience, the feature is more annoying than it is useful in most cases.
Ref: https://github.com/netlify/cli/issues/615
Ref: https://github.com/netlify/cli/issues/608#issuecomment-573114582

**- Test plan**
Run `netlify dev` in any project

Change the `autoLaunch` parameter logic, so that, it requires `autoLaunch` to be `truthy` to open the browser automatically.

🐰

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -459,11 +459,7 @@ class DevCommand extends Command {
       }
     })
 
-    if (
-      isEmpty(config.dev) ||
-      !config.dev.hasOwnProperty('autoLaunch') ||
-      (config.dev.hasOwnProperty('autoLaunch') && config.dev.autoLaunch !== false)
-    ) {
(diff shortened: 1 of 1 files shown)


erquhart:
@RaeesBhatti we should probably have some product agreement before changing current behaviors. The point was made in #608 that CRA also does this and some consider it a positive for DX. This also is a breaking change for folks who expect it. Since we have the `autoLaunch` config for turning this off, let's leave as is for now.

erquhart:
Going to close this so it's status is clear, @RaeesBhatti please comment if you want to discuss further.

Issue #734, opened 2020-03-04
Netlify Dev (Functions): Allow overriding env variables set in app.netlify.com with .env file


RaeesBhatti:
**- Do you want to request a _feature_ or report a _bug_?**
Bug

**- What is the current behavior?**
Env variables set in [app.netlify.com](https://app.netlify.com) override variables in `.env` files for functions when running with `netlify dev`.
**- If the current behavior is a bug, please provide the steps to reproduce.**
* Link your project to a Netlify site using `netlify link`
* Create an env variable in [app.netlify.com](https://app.netlify.com) for your site
* Create a `.env` file in your function directory with the same variable name (access that env variable in your function)
* Run `netilfy dev` and access your function, see the value for the env var

**- What is the expected behavior?**
* `.env` file in function directory overrides env vars set in [app.netlify.com](https://app.netlify.com)

**- Local Environment Information**

```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: macOS 10.15.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 12.16.1 - ~/.n/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.n/bin/npm
  Browsers:
    Chrome: 80.0.3987.122
    Firefox: 72.0.2
    Safari: 13.0.5
```
```
netlify-cli/2.37.0 darwin-x64 node-v12.16.1
```

RaeesBhatti:
This was initially reported here: https://github.com/netlify/cli/pull/714#issuecomment-594832453

PR #735, opened 2020-03-04
Allow overriding functions env vars with .env file

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/734

**- Test plan**
Env variables set in [app.netlify.com](https://app.netlify.com) override variables in `.env` files for functions when running with `netlify dev`.
* Link your project to a Netlify site using `netlify link`
* Create an env variable in [app.netlify.com](https://app.netlify.com) for your site
* Create a `.env` file in your function directory with the same variable name (access that env variable in your function)
* Run `netilfy dev` and access your function, see the value for the env var

* Record `process.env` before creating functions handler
* Provide variables from `.env` file to `lambda-local` (will result in overriding `process.env`)
* Restore `process.env` from earlier recorded object after `lambda-local` has finished executing

🦁

### DIFF
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -157,8 +157,9 @@ function createHandler(dir) {
     const callback = createCallback(response)
     // we already checked that it exports a function named handler above
     let envConfig = {}
-    if (fs.existsSync(path.resolve(path.dirname(functionPath)))) {
-      envConfig = dotenv.parse(fs.readFileSync(path.resolve(path.dirname(functionPath), '.env')))
+    const envPath = path.resolve(path.dirname(functionPath), '.env')
+    if (fs.existsSync(envPath)) {
+      envConfig = dotenv.parse(fs.readFileSync(envPath))
     }
 
     return lambdaLocal.execute({
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -208,17 +208,7 @@ async function serveFunctions(settings) {
     return r
   })
 
-  app.listen(settings.functionsPort, function(err) {
-    if (err) {
-      console.error(`${NETLIFYDEVERR} Unable to start lambda server: `, err) // eslint-disable-line no-console
-      process.exit(1)
-    }
-
-    // add newline because this often appears alongside the client devserver's output
-    console.log(`\n${NETLIFYDEVLOG} Lambda server is listening on ${settings.functionsPort}`) // eslint-disable-line no-console
-  })
-
-  return Promise.resolve()
+  return app
 }
 
 module.exports = { serveFunctions }
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -125,7 +125,7 @@ function createHandler(dir) {
       response.end('Function not found...')
       return
     }
-    const { functionPath, moduleDir } = functions[func]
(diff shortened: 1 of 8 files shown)


src/utils/serve-functions.js
```diff
@@ -154,14 +156,18 @@ function createHandler(dir) {
 
     const callback = createCallback(response)
     // we already checked that it exports a function named handler above
+    let envConfig = {}
+    if (fs.existsSync(path.resolve(path.dirname(functionPath)))) {
+      envConfig = dotenv.parse(fs.readFileSync(path.resolve(path.dirname(functionPath), '.env')))
```
ehmicky: If the `.env` file does not exist, `fs.readFileSync()` will throw. Is this ok?

RaeesBhatti: We're checking for `fs.fs.existsSync` first and then calling `fs.readFileSync`

ehmicky: We are checking if `{functionPath}` exists but not if `{functionPath}/.env` does.

RaeesBhatti: Oh good catch! Thanks

RaeesBhatti:
Actually on the second note, I made sure that the app server env variables are [not set on the main process](https://github.com/netlify/cli/blob/57e49745adb64ec0d6b59b18d4f5cc167faccb23/src/commands/dev/index.js#L320-L328), they're instead only set on the child process on app server.

Similarly in the function server, since we're not creating a child process to serve functions, I record the main process variables [before any function is run](https://github.com/netlify/cli/pull/735/files#diff-cf2bd94d7688f0711bbd931e6c5e0397R199-R200), run the function with [its specific env vars](https://github.com/netlify/cli/pull/735/files#diff-cf2bd94d7688f0711bbd931e6c5e0397R159-R170), then [destroy those functions specific vars](https://github.com/netlify/cli/pull/735/files#diff-cf2bd94d7688f0711bbd931e6c5e0397R171) after function execution is complete and [reset the vars from main process](https://github.com/netlify/cli/pull/735/files#diff-cf2bd94d7688f0711bbd931e6c5e0397R206-R207).

On the other hand, if app server or any of the functions is compromised, they have access to file system and can read anything.

erezrokah:
@RaeesBhatti, sorry I should have been more clear with my note. It wasn't related to the cli itself, was referring to deployed functions in general.
Each site function inherits all env. variables configured for that site and not just the ones it should have access to.

RaeesBhatti:
Ah I see. I think we implemented the wrong behavior in the CLI then.

erezrokah:
I thought this is done here https://github.com/netlify/cli/blob/0d076ac7f02fe296ea2623cbe26496fd506bebcf/src/commands/dev/exec.js#L17
https://github.com/netlify/cli/blob/0d076ac7f02fe296ea2623cbe26496fd506bebcf/src/utils/dev.js#L82

RaeesBhatti:
Yes that adds env variables from the production to Netlify Dev, this PR is specifically about supporting `.env` files for Netlify Dev specific functions. e.g. If you want to override prod env vars.

erezrokah:
I think we're agreeing on the same thing :) I don't think I under your comment about implementing the wrong thing. Looks like that with your fix the CLI behaves like it should - overriding prod env with ones from `.env` files

PR #740, opened 2020-03-07
fix(telemetry): do not track disabling of telemetry

### LINKED ISSUE #739: netlify still connects to telemetry server when user does not consent to telemetry!
# Bug

I ran `netlify --telemetry-disable` on a fresh install (`rm -rf ~/.netlify`) and it connected to `cli.netlify.com`:

Screen Shot 2020-03-07 at 07 24 56

1. install netlify cli via yarn/npm
1. run `netlify --telemetry-disable`

Expected: I'm not spied on.

Actual: Your app violates my consent.

**- Local Environment Information**

Paste the results of `netlify status --verbose` and `netlify version` here

```
sneak@nostromo:~$ netlify status --verbose
Not logged in. Please log in to see site status.

Login with "netlify login" command
sneak@nostromo:~$
```

```
sneak@nostromo:~$ netlify version
netlify-cli/2.32.0 darwin-x64 node-v13.1.0
sneak@nostromo:~$
```

# Dear Netlify

Dear Netlify:

I use and really like your service.  I was happy to hear you guys raised some money recently.  I was deeply disappointed to see that you have telemetry enabled in your app; I disabled it on all my systems using it but apparently it's still spying on me.

This sort of spying is deeply unethical.  You are not entitled to silently spy on your users simply because they didn't realize that your app is spyware.  You are not entitled to *assume* consent just because you mentioned on your webpage that your apps will silently spy on your users.  This makes your app **malware**.

"We're going to silently spy on you unless you read our whole website, notice that it's a spyware app, and then disable the spyware part" is not an ethical business practice.  You must **OBTAIN CONSENT** from your users to use their personal data.  You may not *assume consent*.

What you're doing now is unethical and abusive.  That’s simply not affirmative consent. Imagine if you tried that in life! “Anyone who stays in this room after 5pm is consenting to be groped! Proceed at your own risk.”  Don't be that creep.  Please don't ship spyware.

Note that fixing only this bug (the issue of sending telemetry when the user explicitly opts out) is not a complete solution.  You must **stop spying on your users** unless they have *explicitly agreed* to permit you to their usage information.

Best,
-@sneak

PS: "everyone else is doing it" is *not* an acceptable excuse for unethical behavior.

PPS: "but the collection is anonymous!" is not an excuse either: transmitting a user's IP to your telemetry server leaks that user's location, which is a violation of their privacy.  IPs uniquely identify many, many users, and SNI is not yet encrypted.

erquhart:
Fixes #739. The line removed in this PR was added long ago without an issue or PR, so it isn't clear why we need it. Since disabling telemetry via config bypasses this call altogether, we should bypass it for users that disable via flag as well. `track()` shouldn't be called at all for users who disable telemetry.

### DIFF
--- a/src/hooks/init.js
+++ b/src/hooks/init.js
@@ -8,9 +8,6 @@ module.exports = async context => {
     globalConfig.set('telemetryDisabled', true)
     console.log('Netlify telemetry has been disabled')
     console.log('You can renable it anytime with the --telemetry-enable flag')
-    track('user_telemetryDisabled', {
-      force: true
-    })
     process.exit() // eslint-disable-line
   }
   if (context.id === '--telemetry-enable') {


sneak:
`force: true` says it all.

Issue #747, opened 2020-03-13
Local environment variables not loaded in CLI 2.38.0+


cbetta:
**- Do you want to request a _feature_ or report a _bug_?**

A bug

**- What is the current behavior?**

Environment variables in a local `.env` file should be loaded.

**- If the current behavior is a bug, please provide the steps to reproduce.**

Although the CLI detect the `.env` file and prints out all the variables being overriden, none actually are. This behaviour works fine in version 2.37.0 and before. 

**- What is the expected behavior?**

It should load the local `.env` file

**- Local Environment Information**

```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: macOS Mojave 10.14.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
  Binaries:
    Node: 13.8.0 - ~/.nvm/versions/node/v13.8.0/bin/node
    Yarn: 1.22.4 - ~/code/box/developer.box.com-framework/node_modules/.bin/yarn
    npm: 6.13.6 - ~/.nvm/versions/node/v13.8.0/bin/npm
  Browsers:
    Chrome: 82.0.4083.0
    Firefox: 72.0.2
    Safari: 13.0.5
  npmGlobalPackages:
    netlify: 3.0.0
```

```
netlify-cli/2.38.0 darwin-x64 node-v13.8.0
```

RaeesBhatti:
Thank you for reporting this @cbetta. The issue has been identified and a fix will be released soon https://github.com/netlify/cli/pull/748

RaeesBhatti:
A fix for this has been released in `netlify-cli@2.39.0`

DarthOstrich:
This is still broken. Do you want me to open up a new issue?

RaeesBhatti:
@DarthOstrich Can you please explain the issue you're facing!

DarthOstrich:
Hi! When I use CLI v 2.37 my env vars from my local file show up in my react app. When I upgrade past that version (including 2.40) my env variables are the ones in the Netlify dashboard. They are not overwritten by my local file

RaeesBhatti:
Is the `.env` file in your project root directory or in your functions directory? The latest Netlify CLI only supports `.env` files in project root directory. We had added support for per-function `.env` files but it was removed recently because you can't set per-function env vars in production.

DarthOstrich:
Root directory.

RaeesBhatti:
[...]

PR #748, opened 2020-03-13
🐛 Fix a loop for setting env vars

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/747

**- Test plan**
* Added tests
* Have a `.env` file in your project root
* Access those vars in Netlify Functions via `netlify dev`

* Use `Object.entries` instead of `Object.values` 🤦🏽‍♂️

🙈

### DIFF
--- a/src/tests/dev.test.js
+++ b/src/tests/dev.test.js
@@ -15,7 +15,7 @@ test.before(async t => {
     ['dev'],
     {
       cwd: sitePath,
-      env: Object.assign({}, process.env),
+      env: Object.assign({}, process.env, { DUMMY_VAR: "true" }),
       detached: true,
       shell: true,
     }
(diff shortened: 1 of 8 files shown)


src/tests/dev.test.js
```diff
@@ -64,6 +64,49 @@ test.serial('netlify dev functions timeout', async t => {
   })
 })
 
+test('netlify dev env file', async t => {
+  let data = ""
+  await new Promise((resolve, reject) => {
```
ehmicky: Small detail: using a HTTP client library like `got` might help remove some lines of code.

RaeesBhatti: One thing at a time 😊

Issue #751, opened 2020-03-16
Netlify Dev runs into fatal error if no .netlify directory present


RaeesBhatti:
**- Do you want to request a _feature_ or report a _bug_?**
Bug
**- What is the current behavior?**
If no `.netlify` directory is present, `netlify dev` dies with following error:
```
TypeError: Cannot read property 'startsWith' of undefined
    at isNetlifyDir (~/Projects/netlify/cli/src/commands/dev/index.js:55:14)
    at DevCommand.run (~/Projects/netlify/cli/src/commands/dev/index.js:381:31)
    at <anonymous>
```
**- If the current behavior is a bug, please provide the steps to reproduce.**
* Create an empty project
* Run `netlify dev`

**- What is the expected behavior?**
`netlify dev` runs without .netlify directory.
**- Local Environment Information**
Paste the results of `netlify status --verbose` and `netlify version` here
```
netlify-cli/2.39.0 darwin-x64 node-v8.17.0
```

PR #752, opened 2020-03-16
Netlify Dev: Verify publish dir before comparing

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/751

**- Test plan**
Try running `netlify dev` in https://github.com/tamouse/jamstack-intro-basic

Make sure `config.build.publish` is defined before string comparison.

🦁


RaeesBhatti:
@ehmicky Can you shed some light on why `config.build.publish` is not defined in this case! It should be the project root directory by default, right?

ehmicky:
When `publish` is `undefined`, the project root directory is used. However I removed assigning the default value in that case because of https://github.com/netlify/build/pull/863.

Issue #446, opened 2019-04-22
Add support for custom headers


fool:
**- Do you want to request a *feature* or report a *bug*?**

*feature*

**- What is the current behavior?**

*netlify custom headers (cf https://www.netlify.com/docs/headers-and-basic-auth/#custom-headers) are not supported*

**- If the current behavior is a bug, please provide the steps to reproduce.**

*attempt to use a custom header from a dev instance, do not see the expected value*

**- What is the expected behavior?**

*attempting to use a path covered by a header from a dev instance should return contents with that custom header*

**- Please mention your node.js, and operating system version.**

*all*

MichalStaruch:
Honoring headers like CSP defined in _headers file would be a great feature to see in netlify dev. Without that development pipeline loses a lot of consistency. Site that looks good on localhost may be a total mess when deployed on production (strict CSP won't allow inline JS and CSS to run, for example).

PR #754, opened 2020-03-16
Add support for _headers files

RaeesBhatti:
**- Summary**

Fixes: https://github.com/netlify/cli/issues/446

**- Test plan**
Have the following file in your project root:
```
# a path:
/templates/index.html
  # headers for that path:
  X-Frame-Options: DENY
  X-XSS-Protection: 1; mode=block
# another path:
/templates/index2.html
  # headers for that path:
  X-Frame-Options: SAMEORIGIN
/*
  cache-control: max-age=0
  cache-control: no-cache
  cache-control: no-store
  cache-control: must-revalidate
```
Run `netlify dev`, make an http request to `/` and observe response headers.

* Add new function `parseHeadersFile`
* Watch for changes in `_headers` files in project root and dist directory
* Set headers defined in `_headers` files to responses

🦁

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -100,8 +100,8 @@ function initializeProxy(port, distDir, projectDir) {
   })
 
   const headersFiles = Array.from(new Set([
+    path.resolve(projectDir, '_headers'),
     path.resolve(distDir, '_headers'),
-    path.resolve(projectDir, '_headers')
   ]))
 
   let headerRules = {}
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -30,6 +30,8 @@ const {
 const boxen = require('boxen')
 const { createTunnel, connectTunnel } = require('../../utils/live-tunnel')
 const createRewriter = require('../../utils/rules-proxy')
+const { onChanges } = require('../../utils/rules-proxy')
+const parseHeadersFile = require('../../utils/headers')
 
 function isFunction(functionsPort, req) {
   return functionsPort && req.url.match(/^\/.netlify\/functions\/.+/)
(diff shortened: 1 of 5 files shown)


RaeesBhatti:
Just realized that this is missing support for `*` and `:placeholder`.

RaeesBhatti:
Specification at https://docs.netlify.com/routing/headers/

src/commands/dev/index.js
```diff
@@ -97,6 +99,17 @@ function initializeProxy(port) {
...
 
+  const headersFiles = Array.from(new Set([
+    path.resolve(projectDir, '_headers'),
+    path.resolve(distDir, '_headers'),
+  ]))
+
+  let headerRules = {}
+  onChanges(headersFiles, () => {
+    console.log(`${NETLIFYDEVLOG} Reloading headers files`, headersFiles.map(p => path.relative(projectDir, p)))
+    headerRules = headersFiles.reduce((prev, curr) => Object.assign(prev, parseHeadersFile(curr)), {})
```
erezrokah: Should we catch `parseHeadersFile` errors in this callback and log them?

erezrokah: Should we catch `parseHeadersFile` errors in this callback and log them?

RaeesBhatti: We're throwing those errors without catching because we want to halt the execution if a rule rule is not correct.

erezrokah:
With this PR merged the cli fails on `gatsby-starter-netlify-cms`.
Looks like the header file in that repo is not in the expected format:
https://github.com/netlify-templates/gatsby-starter-netlify-cms/blob/9d00ee28c82519f331eceff3edfdfabdff051ff9/_headers

Deploy succeeds though and shows 41 header rules processed. 
After the CLI fails, Gatsby server remains running in the background.
Also a browser window is still opened regardless of the failure.

@RaeesBhatti thoughts?

RaeesBhatti:
The headers file you provided doesn't look correct according to the specification: https://docs.netlify.com/routing/headers/#syntax-for-the-headers-file. The syntax used there is of a TOML file. I would suggest using the correct syntax for `_headers` or putting those rules in `netlify.toml` instead.
I think the production is being more lenient on this by trying to parse it as TOML if _headers fails. IMO we shouldn't implement that behavior in the CLI (not right now at least) since the syntax for an `_headers` file is clearly specified in the docs.

RaeesBhatti:
As for the CLI failing because of syntax errors and Gatsby server staying alive is an issue that we should definitely look at.

erquhart:
@rybit it looks like we allow `_headers` files to have TOML content, but this behavior isn't documented. I'm guessing this is a backwards compatibility choice - can you confirm, or mention someone who can?

@RaeesBhatti if this is indeed something we're intentionally supporting in the platform, we should mirror it in the CLI. But let's see where this discussion goes.

Issue #563, opened 2019-10-07
standardize port terminology


sw-yx:
**- Do you want to request a *feature* or report a *bug*?**
bug

**- What is the current behavior?**

it is unnecessarily confusing to teach/understand/document what `port` means in netlify dev. it causes us to shoot ourselves in the foot and likely has caused bugs as well as user confusion.

there are 3 kinds of ports.

- the dev server port we are proxying, which is what we spin up or the project spins up via our detector
- the functions server port
- the final port that we expect netlify users to use

the codebase is a mess when it comes to this.

for example, what do you expect netlify dev to do when someone sets `port` in the `[dev]` block of `netlify.toml`?

[it sets the netlify dev port, right? so obvious?](https://github.com/netlify/cli/blob/8ebf422ee4400c2ae74eaec8e72a057827655ad1/src/commands/dev/index.js#L396-L399)

[umm idk](https://github.com/netlify/cli/blob/8ebf422ee4400c2ae74eaec8e72a057827655ad1/src/commands/dev/index.js#L173) maybe it also sets the final port too?

i think we owe it to ourselves and our users to clarify all this. in the process we may uncover some bugs.

**- What is the expected behavior?**

clarity.

sw-yx:
cc @davidwells

DavidWells:
Whats the proposal for the new names?

sw-yx:
up to you

Issue #438, opened 2019-05-07
Support non-random port assignment?


mattlenz:
Is it possible to run netlify dev and specify which port the dev server will listen on?

At the moment it seems to select a random open port which makes it difficult to integrate with existing development processes.

The `--port` argument and config currently relate to the application port that will be proxied by netlify dev, but it looks like there is no way to configure netlify dev's port.

oles:
@sw-yx this still seems to be a lovely mess.

as far as I can tell, none of the port configs even work or do anything.

I wanted to change which port the "Server now ready on ..." uses.

Neither --port flag or lines in the netlify.toml [dev] block have seemingly any effect.

Frustrating.

oles:
Great! Thank you, @RaeesBhatti 🎉

PR #755, opened 2020-03-16
Make sure to respect port flags and configuration

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/438
Fixes: https://github.com/netlify/cli/issues/563

**- Test plan**
Specify `port` in your `netlify.toml`, like so:
```
[dev]
    port = 2021
```
Run `netlify dev` and make sure it is running on port 2021 (if is is not in use by another process).
Run `netlify dev --port=2025` and verify that it's running on port 2025.

Reset `port`, `targetPort` and `functionsPort` on `settings` object with values from flags having the highest priority, then configuration file (`netlify.toml` etc) and then site detectors.

🦁

### DIFF
--- a/docs/netlify-dev.md
+++ b/docs/netlify-dev.md
@@ -136,10 +136,7 @@ Netlify Dev will attempt to detect the SSG or build command that you are using, 
   publish = "dist" # If you use a _redirect file, provide the path to your static content folder
 ```
 
-<details>
-<summary>
-<b>Explanation of ports in Netlify Dev</b>
-</summary>
+## Explanation of ports in Netlify Dev
 
 There will be a number of ports that you will encounter when using Netlify Dev, especially when running a static site generator like Gatsby which has its own dev server. All the port numbers can be a bit confusing, so here is a brief explainer.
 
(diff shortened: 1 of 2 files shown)


RaeesBhatti:
I'm thinking of removing the option to specify `functionsPort` since that is used internally only, and not accessed by users directly. I think it just adds to the confusion. What do you think @ehmicky @erquhart

docs/netlify-dev.md
```diff
@@ -157,7 +154,14 @@ As for which port to use while doing local development in Netlify Dev, always lo
...
 
-</details>
+**Specifying custom ports for Netlify Dev**
+
+Netlify Dev allows you to specify the following parameters for port as both flags and in config file (`netlify.toml` etc.):
+* `port`: The port for the main Netlify Dev server, the one you'll open in browser.
+* `targetPort`: The port for your application server or site generator.
+* `functionsPort`: The port for Netlify Functions server. This server is accessed internally within Netlify Dev, you shouldn't need to access it directly. And thus shouldn't need to change this port. You can access your functions at Netlify Dev main server port like so: `http://localhost:<port>/.netlify/functions/<your-function-slug>`.
+
+Netlify Dev tries to acquire these ports but if any of them is not available (already in use by another application), it assigns a random port instead of that.
```
erezrokah: Shouldn't we fail if the user explicitly sets a port but it is not available? 
I'm assuming users set a specific port since they want to use it later on in their code/login, so assigning a random one might create issues.

Don't think it is a blocker for the PR, just something to consider

src/commands/dev/index.js
```diff
@@ -402,6 +402,21 @@ class DevCommand extends Command {
...
+          ...(config.dev.port && { port: config.dev.port }),
+          ...(config.dev.targetPort && { proxyPort: config.dev.targetPort }),
+          ...(config.dev.functionsPort && { functionsPort: config.dev.functionsPort })
+        },
+        {
+          ...(flags.port && { port: flags.port }),
+          ...(flags.targetPort && { proxyPort: flags.targetPort }),
+          ...(flags.functionsPort && { functionsPort: flags.functionsPort })
+        },
+    )
```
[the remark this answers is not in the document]
RaeesBhatti: I was trying to avoid an import but I think we can make it a bit more concise without `pickBy`.

docs/netlify-dev.md
oles: Yes please. Fail fast when the port is taken.

Else most people will start their server, try to use the port they specify, get confused on why it is not working, try clearing their cache or something, and then finally check what port it actually started on.

erquhart: Can be discussed under #767.

PR #764, opened 2020-03-24
Dev: Pass .env vars as an object to child process

RaeesBhatti:
**- Summary**
This is to fix an inconsistency in NodeJS which results in `process.env` not realizing changes if they're accessed shortly after.
Fixes: https://github.com/netlify/cli/issues/747

**- Test plan**
* Have your `create-react-app` project connected to Netlify production with `netlify link`.
* Have some env variable in Netlify production
* Have a `.env` file in project root locally that overrides that env variable
* Run `netlify dev` and observe the value of that env variable

Pass the variables from `.env` file as `setting.env` object to `startDevServer`, so that, they're set explicitly on the new process.

🐳

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -380,7 +380,15 @@ class DevCommand extends Command {
       const { addEnvVariables } = require('../../utils/dev')
       addonUrls = await addEnvVariables(api, site, accessToken)
     }
+
     process.env.NETLIFY_DEV = 'true'
+    // Override env variables with .env file
+    const envFile = path.resolve(site.root, '.env')
+    if (fs.existsSync(envFile)) {
+      const vars = dotenv.parse(fs.readFileSync(envFile)) || {}
+      console.log(`${NETLIFYDEVLOG} Overriding the following env variables with ${chalk.blue('.env')} file:`, chalk.yellow(Object.keys(vars)))
+      Object.entries(vars).forEach(([key, val]) => process.env[key] = val)
+    }
 
     let settings = await serverSettings(Object.assign({}, config.dev, flags))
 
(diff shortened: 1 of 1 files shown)


src/commands/dev/index.js
```diff
@@ -424,6 +424,7 @@ class DevCommand extends Command {
     const envFile = path.resolve(site.root, '.env')
     if (fs.existsSync(envFile)) {
       const vars = dotenv.parse(fs.readFileSync(envFile)) || {}
+      settings.env = { ...settings.env, ...vars }
       console.log(`${NETLIFYDEVLOG} Overriding the following env variables with ${chalk.blue('.env')} file:`, chalk.yellow(Object.keys(vars)))
       Object.entries(vars).forEach(([key, val]) => process.env[key] = val)
```
RaeesBhatti: `process.env[key]` being set here. Is not accessible two lines down when accessed in [`startDevServer`](https://github.com/netlify/cli/pull/764/files#diff-8a2f65e3aede3ea9696e546766f50f03R319).
cc @erquhart

erezrokah: The cause for this bug is the spread order of properties here:
https://github.com/netlify/cli/blob/3085892a7fb54f46c954ab82839dce605cf44cba/src/commands/dev/index.js#L341

`settings.env` properties override `process.env` properties.

Changing that line to `env: {  ...settings.env,...process.env, FORCE_COLOR: 'true' },`  fixes it

RaeesBhatti:
Ah I see. Thanks a bunch @erezrokah. This was a head scratcher. I never looked at the detectors logic, I thought they only returned application specific env variables.

RaeesBhatti:
Can you please take another look @erezrokah

erezrokah:
Sure, reviewing now

RaeesBhatti:
Yeah. All of the detectors inherit `process.env`. And it is part of the template, so, the future detectors will do the same.
I suggest that we don't change this since it's explained in the docs and example. But if we do need to, we can do that later.

PR #768, opened 2020-03-26
Upgrade `zip-it-and-ship-it` and `js-client`

ehmicky:
This upgrades `zip-it-and-ship-it` to the latest version used inside Netlify Build.
This new version should be backward compatible, although it has not been tested with many real users yet.

This will ensure Netlify Build users are using consistent versions for `zip-it-and-ship-it` since that library is used separately by several projects: Netlify Build, Netlify CLI and the JavaScript client, which are requiring each other.

If there are issues in the beta version of `zip-it-and-ship-it`, it is better for us to catch those with local builds than with production builds.

This PR also upgrades to the latest major of the `js-client`. That major version only adds one thing: upgrading to the latest `zip-it-and-ship-it` as well (see https://github.com/netlify/js-client/pull/90). This PR put both upgrades together to ensure the same version of `zip-it-and-ship-it` is used whether used directly or through the `js-client`.

Before merging this, @RaeesBhatti are there any chances you could manually test it locally (anything related to Functions, such as deploying then, etc.)? Thanks a lot!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1039,9 +1039,9 @@
       }
     },
     "@netlify/open-api": {
-      "version": "0.9.0",
-      "resolved": "https://registry.npmjs.org/@netlify/open-api/-/open-api-0.9.0.tgz",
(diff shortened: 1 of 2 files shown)


RaeesBhatti:
Tested and working fine. Thanks for the PR @ehmicky

Issue #769, opened 2020-03-27
Fails to install on Node `>=12` on Linux


ehmicky:
**- Do you want to request a _feature_ or report a _bug_?**

A bug.

**- What is the current behavior?**

When running `git clone` then `npm install`, it fails with:

```
npm i

> grpc@1.20.0 install /home/ether/netlify/cli/node_modules/grpc
> node-pre-gyp install --fallback-to-build --library=static_library

node-pre-gyp WARN Using request for node-pre-gyp https download 
node-pre-gyp WARN Tried to download(404): https://node-precompiled-binaries.grpc.io/grpc/v1.20.0/node-v79-linux-x64-glibc.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for grpc@1.20.0 and node@13.11.0 (node-v79 ABI, glibc) (falling back to source compile with node-gyp) 
make: Entering directory '/home/ether/netlify/cli/node_modules/grpc/build'
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/surface/init.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/avl/avl.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/backoff/backoff.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channel_args.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channel_stack.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channel_stack_builder.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channel_trace.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channelz.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/channelz_registry.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/connected_channel.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/handshaker.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/handshaker_registry.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/channel/status_util.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/compression/compression.o
[...]

PR #771, opened 2020-03-27
Upgrade `@netlify/build` and `@netlify/config`

ehmicky:
This upgrades `@netlify/build`. This should not be necessary since we use semver. However semver only works when users are updating their `package-lock.json`. Making a dummy new release of the CLI (which is essentially the goal of this PR) will update users lock files (if they upgrade Netlify CLI). 

This is important as part of the MVP, to make sure users get the same experience on Netlify CLI than in the UI.

Important: I could not update `package-lock.json` since Netlify CLI fails to install on my machine due to #769. Could someone not on Linux help me run `npm install` to update that file then push to this PR? Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -691,22 +691,21 @@
       }
     },
     "@netlify/build": {
-      "version": "0.1.98",
-      "resolved": "https://registry.npmjs.org/@netlify/build/-/build-0.1.98.tgz",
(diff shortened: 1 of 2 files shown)


RaeesBhatti:
@ehmicky Can you please fix these conflicts!

ehmicky:
Done.

Please note you would need to update the `package-lock.json` on this PR before merging it. I could not do it since `npm install` does not work on Linux with Netlify CLI anymore (#769) it seems.

RaeesBhatti:
It seems like GRPC build fails on Node 12. Can you try using Node 8 in the future while we fix this!
https://github.com/grpc/grpc-node/issues/834

PR #775, opened 2020-03-30
Make `netlify build` visible

ehmicky:
Now that Netlify Build is officially in Beta, we should make the `netlify build` CLI command visible.

This also adjusts the command's description.


RaeesBhatti:
Thanks for the PR @ehmicky

Issue #705, opened 2020-02-09
Some rewrites shadow static files by default


LynnKirby:
**- Do you want to request a _feature_ or report a _bug_?**

Bug.

**- What is the current behavior?**

Rewrites shadow static files (i.e. serve the rewrite rule instead of the file) without the `!` or force setting when rewriting to non-HTML files or to `/404.html`. For example:

```
/*    /style.css                              200
/*    /404.html                               200
/*    /.netlify/functions/my-function/:splat  200
/:id  /.netlify/functions/my-function?id=:id  200
```

**- If the current behavior is a bug, please provide the steps to reproduce.**

Add `index.html` and `404.html` as static files. Set `_redirects` to:

```
/*  /404.html  200
```

Navigate to `/index.html`. `404.html` will be displayed instead.

**- What is the expected behavior?**

Rewrites don't shadow unless forced. The rewrites and proxies documentation says:

> By default, you can’t shadow a URL that actually exists within the site when using a splat or dynamic path segment

RaeesBhatti:
This seems like a bug. Thank you for reporting this @LynnKirby .

PR #776, opened 2020-03-30
Fix rewrite force behavior for static files

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/705

**- Test plan**

Add `index.html` and `404.html` as static files. Set `_redirects` to:

```
/*  /404.html  200
```

Navigate to `/index.html`. `404.html` will **not be displayed**.

Make sure to trim `/` from pathname when resolving path to check for static files.

🐵

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -45,7 +45,7 @@ function addonUrl(addonUrls, req) {
 
 function notStatic(pathname, publicFolder) {
   return alternativePathsFor(pathname)
-    .map(p => path.resolve(publicFolder, p.startsWith('/') ? p.substr(1) : p))
+    .map(p => path.resolve(publicFolder, p.substr(1)))
     .every(p => !fs.existsSync(p))
 }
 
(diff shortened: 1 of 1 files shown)


sanderfish:
I think I was running into the issue fixed in this PR, but I'm unable to upgrade netlify-cli to 2.41.0. Would it be possible to publish the latest version on NPM?

RaeesBhatti:
@sanderfish `2.41.0` has been published now. Please try again

sanderfish:
@RaeesBhatti that was quick, thank you! I'm getting some odd behaviour with this latest version. Specifically:

- running `netlify dev` starts the server then starts displaying .......... and adds a dot for every second it's running
- running `netlify dev --dir=public` starts the server at port 3999 and then says `Waiting for localhost:34567....................................................`. It keeps adding a dot for every second it's running.
- _redirects rule of `/* https://othersite.com/:splat 200!` renders "Forbidden"

Should I move this to an issue?

RaeesBhatti:
@sanderfish that seems like a bug 😅. Please open an issue for it and add steps to reproduce while I try to hunt that down

Issue #447, opened 2019-04-22
Add support for forms


fool:
**- Do you want to request a *feature* or report a *bug*?**

*feature*

**- What is the current behavior?**

*netlify forms are not supported*

**- If the current behavior is a bug, please provide the steps to reproduce.**

*attempt to use a netlify form from a dev instance, get 404*

**- What is the expected behavior?**

*attempting to use a netlify form from a dev instance should return successful response and further run the event-triggered function, if any*

**- Please mention your node.js, and operating system version.**

*all*

cannikin:
I could see Phase 1 being as simple as creating a `forms.txt` file in the root of the app and appending the raw body from any POSTs to it. But the browser should display the same "Thank you" page that you get in production. Likewise an AJAX submit to the form should behave like production.

Phase 2 lets you browse to /.netlify/forms and you see the real Forms UI just like on app.netlify.com 😉

erikwoods:
For me personally, I would at least like to have the Netlify Form use the action attribute to redirect a Dev user to the appropriate page. Otherwise it just looks like the form/page is busted. I don't want my engineering friends to get hung up on this - and they will as things are today. :(

PR #780, opened 2020-03-31
Add support for form submissions

RaeesBhatti:
**- Summary**
Add support for form submissions.
Fixes: https://github.com/netlify/cli/issues/447

**- Test plan**
* Run `netlify dev`
* Make a POST request with url encoded data

Transform the body of POST requests with Form data to match a `submission-created` event in Production.

🐰


src/commands/dev/index.js
```diff
@@ -172,7 +175,25 @@ async function startProxy(settings, addonUrls, configPath, projectDir) {
     configPath,
   })
 
+  mkdirp.sync(path.resolve(projectDir, '.netlify'))
```
ehmicky: Would `await util.promisify(mkdirp)(path.resolve(projectDir, '.netlify'))` work instead?

src/utils/forms.js
```diff
@@ -0,0 +1,17 @@
+const FORM_URLENCODED = 'application/x-www-form-urlencoded'
+
+function collectRequestBody(request, callback) {
```
ehmicky: I think you might want to use a library to read `request` body instead? Libraries like `raw-body` adds lots of tricks and handle some edge cases when doing this.

```js
const getRawBody = require('raw-body')
const body = String(await getRawBody(request))
```

src/commands/dev/index.js
```diff
@@ -172,7 +175,25 @@ async function startProxy(settings, addonUrls, configPath, projectDir) {
...
+  const csvWriter = csvWriterPkg.createArrayCsvWriter({
+    alwaysQuote: true,
+    header: ['timestamp', 'path', 'body'],
+    path: formSubmissionsFile,
+    append: true,
+  })
+
   const server = http.createServer(function(req, res) {
+    if (req.method === 'POST') {
+      collectRequestBody(req, async body => {
```
ehmicky: The `Content-Type` response header should be set to `text/csv`?

src/commands/dev/index.js
```diff
@@ -172,7 +175,25 @@ async function startProxy(settings, addonUrls, configPath, projectDir) {
     configPath,
   })
 
+  await mkdirp(path.resolve(projectDir, '.netlify'))
```
ehmicky: `mkdirp` requires `util.promisify()` to be used with `await` I believe.

RaeesBhatti: The latest version has supported for promises by default

ehmicky: Oh right, I missed that you also upgraded `mkdirp` :+1:

erquhart:
@fool correct if I'm wrong but you weren't looking for actual submissions here, just a successful response and trigger the form submission triggered function if it exists. The form submission itself would not be retained. Sound about right?

src/commands/dev/index.js
fool: most folks return only html from their form submission, so probably don't need to get too fancy.

fool:
yes, that was indeed my thinking, @erquhart

src/utils/forms.js
RaeesBhatti: Oh nice, thanks!

RaeesBhatti:
@fool Do you have any advice on where to look for sample `submission-created ` event payload? I need the event shape to emulate that in `netlify dev`.

fool:
I don't have good examples of the payloads, aside from what is in the public doc:

https://docs.netlify.com/functions/trigger-on-events/#payload

fool:
Would be good to get a "dump the input" function example spun up.  I bet @depadiernos could write one in about 5 minutes, but it's not something I can just toss out since I'm no javascript pro

PR #781, opened 2020-04-01
fix: verbose flag displays the necessary information

### LINKED ISSUE #778: verbose flag doesn't serve the purpose
**- Do you want to request a _feature_ or report a _bug_?**
bug report

We've the `--verbose` flag as part of `status` command which shows up local environment information useful while submitting bug reports. But it doesn't display the information as expected.

https://github.com/netlify/cli/blob/d54b500c548c0e45d9a496eec988726e23cb67da/src/commands/status/index.js#L78
Moving the above block just below the respective line would serve the purpose.
https://github.com/netlify/cli/blob/d54b500c548c0e45d9a496eec988726e23cb67da/src/commands/status/index.js#L12

Type in `netlify status --verbose`
If you're logged in, the respective information is being displayed or else an appropriate warning shows up.

`netlify status --verbose` displays local environment information.

jamesgeorge007:
**- Summary**
Made `verbose` flag to show up local environment information which wasn't the case before.
closes #778

**- Test plan**

`--verbose` flag shows up information regarding the local environment when passed along with the `status` command.

image

### DIFF
--- a/src/commands/status/index.js
+++ b/src/commands/status/index.js
@@ -10,6 +10,22 @@ class StatusCommand extends Command {
   async run() {
     const { globalConfig, api, site } = this.netlify
     const { flags } = this.parse(StatusCommand)
+
+    if (flags.verbose) {
+      this.log()
+      this.log(`────────────────────┐
+ Environment Info   │
+────────────────────┘`)
+      const data = await envinfo.run({
+        System: ['OS', 'CPU'],
+        Binaries: ['Node', 'Yarn', 'npm'],
+        Browsers: ['Chrome', 'Edge', 'Firefox', 'Safari'],
+        npmGlobalPackages: ['netlify']
+      })
+      this.log(data)
+      this.exit()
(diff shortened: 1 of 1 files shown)


src/commands/status/index.js
```diff
@@ -10,6 +10,22 @@ class StatusCommand extends Command {
...
+ Environment Info   │
+────────────────────┘`)
+      const data = await envinfo.run({
+        System: ['OS', 'CPU'],
+        Binaries: ['Node', 'Yarn', 'npm'],
+        Browsers: ['Chrome', 'Edge', 'Firefox', 'Safari'],
+        npmGlobalPackages: ['netlify-cli']
+      })
+      this.log(data)
+      this.exit()
```
RaeesBhatti, line 26: Why exit here?

jamesgeorge007: Yeah, It would make sense to log rest of the status information.

PR #799, opened 2020-04-08
Revert logs tailing because of incompatibilities

RaeesBhatti:
**- Summary**
* The `firebase` dependency is breaking install for some users

**- Test plan**
That's the problem

* Revert changes related to build log tailing

🐡


ehmicky:
Would it be possible to split this into two PR, one for `firebase` and one for `@netlify/config`?

Downgrading `@netlify/config` would break running `netlify build` locally, so ideally we should find a fix for the issue introduced by `@netlify/config` instead. Do you have more information about that issue?

RaeesBhatti:
I've updated the PR and reverted only the changes related to logs tailing and `firebase` package, which is the main culprit of this incompatibility.

PR #806, opened 2020-04-08
Move --verbose flag to netlify version command

RaeesBhatti:
**- Summary**
`netlify status` shows information about current signed-in user and their linked site which is private information. `--verbose` flag shows information about current system and installed `netlify-cli` package.
IMO it makes more sense to move it to `netlify version` command. This way users can paste results of `netlify version --verbose` and it will give us a complete picture of their environment without a possibility of accidentally logging private information.

**- Test plan**
`netlify version --verbose`

Move `--verbose` flag logs to `netlify version` command.

### DIFF
--- a/src/hooks/init.js
+++ b/src/hooks/init.js
@@ -1,3 +1,4 @@
+const envinfo = require('envinfo')
 const header = require('../utils/header')
 const globalConfig = require('../utils/global-config')
 const { track } = require('../utils/telemetry')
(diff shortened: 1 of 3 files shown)


src/hooks/init.js
```diff
@@ -20,5 +21,19 @@ module.exports = async context => {
     process.exit() // eslint-disable-line
   }
 
+  if (process.argv.length > 3 &&
+      ['-v', '--version', 'version'].includes(process.argv[2]) && process.argv[3] === '--verbose') {
```
ehmicky, line 25: Should the order matter?

```
if (process.argv.includes(['-v', '--version', 'version'] && process.argv.includes('--verbose')) {
```

RaeesBhatti: Yes we want to make sure it is the `netlify version` command because this logic runs on everything.

RaeesBhatti: Another command could have `version` flag as well. We only want to run this logic specifically for `netlify version`

Issue #817, opened 2020-04-13
netlify dev: 'C:\Program' is not recognized as an internal or external command


teachtyler:
**- Do you want to request a _feature_ or report a _bug_?**
report a bug

**- What is the current behavior?**
netlify dev fails on create-react-app

```
$ ntl dev
◈ Netlify Dev ◈
◈ Starting Netlify Dev with create-react-app
Waiting for localhost:3000.'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
◈ Error while killing child process: kill ESRCH
```

**- Local Environment Information**
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: Windows 10 10.0.17763
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  Binaries:
    Node: 13.3.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.0 - C:\Program Files\nodejs\yarn.CMD
    npm: 6.13.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.831.0

teachtyler:
recreation steps:
```
npx create-react-app netlify-react-app
cd netlify-react-app
netlify dev
```
as well as 
https://github.com/teachtyler/create-react-app-lambda
https://github.com/teachtyler/deploy-create-react-app

RaeesBhatti:
Thank you for reporting this issue @teachtyler. The problem has been identified and fixed here https://github.com/netlify/cli/pull/818 and the fix will be released soon. In the mean time, you can try out the PR branch and provide feedback on if it fixes your problem.

PR #818, opened 2020-04-13
Improve child process spawning

RaeesBhatti:
**- Summary**
If the path to binary is not quoted on Windows and contains a space, it will get the path wrong and fail. This change makes sure that the path to app binary is quoted.
Fixes: https://github.com/netlify/cli/issues/817

**- Test plan**
* Have your app binary in path with spaces
* Run `netlify dev`

Quote app binary path before spawning it.

🐬

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -342,7 +342,8 @@ async function startDevServer(settings, log) {
 
   log(`${NETLIFYDEVLOG} Starting Netlify Dev with ${settings.type}`)
   const args = settings.command === 'npm' ? ['run', ...settings.args] : settings.args
-  const ps = child_process.spawn(await which(settings.command), args, {
+  const commandBin = await which(settings.command)
+  const ps = child_process.spawn(`"${commandBin}"`, args, {
     env: { ...settings.env, FORCE_COLOR: 'true' },
     stdio: settings.stdio || 'inherit',
     detached: true,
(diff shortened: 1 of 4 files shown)


ehmicky:
Note: an alternative would be not to use `shell: true`. This means users won't be able to use shell features like `$variable` expansion or globbing, but it also means no escaping issues. It also is safer. See [more information](https://github.com/sindresorhus/execa#shell).

RaeesBhatti:
Thanks for the tip @ehmicky. I attempted that yesterday but could make the addons tests work on Windows. Found out the problem today, we were trying to invoke a JS file without shell, which isn't supported. Updated now. Please take another look!

src/commands/dev/index.js
```diff
@@ -342,14 +342,13 @@ async function startDevServer(settings, log) {
...
   log(`${NETLIFYDEVLOG} Starting Netlify Dev with ${settings.type}`)
   const args = settings.command === 'npm' ? ['run', ...settings.args] : settings.args
-  const ps = child_process.spawn(await which(settings.command), args, {
+  const commandBin = await which(settings.command)
+  const ps = child_process.spawn(commandBin, args, {
     env: { ...settings.env, FORCE_COLOR: 'true' },
-    stdio: settings.stdio || 'inherit',
-    detached: true,
-    shell: true,
+    stdio: 'pipe',
```
ehmicky, line 348: Is it possible for `settings.command` to ask for interactive input from the user?
If so, `stdio: 'pipe'` would not work.

RaeesBhatti: How about we do `process.stdin.pipe(ps.stdin)`?

RaeesBhatti: Implemented here: https://github.com/netlify/cli/pull/823/files

Issue #374, opened 2019-07-24
Prevent terminal from being cleared in create-react-app project


raeesbhatti:
**- Do you want to request a *feature* or report a *bug*?**
Bug
**- What is the current behavior?**
When you run `netlify dev` from a `create-react-app` project, the terminal is cleared by `create-react-app` script.

**- If the current behavior is a bug, please provide the steps to reproduce.**
* `cd` into a `create-react-app` project
* Run `netlify dev`
* The terminal will be cleared by `create-react-app` script removing information by `netlify dev`
**- What is the expected behavior?**
* When you run `netlify dev`, the information would stay in the terminal and not be cleared
**- Please mention your node.js, and operating system version.**
Not relevant

raeesbhatti:
Fixed by: https://github.com/netlify/netlify-dev-plugin/pull/227

PR #823, opened 2020-04-13
Dev: Pipe stdin from parent to app process

RaeesBhatti:
**- Summary**
Pipe `stdin` of parent process to `stdin` of spawned app process.
Ref: https://github.com/netlify/cli/pull/818#discussion_r407461613

**- Test plan**


RaeesBhatti:
We're actually hoping to that child processes will not consider it a `tty` because this prevents some frameworks from clearing the screen, which removes useful information logged by Netlify CLI. https://github.com/netlify/cli/issues/374

Issue #827, opened 2020-04-13
Redirects tag in netlify.toml causes bundle error


mrrosoff:
**- Do you want to request a _feature_ or report a _bug_?**

Report a bug.

**- What is the current behavior?**

When using a redirects tag in netlify.toml (intended for React Router), netlify dev finds bundle.js from webpack-dev-server as index.html.

This is similar to this issue: https://github.com/netlify/cli/issues/443

**- If the current behavior is a bug, please provide the steps to reproduce.**

The repository is here: https://github.com/tc30/blockchain-document-dapp

Running npm run dev will successfully link the files together under webpack-dev-server. When built for production, netlify handles redirects correctly. Issue only exists in netlify dev.

**- What is the expected behavior?**

It redirects correctly and finds bundle.js in dev mode.

**- Local Environment Information**

System:
    OS: Linux 5.3 elementary OS 5.1.3 Hera
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    Yarn: 1.22.4 - /usr/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.14.1/bin/npm
  Browsers:
    Chrome: 81.0.4044.92
  npmGlobalPackages:
    netlify-cli: 2.45.0

netlify-cli/2.45.0 linux-x64 node-v12.14.1

RaeesBhatti:
@mrrosoff Thanks for opening this ticket! Can you please update to `netlify-cli@2.46.0` and see if that fixes this problem.

mrrosoff:
@RaeesBhatti No problem. Updated, two interesting behaviors now occur. 

Netlify and Webpack port now both open, and the Netlify port has no data. Some communication link has been changed. This removes the core behavior of Netify-dev. This seems to occur with or without the redirect field in netlify.toml. As the core functionality has been removed, it is impossible to test the redirect functionality.

The webpack-dev-server port does not have lambda function access. This is expected.

Thank you for your help!

mrrosoff:
This was on netlify-cli@2.47.0

mrrosoff:
@RaeesBhatti Tested netlify@2.46.0, different issue than 2.47.0.

No discernible change between 2.46.0 and 2.45.0.

RaeesBhatti:
[...]

Issue #441, opened 2019-04-24
"Starting Netlify Dev with gatsby" and then nothing


johnmryan:
**- Do you want to request a *feature* or report a *bug*?**
Bug

**- What is the current behavior?**
Followed the prerequisites section, ran `netlify dev` in a Gatsby project with netlify functions. Receive the following output:
```
◈ Netlify Dev ◈
◈ Injected build setting env var:  CLIENT_ID
... other injects
◈ Starting Netlify Dev with gatsby
```
Nothing else happens, it just exits after this with no warning. And nothing is running on local. Will try to step through the `netlify dev` code locally in the meantime to see if something jumps out at me.

**- Please mention your node.js, and operating system version.**
Ubuntu 18.04, Node 10.15.2

meronek:
Hi, amateur here also having this issue. Have been using netlify dev for a few weeks now on other projects but on this new one using react-static, netlify dev won't start up.

OSX, VS Code.

Just ran npm install netlify-cli -g
then netlify dev
and I get:

◈ Netlify Dev ◈
◈ Starting Netlify Dev with react-static
Waiting for localhost:3000.
◈ Lambda server is listening on 34567
And it exits back to terminal prompt.

Repo:
https://github.com/meronek/gfl-series-react-static

Thanks for any help or insight you can provide.

PR #843, opened 2020-04-18
Netlify Dev: Add framework = "#custom" option

RaeesBhatti:
**- Summary**
This is to clear confusion around usage.
If `framework = "#custom"` in `netlify.toml` `dev` section, the user must specify `command` and `targetPort` options as well and we'll throw an error if they're missing any of these.
If `framework` is not `#custom` and user has specified `command` and `targetPort` options, we'll throw an error and ask them set `framework` to `#custom`.
If `dir` flag is specified, we'll ignore all detectors and throw an error if `targetPort` or `command` is specified.

Fixes: https://github.com/netlify/cli/issues/827
Fixes: https://github.com/netlify/cli/issues/441
Fixes: https://github.com/netlify/cli/issues/851
Fixes: https://github.com/netlify/cli/issues/862
Fixes: https://github.com/netlify/cli/issues/850
Fixes: https://github.com/netlify/cli/issues/847

**- Test plan**
* New tests added.
* Run `netlify dev` in your project

* Add `framework = "#custom"` option for `netlify.toml`'s `dev` section
* Don't allow usage of `--dir` flag with `command` or `targetPort` config options
* Add more test
* Throw lots of errors for invalid config

👻


erquhart:
It sounds like our heuristics are getting thrown off. Do we know specific reasons why inferred build command and port didn’t work for the linked issues?

It seems ideal to address this by being very clear about what configuration has been inferred. If we do go the disabling route, consider adding an option for disabling heuristics (would need a sensible name not involving the word “heuristics”) rather than accepting a special framework name.

RaeesBhatti:
**Highlights**
* Fixing the confusion around detectors and `command`, `port`, `targetPort` options
* When we're running with custom `command` and `targetPort` we don't know at the stage of redirect routing if it's a static server or a framework server, passing down value of framework and making sure it's present even when someone is using custom framework enables us to route to correct endpoint
* Making absolutely sure that the user can't get Netlify Dev config wrong. We're throwing errors with helpful and instructive messages to tell the users when they have a problematic config
* The static server settings were being overwritten and this was leading to confusion, we're making sure that the critical settings can't be overwritten in a way that causes confusion
* Mores tests, both unit tests and integration tests with larger overview. We're also testing `create-react-app` now to make sure that config works perfectly in a dynamic app, not just a static one.

PR #853, opened 2020-04-23
Upgrade Netlify Build

ehmicky:
This upgrades `@netlify/build` and `@netlify/config` to the latest version.

One major change is that the `build.lifecycle` property is now deprecated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -503,12 +503,12 @@
       }
     },
     "@netlify/build": {
-      "version": "0.3.6",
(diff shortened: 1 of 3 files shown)


RaeesBhatti:
It seems like the same test is failing on Windows again and again. Can you take a look!

ehmicky:
I have opened #854 for this. The tests are failing on Windows, but this does not appear to be related to this PR.

ehmicky:
Looking into it now.

ehmicky:
Fixed!

Issue #847, opened 2020-04-21
Unsupported value provided for "framework" option in config


pawel2105:
**- Do you want to request a _feature_ or report a _bug_?**

Bug

**- What is the current behavior?**

Running `netlify dev` with a framework config added to `netlify.toml` results in the following error:

```
Error: Unsupported value provided for "framework" option in config
    at module.exports.serverSettings (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/src/utils/detect-server.js:59:30)
    at DevCommand.run (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/src/commands/dev/index.js:413:26)
    at async DevCommand._run (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/node_modules/@oclif/command/lib/command.js:42:20)
    at async Config.runCommand (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/node_modules/@oclif/config/lib/config.js:172:9)
    at async Main.run (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/node_modules/@oclif/command/lib/main.js:21:9)
    at async Main._run (~/.nvm/versions/node/v13.13.0/lib/node_modules/netlify-cli/node_modules/@oclif/command/lib/command.js:42:20)
```

**- If the current behavior is a bug, please provide the steps to reproduce.**

My `netlify.toml` file looks like:

```
[build]
  publish = "out"
  command = "node_modules/next/dist/bin/next build"
  environment = { NODE_ENV = "production" }

[dev]
  publish = "out"
  command = "node_modules/next/dist/bin/next build"
  environment = { NODE_ENV = "production" }
  port = 8888
  targetPort = 3000
  framework = "next.js"
```

**- What is the expected behavior?**

framework config should not be throwing an error.

**- Local Environment Information**

```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  Binaries:
    Node: 13.13.0 - ~/.nvm/versions/node/v13.13.0/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v13.13.0/bin/npm
  Browsers:
    Chrome: 80.0.3987.163
    Firefox: 75.0
[...]

PR #857, opened 2020-04-24
Dev framework matching and functions:invoke bug fixes

MuffinTheMan:
**- Summary**

Fixes https://github.com/netlify/cli/issues/847
I believe this might also fix https://github.com/netlify/cli/issues/441

When running `netlify dev` with `framework = "svelte"` in the `[dev]` section of `netlify.toml`, I got the error: `Unsupported value provided for "framework" option in config`.

It turns out the logic would've required me to use `framework = "svelte.js.js"`, because `.js` was being appended to the file name we wanted to check `"svelte"` against instead of appending `.js` to `"svelte" to check against the file name. This has been fixed.

I also noticed that the `svelte.js` settings weren't actually being used (I noticed because I couldn't run `netlify dev`, since I wasn't providing a `targetPort`, which should've been pulled from `proxyPort` in `svelte.js`). I ended up merging the defaults from `svelte.js` (or the corresponding detector for whichever framework is being used) into `settings` to make use of these defaults.

Finally, I noticed that parameters were not being passed properly into the `serverSettings` function from `FunctionsInvokeCommand.run()`, which was resulting in NPEs in my situation when calling `serverSettings` (when running `netlify functions:invoke`). The proper parameters (hopefully) are now being passed, and `console.log` is being sent through as the `log` function (seemed like the simplest solution at this time).

**- Test plan**

Run `netlify dev` with some option `framework = "svelte"` (or some other valid framework) and verify the dev server is run and can be reached properly.

In my case, `netlify.toml`:

```
[build]
    command = "npm run build"
    functions = "./functions"
    publish = "public"

[dev]
    framework = "svelte"
    command = "npm run start"
    publish = "public"
```

Verify you can invoke a function via `netlify function:invoke` with some test function like:

```
exports.handler = function(event, context) {
  return { statusCode: 200, body: 'hello' }
}
```

**Note:** this isn't working for me (but it makes it further than it did before the fixes in this PR, since it at least doesn't die from an NPE); there appears to be some other bug that I would appreciate feedback on so I can perhaps track it down. I *can* run the function properly by directly hitting the localhost url when running `netlify dev`. Here's what I get:

```
? Invoke with emulated Netlify Identity authentication headers? (pass --identity/--no-identity to override) No // Note: this also fails with Yes
ran into an error invoking your function
FetchError: request to http://localhost:8888/.netlify/functions/test-function failed, reason: connect ECONNREFUSED 127.0.0.1:8888
    at ClientRequest.<anonymous> (/Users/caleb/git-other/cli/node_modules/node-fetch/lib/index.js:1455:11)
    at ClientRequest.emit (events.js:209:13)
    at Socket.socketErrorListener (_http_client.js:406:9)
    at Socket.emit (events.js:209:13)
    at emitErrorNT (internal/streams/destroy.js:91:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:59:3)
    at processTicksAndRejections (internal/process/task_queues.js:77:11) {
  message: 'request to http://localhost:8888/.netlify/functions/test-function failed, reason: connect ECONNREFUSED 127.0.0.1:8888',
  type: 'system',
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED'
}
```

netlify dev framework config and function:invoke bug fixes

close-up-photography-of-fawn-pug-covered-with-brown-cloth-374898


src/commands/functions/invoke.js
```diff
@@ -41,7 +41,8 @@ class FunctionsInvokeCommand extends Command {
       process.exit(1)
     }
 
-    let settings = await serverSettings(Object.assign({}, config.dev, flags))
+    // Pass in console.log as the logger for now.
+    let settings = await serverSettings(config.dev, flags, console.log)
```
MuffinTheMan, line 45: I could be wrong on this; but it didn't appear that `config.dev` and `flags` should be getting merged here and passed as the `devConfig` argument to `serverSettings` when it takes in `flags` as the second argument. Also, we'll get an NPE with nothing passed for the third argument (`log`).

Also, it appears there will be some conflicts with https://github.com/netlify/cli/pull/843.

src/utils/detect-server.js
```diff
@@ -55,14 +55,17 @@ module.exports.serverSettings = async (devConfig, flags, log) => {
   } else if (devConfig.framework === '#static') {
     settings.framework = devConfig.framework
   } else {
-    const detectorName = detectorsFiles.find(dt => `${dt}.js` === devConfig.framework)
-    if (!detectorName) throw new Error('Unsupported value provided for "framework" option in config')
-
+    // detectorsFiles end in '.js`; but devConfig.framework shouldn't. To test for match, append '.js' to devConfig.framework.
+    const detectorName = detectorsFiles.find(dt => dt === `${devConfig.framework}.js`)
+    if (!detectorName) throw new Error(`Unsupported value '${devConfig.framework}' provided for "framework" option in config`)
```
MuffinTheMan, line 60: It appears `.js` was being appended to the wrong value here. Also, I added the actual unsupported value to the error so it's a more helpful error (hopefully).

src/utils/detect-server.js
```diff
@@ -55,14 +55,17 @@ module.exports.serverSettings = async (devConfig, flags, log) => {
...
+    
     const detector = loadDetector(detectorName)
     const detectorResult = detector()
     if (!detectorResult) throw new Error(`Specified "framework" detector "${devConfig.framework}" did not pass requirements for your project`)
 
     settings.args = chooseDefaultArgs(detectorResult.possibleArgsArrs)
+    // Merge detectorResult default settings into settings Object.
+    Object.assign(settings, detectorResult)
```
MuffinTheMan, line 68: Not sure if this is right; but it didn't appear we were really using the `detectorResult` fully (`chooseDefaultArgs()` looks like it doesn't pull in all the defaults that we want).

MuffinTheMan:
I'm not sure why the checks are failing, because `npm test` shows all pass when I run locally. Could it be that I pushed another commit before the initial checks had completed? Is there a way to trigger them again without pushing another change?

JonSullivanDev:
Confirm that I've seen most of these things pop up tonight after updating as well. Ended up rolling back to 2.46.0 because even when spoofing the framework `jekyll.js.js` it was still erroring in trying to call `flags.dir` a little lower in that `detect` file. 2.46 is working great for me😅 just need to manually invoke some functions tonight 😜

erezrokah:
Is this still relevant? Looks like both of the issues in this PR are resolved by https://github.com/netlify/cli/pull/843
@RaeesBhatti ?

RaeesBhatti:
Yes, we can close this. Thanks for the hints @MuffinTheMan

PR #864, opened 2020-04-27
Use .app for the site's name suggestion message.

afuh:
**- Summary**
Updates the default domain for the site's name suggestion message during the creation process.


RaeesBhatti:
Thanks for the PR @afuh

PR #865, opened 2020-04-27
fix the dev server issue when proxy server returns 304 while redirect status is force set to 200

bencao:
**- Summary**

currently if we have a redirect rule in netlify.toml with
force = true and status = 200
the response from proxy server will always be 200
this behavior is problematic when the proxy returns 304 with empty body
the browser will always render a blank page
the change proposed in this PR is to leave 3xx code back to the browser and never override

**- Test plan**

Before: Page Request return Status 200 with empty Body

Screen Shot 2020-04-27 at 1 43 26 PM
Screen Shot 2020-04-27 at 1 43 34 PM

After: Page Request return Status 304

Screen Shot 2020-04-27 at 1 42 10 PM

when proxy server returns status code between 300 and 399, the return status will honor proxy server status instead of using the one designated in redirection rule.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -132,7 +132,12 @@ function initializeProxy(port, distDir, projectDir) {
     if (!isEmpty(pathHeaderRules)) {
       Object.entries(pathHeaderRules).forEach(([key, val]) => res.setHeader(key, val))
     }
-    res.writeHead(req.proxyOptions.status || proxyRes.statusCode, proxyRes.headers)
+    // leave 3xx redirects handling to browser
+    if (proxyRes.statusCode >= 300 && proxyRes.statusCode < 400) {
+      res.writeHead(proxyRes.statusCode, proxyRes.headers);
+    } else {
+      res.writeHead(req.proxyOptions.status || proxyRes.statusCode, proxyRes.headers)
+    }
     proxyRes.on('data', function(data) {
       res.write(data)
     })


src/commands/dev/index.js
```diff
@@ -132,7 +132,12 @@ function initializeProxy(port, distDir, projectDir) {
     if (!isEmpty(pathHeaderRules)) {
       Object.entries(pathHeaderRules).forEach(([key, val]) => res.setHeader(key, val))
     }
-    res.writeHead(req.proxyOptions.status || proxyRes.statusCode, proxyRes.headers)
+    // leave 3xx redirects handling to browser
+    if (proxyRes.statusCode >= 300 && proxyRes.statusCode < 400) {
```
bencao, line 136: In fact, I'm considering we probably should always return the proxyRes.statusCode, no matter it is 3xx or 4xx or 5xx.

RaeesBhatti: Thank you for opening this PR @bencao. The reason we have `req.proxyOptions.status` is to force a certain status code, for example, `200!` or `404!`

bencao: configuration like this is a bit vague, what if the api server returns http 500?

```
[[redirects]]
  from = "/search"
  to = "https://api.mysearch.com"
  status = 200
  force = true
```

bencao: as a user, my actual intention is basically requesting a server side transparent proxy, not enforcing the status code to 200

RaeesBhatti: I think we would still enforce `200` as status

bencao: ok.. do you have some ideas about how we can gracefully solve the above proxy server 304 resulted in blank page issue?

RaeesBhatti: Can you try out this branch https://github.com/netlify/cli/pull/843 and see if that fixes the problem for you!

bencao: I checked the branch, it seems to be a bit too complicated to solve my very specific problem..

bencao:
close and waiting for a better approach

Issue #884, opened 2020-05-08
30x redirects are broken


RaeesBhatti:
**- Do you want to request a _feature_ or report a _bug_?**
Bug
**- What is the current behavior?**
The following URL redirect doesn't work for project using a framework.
```
/someurl        /otherurl
```
**- If the current behavior is a bug, please provide the steps to reproduce.**
Originally [posted on the community](https://community.netlify.com/t/netlify-dev-error/13735/22?u=raeesbhatti).
**- What is the expected behavior?**
The redirect rule above works correctly with framework projects.
**- Local Environment Information**

```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: macOS 10.15.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 10.20.1 - ~/.n/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.n/bin/npm
  Browsers:
    Chrome: 81.0.4044.138
    Firefox: 76.0
    Safari: 13.1
  npmGlobalPackages:
    netlify-cli: 2.50.0

netlify-cli/2.50.0 darwin-x64 node-v10.20.1
```

Issue #848, opened 2020-04-21
netlify dev: default 404 page handling, file/url shadowing not functioning


moop-moop:
**- Do you want to request a _feature_ or report a _bug_?**

bug

**- What is the current behavior?**

When using `netlify dev` the documented default 404 and file shadowing behavior does not work.

- https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling

- https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing

**- If the current behavior is a bug, please provide the steps to reproduce.**

1. Create a simple static site folder with a 404.html at the root and \index.html, \foo.html, \foo\bar.html or \not-foo\index.html or similar file patterns.
2. create a `_redirects` files with:
`/foo /not-foo 200`
3. Test 404, go to any non-existent url (http://localhost:8888/nothinghere) and you will receive a  default 404 status, but not the \404.html
4. Test file/url shadowing, go to http://localhost:8888/foo and you will receive the page from \not-foo\index.html
5. Add `/foo.html /not-foo 200` to the `_redirects` file
6. Test file/url shadowing, go to http://localhost:8888/foo.html and you will still receive the page from \not-foo\index.html

**- What is the expected behavior?**

1. the \404.html page should be used for 404 status as per documentation
2. when a request matches an existing static file, that should be returned and the proxying rule ignored as per documentation

**- Local Environment Information**

────────────┐
 Environment Info
────────────┘
  System:
    OS: Windows 10 10.0.17763
    CPU: (12) x64 Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
  Binaries:
    Node: 12.16.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.831.0

netlify-cli/2.47.0 win32-x64 node-v12.16.2

────────────┐ 
        netlify.toml
────────────┘
```
[dev]
  functions = "functions"
  publish = "public"
  framework= "#static"
```

RaeesBhatti:
Thank you for reporting this @moop-moop and providing references. There's a PR to fix this behavior https://github.com/netlify/cli/pull/885. If you can, please try it out and let us know there's anything missing or wrong!

PR #885, opened 2020-05-08
More redirect and rewrite fixes

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/884
Fixes: https://github.com/netlify/cli/issues/848
Fixes: https://github.com/netlify/cli/issues/890

**- Test plan**
Added new integration tests for these cases.

* Improve shadowing on request url
* Implement shadowing on destination URL's
* Fix condition for redirect rule processing when using a framework
* Add logic for 404 rewrites and shadowing
* Fix function event body when request has no body
* Add a bunch of tests

🐣


RaeesBhatti:
Seems like partially blocked on https://github.com/netlify/bitballoon/issues/5624

src/utils/detect-server.js
```diff
@@ -126,7 +126,7 @@ module.exports.serverSettings = async (devConfig, flags, projectDir, log) => {
   settings.port = port
 
   settings.jwtRolePath = devConfig.jwtRolePath || 'app_metadata.authorization.roles'
-  settings.functionsPort = await getPort({ port: settings.functionsPort || 34567 })
```
RaeesBhatti, line 129: When running two instance of `netlify dev` around the same time `getPort` things that port `34567` is available which causes one of them to fail.
Falling back to `0` port makes sure that we get a random port for `functionPort` on every instance.

src/commands/dev/index.js
```diff
@@ -283,8 +285,9 @@ async function serveRedirect(req, res, proxy, match, options) {
     return render404(options.publicFolder)
   }
 
-  if (match.force || (!(await isStatic(reqUrl.pathname, options.publicFolder) || options.framework) && match.status !== 404)) {
```
RaeesBhatti, line 286: There are two main changes here. `|| options.framework` should've been `&& options.framework` which was a mistake and we've added tests to catch it in the future. And `match.status !== 404` shouldn't really be here. I'm not sure why exactly it is here because we want to overwrite `status` for `404` rewrites but we're added tests to make sure the intended functionality is preserved.

src/commands/dev/index.js
```diff
@@ -273,6 +273,8 @@ async function serveRedirect(req, res, proxy, match, options) {
     }
   }
 
+  const staticFile = await getStatic(req.url, options.publicFolder)
+  if (staticFile) req.url = staticFile
```
RaeesBhatti, line 277: This is done to implement https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing behavior on request URL's. If there is a static file, we want to rewrite request URL to match that request. Since we already have `match` object which contains `force` and `to`, we don't need to worry about destination at this point.

src/commands/dev/index.js
```diff
@@ -45,14 +45,14 @@ function addonUrl(addonUrls, req) {
...
-async function isStatic(pathname, publicFolder) {
+async function getStatic(pathname, publicFolder) {
   const alternatives = alternativePathsFor(pathname).map(p => path.resolve(publicFolder, p.substr(1)))
 
   for (const i in alternatives) {
     const p = alternatives[i]
     try {
       const pathStats = await stat(p)
-      if (pathStats.isFile()) return true
+      if (pathStats.isFile()) return '/'+path.relative(publicFolder, p)
```
RaeesBhatti, line 55: Instead of just checking for static file, we're retuning the path to file that exists, so that, we can use it `serveRedirect` function.

src/utils/serve-functions.js
```diff
@@ -121,27 +121,24 @@ function createHandler(dir) {
...
-      .split(remoteAddress.includes('.') ? ':' : ',')
-      .pop()
-      .trim()
+    let remoteAddress = request.get('x-forwarded-for') || request.connection.remoteAddress || ''
+    remoteAddress = remoteAddress.split(remoteAddress.includes('.') ? ':' : ',').pop().trim()
 
     let requestPath = request.path
-    if (request.headers['x-netlify-original-pathname']) {
-      requestPath = request.headers['x-netlify-original-pathname']
+    if (request.get('x-netlify-original-pathname')) {
```
RaeesBhatti, line 132: A better way to access headers that makes sure that our access is case-insensitive.

src/utils/serve-functions.js
```diff
@@ -121,27 +121,24 @@ function createHandler(dir) {
     }
     const { functionPath } = functions[func]
 
-    const body = request.body.toString()
-    var isBase64Encoded = Buffer.from(body, 'base64').toString('base64') === body
+    const body = request.get('content-length') ? request.body.toString() : undefined
```
RaeesBhatti, line 124: This makes sure that we're not sending `body` if there is no body. Calling `request.body.toString()` was causing even empty body to produce `[Object]` as string.

tests/dev.test.js
```diff
@@ -57,24 +58,155 @@ test('netlify functions:invoke', async t => {
...
 
-test('netlify dev: functions env file', async t => {
+test('functions env file', async t => {
   const response = await fetch(`http://${host}:${port}/.netlify/functions/env`).then(r => r.text())
 
   t.is(response, 'true')
 })
 
-test('netlify dev: functions env file overriding prod var', async t => {
+test('functions rewrite echo without body', async t => {
```
RaeesBhatti, line 67: Adds tests for `/api/*` to `/.netlify/functions/:splat` with a static server.

src/utils/serve-functions.js
```diff
@@ -121,27 +121,24 @@ function createHandler(dir) {
...
+      requestPath = request.get('x-netlify-original-pathname')
       delete request.headers['x-netlify-original-pathname']
     }
 
     const event = {
       path: requestPath,
       httpMethod: request.method,
       queryStringParameters: queryString.parse(request.url.split(/\?(.+)/)[1]),
-      headers: Object.assign({}, request.headers, { 'client-ip': remoteAddress }),
+      headers: { ...request.headers, 'client-ip': remoteAddress },
```
RaeesBhatti, line 141: Updated syntax. Nothing changed.

src/utils/serve-functions.js
RaeesBhatti: Nothing changed here interns of functionality.

src/utils/serve-functions.js
```diff
@@ -121,27 +121,24 @@ function createHandler(dir) {
...
+    let isBase64Encoded = false
+    if (body) isBase64Encoded = Buffer.from(body, 'base64').toString('base64') === body
 
-    let remoteAddress =
-      request.headers['x-forwarded-for'] || request.headers['X-Forwarded-for'] || request.connection.remoteAddress || ''
-    remoteAddress = remoteAddress
-      .split(remoteAddress.includes('.') ? ':' : ',')
-      .pop()
-      .trim()
+    let remoteAddress = request.get('x-forwarded-for') || request.connection.remoteAddress || ''
```
RaeesBhatti, line 128: Using `header.get` to make sure we're getting the headers in case-insensitive way. Then some syntax improvement to make it more legible

src/utils/rules-proxy.test.js
```diff
@@ -19,11 +19,38 @@ test('parseFile: netlify.toml', async t => {
...
     },
     {
-      path: '/*',
-      to: '/index.html',
+      path: '/foo',
+      to: '/not-foo',
       status: 200,
       force: false,
     },
+    {
```
RaeesBhatti, line 27: Changed because new rules were added.

tests/cra.test.js
```diff
@@ -1,16 +1,12 @@
...
 const { spawn } = require('child_process')
+const url = require('url')
 const test = require('ava')
 const fetch = require('node-fetch')
-const mkdirp = require('mkdirp')
 const cliPath = require('./utils/cliPath')
 const { randomPort } = require('./utils/')
 const sitePath = path.join(__dirname, 'site-cra')
 
-const fileWrite = util.promisify(fs.writeFile)
```
RaeesBhatti, line 12: No need to write files every time since they're already in `git`

tests/cra.test.js
```diff
@@ -95,6 +100,50 @@ test('netlify dev cra: robots.txt', async t => {
   t.regex(body, /# https:\/\/www.robotstxt.org\/robotstxt.html/)
 })
 
+
+test('functions rewrite echo without body', async t => {
```
RaeesBhatti, line 104: New tests to make sure `/api/*` to `/.netlify/functions/:splat` work with framework servers as well.

PR #892, opened 2020-05-11
Update syntax for functions:invoke --payload

cdepman:
As written, gives an empty object as the body of the received event. With the fix, will produce the desired result.

**- Summary**

Was trying to test lambda function invocations locally, would not work with documentation as written.

**- Test plan**

create a simple `testFunction`, expect a certain <payload>, attempt to pass that payload with `netlify functions:invoke testFunction --payload <payload>`

Update documentation for CLI function invocation with --payload flag

🐼


RaeesBhatti:
Thanks for the PR @cdepman

PR #910, opened 2020-05-20
fix: use named Octokit import

jlengstorf:
**- Summary**

Octokit now uses named imports and prints a deprecation notice if you use the default import 

```
[@octokit/rest] `const Octokit = require("@octokit/rest")` is deprecated. Use `const { Octokit } = require("@octokit/rest")` instead
```

this change swaps out to the named import to remove this deprecation notice

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,4 +26,7 @@ jobs:
       - run: npm ci --prefix tests/site-cra
       - run: npm test
         env:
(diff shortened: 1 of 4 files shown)


erezrokah:
@RaeesBhatti, @jlengstorf I added another commit to sort out the tests.
A bit crude, but we can improve on it.
WDTY?

RaeesBhatti:
Thanks @erezrokah

PR #924, opened 2020-05-24
Improve dotenv support

JLarky:
**- Summary**

* previous implementation wasn't really useful as it was limited to just two
files `.env.development` and `.env` (and not even merging them)
* so we are bringing `.env.local` and `.env.developepment.local` file support and so on
* as well as `dotenv-expand` that can be pretty useful as well as just there
to be consistent with cra and dotenv-load
* most egregious of them was that if you have `ABC=0` in `.env` you won't be able
to change it by running `ABC=123 netlify dev` as it will continue to have
`process.env.ABC=0` inside `.netlify/functions`

**- Test plan**

I made it so it passes current unit tests, but the scenario that I was testing was netlify website with this kind of configuration:

```
# .env
FUNCTIONS_A="from .env"
FUNCTIONS_B="from .env"
FUNCTIONS_C="from .env"
```
 
```
# .env.local
FUNCTIONS_A="from .env.local"
FUNCTIONS_B="from .env.local"
```
 
```js
// functions/test.js

const headers = {
  "Access-Control-Allow-Origin": "*",
  "Access-Control-Allow-Headers": "Content-Type",
};

exports.handler = function (event, context, callback) {
  callback(null, {
    statusCode: 200,
    headers: headers,
    body: JSON.stringify(getDebugEnv()),
  });
};

function getDebugEnv() {
  return Object.keys(process.env)
    .filter(key => key.startsWith("FUNCTIONS_"))
    .reduce((acc, key) => ({ ...acc, [key]: process.env[key] }), {});
}
```

Now if we run

    FUNCTIONS_A="from cli" netlify dev

it will print

    ◈ Overriding the following env variables with .env.local,.env files: FUNCTIONS_A,FUNCTIONS_B,FUNCTIONS_C

and `curl http://localhost:3200/.netlify/functions/hash` produces

```json
{
  "FUNCTIONS_A": "from cli",
  "FUNCTIONS_B": "from .env.local",
  "FUNCTIONS_C": "from .env"
}
```

just to compare to what I was fixing, here's result that `v2.51.0` produces
```
{
  "FUNCTIONS_A": "from .env",
  "FUNCTIONS_B": "from .env",
  "FUNCTIONS_C":"from .env"
}
```

Improved support for `.env` files, now it behaves like [create-react-app](https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used) which means it will use files `.env.development.local`, `.env.development`, `.env.local` and `.env`; as well as it will expand environment variables

image


JLarky:
ok, I see all of those tests are failing :) it looks like this is because of `NODE_ENV=test` again, I'm open to suggestions on what is the right way to deal with it :)

JLarky:
And now it fails without any indications of why :) I can infer from the name that it wants me to add labels to RP, but I can't edit it.

JLarky:
I was working on a project that would have been benefited from you fixing this... maybe one more picture will help
image

erezrokah:
Sorry for the late follow up @JLarky. Will try to review this week.

JLarky:
> Sorry for the late follow up @JLarky. Will try to review this week.

no worries :) just wanted to hear at least something back :) let me know how I should update this PR so it can be merged :) should I rebase it on the latest master? :) as you can imagine last time I had this repo set up was half a year ago :) I don't want to spend too much time before knowing if this PR is going through or not :)

erezrokah:
> should I rebase it on the latest master

No worries, I'll do the rebase and ping you for approval

erezrokah:
@JLarky I followed up with some changes:
1. Dropped `dotenv-expand` as it [doesn't seem to be maintained](https://github.com/motdotla/dotenv-expand/pulls) and I couldn't find a way to use interpolation without mutating current `process.env`:
https://github.com/motdotla/dotenv-expand/blob/de9e5cb0215495452f475f5be4dea1580b8217cd/lib/main.js#L5
I'm open to adding this feature if another lib if available/in another PR
2. Changed the code to read all `.env` files in parallel.
3. Changed `vars` to be the result of `Object.entries` to make it easier to use.

To be done:
1. Add some more tests.
2. Update the logs as I believe they don't reflect the code anymore.

JLarky:
> 1. Dropped `dotenv-expand` as it [doesn't seem to be maintained](https://github.com/motdotla/dotenv-expand/pulls) and I couldn't find a way to use interpolation without mutating current `process.env`:
>    https://github.com/motdotla/dotenv-expand/blob/de9e5cb0215495452f475f5be4dea1580b8217cd/lib/main.js#L5
>    I'm open to adding this feature if another lib if available/in another PR

This is to be consistent with CRA and dotenv-load. https://www.npmjs.com/package/react-scripts?activeTab=dependencies so my opinion is "once those packages use something else you can change it to that thing to be consistent"

> 2. Changed the code to read all `.env` files in parallel.

as long as it doesn't mess with the order I guess? Do you plan to move your changes upstream? :) https://github.com/facebook/create-react-app/blob/5bd6e73047ef0ccd2f31616255c79a939d6402c4/packages/react-scripts/config/env.js

> 3. Changed `vars` to be the result of `Object.entries` to make it easier to use.

I have no opinion on that

> To be done:
> 
> 1. Add some more tests.

I only changed the test so they don't fail CI

> 2. Update the logs as I believe they don't reflect the code anymore.

I don't know what you mean

erezrokah:
> This is to be consistent with CRA and dotenv-load.

We can re-visit this in another separate PR. Don't think we should add an unmaintained library.

> Do you plan to move your changes upstream? :)

Oh, didn't realise this code was based on that. I think it would be not that straightforward as the code there is synchronous to begin with. 

> I only changed the test so they don't fail CI

I'm going to add those

> I don't know what you mean

Sorry, clarifying - we are no longer overriding env variables as stated here:
https://github.com/netlify/cli/pull/924/files#diff-dd36b1f406f5fb77839467571252f745R29
we're just sourcing/adding them.

JLarky:
> Don't think we should add an unmaintained library.

as opposed to this one?

erezrokah:
Thank you for your feedback, @JLarky. I'm going to follow up with these changes in another PR as I believe the current handling of `.env` files should be improved

PR #944, opened 2020-06-07
Update Functions Docs to Include Functions Domain

alteredorange:
**- Summary**

I was unaware the functions were actually being served through the browser, I thought you had to interact with them through the command line.  While trying to find a way to interact with them through the browser I read the netlify-lambda docs and realized that netlify dev does serve the files in the browser, it's just not mentioned anywhere.

**- Test plan**
No need for tests, just updating the docs.

Include url where functions are served (/.netlify/functions/your-function)


RaeesBhatti:
Thank you for the PR @alteredorange!

PR #950, opened 2020-06-19
Get functionsPort from devConfig

luis-agm:
**- Summary**

In an app using react with webpack-dev-server I noticed the functionsPort option in dev config was being ignored, and apparently it was trying to set it from an undefined value and not from the dev config itself.

**- Test plan**

 * Have `functionsPort` in `netlify.toml [dev]` block.

Before:
Screenshot from 2020-06-18 22-53-55

After:

Screenshot from 2020-06-18 22-54-59

 * Get functionsPort config from the dev config block in `netlify.toml` 

doggos

### DIFF
--- a/src/utils/detect-server.js
+++ b/src/utils/detect-server.js
@@ -154,7 +154,7 @@ module.exports.serverSettings = async (devConfig, flags, projectDir, log) => {
   settings.port = port
 
   settings.jwtRolePath = devConfig.jwtRolePath || 'app_metadata.authorization.roles'
-  settings.functionsPort = await getPort({ port: settings.functionsPort || 0 })
+  settings.functionsPort = await getPort({ port: devConfig.functionsPort || 0 })
   settings.functions = devConfig.functions || settings.functions
 
   return settings


RaeesBhatti:
@luis-agm `functionsPort` configurability was disabled intentionally. Can you please explain your use case for making this configureable!

luis-agm:
@RaeesBhatti I guess the problem came from another one. I noticed that I was not able to reach my functions from my app, because the request was being made to the same port of the app, and since the functions port appeared to be random it was hard for me to implement a manual proxy. But I noticed `netlify dev` is supposed to be proxying the requests, so if that's the case then I really see no use case for configuring the functionsPort. I just wonder why that is not working since the requests are being made to a url that matches the RegExp made to identify them. Also the output from `webpack-dev-server` seems to override the `netlify dev` output so old-school debugging is hard, and the `port` option seems to be ignored as the application is only accessible through the port I set on webpack-dev-server config.

RaeesBhatti:
@luis-agm The usage of `port` and `targetPort` is explained in [Netlify Dev docs page](https://cli.netlify.com/netlify-dev#explanation-of-ports-in-netlify-dev). In your case, if `webpack-dev-server` is running on port `3000`, you should set `targetPort = 3000` and `port = 8888`. Then you'll be able to access Netlify Dev on `http://localhost:8888/`.
In short, `targetPort` is the port your site generator is running on. `port` is the port you want to access Netlify Dev on.

Issue #859, opened 2020-04-26
Bug: deploy command regression


DavidWells:
The `netlify deploy` command used to deploy one of my sites has stopped working in `netlify-cli@2.48.0` 

The command `netlify deploy` fails with a "Base directory does not exist" error.

```
> netlify deploy -p --dir public

Base directory does not exist: /Users/davidwells/David/analytics-project/analytics/site/main/site
```

Was something changed recently with the `netlify deploy` command?

This is the site in question: https://github.com/DavidWells/analytics/tree/master/site/main and the previously working deploy script is here:  https://github.com/DavidWells/analytics/blob/master/site/main/package.json#L14

DavidWells:
It looks like this is an error coming from netlify/config https://github.com/netlify/build/blob/b6006915061ecd890a7d4f8b857cddc22e778f1a/packages/config/src/files.js#L65 cc @ehmicky

DavidWells:
Reverting to `netlify-cli@2.40.0` fixes this error

ehmicky:
Just to confirm, I think the error message should be `
Base directory does not exist: /Users/davidwells/David/analytics-project/analytics/site/main/public`, not `
Base directory does not exist: /Users/davidwells/David/analytics-project/analytics/site/main/site` if you passed `--dir public`, please confirm if that's not the case.

It looks like the issue is that the `public` directory does not exist and you're trying to deploy it. Did you run `npm run build` first?

DavidWells:
The build folder is there locally. This is a regression in how the CLI works. Possibly related to how this site uses yarn workspaces? IDK

`netlify deploy -p --dir public` should deploy my site, instead it results in "Base directory does not exist: /xyz/site" Error with latest versions of CLI

ehmicky:
I am having issues reproducing this problem. When I make sure `npm run build` has been run first, `npm run deploy` seems to work fine for me with the latest version of `netlify-cli`. @RaeesBhatti Do you manage to reproduce this issue?

DavidWells:
Related https://github.com/netlify/cli/issues/862

RaeesBhatti:
@DavidWells Can you try the following branch! see if you can still reproduce this issue there: https://github.com/netlify/cli/pull/843

timchambers:
I'm also hitting this now. Did not occur in 2.41.0, does occur in 2.42.0 and later. I'm also using yarn workspaces with this repository, in case that helps, since @DavidWells mentiojned he is as well.

[...]

PR #953, opened 2020-06-24
Improve how the project root is resolved

ehmicky:
As described in https://github.com/netlify/cli/issues/859, this improves how the project root is resolved, by moving the `findRoot` logic to `@netlify/config` instead.

Note: this PR is a WIP since it requires https://github.com/netlify/build/pull/1579 and https://github.com/netlify/build/pull/1580 to be merged first. This is why the tests are currently failing.


src/utils/command.js
```diff
@@ -23,20 +22,19 @@ class BaseCommand extends Command {
...
     // Grab netlify API token
     const authViaFlag = getAuthArg(argv)
 
     const [token] = this.getConfigToken(authViaFlag)
 
+    const cachedConfig = await this.getConfig(cwd, state, token)
+    const { configPath, config, repositoryRoot: projectRoot } = cachedConfig
+
     // Get site id & build state
     const state = new StateConfig(projectRoot)
```
erezrokah, line 36: Let's say I have the following directory structure:
```
repo/
  .git/
  site/
    .netlify/state.json
    netlify.toml
```

`repositoryRoot` will evaluate to `repo`, am I correct?
That makes sense.

But `StateConfig` tries to read `<projectRoot>/.netlify/state.json` which will fail.

The CLI seems to link a folder to a site (and not a project), thus I'll need to link both `repo` and `repo/site` folders for that to work.

Looks like we should separate concerns between evaluating the repository root and finding the location of `state.json`

ehmicky: This should now work with https://github.com/netlify/build/pull/1581.

Re-writing the PR accordingly right now.

ehmicky:
I have updated this PR based on https://github.com/netlify/cli/issues/859#issuecomment-649639490.

ehmicky:
There seem to be still be some issues, debugging right now.

ehmicky:
Problems fixed!

erezrokah:
Tested and the fix works.
Also - the code looks much simpler now, thanks @ehmicky!

JamesMcMahon:
Does this change result in breaking existing configuration?

My configuration, which was:
```
[build]
  publish = "build"
```

Broke post netlify cli, as it no longer is checking for `build` relative to where I am running the deploy command. It now expects `build` to be off of the repo root.

Not sure if this was the change that broke it, but looking at the changes it looked like the most likely culprit.

ehmicky:
Hi @JamesMcMahon,

Could you please print your logs when running the Netlify CLI, with the environment variable `NETLIFY_BUILD_DEBUG` set to `true`? Thanks!

FinnWoelm:
@JamesMcMahon, I ran into the same issue. I solved it by creating an empty `.git` directory in my subfolder. For some reason, that worked.

My files:

```
+ .git/
+ node_modules/
+ package.json
+ file1.js
+ file2.js
+ subfolder/
  + .git/    <-- completely empty directory
  + package.json
  + netlify.toml
  + publish/
  + ...
```

My `netlify.toml` configuration:

```
[build]
  publish = "publish"
```

My subfolder package.json:
```
{
  "name": "test",
  "scripts": {
    "deploy": "../../../node_modules/.bin/netlify deploy"
  }
}
```

Perhaps it helps someone :blush:

JamesMcMahon:
Just to clarify, I had no problems fixing the issue. I was reporting it hear because it was a breaking change that got bundled into a seem-lying innocuous minor release.

In my case, I fixed it by changing

```toml
[build]
publish = "build"
```

To:

```toml
[build]
publish = "frontend/build"
```

This is maybe the wrong place to report this, also maybe kind of pointless given that release has already happened. I wanted to record the change for posterity sake.

@ehmicky If you still need any of that information let me know.

ehmicky:
@JamesMcMahon If you could post any of the information listed above, this could fix this problem for other users without them having to modify their `netlify.toml` nor add a `.git` directory. Otherwise, I'm glad your problem is solved :)

Issue #745, opened 2020-03-11
CLI seems to use only synchronous deploys even for sites that are over the "syncfileslimit" and this can lead to upload failures


fool:
**- Do you want to request a _feature_ or report a _bug_?**

bug

**- What is the current behavior?**

currently, the CLI always attempts to deploy synchronously, which even for sites with only a few thousand files, can lead to 502 gateway errors during upload, and this causes deploys to fail.

**- If the current behavior is a bug, please provide the steps to reproduce.**

Try to deploy a site with 10k new files via CLI.  It may occasionally work, but often will not.  This has been reproduced with sites as small as 4500 files and as big as 60k files.  This happens even with high `-p X` values to set a large timeout.

The setting that was proposed to change that controls this:

https://github.com/netlify/js-client/blob/master/src/deploy/index.js#L25

does not seem to support being changed (or changes are ignored) as we've tried setting it to smaller numbers including 1 with the behavior not changing according to our internal logs AFAICT.

**- What is the expected behavior?**

Deployment works and is handled asynchronously on our servers to allow success no matter how large the site.

**- Local Environment Information**
Happens with v2.37.0 on linux

** Additional netlify-private context**
There was a long thread discussing this here in slack:  https://netlify.slack.com/archives/C0QFKKY2U/p1582742724079200 .  See especially comments from @rybit and @biilmann that we should consider having the CLI use "always async"

fool:
This has affected at least one enterprise and one pro customer:

- https://netlify.zendesk.com/agent/tickets/18594 (enterprise)
- https://netlify.zendesk.com/agent/tickets/20209 (Pro)

RaeesBhatti:
Thanks for reporting this @fool. It's in the backlog we'll take a look at this soon.

RaeesBhatti:
@ingride Can you please elaborate on the `async` property in [`createSiteDeploy`](https://open-api.netlify.com/#operation/createSiteDeploy) API call. What does that mean exactly?

RaeesBhatti:
Seems like partially blocked on https://github.com/netlify/bitballoon/issues/5624

stephencweiss:
@RaeesBhatti - is #5624 supposed to be publicly viewable? If so, it might be a broken link (or it is for me at a minimum :)).

Thanks!

RaeesBhatti:
@stephencweiss That is an internal issue with out API engine. Please hold on this problem is on our priority list.

stephencweiss:
[...]

PR #966, opened 2020-06-29
Deploy: Force async on large deployments

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/745

**- Test plan**
Test manually with large scale sites that fail with timeout

Set deployment `syncFileLimit: 1` to force an async deployment

🐠


RaeesBhatti:
Reasoning is explained here: https://netlify.slack.com/archives/C9D3B8RGF/p1593767386073000

Issue #904, opened 2020-05-13
netlify dev opens up node instances seemingly recursively until PC is shut down.


EmNudge:
**- Do you want to request a _feature_ or report a _bug_?**
report a bug
**- What is the current behavior?**
`netlify dev` will open node instances every second or so until the computer is shut down. CPU is maxed out within seconds.
**- If the current behavior is a bug, please provide the steps to reproduce.**
I honestly do not know how to reproduce it as I have tried and failed, but it acts as such on a project of mine [here](https://github.com/EmNudge/Open-Fortress-Site). I was told to post a bug here after posting on netlify's community with no resolution.
I moved to just use dummy data while developing as it seems to run fine when Netlify runs the site.
**- What is the expected behavior?**
to not open up node instances. My best guess is that this is some kind of recursive bug where something is being checked over and over again, but I really don't know much else.
**- Local Environment Information**
```
  System:
    OS: Windows 10 10.0.18362
    CPU: (16) x64 AMD Ryzen 7 2700X Eight-Core Processor
  Binaries:
    Node: 12.8.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.10.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0

netlify-cli/2.50.0 win32-x64 node-v12.8.1
```

RaeesBhatti:
Thank you for opening this ticket @EmNudge. I tried to run your site locally but couldn't because there's a Webpack configuration problem with netlify-lambda. It could be that your Webpack configuration is not committed to Git.
```
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be an non-empty object.
   -> Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.
```

erezrokah:
I was able to reproduce the issue by reverting https://github.com/EmNudge/Open-Fortress-Site/commit/02eac5461f69aca5194145e630bb8976cc057f76.
Looks like that commit moves the functions to sub directories which is not supported by `netlify-lambda`.

After reverting the commit and running `netlify dev` it seems that there are multiple processes of `npm run lambda`

PR #974, opened 2020-07-01
fix: wait for chokidar ready event before listening to function builder src directory

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/904

**- Test plan**

You can follow my comment [here](https://github.com/netlify/cli/issues/904#issuecomment-652421285) to reproduce

[chokidar](https://github.com/paulmillr/chokidar) sends an initial `add` event for each existing file in the watched directory.
In the case of the bug, that event was fired for each file under `functions/node_modules` spawning multiple build commands.

You can ignore those initial events by waiting for the `ready` event or passing `ignoreInitial`:
https://github.com/paulmillr/chokidar#path-filtering

Also debounced the build command to make sure only one instance of it executes at a single point in time.

🐈

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10048,8 +10048,7 @@
     "lodash.debounce": {
       "version": "4.0.8",
       "resolved": "https://registry.npmjs.org/lodash.debounce/-/lodash.debounce-4.0.8.tgz",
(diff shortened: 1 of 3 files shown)


RaeesBhatti:
Thank you for this PR @erezrokah

erezrokah:
My pleasure 😄

PR #975, opened 2020-07-03
Use base64 encoding for binary types

ruiramos:
This might not be an exhaustive list of all the mimetypes we'll have to support, but it solves the problem I was running into (uploading images through a Netlify function locally using `netlify dev`)

**- Summary**

I've been trying to use a Netlify function as an image/file uploading proxy and running into problems, both locally using `netlify dev` and in production.

I managed to fix the problems in production after looking at [this thread](https://community.netlify.com/t/baffled-by-binary-event-body-in-lambda-function/17395). I believe this is something you're currently looking into.

Going over the `netlify-cli` code, I saw there was no way of changing the default 'utf8' decoding format for the request `body` Buffer, which won't work for binary data. As I say above, I'm unsure if there are any other mime-types we'll want to add to the list, I couldn't find a definitive list for this (if there is one). 

**- Test plan**

You can now upload images running `netlify dev` locally.

Uses base64 encoding for `body` Buffers when the `content-type` suggests binary data.

### DIFF
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -130,15 +130,13 @@ function createHandler(dir) {
     }
     const { functionPath } = functions[func]
 
-    const encoding = (request.headers["content-type"] || "").match(
+    const isBase64Encoded = !!(request.headers['content-type'] || '').match(
       /image|audio|video|application\/pdf|application\/zip|applicaton\/octet-stream/
-    )
-      ? "base64"
-      : "utf8";
+    );
 
-    const body = request.get('content-length') ? request.body.toString(encoding) : undefined
-    let isBase64Encoded = false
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "netlify-cli",
+  "name": "ruiramos-netlify-cli",
```
RaeesBhatti, line 2: Are you sure you mean to push this change to this PR?

ruiramos: Apologies @RaeesBhatti. 
This is actually blocking me so I'm trying to publish a version of the cli with this fixed, obviously I screwed up!
The PR should be good now.

RaeesBhatti:
Thank you for this PR @ruiramos!

Issue #970, opened 2020-06-30
querystring no longer parsed as expected


smakinson:
**- Do you want to request a _feature_ or report a _bug_?**
bug for netlify dev

**- What is the current behavior?**
a query string param such as: `category[SOMETHING][]=something` is not longer coming through properly so it can be proxied to another api. It looks like it converts to: `category: '[object Object]'`, stringfied: `category%21=%5Bobject%20Object%5D`

**- If the current behavior is a bug, please provide the steps to reproduce.**

**- What is the expected behavior?**
The query string param to remain untouched as it did in a prior recent version so it can be proxied locally as-is

**- Local Environment Information**

 System:
    OS: macOS 10.15.5
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  Binaries:
    Node: 12.15.0 - ~/.asdf/installs/nodejs/12.15.0/bin/node
    Yarn: 1.22.4 - ~/yarn/global/bin/yarn
    npm: 6.13.4 - ~/.asdf/installs/nodejs/12.15.0/bin/npm

netlify-cli/2.56.0 darwin-x64 node-v12.15.0

RaeesBhatti:
Thank you for reporting this issue @smakinson . This will be fixed by https://github.com/netlify/cli/pull/976

If you can please try out the PR and let us know if it fixes the problem for you!

smakinson:
@RaeesBhatti It looks like that fixes it. Thank you for doing this so quickly!

PR #976, opened 2020-07-03
Functions: Handle query parameters correctly

RaeesBhatti:
**- Summary**
Fixes: https://github.com/netlify/cli/issues/970
We were previously using `query` object provided by Express which turns out to parse URL query parameters incorrectly. The new approach uses native `URL` class to get the query parameters from the original request URL.

**- Test plan**
Test added

The new approach uses native `URL` class to get the query parameters from the original request URL.

🐥


RaeesBhatti:
Thanks for the tip @erezrokah

Issue #959, opened 2020-06-25
Pass the `--site` CLI flag to `@netlify/config`


ehmicky:
The `deploy` and `open` commands use a `--site` CLI flag. 

When defined, its value (instead of the `siteId` inside `.netlify/state.json`) should be [passed to `@netlify/config`](https://github.com/netlify/cli/blob/93840689c0e7dadbd441fb4a7729fd8da25b1ddb/src/utils/command.js#L81).

PR #1008, opened 2020-07-20
Pass site or siteId to @netlify/config

RaeesBhatti:
**- Summary**

Fixes: https://github.com/netlify/cli/issues/959

**- Test plan**

Use `siteId` or `site` flags as `siteId` as parameters to `resolveConfig`

🐠

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -32,7 +32,7 @@ class BaseCommand extends Command {
     // Get site id & build state
     const state = new StateConfig(cwd)
 
-    const cachedConfig = await this.getConfig(cwd, state, token)
+    const cachedConfig = await this.getConfig(cwd, state, token, argv)
     const { configPath, config, buildDir } = cachedConfig
 
     const apiOpts = {}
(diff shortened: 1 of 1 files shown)


src/utils/command.js
```diff
@@ -69,14 +69,14 @@ class BaseCommand extends Command {
...
-  async getConfig(cwd, state, token) {
+  async getConfig(cwd, state, token, argv) {
     try {
       return await resolveConfig({
         config: argv.config,
         cwd: cwd,
         context: argv.context,
         debug: argv.debug,
-        siteId: state.get('siteId'),
+        siteId: argv.siteId || typeof argv.site === 'string' && argv.site || state.get('siteId'),
```
ehmicky, line 79: Do we have any CLI flag called `siteId` instead of `site`?

ehmicky:
CI tests seem to be failing due to linting.

src/utils/command.js
RaeesBhatti: Yes, we have a `site` flag in `netlify deploy`

ehmicky: We have a `site` CLI flag in other commands as well. But I was wondering about a `siteId` CLI flag?

RaeesBhatti: The other `site` flag is a boolean. That's why I'm checking for `typeof`. We're checking for both `siteId` and a string `site` flag now.

ehmicky: I am wondering if `argv.siteId ||` could be removed since we don't have any `siteId` CLI flag at the moment?

PR #1016, opened 2020-07-23
Revert "Test on Node 14"

erquhart:
Reverts netlify/cli#1014


ehmicky:
Do we remember why we had to revert this PR?

Issue #1031, opened 2020-07-27
Test: make it easier to add tests


erezrokah:
At the moment most of our tests share the same site:
https://github.com/netlify/cli/blob/aa12f8ab08a1db3f72a82b14a93476a38be6ef84/tests/dev.test.js#L9
Or use pre-existing directories of sites:
https://github.com/netlify/cli/blob/aa12f8ab08a1db3f72a82b14a93476a38be6ef84/tests/build.test.js#L32

This makes the tests hard to follow and also adding new tests requires setting up a site manually.

Ideally we would have a test setup utility that allows configuring a site with different configurations (UI settings, `netlify.toml` settings, functions, etc).

Issue #684, opened 2020-01-15
kill call targets a non-existing PID: causes `env: node: No such file or directory`


sw-yx:
**- Do you want to request a _feature_ or report a _bug_?**
bug

**- What is the current behavior?**
when we run netlify dev on https://github.com/netlify/netlify-react-ui/, with CLI version 2.29, it results in this

![image](https://user-images.githubusercontent.com/6764957/72442828-351fc100-377b-11ea-9457-1ee8d9bf74da.png)

(ignore the console.log of settings, thats me trying to debug)

**- If the current behavior is a bug, please provide the steps to reproduce.**
as above

**- What is the expected behavior?**
netlify dev should work on netlify app

@ehmicky thinks it is a recent regression - we are killing a non existent PID here, causing the [Kill ESRCH](https://unix.stackexchange.com/questions/258955/what-does-esrch-mean) error https://github.com/netlify/cli/blob/a1151afbefb9599052fc104b814270493eceec14/src/commands/dev/index.js#L327

however when i tweak that line i still cant get it to work

RaeesBhatti:
Hey @sw-yx , thanks for reporting this issue. I've identified the problem and provided a fix in https://github.com/netlify/cli/pull/685. Please try it out and let me know.

PR #1040, opened 2020-07-29
test: refactor tests

erezrokah:
Fixes https://github.com/netlify/cli/issues/1031

Also renamed many tests so their description matches what we're testing.
Also renamed `dev.test.js` to `command.dev.test.js` (soon to be followed in another PR for `build.test.js`)


tests/addons.test.js
```diff
@@ -1,16 +1,8 @@
-const path = require('path')
 const test = require('ava')
 const stripAnsi = require('strip-ansi')
 const cliPath = require('./utils/cliPath')
-const exec = require('./utils/exec')
```
erezrokah, line 5: `./utils/exec` is a promisified version of node `exec`.
We were actually spawning commands in 3 different ways in the tests.
1. Using `./utils/exec` (addons tests)
2. Using a promisified `exec` duplicating `./utils/exec` logic (dev command tests)
3. Using `execa` (build command tests).

I aligned the places I could find to use `execa` and removed `./utils/exec`

tests/command.dev.test.js
```diff
@@ -0,0 +1,793 @@
+const test = require('ava')
```
erezrokah, line 1: This file was refactored and renamed from `dev.test.js`.

Each test creates a dedicated site with the relevant settings just for that test

tests/command.dev.test.js
```diff
@@ -0,0 +1,793 @@
...
+      })
+
+    await builder.buildAsync()
+
+    const server = await startDevServer({ cwd: builder.directory })
+
+    const response1 = await fetch(`${server.url}/not-foo`).then(r => r.text())
+    const response2 = await fetch(`${server.url}/not-foo/`).then(r => r.text())
+
+    // TODO: check why this doesn't redirect
```
erezrokah, line 680: Kept this from the original test, still not sure what I'm missing here

tests/command.functions.test.js
```diff
@@ -0,0 +1,29 @@
+const test = require('ava')
```
erezrokah, line 1: This test was inside `dev.test.js`.
Since it tests a different command I moved it to a separate file.

tests/utils/siteBuilder.js
```diff
@@ -0,0 +1,107 @@
+const { toToml } = require('tomlify-j0.4')
+const fs = require('fs-extra')
+const path = require('path')
+const tempy = require('tempy')
+const os = require('os')
+
+const createSiteBuilder = ({ siteName }) => {
```
erezrokah, line 7: At some point I would like to add `withUISetting` which will use the API to create the site.

tests/utils/devServer.js
```diff
@@ -0,0 +1,92 @@
...
+  })
+  return new Promise((resolve, reject) => {
+    ps.stdout.on('data', data => {
+      if (data.toString().includes('Server now ready on')) {
+        resolve({
+          url,
+          host,
+          port,
+          close: async () => {
+            const pids = await pidtree(ps.pid).catch(() => [])
```
erezrokah, line 36: https://github.com/sindresorhus/execa/issues/96#issuecomment-612834665 ❤️

erezrokah:
@ehmicky, sorry for the site of this PR, got too deep and had to dig my way out of it by doing more refactoring 😄

tests/addons.test.js
ehmicky: YES! :tada: 

Child process spawning is complex :P

tests/utils/devServer.js
```diff
@@ -0,0 +1,92 @@
...
+function getRandomPortStart(rng) {
+  const startPort = Math.floor(rng() * 10000) + 10000 // 10000 to avoid collisions with frameworks ports
+  return startPort
+}
+
+let currentPort = getRandomPortStart(rng)
+
+const startServer = async ({ cwd, env = {} }) => {
+  const tryPort = currentPort++
+  const port = await getPort({ port: tryPort })
```
ehmicky, line 19: Can we use port `0` instead?
It [uses a randomly available port](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback).
Then the port can be retrieved with `server.address()`. Not sure if that would work here though.

tests/utils/devServer.js
```diff
@@ -0,0 +1,92 @@
...
+  const host = 'localhost'
+  const url = `http://${host}:${port}`
+  console.log(`Starting dev server on port: ${port} in directory ${path.basename(cwd)}`)
+  const ps = execa(cliPath, ['dev', '-p', port], {
+    cwd,
+    env: { ...process.env, ...env },
```
ehmicky, line 25:
```suggestion
    env,
```

Execa appends to `env` by default (instead of overwriting it), as opposed to `child_process.spawn()`.

tests/utils/devServer.js
```diff
@@ -0,0 +1,92 @@
...
+          },
+        })
+      }
+    })
+
+    let error = ''
+    ps.stderr.on('data', data => {
+      error = error + data.toString()
+    })
+    ps.on('close', code => {
```
ehmicky, line 55: You might want to look for `ps` rejection instead (`ps` is a promise as well). A child process can fail for more reasons than exit code non-0:
  - signal termination (exit code will be `undefined`)
  - `error` event on `stdin`, `stdout` or `stderr`
  - child spawning error

tests/utils/devServer.js
```diff
@@ -0,0 +1,92 @@
...
+        resolve({
+          url,
+          host,
+          port,
+          close: async () => {
+            const pids = await pidtree(ps.pid).catch(() => [])
+            pids.forEach(pid => () => {
+              try {
+                process.kill(pid)
+              } catch (e) {
```
ehmicky, line 40: `process.kill()` never throws (I had to look into the Node.js internal code once :) ). In particular, it does not throw when the PID does not exist or has already been killed.

tests/utils/devServer.js
erezrokah: Tried using random port and it lead to collisions.

Things I can think of that contributed to the collisions:

1. There is a delay between the time `getPort` returns the free port to the time the dev server actually acquires it (since we're spawning it via a child process). That leads to race conditions.
2. The dev server also the underlying framework server on a port which is not randomised yet. For example the create react app server runs on port 3000. Tried using `targetPort` for the `cra` test, but that didn't seem to work.

The current solutions seeds a random port generator for each process based on the process id and increment the port per each test in the same process. Also we add 10000 to the starting port to avoid collisions with frameworks.

tests/utils/devServer.js
erezrokah: Cool, I'll fix it

tests/utils/devServer.js
erezrokah: I actually had it without the `try/catch` initially but added it later since it did throw.
Also https://github.com/netlify/cli/issues/684 and https://github.com/sindresorhus/execa/issues/96#issuecomment-612787831

tests/utils/devServer.js
erezrokah: That's a good point.
We can't `await` on `ps` since the dev process never exits until we kill it (unless there was an unexpected error).
The code here is mostly intended to catch errors with the dev server not initialising properly (and prevent waiting indefinitely for the `Server now ready on` message).

ehmicky: We could do something like: 

```js
ps.catch(error => {
  reject(error)
})
```

tests/utils/devServer.js
ehmicky: Oh, sorry, my bad! :+1:

tests/utils/devServer.js
erezrokah: Fixed in https://github.com/netlify/cli/pull/1040/commits/0f0d6bdb0c2a10dcde6e2b51cdfdfeadc333a9f2

Issue #1052, opened 2020-08-02
chore: update docs site dependencies


erezrokah:
The docs site code needs some cleanup.
Initially we should update dependencies, enforce linting and run the build on a recent node version

PR #1053, opened 2020-08-02
chore(docs-site): update site dependencies and enforce linting

erezrokah:
Fixes https://github.com/netlify/cli/issues/1052


site/package.json
```diff
@@ -20,19 +20,22 @@
...
     "clean": "rimraf dist"
   },
+  "engines": {
+    "node": ">=v12"
+  },
   "license": "MIT",
   "dependencies": {
-    "@compositor/x0": "^6.0.5",
-    "@rebass/mdx": "^1.0.0-1",
+    "@compositor/x0": "^6.0.7",
```
erezrokah, line 28: We should think about removing the current toolset as `@compositor/x0` hasn't been updated in a while

site/src/_layout.js
```diff
@@ -2,28 +2,12 @@ import React from 'react'
...
-  Divider,
-  Heading,
-  NavLink,
-  BlockLink,
-  Button,
-  ButtonTransparent,
-} from 'rebass'
-import { InstantSearch, Hits, SearchBox, createConnector, Configure, Highlight } from 'react-instantsearch-dom'
-import { borderColor, themeGet } from 'styled-system'
+import styled from 'styled-components'
```
erezrokah, line 5: Changes in this file are mostly about removing unused vars

site/package.json
```diff
@@ -20,19 +20,22 @@
     "sync": "node ./sync.js",
     "clean": "rimraf dist"
   },
+  "engines": {
+    "node": ">=v12"
```
ehmicky, line 24: We might want to specify the minor/patch version too.
This is because tools are quite ambiguous in their behavior there. `semver` interprets this as `>= 12.0.0`, but `npm` interprets this as `>= last 12.*.*`, i.e. `=== last 12.*.*`.

erezrokah: Good point - fixed

XhmikosR:
Would be nice to see the lgtm alerts sorted too https://lgtm.com/projects/g/netlify/cli/?mode=list and eventually adding CodeQL since it's the successor to LGTM :)

PR #1066, opened 2020-08-03
Test: refactor build tests

erezrokah:
This removes the existing static directories used for the build command tests in favour of creating them as a part of each test setup.


erezrokah:
> This might be a little slower, since fixture files are created/deleted each time the tests are run, as opposed to only once when the git repository is being cloned.

True, but I would take slower in this case. Also this is less prone to abuse - for example we had multiple tests using the same `dummy-site` which make it very hard to understand the setup of each test.


> Also, this requires new developers to learn the builder API. It also requires maintaining and extending it. I would personally think that would add some complexity to the tests setup, as opposed to using plain old files for the fixtures directly.

Also true, but I think most of the time devs will copy paste an existing tests so that won't be much of an issue.
The previous approach led to devs copying tests, but keeping the same directory (just adding stuff to the same site).

> That being said, it keeps the fixture content closer to the test assertions, so there is less context switching / multiple tabs required, which is good.

⬆️ This is the main reason for the change

ehmicky:
Good point that it encourages decoupling test fixtures. :+1:

PR #1068, opened 2020-08-03
Remove detectors code that is not currently working

ehmicky:
When we fail to detect the build command with Netlify Dev, many detectors are adding a default command.

The way this is currently done does not work. This is because the `possibleArgsArr` are the arguments passed to `command` which is `npm run` or `yarn`. For example, if `possibleArgsArr` is `['gatsby', 'develop']`, then `npm run gatsby develop` will be run instead of simply `gatsby develop`. 

A follow-up PR will fix how we assign a default build command. This PR just removes the code that is currently failing.

### DIFF
--- a/src/detectors/angular.js
+++ b/src/detectors/angular.js
@@ -14,11 +14,6 @@ module.exports = function() {
     preferredCommand: 'ng build --prod',
   })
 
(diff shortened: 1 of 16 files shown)


erezrokah:
> The way this is currently done does not work.
🤦

PR #1070, opened 2020-08-03
Simplify detectors watch command

ehmicky:
We currently split the Netlify Dev watch command into `settings.command` (the binary) and `settings.args` (the arguments array). Splitting those is usually done to escape whitespaces and delimitate tokens between the command and the arguments. However, in our case, it does not make much sense considering the watch command can only be either:
  - `npm run scriptName`
  - some framework-specific hardcoded string like `hugo server -w`, just never includes whitespaces in our frameworks

This also prevents us from potentially spawning the watch command in a subshell (like Bash), which prevents users from using shell features like variable expansion or globbing in their `dev.command`.

Finally, splitting those makes the code logic more complicated.

This PR merges those into a single `settings.watchCommand` instead.

It also simplifies some logic which was first passing `npm` as the watch command, then later on transforming it to `npm run`, by passing it as `npm run` directly instead.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -389,15 +389,16 @@ async function startDevServer(settings, log) {
   }
 
   log(`${NETLIFYDEVLOG} Starting Netlify Dev with ${settings.framework || 'custom config'}`)
-  const commandBin = await which(settings.command).catch(err => {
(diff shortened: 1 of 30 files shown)


erezrokah:
Closing this in favor of doing it as a part of https://github.com/netlify/cli/issues/1495.
@ehmicky please let me know if you think otherwise

PR #1071, opened 2020-08-03
Improve detectors loading

ehmicky:
We are lazy loading all detectors by using `fs.readdir()` on their directory, then `require()` them at runtime.

Detectors are very small, simple and all use the same common logic. There is no need to lazy load them nor provide additional error handling.

This PR simplifies the way detectors are loaded, by using an `index.js` that `require()` them at load-time instead.

This will make it easier to integrate with `@netlify/framework-info` as well.


src/utils/detect-server.js
```diff
@@ -22,19 +23,10 @@ module.exports.serverSettings = async (devConfig, flags, projectDir, log) => {
...
-    const detectors = detectorsFiles.map(det => {
-      try {
-        return loadDetector(det)
-      } catch (err) {
-        console.error(err)
-        return null
-      }
-    })
-    for (const detector of detectors) {
-      const detectorResult = detector(projectDir)
```
erezrokah, line 35: Not sure why we were passing `projectDir` to `detector`

ehmicky: This was dead code (not used in any detector).

erezrokah:
Closing this in favor of doing it as a part of https://github.com/netlify/cli/issues/1495.
@ehmicky please let me know if you think otherwise

Issue #890, opened 2020-05-09
function redirects are not working with netlify dev


bradgarropy:
*Note: I'm still working on a minimal reproduction, but you can check on the [live site](https://master--dailytexascountry.netlify.app/store).*

---

### Do you want to request a _feature_ or report a _bug_?
🐛 bug

### What is the current behavior?
First, see this [Twitter discussion](https://twitter.com/bradgarropy/status/1258993787990409217) with @jlengstorf.

I defined the following redirect in `netlify.toml`.

```
[[redirects]]
    from = "/api/*"
    to = "/.netlify/functions/:splat"
    status = 200
```

Then in my application I have a serverless function located at `/.netlify/functions/subscribe`.
I hit it via a `POST` request using the browser's `fetch` API like this.

```
const options = {
    method: "POST",
    body: JSON.stringify(data),
    headers: {"Content-Type": "application/json"},
}

const response = await fetch("/api/subscribe", options)
```

It returns an HTML document rather than the expected JSON output.

```
<!DOCTYPE html><html><head><meta charSet="utf-8"/><meta http-equiv="x-ua-compatible" content="ie=edge"/><meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"/><meta name="note" content="environment=development"/><title data-react-helmet="true"></title><link href="https://fonts.googleapis.com/css?family=Inter|Roboto|Patua+One" rel="stylesheet"/><style type="text/css">
    .anchor.before {
      position: absolute;
      top: 0;
      left: 0;
      transform: translateX(-100%);
      padding-right: 4px;
    }
    .anchor.after {
      display: inline-block;
      padding-left: 4px;
    }
    h1 .anchor svg,
    h2 .anchor svg,
    h3 .anchor svg,
    h4 .anchor svg,
    h5 .anchor svg,
    h6 .anchor svg {
      visibility: hidden;
    }
    h1:hover .anchor svg,
    h2:hover .anchor svg,
    h3:hover .anchor svg,
    h4:hover .anchor svg,
    h5:hover .anchor svg,
    h6:hover .anchor svg,
    h1 .anchor:focus svg,
    h2 .anchor:focus svg,
    h3 .anchor:focus svg,
    h4 .anchor:focus svg,
    h5 .anchor:focus svg,
    h6 .anchor:focus svg {
      visibility: visible;
    }
  </style><script>
    document.addEventListener("DOMContentLoaded", function(event) {
[...]

PR #1085, opened 2020-08-11
fix(dev-external-redirects): re-stream request body

erezrokah:
Fixes https://github.com/netlify/cli/issues/890, specifically the comment here:
https://github.com/netlify/cli/issues/890#issuecomment-670989422

Since we consume the request body stream here:
https://github.com/netlify/cli/blob/master/src/commands/dev/index.js#L192
we need to recreate the stream.

### DIFF
--- a/src/utils/stream.js
+++ b/src/utils/stream.js
@@ -1,3 +1,5 @@
+const { Readable } = require('stream')
+
 function createStreamPromise(stream, timeoutSeconds, bytesLimit = 1024 * 1024 * 6) {
   return new Promise(function(resolve, reject) {
     let data = []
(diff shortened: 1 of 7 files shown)


src/utils/stream.js
```diff
@@ -39,4 +41,15 @@ function createStreamPromise(stream, timeoutSeconds, bytesLimit = 1024 * 1024 *
   })
 }
 
-module.exports = { createStreamPromise }
+function bufferToStream(buffer) {
```
ehmicky, line 44: Should we use the dependency directly?

https://github.com/sindresorhus/to-readable-stream/blob/master/index.js

This will ensure they update that code for us :)

erezrokah: Good point. Actually found a couple of other unmaintained libs that does this, but didn't find this one.

erezrokah: Fixed in 646288ffc18a924eba67028071071e302144799f

ehmicky:
Looks good! :+1:

PR #1153, opened 2020-08-26
Remove v from node engine version string

himynameisjonas:
**- Summary**

Trying to install the cli with [Volta](https://volta.sh) using `volta install netlify-cli` failed with `error: Could not parse version ">=v8.17.0"`. Removing the `v` from the version string should fix that.

**- Test plan**
I could install version `2.58.0` without any problems and when comparing diff of 2.59.0 I noticed the node engine change that added the `v`. 

Remove unnecessary `v` from node engine version string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "Bret Comnes <bcomnes@gmail.com> (https://bret.io)"
   ],
   "engines": {
-    "node": ">=v8.17.0"
+    "node": ">=8.17.0"
   },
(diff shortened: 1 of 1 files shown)


ehmicky (approved):
Thanks a lot @himynameisjonas!

erezrokah:
Looks like this was created by renovate bot https://github.com/netlify/cli/pull/994/files

Issue #1017, opened 2020-07-26
403 Forbidden when Folder and File share name (e.g. hello/ and hello.html)


FinnWoelm:
Hey Netlify :heart:,

Thanks for all the amazing work you do! I found a small bug with netlify dev — its behavior is inconsistent with deployed Netlify when you have a folder and file with the same name.

**- Do you want to request a _feature_ or report a _bug_?**

Report a bug.

**- What is the current behavior?**

When you have a folder and a file with the same name and visit the path, you see the file contents when deployed on Netlify, but you get a 403 Forbidden error with netlify dev.

For example, let's say we have a folder `hello/` and a file `hello.html`. When requesting `/hello`,

- Netlify displays the contents of `hello.html`
- `netlify-dev` responds with 403 Forbidden

**- If the current behavior is a bug, please provide the steps to reproduce.**

*For netlify dev:*

```
git clone https://github.com/FinnWoelm/netlify-same-name-for-folder-and-file
netlify dev
```

Visit localhost:8888/static and see 403 Forbidden.

![Screenshot from 2020-07-26 11-34-25](https://user-images.githubusercontent.com/7606194/88475870-22f5bb00-cf34-11ea-9ea9-7ada479d4c05.png)

*For deployed Netlify:*

Visit https://same-name-for-folder-and-file.netlify.app/hello and see contents of hello.html

![Screenshot from 2020-07-26 11-35-12](https://user-images.githubusercontent.com/7606194/88475866-1d987080-cf34-11ea-8968-86a043710312.png)

**- What is the expected behavior?**

netlify dev should match the behavior of deployed Netlify and display the contents of `hello.html`.

**- Local Environment Information**
```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: Linux 5.4 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    npm: 6.13.7 - ~/.nvm/versions/node/v12.14.1/bin/npm
  Browsers:
    Firefox: 78.0.2
  npmGlobalPackages:
    netlify-cli: 2.58.0

netlify-cli/2.58.0 linux-x64 node-v12.14.1
```

**How to Fix:**

[...]

PR #1155, opened 2020-08-27
Fix: Return .html file when file and folder share name

FinnWoelm:
Hi Netlify :heart:,

I'm submitting this PR as suggested by @erezrokah in order to fix issue #1017.

**- Summary**

When you have a folder and a file with the same name (e.g., foo.html and foo/), Netlify will serve the contents of foo.html.

Until now, netlify dev was inconsistent with that behavior and returned a 403 Forbidden error in this scenario (because directory listings are forbidden).

This PR makes netlify dev behave consistent with Netlify by serving the .html (or .htm) file instead of returning 403 Forbidden.

Fixes: #1017 

**- Test plan**

I first added a test to `tests/command.dev.test.js`. I ran the test and confirmed that it failed. Then I implemented the fix. I reran the test and the test passed.

I ran `npm run test` and everything came back green.

Fix: Return .html file when file and folder share name (#1017)

WhatsApp Image 2019-12-25 at 2 04 28 PM (1)

Happy to make any changes/improvements you deem necessary! Thanks for all your great work :raised_hands:

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -128,7 +128,7 @@ function initializeProxy(port, distDir, projectDir) {
     }
   })
   proxy.on('proxyRes', (proxyRes, req, res) => {
-    if (proxyRes.statusCode === 404) {
+    if (proxyRes.statusCode === 404 || proxyRes.statusCode === 403) {
       if (req.alternativePaths && req.alternativePaths.length > 0) {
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
This is great @FinnWoelm!
Thank you for adding the test case for it too ❤️

Issue #911, opened 2020-05-20
port and targetPort options are confusing


RaeesBhatti:
**- Do you want to request a _feature_ or report a _bug_?**
Enhancement?
**- What is the current behavior?**
People often get confused about `port` and `targetPort` options because they sound similar and don't convey enough obvious meaning to new users.
We should rename either `targetPort` or both of them to something more obviously meaningful.
 
**- If the current behavior is a bug, please provide the steps to reproduce.**

**- What is the expected behavior?**
People should be able to guess right away, which port means what.
**- Local Environment Information**

```
────────────────────┐
 Environment Info   │
────────────────────┘

  System:
    OS: macOS 10.15.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 10.20.1 - ~/.n/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.n/bin/npm
  Browsers:
    Chrome: 81.0.4044.138
    Firefox: 76.0
    Safari: 13.1
  npmGlobalPackages:
    netlify-cli: 2.51.0

netlify-cli/2.51.0 darwin-x64 node-v10.20.1
```

DavidSabine:
"target" can be interpreted too many ways, so causes confusion. I suggest the variables be named as follows:

`port` to be renamed:
> `netlifydev_will_publish_the_build_on_this_port`

and `targetPort` be renamed:
> `netlifydev_will_listen_for_your_app_on_this_port`

Those names are descriptive, hard to misinterpret, and therefore good. So what if they are sentences if it eliminates confusion and makes the code easier to read.

Thank you for creating this issue. The port stuff has been confusing since my first use of the cli.

esquaredly:
For someone who ended up on this issue due to being confused, I believe:
`port` and `netlifyDevPort`
would been enough of a hint.

DavidSabine:
> `port` and `netlifyDevPort` would been enough of a hint for me to avoid the confusion.

Sooo... `port` would be the port of my app? And `netlifyDevPort` would be the port I visit after running `netlify dev` command?

RaeesBhatti:
`port` currently means the port of Netlify Dev server. I think changing the meaning of that will lead to things breaking and confusion. We should move away from `port` and `targetPort` entirely

ksnyde:
[...]

PR #1157, opened 2020-08-27
Make sure scheme is formatted right for redirector

biilmann:
**- Summary**

Make host based redirect rules work with `netlify dev`

**- Test plan**

Add a rule like:

```
http://testing.dev/* /testing.html 200!
```

to a local project and test with:

```
curl http://locahost:8888 -H "Host: testing.dev"
```

Without this fix, the redirect won't work, with this fix the redirect will work.

Make host based rules like:

```
http://localhost/* /:splat 200
```

Work in Netlify dev


erezrokah (approved):
This fix looks good to me.
~~I added a test case, which initially failed on CI due to unrelated collisions between tests.
The solution was to add a hidden flag so we can control the static server port from the tests (`frameworkPort` is only used for frameworks. we should probably improve this in https://github.com/netlify/cli/issues/911)~~
Removed the test until I can get it to a stable state

Issue #1158, opened 2020-08-27
Custom 404.html in publish folder is ignored


FinnWoelm:
Hey Netlify :heart:,

As I've said before, thanks for all your work! :raised_hands: Here is a tiny issue. I'm going to submit a corresponding PR in a few minutes.

**Describe the bug**

In Netlify, you can set a custom 404 page by putting a `404.html` file into the publish folder (e.g., `public` or whatever is specified in `netlify.toml`). This custom 404 page is not picked up by netlify dev. For it to work in netlify dev, the custom `404.html` file must be in the root folder. This is inconsistent with Netlify in production.

**To Reproduce**

```
git clone https://github.com/FinnWoelm/netlify-404
cd netlify-404
netlify dev
```

Visit http://localhost:8888/this-file-does-not-exist and see `Not found` instead of the custom 404.html page:

![Screenshot_2020-08-27 Screenshot](https://user-images.githubusercontent.com/7606194/91483129-806c8700-e8a7-11ea-8ff4-f412337701e6.png)

**Configuration**

- If possible, please copy/paste below your `netlify.toml`.

```
[build]
  publish     = "public"
```

- Please enter the following command in a terminal and copy/paste its output:

```bash
  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
    Memory: 256.93 MB / 7.68 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 12.14.1 - ~/.nvm/versions/node/v12.14.1/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 6.14.7 - ~/.nvm/versions/node/v12.14.1/bin/npm
    Watchman: 4.7.0 - /usr/local/bin/watchman
  npmGlobalPackages:
    netlify-cli: 2.59.1
```

**Expected behavior**

My custom error page should be visible, like it is on Netlify when visiting https://netlify-404.netlify.app/this-file-does-not-exist:

![Screenshot_2020-08-27 404 - Not Found](https://user-images.githubusercontent.com/7606194/91483438-0092ec80-e8a8-11ea-9256-b450bdfb2886.png)

**Additional context**

Working on a PR right now. This issue is just here for context.

The PR should only be a single change:

[...]

PR #1159, opened 2020-08-27
Fix: Use 404.html page in publish folder

FinnWoelm:
**- Summary**

In Netlify, you can set a custom 404 page by putting a 404.html file inside the publish folder (e.g., public or whatever is specified in netlify.toml).

Until now, this custom 404 page was not picked up by netlify dev. For it to work in netlify dev, the custom 404.html file had to be in the root folder.

This PR makes the behavior of netlify dev consistent with Netlify in production by using the custom 404.html page in the publish folder.

Fixes: https://github.com/netlify/cli/issues/1158

**- Test plan**

I first added a test to tests/command.dev.test.js. I ran the test and confirmed that it failed. Then I implemented the fix. I reran the test and the test passed.

Screenshot from 2020-08-27 21-09-15

Screenshot from 2020-08-27 21-17-13

I ran npm run test and everything came back green.

Fix: Use 404.html page in publish folder (#1158)

WhatsApp Image 2019-12-25 at 2 04 28 PM (3)

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -380,7 +380,7 @@ async function startDevServer(settings, log) {
       name: 'netlify-dev',
       port: settings.frameworkPort,
       templates: {
-        notFound: '404.html',
+        notFound: path.join(settings.dist, '404.html'),
       },
     })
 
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
@FinnWoelm you're on 🔥

PR #1161, opened 2020-08-28
Fix typo

syntaxhacker:
Fixing small typo


ehmicky (approved):
Thanks for the fix @syntaxhacker!

erezrokah (approved):
Thanks @syntaxhacker, some md files are actually autogenerated based on the commands.

PR #1188, opened 2020-08-31
test: set random static server port for tests

erezrokah:
This is an attempt to make the tests more stable.
The static server port is currently hard coded (that port is used when no framework is detected and we just server files statically from the publish folder).
Added a hidden flag to be used in tests for the sake of randomising the port.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -623,6 +623,10 @@ DevCommand.flags = {
   targetPort: flags.integer({
     description: 'port of target app server',
   }),
+  staticServerPort: flags.integer({
+    description: 'port of the static app server used when no framework is detected',
+    hidden: true,
+  }),
   dir: flags.string({
     char: 'd',
     description: 'dir with static files',
--- a/src/utils/detect-server.js
+++ b/src/utils/detect-server.js
@@ -188,7 +188,7 @@ async function getStaticServerSettings(settings, flags, projectDir, log) {
   log(`${NETLIFYDEVWARN} Running static server from "${path.relative(path.dirname(projectDir), dist)}"`)
   return {
     noCmd: true,
-    frameworkPort: await getPort({ port: 3999 }),
+    frameworkPort: await getPort({ port: flags.staticServerPort || 3999 }),
     dist,
   }
 }
--- a/tests/utils/devServer.js
+++ b/tests/utils/devServer.js
@@ -20,7 +20,7 @@ const startServer = async ({ cwd, env = {} }) => {
   const host = 'localhost'
   const url = `http://${host}:${port}`
   console.log(`Starting dev server on port: ${port} in directory ${path.basename(cwd)}`)
-  const ps = execa(cliPath, ['dev', '-p', port], {
+  const ps = execa(cliPath, ['dev', '-p', port, '--staticServerPort', port + 1000], {
     cwd,
     env,
   })


ehmicky:
Note: if two tests are started in parallel, they most likely will receive different port thanks to `get-port`:

https://github.com/netlify/cli/blob/b83885e4be7e1026ee9386aeac7bcf91d64f733c/src/utils/detect-server.js#L159

However, looking at `get-port` code, this is not concurrecy-safe, so there would be a chance of sequences like:
  - Test A starts
  - Test B starts
  - Test A starts and stops a TCP server (using `get-port`), confirming `3999` is not taken
  - Test B starts and stops a TCP server (using `get-port`), confirming `3999` is not taken
  - Test A finally starts Netlify Dev server at `3999`
  - Test B starts another server at `3999`, which fails

Injecting a test port is one solution to this problem :+1: 

I am wondering what are your thoughts about those alternatives?
  - A boolean CLI flag to set a random port (instead of having to specify the port)
  - If Netlify Dev starts and the port is taken, retry with `port + 1`

erezrokah:
This is confusing since we have different ports we need to acquire.
The dev server port in the code you linked:
https://github.com/netlify/cli/blob/b83885e4be7e1026ee9386aeac7bcf91d64f733c/src/utils/detect-server.js#L159
which can be overridden using the `--port` flag.

The framework port - the port the underlying framework server, e.g. Create React App app runs on,
which can be overridden using the `--targetFramework` flag.

A static server port which is used when no framework is detected:
https://github.com/netlify/cli/blob/b83885e4be7e1026ee9386aeac7bcf91d64f733c/src/utils/detect-server.js#L144
This port is currently hardcoded and can't be overridden (this PR fixes that).

> However, looking at `get-port` code, this is [not concurrency-safe](https://github.com/sindresorhus/get-port/blob/a9b445ea0afbd75b81b7b9011c896e111e2d4aee/index.js#L24), so there would be a chance of sequences like:
> 
> * Test A starts
> * Test B starts
> * Test A starts and stops a TCP server (using `get-port`), confirming `3999` is not taken
> * Test B starts and stops a TCP server (using `get-port`), confirming `3999` is not taken
> * Test A finally starts Netlify Dev server at `3999`
> * Test B starts another server at `3999`, which fails
> 
> Injecting a test port is one solution to this problem 👍

Exactly with the note that `get-port` locks returned ports for about 15-30 seconds, so hitting the race condition is rare when calling `get-port` in the same process.
When calling from different processes, the port can be considered free since there is a delay between when `get-port` returns the port until `netlify dev` acquires it.
This was happening in our tests with the dev server port (thus the current usage of random port ranges in the tests seeded by the process id).


> I am wondering what are your thoughts about those alternatives?
> 
> * A boolean CLI flag to set a random port (instead of having to specify the port)
> * If Netlify Dev starts and the port is taken, retry with `port + 1`

We are actually doing both of these at the tests level:
1. Each spec gets a random port range seeded from the PID (since specs run in different processes). Each test in a specific spec gets and incrementing port to try. Ranges start high enough to avoid collisions with framework ports (at some point we would need to coordinate those too when we add frameworks tests that run in parallel).
2. Starting the dev server is retried:
https://github.com/netlify/cli/blob/b83885e4be7e1026ee9386aeac7bcf91d64f733c/tests/utils/devServer.js#L59

Doing those at the CLI level would require parsing the CLI output for the port that was eventually acquired (our users were doing that before `--port` was introduced).

The main issue is that the CLI always uses port 3999 when starting its own static server. I think the only reason it is working with the current tests is due to the retry mechanism.

ehmicky (approved):
Thanks for the breakdown @erezrokah, this makes sense!

PR #1190, opened 2020-09-01
fix(docs): hide hidden command flags from docs

erezrokah:
The `staticServerPort` flag added to the `dev` command [here](https://github.com/netlify/cli/pull/1188) is shown in the docs.

This PR fixes the code that generates the docs to remove hidden flags.

### DIFF
--- a/scripts/generateCommandData.js
+++ b/scripts/generateCommandData.js
@@ -1,4 +1,7 @@
 const path = require('path')
+// This dependency is installed in `site/package.json`
+// eslint-disable-next-line node/no-extraneous-require
+const filterObj = require('filter-obj')
 const globby = require('markdown-magic').globby
 
 module.exports = function generateCommandData() {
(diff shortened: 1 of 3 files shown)


ehmicky (approved):

scripts/generateCommandData.js
```diff
@@ -1,4 +1,7 @@
 const path = require('path')
+// This dependency is installed in `site/package.json`
+// eslint-disable-next-line node/no-extraneous-require
```
ehmicky, line 3: Should we add some `override` in `.eslintrc.json` to omit that ESLint rule inside `site/**`?

erezrokah: I think we should move this script to `site/scripts`, then we won't need the override/disable comment.

I'll open a new issue for that

erezrokah: https://github.com/netlify/cli/issues/1191

PR #1196, opened 2020-09-02
Feat: add traffic mesh proxy flag

erezrokah:
Fixes https://github.com/netlify/traffic-mesh/issues/806

This is a big PR since I did some refactoring to the dev command in the process since it was very hard to follow.
The commits are small though.

I'll add comments to clarify the changes.

This PR consists of 3 main changes:
1. Extract common logic to download binary clients (this was used in the tunnel client and now shared with traffic mesh agent).
2. Extract the different dev command responsibilities (starting functions server, framework server, Netlify dev proxy, tunnel client, etc) into separate functions/files.
3. Traffic Mesh flag implementation which was quite small.

> When enabling traffic mesh some test are still failing - looking into those.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -1,375 +1,21 @@
-const url = require('url')
-const util = require('util')
-const { URLSearchParams } = require('url')
 const path = require('path')
-const fs = require('fs')
 const { flags } = require('@oclif/command')
 const child_process = require('child_process')
-const http = require('http')
-const httpProxy = require('http-proxy')
 const waitPort = require('wait-port')
 const stripAnsiCc = require('strip-ansi-control-characters')
 const which = require('which')
-
-const { createProxyMiddleware } = require('http-proxy-middleware')
-const cookie = require('cookie')
-const get = require('lodash.get')
-const isEmpty = require('lodash.isempty')
 const { serverSettings } = require('../../utils/detect-server')
 const { startFunctionsServer } = require('../../utils/serve-functions')
 const Command = require('../../utils/command')
 const chalk = require('chalk')
-const jwtDecode = require('jwt-decode')
 const open = require('open')
-const contentType = require('content-type')
 const { NETLIFYDEV, NETLIFYDEVLOG, NETLIFYDEVWARN, NETLIFYDEVERR } = require('../../utils/logo')
 const boxen = require('boxen')
-const { createTunnel, connectTunnel } = require('../../utils/live-tunnel')
-const { createRewriter } = require('../../utils/rules-proxy')
-const { onChanges } = require('../../utils/rules-proxy')
-const { parseHeadersFile, objectForPath } = require('../../utils/headers')
+const { startLiveTunnel } = require('../../utils/live-tunnel')
 const { getEnvSettings } = require('../../utils/env')
-const { createStreamPromise } = require('../../utils/create-stream-promise')
-const toReadableStream = require('to-readable-stream')
+const { startProxy } = require('../../utils/proxy')
 
-const stat = util.promisify(fs.stat)
-
-function isInternal(url) {
-  return url.startsWith('/.netlify/')
-}
-function isFunction(functionsPort, url) {
-  return functionsPort && url.match(/^\/.netlify\/functions\/.+/)
-}
-
-function addonUrl(addonUrls, req) {
-  const m = req.url.match(/^\/.netlify\/([^\/]+)(\/.*)/) // eslint-disable-line no-useless-escape
-  const addonUrl = m && addonUrls[m[1]]
-  return addonUrl ? `${addonUrl}${m[2]}` : null
-}
-
-async function getStatic(pathname, publicFolder) {
-  const alternatives = [pathname, ...alternativePathsFor(pathname)].map(p => path.resolve(publicFolder, p.substr(1)))
-
-  for (const i in alternatives) {
-    const p = alternatives[i]
-    try {
-      const pathStats = await stat(p)
-      if (pathStats.isFile()) return '/' + path.relative(publicFolder, p)
-    } catch (err) {
-      // Ignore
-    }
-  }
-  return false
-}
-
-function isExternal(match) {
-  return match.to && match.to.match(/^https?:\/\//)
-}
-
-function isRedirect(match) {
-  return match.status && match.status >= 300 && match.status <= 400
-}
-
-function render404(publicFolder) {
-  const maybe404Page = path.resolve(publicFolder, '404.html')
-  if (fs.existsSync(maybe404Page)) return fs.readFileSync(maybe404Page)
-  return 'Not Found'
-}
-
-// Used as an optimization to avoid dual lookups for missing assets
-const assetExtensionRegExp = /\.(html?|png|jpg|js|css|svg|gif|ico|woff|woff2)$/
-
-function alternativePathsFor(url) {
-  const paths = []
-  if (url[url.length - 1] === '/') {
-    const end = url.length - 1
-    if (url !== '/') {
-      paths.push(url.slice(0, end) + '.html')
-      paths.push(url.slice(0, end) + '.htm')
-    }
-    paths.push(url + 'index.html')
-    paths.push(url + 'index.htm')
-  } else if (!url.match(assetExtensionRegExp)) {
-    paths.push(url + '.html')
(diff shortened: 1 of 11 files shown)


src/lib/exec-fetcher.js
```diff
@@ -0,0 +1,90 @@
+const fs = require('fs-extra')
```
erezrokah, line 1: Most of the logic here was extracted from https://github.com/netlify/cli/blob/93468a3fc43db5d637b30b9cbb21042da8626962/src/utils/live-tunnel.js

And made generic so we can re-use it.

src/lib/exec-fetcher.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  try {
+    const outdated = await isVersionOutdated({
+      packageName,
+      currentVersion: match[1],
+    })
+    return outdated
+  } catch (e) {
+    if (exists) {
```
erezrokah, line 68: Added a small fix to use existing version if checking for updates failed.
That can happen when reaching GitHub API limit:
https://github.com/netlify/gh-release-fetch/pull/3

src/lib/exec-fetcher.test.js
```diff
@@ -0,0 +1,70 @@
...
+test(`should postix exec with .exe on windows`, t => {
+  const execName = 'some-binary-file'
+  if (process.platform === 'win32') {
+    t.is(getExecName({ execName }), `${execName}.exe`)
+  } else {
+    t.is(getExecName({ execName }), execName)
+  }
+})
+
+const packages = [
```
erezrokah, line 22: Feels a bit strange to have this here and not in a dedicated test file (one for the tunnel client and one for the traffic mesh).
However I prefer it this way since we're actually testing the `exec-fetcher`

src/lib/settings.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path')
```
erezrokah, line 1: I've started a new concept of `libs` instead of `utils` that we currently have.
It seems that most of our current `utils` are not used for sharing code, but just to extract functionality from the different commands.
However some commands/utils need to share the same code - so I started putting those parts under lib.

src/utils/live-tunnel.js
```diff
@@ -58,70 +55,38 @@ async function connectTunnel(session, netlifyApiToken, localPort, log) {
...
-  return stat && stat.isFile() && isExe(stat.mode, stat.gid, stat.uid)
+    extension: process.platform === 'win32' ? 'zip' : 'tar.gz',
+  })
 }
 
-function isExe(mode, gid, uid) {
-  if (isWindows()) {
-    return true
-  }
+const startLiveTunnel = async ({ siteId, netlifyApiToken, localPort, log }) => {
```
erezrokah, line 80: `createTunnel` and `connectTunnel` were called separately in the dev command.
I grouped them into  `startLiveTunnel`

src/utils/proxy.js
```diff
@@ -0,0 +1,346 @@
+const path = require('path')
```
erezrokah, line 1: This was mostly copied as is from the `dev` command

src/commands/dev/index.js
```diff
@@ -384,8 +29,11 @@ async function startDevServer(settings, log) {
       },
     })
 
-    server.start(function() {
-      log(`\n${NETLIFYDEVLOG} Server listening to`, settings.frameworkPort)
+    await new Promise(resolve => {
```
erezrokah, line 32: Added some code to wait until the server is ready

src/utils/proxy.js
```diff
@@ -0,0 +1,346 @@
...
+      return proxy.web(req, res, options)
+    },
+    ws: (req, socket, head) => proxy.ws(req, socket, head),
+  }
+
+  return handlers
+}
+
+async function startProxy(settings = {}, addonUrls, configPath, projectDir) {
+  const functionsServer = settings.functionsPort ? `http://localhost:${settings.functionsPort}` : null
```
erezrokah, line 286: The original `startProxy` had some code to wait for the framework server port and functions server port:
https://github.com/netlify/cli/blob/93468a3fc43db5d637b30b9cbb21042da8626962/src/commands/dev/index.js#L170
https://github.com/netlify/cli/blob/93468a3fc43db5d637b30b9cbb21042da8626962/src/commands/dev/index.js#L178

I removed it in favour of:
1. Using `waitPort` in the function that starts the framework server: https://github.com/netlify/cli/pull/1196/files#diff-8a2f65e3aede3ea9696e546766f50f03R82
2. When starting the functions server - wait on a promise until the server is ready: https://github.com/netlify/cli/pull/1196/files#diff-cf2bd94d7688f0711bbd931e6c5e0397R373

src/utils/proxy.js
```diff
@@ -0,0 +1,346 @@
...
+
+      proxy.web(req, res, options)
+    })
+  })
+
+  server.on('upgrade', function(req, socket, head) {
+    proxy.ws(req, socket, head)
+  })
+
+  return new Promise(resolve => {
```
erezrokah, line 339: Also added this code to wait for the dev server to be ready

src/utils/proxy.js
```diff
@@ -0,0 +1,346 @@
...
+    server.listen({ port: settings.port }, () => {
+      resolve(`http://localhost:${settings.port}`)
```
erezrokah, line 341: The original `startProxy` returned an object with `url` and `port` as properties:
https://github.com/netlify/cli/blob/93468a3fc43db5d637b30b9cbb21042da8626962/src/commands/dev/index.js#L235
but only `url` was used.
Simplified the function to return just the `url` as a string

src/utils/serve-functions.js
```diff
@@ -318,4 +310,77 @@ async function serveFunctions(dir, siteInfo = {}) {
   return app
 }
 
-module.exports = { serveFunctions }
+const getBuildFunction = ({ functionBuilder, log }) => {
```
erezrokah, line 313: Moved this chunk of code from the `dev` command

src/utils/serve-functions.js
```diff
@@ -318,4 +310,77 @@ async function serveFunctions(dir, siteInfo = {}) {
...
+        functionWatcher.on('add', debouncedBuild)
+        functionWatcher.on('change', debouncedBuild)
+        functionWatcher.on('unlink', debouncedBuild)
+      })
+    }
+
+    const functionsServer = await serveFunctions(settings.functions, siteInfo)
+
+    await new Promise(resolve => {
+      functionsServer.listen(settings.functionsPort, err => {
```
erezrokah, line 374: Added this to wait until the server is ready

src/utils/traffic-mesh.js
```diff
@@ -0,0 +1,70 @@
+const path = require('path')
```
erezrokah, line 1: This is very similar to the tunnel client code

src/utils/traffic-mesh.js
```diff
@@ -0,0 +1,70 @@
...
+    process.on(signal, () => {
+      const sig = signal == 'exit' ? 'SIGTERM' : signal
+      subprocess.kill(sig, {
+        forceKillAfterTimeout: 2000,
+      })
+    })
+  )
+
+  try {
+    const open = await waitPort({ port, output: 'silent', timeout: 30 * 1000 })
```
erezrokah, line 60: Wait for the agent to be ready (the traffic mesh agent writes to a log file and not stdout) so waiting on the port works well.

tests/command.dev.test.js
```diff
@@ -6,118 +6,104 @@ const FormData = require('form-data')
 const { withSiteBuilder } = require('./utils/siteBuilder')
 const { startExternalServer } = require('./utils/externalServer')
 
-test('should return index file when / is accessed', async t => {
-  await withSiteBuilder('site-with-index-file', async builder => {
-    builder.withContentFile({
-      path: 'index.html',
-      content: '<h1>⊂◉‿◉つ</h1>',
-    })
+const testMatrix = [
```
erezrokah, line 9: The changes in this file is for support a test matrix with different args (and creating unique test names)

ehmicky (approved):
:rocket:

erezrokah:
@ehmicky wow that was fast. Going to disable the new failing test (they are failing to due GitHub API limit reached on CI).

PR #1201, opened 2020-09-06
fix(command-dev): prevent Docusaurus from launching browser

nickmccurdy:
**- Summary**

Docusaurus automatically launches a browser, but Netlify Dev does not prevent it, leading to confusing scenarios where you'll have the wrong server open in your browser and Netlify options like redirects won't work. This also brings the functionality in line with similar tools like create-react-app.

**- Test plan**

I ran `bin/run` and Docusaurus did not launch the browser as I expected.

Disable Docusaurus' `BROWSER` environment variable.


erezrokah (approved):

erezrokah:
Thanks @nickmccurdy!

Issue #492, opened 2019-08-29
Event triggered functions testing help


wking-io:
Looking for some guidance here:

When trying to debug event triggered functions using `build:invoke` how can I mimic the exact data the production function will have? Specifically I am trying to use the `submission-created` trigger.

I even started trying to log the data in production for the submission, but logging for event triggered functions are broken?

Any help would be appreciated. 😊

sw-yx:
ok first of all no such thing as build:invoke. i think you mean functions:invoke

second of all, thanks for trying it out! it is a relatively new feature. 

the straight answer is we havent coded it yet. i just didnt get round to it and shipped it first to get feedback like this.

if you want to take a crack, the relevant code is here: https://github.com/netlify/cli/blob/master/src/commands/functions/invoke.js#L75-L81

we just need to specialcase submission-created and then figure out what kind of mock data you want to send.

as per [the docs](https://github.com/netlify/cli/blob/master/docs/commands/functions.md#functionsinvoke), you are always welcome to send your own custom payload using the flags described. try it! its a nice feature @DavidWells suggested.

cc @raeesbhatti to handle this, i'm unsubscribing

toobulkeh:
@sw-yx the custom payloads don't work, they're always generating "TODOs" since those are hard coded.

erezrokah:
@toobulkeh looks like there is a bug in the docs. Doubles quotes should be escaped.
`netlify functions:invoke submission-created --payload "{\"foo\": 1}"`.

Closing as the original issue seems to have gone stale.

PR #1209, opened 2020-09-07
docs(command-functions): fix invoke example payload

erezrokah:
See https://github.com/netlify/cli/issues/492#issuecomment-687957987


mraerino (approved):
📝

PR #1215, opened 2020-09-08
Feat: proxy support

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/170

**- Test plan**

Added some units tests.

Tested with https://mitmproxy.org/ and the commands (after copying the certificate `~/.mitmproxy/mitmproxy-ca-cert.pem`):
```
mitmproxy -p 6666
netlify sites:list --httpProxy=http://localhost:6666 --httpProxyCertificateFilename=mitmproxy-ca-cert.pem
```

Allows setting a proxy server (and certificate to allow secure connections)

🐼


docs/commands/status.md
```diff
@@ -38,6 +40,8 @@ netlify status:hooks
 **Flags**
 
 - `debug` (*boolean*) - Print debugging information
+- `httpProxy` (*option*) - Proxy server address to route requests through.
```
ehmicky, line 43: For the other flags, the parenthesis is for the type, not the optionality, is it? I.e. should it say `(*string*)` instead.
Also, I would expect flags to be optional by default, and have `(required)` when not. What do you think?

package.json
```diff
@@ -173,6 +174,7 @@
     "pidtree": "^0.5.0",
     "prettier": "^1.18.2",
     "seedrandom": "^3.0.5",
+    "sinon": "^9.0.3",
```
ehmicky, line 177: Sinon `9` dropped support for Node 8: https://github.com/sinonjs/sinon/blob/master/CHANGELOG.md#900--2020-02-19

src/lib/http-agent.js
```diff
@@ -0,0 +1,77 @@
...
+      port: parseInt(proxyUrl.port) || (scheme === 'http' ? 80 : 443),
+      host: proxyUrl.hostname,
+      timeout: 50,
+      output: 'silent',
+    })
+    if (!open) {
+      throw new Error(`Can't connect to '${httpProxy}'`)
+    }
+  } catch (e) {
+    log(NETLIFYDEVERR, `${httpProxy} is not available`)
```
ehmicky, line 52: Should `e` be printed? Otherwise, the `Can't connect to ...` error message above is not printed.

src/lib/http-agent.js
```diff
@@ -0,0 +1,77 @@
...
+    exit(1)
+  }
+
+  let certificate
+  if (certificateFile) {
+    try {
+      certificate = await fs.readFile(certificateFile)
+    } catch (error) {
+      log(NETLIFYDEVWARN, `could not read certificate file '${certificateFile}'`)
```
ehmicky, line 61: Should `error` be printed here as well? To get some additional context: permission error, does not exist, etc.

docs/commands/status.md
erezrokah: I found it confusing too, but it happens with other string options:
https://cli.netlify.com/commands/deploy/

I think `option` is the type.

I'll open a new issue to sort his.

package.json
erezrokah: 🤦  Good catch. I guess tests are passing on node 8 since I didn't use any API that requires a newer version.
I'll update it

src/lib/http-agent.js
erezrokah: Good point, I'll fix it

src/lib/http-agent.js
erezrokah: I'll separate the error reporting between `wait-port` failing (unknown error) and reaching the timeout.

erezrokah: Fixed in https://github.com/netlify/cli/pull/1215/commits/7a41098f5c04bcf918f3c42e70556ecb4d1c8b56#diff-4c26db14dfb469fcc316ade4b76309c0R41

src/lib/http-agent.js
erezrokah: Fixed in https://github.com/netlify/cli/pull/1215/commits/7a41098f5c04bcf918f3c42e70556ecb4d1c8b56#diff-4c26db14dfb469fcc316ade4b76309c0R66

package.json
erezrokah: Fixed it as a part of the rebase https://github.com/netlify/cli/pull/1215/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R176

ehmicky (approved):
:rocket:

Issue #1191, opened 2020-09-01
Chore: cleanup docs sites scripts


erezrokah:
Related to https://github.com/netlify/cli/issues/304

Some docs site related scripts are under the root `scripts` directory:
https://github.com/netlify/cli/blob/b311b58253e4ac5840dbdd63848d28c4a467f275/scripts/docs.js#L8

We should move them to `site/scripts` for clarity (and also https://github.com/netlify/cli/pull/1190/files#r481076730)

PR #1216, opened 2020-09-08
chore: cleanup site scripts

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1191

**- Test plan**

Not relevant

Moved relevant site scripts under `site/scripts` and removed the unused (so it seems) `algolia.js` script

🐈


package.json
```diff
@@ -38,9 +38,7 @@
     "url": "https://github.com/netlify/cli/issues"
   },
   "scripts": {
-    "build": "run-s build:*",
-    "build:site": "cd site && npm install && npm run build",
-    "ci": "run-s test build",
+    "build:site": "cd site && npm ci && npm run build",
```
ehmicky, line 41: Would `npm install` be better than `npm ci` when running this locally?

erezrokah: I would actually always opt for `npm ci` (so developers always have the same dependencies installed), unless one is explicitly trying to update dependencies.
`npm install` will update lock file if it doesn't match with `package.json`. I would prefer to fail in that case.

ehmicky: This makes sense. :+1:

ehmicky (approved):

PR #1228, opened 2020-09-12
feat(command-dev): Support Docusaurus V2

### LINKED ISSUE #1199: Docusaurus 2 is not detected by netlify dev
**Describe the bug**

`netlify dev` can detect Docusaurus 1 sites, but does not detect Docusaurus 2 sites and treats them like static sites.

**To Reproduce**

1. `npx @docusaurus/init@next init my-website classic`
2. `cd my-website`
3. `netlify dev`

**Configuration**

```
  System:
    OS: macOS 10.15.6
    CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
    Memory: 167.68 MB / 8.00 GB
    Shell: 3.1.2 - /usr/local/bin/fish
  Binaries:
    Node: 14.9.0 - /usr/local/bin/node
    npm: 6.14.8 - /usr/local/bin/npm
  npmGlobalPackages:
    netlify-cli: 2.61.1
```

**Expected behavior**

Docusaurus 2 sites should be detected like Docusaurus 1 sites. This could be achieved either by adding 2 support to the existing detector with dynamic commands or adding a new detector for only 2 support.

**CLI Output**

```
◈ Netlify Dev ◈
◈ No app server detected and no "command" specified
◈ Running static server from "cli/site/dist"
```

**Additional context**

It seems like `docusaurus-start` has changed to `docusaurus start`.

monsonjeremy:
**- Summary**

This PR closes #1199. With Docusaurus V2, the dependencies and commands have changed. In order to support the new version I've updated the dependency check to look for either the V1 dependency or the V2 dependency. Depending on whether the V1 or V2 dependency is installed, the correct command will be added to the `scanScripts` call.

**- Test plan**

```sh
npx @docusaurus/init@next init my-website classic
cd my-website
netlify dev
```

With these changes, running `netlify dev` from within a Docusaurus V2 website will work as expected.

`feat(command-dev): Support Docusaurus V2`

IMG_1304

### DIFF
--- a/src/detectors/docusaurus.js
+++ b/src/detectors/docusaurus.js
@@ -1,16 +1,32 @@
 const { hasRequiredDeps, hasRequiredFiles, getYarnOrNPMCommand, scanScripts } = require('./utils/jsdetect')
 module.exports = function() {
   // REQUIRED FILES
-  if (!hasRequiredFiles(['package.json', 'siteConfig.js'])) return false
+  if (!hasRequiredFiles(['package.json'])) return false
+
+  const hasV1 = hasRequiredDeps(['docusaurus'])
+  const hasV2 = hasRequiredDeps(['@docusaurus/core'])
+
   // REQUIRED DEPS
-  if (!hasRequiredDeps(['docusaurus'])) return false
+  if (!hasV1 && !hasV2) return false
+
+  // REQUIRED CONFIG FILE
+  const hasConfigFile = hasV1 ? hasRequiredFiles(['siteConfig.js']) : hasRequiredFiles(['docusaurus.config.js'])
+
+  if (!hasConfigFile) return false
 
   /** everything below now assumes that we are within gatsby */
 
-  const possibleArgsArrs = scanScripts({
+  const v1Command = {
     preferredScriptsArr: ['start'],
     preferredCommand: 'docusaurus-start',
-  })
+  }
+
+  const v2Command = {
+    preferredScriptsArr: ['start'],
+    preferredCommand: 'docusaurus start',
+  }
+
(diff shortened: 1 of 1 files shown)


src/detectors/docusaurus.js
```diff
@@ -2,15 +2,25 @@ const { hasRequiredDeps, hasRequiredFiles, getYarnOrNPMCommand, scanScripts } =
 module.exports = function() {
   // REQUIRED FILES
   if (!hasRequiredFiles(['package.json', 'siteConfig.js'])) return false
+
+  const hasV1 = hasRequiredDeps(['docusaurus'])
```
erezrokah, line 6: Looks like `npx @docusaurus/init@next init my-website bootstrap` doesn't generate a `siteConfig.js` file which means we won't reach this code:
https://v2.docusaurus.io/docs/installation#project-structure
https://docusaurus.io/docs/en/installation#verifying-installation

should we check for `docusaurus.config.js` instead?

erezrokah (changes requested):
Thanks @monsonjeremy, I think one more check is needed. Please let me know what you think.

src/detectors/docusaurus.js
monsonjeremy: Nice catch! Updating now!

erezrokah: Thanks @monsonjeremy, should we put the config check under the `hasV1` and `hasV2` flags for clarity?

erezrokah (changes requested):

src/detectors/docusaurus.js
monsonjeremy: Yeah, that's not a bad idea. Let me change that now!

erezrokah (approved):

PR #1229, opened 2020-09-13
README: use Snyk badge for deps security

lirantal:
**- Summary**

Replacing outdated david-dm's badge with Snyk's for security insights.

**- Test plan**

N/A

docs(readme): use Snyk badge for deps security

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Netlify CLI
 
-[![npm version][npm-img]][npm] [![downloads][dl-img]][dl] [![netlify-status][netlify-img]][netlify] [![security][snyk-img]][snyk][![FOSSA Status](https://app.fossa.com/api/projects/custom%2B17679%2Fgit%40github.com%3Anetlify%2Fcli.git.svg?type=shield)](https://app.fossa.com/projects/custom%2B17679%2Fgit%40github.com%3Anetlify%2Fcli.git?ref=badge_shield)
+[![npm version][npm-img]][npm] [![downloads][dl-img]][dl] [![netlify-status][netlify-img]][netlify] [![dependencies][david-img]][david] [![security][snyk-img]][snyk][![FOSSA Status](https://app.fossa.com/api/projects/custom%2B17679%2Fgit%40github.com%3Anetlify%2Fcli.git.svg?type=shield)](https://app.fossa.com/projects/custom%2B17679%2Fgit%40github.com%3Anetlify%2Fcli.git?ref=badge_shield)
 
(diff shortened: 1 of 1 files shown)


erezrokah (commented):
Thanks @lirantal, seems like each badge can serve a different purpose, should we keep `david` and add `snyk`?

lirantal:
It is redundant in a way, since Snyk provides both the value of security fixes, as well as general updates to keep your dependencies from a version drift.

WRT to david-dm, while I don't think there was any formal announcement, the project seems stale and out of date since 2018, with some outages happening, and the `david` code project itself has out of date prod and devDeps so while I haven't explored further, it seems either unmaintained or very lowly so. 

What I could find is a reference such as issue 123 on https://github.com/alanshaw/david:
image

The project being unmaintained isn't a bad thing on its own though, nor do "out of date" deps a bad thing.

erezrokah:
> It is redundant in a way, since Snyk provides both the value of security fixes, as well as general updates to keep your dependencies from a version drift.

Is there a way to relay both security vulnerabilities information and out of date dependencies information in the Snyk badge?
At the moment it only shows vulnerabilities (pending https://github.com/netlify/gh-release-fetch/issues/1) which hides the dependencies information.

> The project being unmaintained isn't a bad thing on its own though, nor do "out of date" deps a bad thing.

Seems like there is some in progress work to bring the project up to speed by the maintainer:
https://github.com/alanshaw/david/pull/172
with alpha versions being released:
https://www.npmjs.com/package/david
https://github.com/alanshaw/david/releases

The repo for the site itself https://github.com/alanshaw/david-www needs some work, maybe it's pending https://github.com/alanshaw/david/pull/172?
Don't think we suffered that much from the downtimes, but I'll probably need to look into some of these:
https://shields.io/category/dependencies
as an alternative first before switching.

lirantal:
Indeed we don't yet have a badge for out of date dependencies. It's not something to planned, so if you particularly need and want to badge this out then david-dm or alternatives. I wasn't really doing any DD on david-dm. Just browsed the repo a bit upon your ask on this and my memory of the project not being actively maintained. Good to keep as is if needed.

I can update the PR to include both david-dm and the snyk badge to get both.

erezrokah:
> I can update the PR to include both david-dm and the snyk badge to get both.

Thank you for clarifying, until I can make a more informed decision it would be great to have both.

lirantal:
Definitely. PR updated to include both 👍

erezrokah (approved):

Issue #1236, opened 2020-09-14
Deploy edge handlers locally


mraerino:
**Is your feature request related to a problem? Please describe.**

Right now, the edge handler plugin will only put the bundled edge handler into a local directory when running via `netlify build` in the CLI. We need to upload bundled edge handlers when deploying locally.

**Describe the solution you'd like**

`netlify deploy` should detect whether there are bundled edge handlers in `.netlify/edge-handlers`

If there is a manifest & a bundle, it should use the upload method from the edge handler plugin to upload the bundle to the api: https://github.com/netlify/netlify-plugin-edge-handlers/blob/master/src/upload.js

- Read the bundle file and pass the file buffer as `buf`
- Read the manifest and parse it as json, pass the object as `info`
- The access token should be valid for the site that is being deployed to

This should happen after the deploy has been created, but before the deploy is complete (and post-processing starts).

**Describe alternatives you've considered**

The edge handler plugin could upload the bundle directly, but when `netlify build` runs locally, we don't have a deploy object yet.

**Additional context**

API endpoint used by the plugin: https://bitballoon-openapi.netlify.app/external#operation/Edge%20Handlers-create

**Can you submit a pull request?**

Yes, but @erezrokah might be helping with this.

erezrokah:
I did some initial work in https://github.com/netlify/cli/tree/feat/edge_handlers_deploy (it is based on https://github.com/netlify/cli/pull/1240) to add deploying edge handlers using the CLI.
The deploy doesn't work at the moment since we need to deploy edge handlers before the deploy is complete.
The CLI uses a [`deploy` function](https://github.com/netlify/js-client#deploy--await-clientdeploysiteid-builddir-opts) from `js-client` which returns after the deploy is complete.

We can:
1. Have `js-client` handle deploying edge handlers (this is the current pattern for functions).
2. Use the `statusCb` mechanism `js-client` has to hook into deploy events. That would require supporting async callbacks in `js-client.`

mraerino:
third option:
patching js-client to accept a `deployId` parameter in the `deploy` function.
if that is passed it can skip creating the deploy.

that way we could create the deploy ourselves, upload handlers and then trigger the js-client function.

mraerino:
this is what buildbot does, since it will be called with an existing deploy: https://github.com/netlify/open-api/blob/master/go/porcelain/deploy.go#L239-L262

mraerino:
[...]

Issue #1227, opened 2020-09-12
Manual deployment hangs with a deployment folder that starts with a dot


d4h0:
**Describe the bug**

Executing 'netlify deploy -d .build' does not find any files, and hangs during the upload.

Here is the output of a tried deployment:

```
Deploying to draft URL...
✔ Finished hashing 0 files
✔ CDN requesting 0 files
✔ Finished uploading 0 assets
⧇ Waiting for deploy to go live...
```

You see, no files were found (even so there are files).

**To Reproduce**

1. Setup a Netlify project
2. on Linux, create a hidden directory (a directory that starts with a dot, i.e. ".deploy")
3. Put some HTML files into the hidden directory
4. Start the deployment of the hidden folder

**Configuration**

> - If possible, please copy/paste below your `netlify.toml`.

I'm using the default `netlify.toml` (I didn't change anything).

> - Please enter the following command in a terminal and copy/paste its output:

```
  System:
    OS: Linux 5.8 Arch Linux
    CPU: (8) x64 Intel(R) Core(TM) i7-4710MQ CPU @ 2.50GHz
    Memory: 1.96 GB / 15.52 GB
    Container: Yes
    Shell: 5.0.18 - /bin/bash
  Binaries:
    Node: 14.10.1 - /usr/bin/node
    npm: 6.14.8 - /usr/bin/npm
    Watchman: 4.9.0 - /usr/bin/watchman
  npmGlobalPackages:
    netlify-cli: 2.61.2
```

**Expected behavior**

The CLI finds files in the deployment folder, and the upload succeeds (at least, if there is a problem, there should be an error message)

**CLI Output**

See above

**Additional context**

After changing the name of the deployment folder from ".build" to "build", everything works as expected

monsonjeremy:
@erezrokah Any advice on how this should be handled? 
Here's the discovery I did:

It looks like this is the desired behavior, here the `hashFiles` function is called using the given `dir`
https://github.com/netlify/js-client/blob/4cb1f240bac527305ff98032a801accf22dea5bf/src/deploy/index.js#L49

The `hashFiles` function uses the `defaultFilter` which ignores hidden files/directories
https://github.com/netlify/js-client/blob/4cb1f240bac527305ff98032a801accf22dea5bf/src/deploy/hash-files.js#L20
https://github.com/netlify/js-client/blob/4cb1f240bac527305ff98032a801accf22dea5bf/src/deploy/util.js#L12

[...]

PR #1240, opened 2020-09-15
refactor: extract deploy command logic into separate functions

erezrokah:
**- Summary**

The PR refactors the `deploy` command code by extracting most of the logic into functions.
Some logic mutates objects so I didn't extract it yet since I wanted the functions to be pure.

**- Test plan**

Tested manually the different deploy flags, and actually exposed old bugs in the process (which I fixed).

Should not change behaviour

🐼 

Related to https://github.com/netlify/cli/issues/1236 and https://github.com/netlify/cli/issues/1227 - I initially started looking into those issues and found the code very hard to follow.

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -4,7 +4,7 @@ const path = require('path')
 const chalk = require('chalk')
 const { flags } = require('@oclif/command')
 const get = require('lodash.get')
-const fs = require('fs')
+const fs = require('fs-extra')
 const prettyjson = require('prettyjson')
 const ora = require('ora')
 const logSymbols = require('log-symbols')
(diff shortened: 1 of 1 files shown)


src/commands/deploy.js
```diff
@@ -4,7 +4,7 @@ const path = require('path')
 const chalk = require('chalk')
 const { flags } = require('@oclif/command')
 const get = require('lodash.get')
-const fs = require('fs')
+const fs = require('fs-extra')
```
erezrokah, line 7: So we can switch to `async` interfaces

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
 const LinkCommand = require('./link')
 const { NETLIFYDEV, NETLIFYDEVLOG, NETLIFYDEVERR } = require('../utils/logo')
 
+const triggerDeploy = async ({ api, siteId, siteData, log, error }) => {
```
erezrokah, line 19: Extracted from https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L89

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+    if (err.status === 404) {
+      error('Site not found. Please rerun "netlify link" and make sure that your site has CI configured.')
+    } else {
+      error(err.message)
+    }
+  }
+}
+
+const getDeployFolder = async ({ flags, config, site, siteData, log }) => {
+  let deployFolder
```
erezrokah, line 35: Extracted from https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L105 and https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L125

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+        filter: input => path.resolve(process.cwd(), input),
+      },
+    ])
+    deployFolder = promptPath
+  }
+
+  return deployFolder
+}
+
+const validateDeployFolder = async ({ deployFolder, error }) => {
```
erezrokah, line 62: Old version was:
https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L463

Which was broken in multiple places:
1. Using `e.status` instead of `e.code` (fixed in the new version)
https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L468
2. Referencing `stat.isDirectory` instead of `stat.isDirectory()` https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L480

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+    throw e
+  }
+
+  if (!stat.isDirectory()) {
+    return error('Deploy target must be a path to a directory')
+  }
+  return stat
+}
+
+const getFunctionsFolder = ({ flags, config, site, siteData }) => {
```
erezrokah, line 84: Extracted from https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L114

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+    functionsFolder = path.resolve(process.cwd(), flags['functions'])
+  } else if (funcConfig) {
+    functionsFolder = path.resolve(site.root, funcConfig)
+  } else if (get(siteData, 'build_settings.functions_dir')) {
+    functionsFolder = path.resolve(site.root, get(siteData, 'build_settings.functions_dir'))
+  }
+  return functionsFolder
+}
+
+const validateFunctionsFolder = async ({ functionsFolder, log, error }) => {
```
erezrokah, line 98: Extracted from https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L114 and fixed to reference `stat.isDirectory()`

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+    error('Functions folder must be a path to a directory')
+  }
+
+  return stat
+}
+
+const validateFolders = async ({ deployFolder, functionsFolder, error, log }) => {
+  const deployFolderStat = await validateDeployFolder({ deployFolder, error })
+  const functionsFolderStat = await validateFunctionsFolder({ functionsFolder, error, log })
+  return { deployFolderStat, functionsFolderStat }
```
erezrokah, line 128: We need the return value of `validateFunctionsFolder`. If it's undefined (since the folder doesn't exist) we don't pass `functionsFolder` to the `js-client`.
This was done here:
https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L167

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+}
+
+const runDeploy = async ({
```
erezrokah, line 131: Extracted from here:
https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L181

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+        return
+      }
+    }
+  }
+
+  const siteUrl = results.deploy.ssl_url || results.deploy.url
+  const deployUrl = get(results, 'deploy.deploy_ssl_url') || get(results, 'deploy.deploy_url')
+  const logsUrl = `${get(results, 'deploy.admin_url')}/deploys/${get(results, 'deploy.id')}`
+
+  return {
```
erezrokah, line 209: This information was used here https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L241 and here https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L263

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
+    siteId: results.deploy.site_id,
+    siteName: results.deploy.name,
+    deployId: results.deployId,
+    siteUrl,
+    deployUrl,
+    logsUrl,
+  }
+}
+
+const printResults = ({ flags, results, deployToProduction, log, logJson, exit }) => {
```
erezrokah, line 220: Extracted from https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L245

src/commands/deploy.js
```diff
@@ -78,216 +321,59 @@ class DeployCommand extends Command {
...
-        fnDir: functionsFolder,
-        statusCb: flags.json || flags.silent ? () => {} : deployProgressCb(),
-        draft: !deployToProduction && !alias,
-        message: flags.message,
-        deployTimeout: flags.timeout * 1000 || 1.2e6,
-        syncFileLimit: 100,
-        branch: alias,
+    log(
+      prettyjson.render({
+        'Deploy path': deployFolder,
```
erezrokah, line 341: We don't need the conditions here:
https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L145
and here https://github.com/netlify/cli/blob/39da8081eaca6e2c8ba58aa782541199e0df48fc/src/commands/deploy.js#L150 and `prettyjson` doesn't print `undefined` values

src/commands/deploy.js
ehmicky: Could we use `utils.promisify` instead?
`fs-extra` includes `graceful-fs` which monkey patches global variables.

src/commands/deploy.js
```diff
@@ -16,20 +16,263 @@ const SitesCreateCommand = require('./sites/create')
...
     const alias = flags.alias || flags.branch
 
     if (flags.branch) {
-      this.warn('--branch flag has been renamed to --alias and will be removed in future versions')
+      warn('--branch flag has been renamed to --alias and will be removed in future versions')
     }
 
     const deployToProduction = flags.prod
-    if (deployToProduction && alias) {
-      this.error(`--prod and --alias flags cannot be used at the same time`)
```
erezrokah, line 31: Replaced this with https://github.com/netlify/cli/pull/1240/files#diff-e9001bf37e31e33db8a717da3ac59a69R474

src/commands/deploy.js
erezrokah: Sure, I can also open an issue to cleanup `fs-extra` from the codebase

erezrokah: See ef9a68ae34cf4357d2bff859671fd22db3c7fdc1 and See https://github.com/netlify/cli/issues/1241

ehmicky: Oh I did not know we used it elsewhere :)

src/commands/deploy.js
ehmicky: :man_facepalming:

src/commands/deploy.js
```diff
@@ -16,20 +17,265 @@ const SitesCreateCommand = require('./sites/create')
...
+
+  if (stat && !stat.isDirectory()) {
+    error('Functions folder must be a path to a directory')
+  }
+
+  return stat
+}
+
+const validateFolders = async ({ deployFolder, functionsFolder, error, log }) => {
+  const deployFolderStat = await validateDeployFolder({ deployFolder, error })
```
ehmicky, line 129: Can we run those two `async` functions in parallel (`Promise.all()`)?

src/commands/deploy.js
```diff
@@ -16,20 +17,265 @@ const SitesCreateCommand = require('./sites/create')
...
+      log('Deploying to draft URL...')
+    }
+
+    results = await api.deploy(siteId, deployFolder, {
+      configPath,
+      fnDir: functionsFolder,
+      statusCb: flags.json || flags.silent ? () => {} : deployProgressCb(),
+      draft: !deployToProduction && !alias,
+      message: flags.message,
+      deployTimeout: flags.timeout * 1000 || 1.2e6,
```
ehmicky, line 177: Should we extract that `1.2e6` into a top-level constant variable?

ehmicky (approved):
Thanks @erezrokah, this is much cleaner! :tada:

src/commands/deploy.js
erezrokah: `validateFunctionsFolder` prints a warning message when the folder doesn't exist, while `validateDeployFolder` exists.
Didn't want to run into a case where we print a warning then exit with another error.

src/commands/deploy.js
erezrokah: Sure going to update it

src/commands/deploy.js
ehmicky: Makes sense :+1: 
This is micro-optimization anyway since those I/O calls are quite fast.

src/commands/deploy.js
erezrokah: Done in https://github.com/netlify/cli/pull/1240/commits/f00db3f9b558720907882f3fe2df9e41e276ea9d#diff-e9001bf37e31e33db8a717da3ac59a69R179

PR #1254, opened 2020-09-19
Fix invalid static parameters

cassidoo:
**- Summary**

There is a [bug in the next-on-netlify](https://github.com/netlify/next-on-netlify/issues/41) package being caused by this. Requests are being routed to a catch-all page, when they shouldn't be. This fixes that!

**- Test plan**

(see linked issue above)

Fix: Request URL for static pathnames decoded properly

image


erezrokah (approved):
Thanks for the fix @cassidoo, just added a test case.

Issue #179, opened 2018-11-15
Creating a deploy should run the command from netlify.toml


sam3d:
**- Do you want to request a *feature* or report a *bug*?**

Feature

**- What is the current behavior?**

When the `netlify.toml` resembles something like this:

```toml
[build]
publish="dist"
command="npm run build"
```

And `netlify deploy` is run, it will use the `publish` key to automatically resolve that deploy directory. However, the `command` value is ignored.

**- What is the desired behavior?**

If the `command` value is present, run it before a deploy to ensure a local build has completed before running a deploy.

**- Please mention your node.js, and operating system version.**
```
➜  node -v
v8.12.0
➜ npm -v
5.10.0
➜ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14
BuildVersion:	18A391
```

bcomnes:
I can see this going both ways,  do one thing well vs try to do the right thing.  I'm open to making this an option.

wafisher:
+1

I'm also confused why that value is ignored. Is it only for Git deploys?

intcreator:
I'm generating static files locally with a WordPress plugin.  That's at least one use case for wanting to configure a build locally (I have a subdirectory with a Node app which requires building)

PR #1266, opened 2020-09-21
Feat: run build command when deploying

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/179

Adds a `--build` flag to run the build command before deploying.

**- Test plan**

- [x] Going to add a test case 

Support running the build command before deploying

😸


ehmicky (approved):
:tada:

Issue #1241, opened 2020-09-15
Chore: remove `fs-extra` as a dependecy


erezrokah:
**Describe the bug**

We can opt out of `fs-extra` by using `util.promisify`

We should remove it as a dependency and update the relevant places in the code.

**Additional context**

Context: https://github.com/netlify/cli/pull/1240#discussion_r488829294

PR #1271, opened 2020-09-23
chore: remove fs-extra

JCMais:
Closes https://github.com/netlify/cli/issues/1241

**- Summary**

See issue above.

I've left a few comments about some changes that need to be addressed before this is merged.

I also wrote this using Node.js 14, so I need to run the tests again using Node.js 8 as it's probably going to yell at me. 😄 I will do that when I find some time tomorrow.

**- Test plan**

This PR adds no new features/changes other than removing all usages of fs-extra, so the existing test suite was sufficient.

There are some tests failing locally here, but it should not be related to these changes, as they were already failing before I started working on this. I will take a look later into this.

remove `fs-extra` and use `util.promisify` instead

baby capybara


site/sync.js
```diff
@@ -1,25 +1,32 @@
...
+  await Promise.all(files.map(async f => (await fs.stat(f)).isDirectory() && readDir(f, allFiles)))
   return allFiles
 }
 
 async function syncLocalContent() {
   const src = path.join(config.docs.srcPath)
   const destination = path.join(config.docs.outputPath)
 
-  await fs.copy(src, destination)
+  await fs.copyFile(src, destination)
```
JCMais, line 29: ⚠ This is not recursive like the one from `fs-extra`, so will not really work. I'm open to suggestions on how to handle this.

site/watch.js
```diff
@@ -41,14 +49,14 @@ function getFullPath(filePath) {
 
 function syncFile(filePath) {
   const { src, destination } = getFullPath(filePath)
-  return fs.copy(src, destination).then(() => {
+  return fs.copyFile(src, destination).then(() => {
```
JCMais, line 52: ⚠ This is not recursive like the one from `fs-extra`

I don't think this is really an issue here as it will only be called with file paths.

src/commands/functions/create.js
```diff
@@ -1,9 +1,10 @@
-const fs = require('fs-extra')
+const fs = require('fs')
```
JCMais, line 1: This file is only using `*Sync`/`createWriteStream` methods, so no need for `util.promisify`.

However, the code is calling `mkdir` with the `recursive` flag which requires Node.js >= v10.12.0.

What would be the best way to handle this?

src/lib/exec-fetcher.js
```diff
@@ -26,12 +27,13 @@ const isExe = (mode, gid, uid) => {
 }
 
 const execExist = async binPath => {
-  const binExists = await fs.exists(binPath)
-  if (!binExists) {
-    return false
+  try {
```
JCMais, line 30: `fs.exists` was not really needed here. If the file does not exist `statSync` will throw an error.

src/lib/exec-fetcher.test.js
```diff
@@ -61,5 +68,7 @@ packages.forEach(({ packageName, execName, execArgs, pattern, extension }) => {
 })
 
 test.afterEach(async t => {
-  await fs.remove(t.context.binPath)
+  await fs.rmdir(t.context.binPath, {
```
JCMais, line 71: ⚠ the `recursive` flag for `rmdir` is experimental and was added only with Node.js v12, so probably out of scope for us, should we use `rimraf` here?

src/utils/proxy.js
```diff
@@ -35,7 +39,7 @@ async function getStatic(pathname, publicFolder) {
   for (const i in alternatives) {
     const p = alternatives[i]
     try {
-      const pathStats = await fs.stat(p)
+      const pathStats = await fsStatAsync(p)
```
JCMais, line 42: This is the only async call to fs.* APIs in this file, it could just become `fs.statSync`, as this should be pretty fast, or we could just convert all other calls to use their async alternative.

tests/utils/siteBuilder.js
```diff
@@ -22,74 +33,75 @@ const createSiteBuilder = ({ siteName }) => {
...
     },
     withEdgeHandlers: ({ handlers }) => {
       const dest = path.join(directory, 'edge-handlers', 'index.js')
       tasks.push(async () => {
-        await fs.ensureFile(dest)
         const content = Object.entries(handlers)
           .map(([event, handler]) => {
             return `export const ${event} = ${handler.toString()}`
           })
           .join('\n')
```
JCMais, line 65: Unrelated, but this is the only join not using `os.EOL`.

tests/utils/siteBuilder.js
```diff
@@ -22,74 +33,75 @@ const createSiteBuilder = ({ siteName }) => {
       const dest = path.join(directory, pathPrefix, 'netlify.toml')
       const content = toToml(config, { space: 2 })
       tasks.push(async () => {
-        await fs.ensureFile(dest)
+        await ensureDir(path.dirname(dest))
```
JCMais, line 36: We only need to ensure that the directory exists, as `writeFile` already creates the file if it does not exist (default mode is 'w').

tests/utils/siteBuilder.js
```diff
@@ -106,7 +118,7 @@ const createSiteBuilder = ({ siteName }) => {
       return builder
     },
     cleanupAsync: async () => {
-      await fs.remove(directory).catch(e => {
+      await fs.rmdir(directory, { recursive: true }).catch(e => {
```
JCMais, line 121: Same than above about the `recursive` flag.

src/lib/exec-fetcher.test.js
ehmicky: Yes (or [`del`](https://github.com/sindresorhus/del) which is based on `rimraf`). The CLI minimum Node.js version is [`8.17.0`](https://github.com/netlify/cli/blob/d949407ef05060aaf2aae8bd227892e4646520ef/package.json#L14).

tests/utils/siteBuilder.js
```diff
@@ -1,9 +1,20 @@
...
+const util = require('util')
+
+const fs = {
+  mkdir: util.promisify(_fs.mkdir),
+  writeFile: util.promisify(_fs.writeFile),
+  rmdir: util.promisify(_fs.rmdir),
+}
+
+const ensureDir = async path => {
+  return fs.mkdir(path, { recursive: true })
```
ehmicky, line 16: [`recursive` was added](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback) in Node `10.12.0`, but we unfortunately need to support Node 8. Would [`make-dir`](https://github.com/sindresorhus/make-dir) work instead?

src/lib/exec-fetcher.test.js
```diff
@@ -1,10 +1,17 @@
...
 const path = require('path')
-const fs = require('fs-extra')
+const _fs = require('fs')
 const tempDirectory = require('temp-dir')
 const { v4: uuid } = require('uuid')
+const util = require('util')
+
 const { getExecName, shouldFetchLatestVersion, fetchLatestVersion } = require('./exec-fetcher')
 
+const fs = {
```
ehmicky, line 10: Would it make sense to group all the `const fs = { ... util.promisify(...), ... }` into a single file?

```js
const fs = require('../utils/fs')
```

src/utils/proxy.js
erezrokah: How about using the sync method and opening a new issue to refactor this file to use `async` functions?

tests/utils/siteBuilder.js
erezrokah: This file is processed only by our infrastructure at the moment (not also locally like other files).

erezrokah (changes requested):
Thanks @JCMais, nice work.
Added some comments.

src/lib/exec-fetcher.test.js
JCMais: Yep, this is much better.

src/lib/exec-fetcher.js
JCMais: Yep, will do it.

src/lib/exec-fetcher.test.js
JCMais: `del` seems like a better alternative.

src/utils/proxy.js
JCMais: Yep, I will open one as soon as I get a chance 👍

tests/utils/siteBuilder.js
JCMais: yep, I will change all calls to `fs.mkdir(path, { recursive: true })` to use `make-dir` instead.

site/sync.js
```diff
@@ -1,25 +1,32 @@
 const path = require('path')
-const fs = require('fs-extra')
-const config = require('./config')
+const _fs = require('fs')
```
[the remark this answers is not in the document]
JCMais: Thanks for the heads up @erezrokah!

I'm going to change these calls to use `fs.promises` directly.

src/lib/http-agent.js
```diff
@@ -1,8 +1,14 @@
...
-const { NETLIFYDEVERR, NETLIFYDEVWARN } = require('../utils/logo')
+const _fs = require('fs')
+const util = require('util')
 const waitPort = require('wait-port')
 const { HttpsProxyAgent } = require('https-proxy-agent')
-const fs = require('fs-extra')
+
+const { NETLIFYDEVERR, NETLIFYDEVWARN } = require('../utils/logo')
+
+const fs = {
```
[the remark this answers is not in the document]
JCMais: Yep, this is much better! I didn't see this file, I will use it instead. 😅

src/utils/proxy.js
JCMais: done: https://github.com/netlify/cli/issues/1275

JCMais:
addressed the requested changes.

`del` could have been a dev dependency, as it's only used in tests, but I decided to create a separated function inside `lib/fs.js` just in case it would be useful later on. This is the reason it is a normal dependency.

Almost all tests are green here now, the tests that were failing before were due to missing the `NETLIFY_AUTH_TOKEN` environment variable. The only one not working now is the one related to Edge Handlers, most probably because I'm don't have access to the preview.

package.json
```diff
@@ -88,14 +88,14 @@
     "cookie": "^0.4.0",
     "copy-template-dir": "^1.4.0",
     "debug": "^4.1.1",
+    "del": "5.1.0",
```
ehmicky, line 91:
```suggestion
    "del": "^5.1.0",
```

tests/utils/siteBuilder.js
```diff
@@ -1,10 +1,15 @@
...
 const path = require('path')
 const { v4: uuidv4 } = require('uuid')
 const tempDirectory = require('temp-dir')
 const os = require('os')
 
+const fs = require('../../src/lib/fs')
+
+const ensureDir = async path => {
+  return fs.mkdirRecursiveAsync(path)
+}
```
JCMais, line 11: Looking at this again now, this function probably can be removed and `fs.mkdirRecursiveAsync` used directly.

site/fs.js
```diff
@@ -0,0 +1,49 @@
...
+const rimraf = require('rimraf')
+const util = require('util')
+
+const rimrafAsync = util.promisify(rimraf)
+
+const copyDirRecursiveAsync = async (src, dest) => {
+  try {
+    fs.mkdir(dest, { recursive: true })
+  } catch (_) {
+    // ignore erros for mkdir
```
erezrokah, line 12:
```suggestion
    // ignore errors for mkdir
```

site/fs.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  const childrenItems = await fs.readdir(src)
```
erezrokah, line 15: This has the potential to create a lot of promises.
I don't think we should optimise this yet since it's only used for our docs site that doesn't contain many of files.

src/commands/functions/create.js
```diff
@@ -216,7 +218,7 @@ async function downloadFromURL(flags, args, functionsDir) {
   }
 
   try {
-    fs.mkdirSync(fnFolder, { recursive: true })
+    mkdirRecursiveSync(fnFolder)
```
erezrokah, line 221: Since this function is already marked `async` should we use the asynchronous version of `mkdir`?

src/lib/exec-fetcher.js
erezrokah, line 30: Nice cleanup!

src/lib/fs.js
```diff
@@ -23,4 +25,18 @@ const fileExistsAsync = async filePath => {
   }
 }
 
-module.exports = { statAsync, readFileAsync, readFileAsyncCatchError, writeFileAsync, fileExistsAsync }
+const mkdirRecursiveSync = path => makeDir.sync(path)
+const mkdirRecursiveAsync = async path => makeDir(path)
```
erezrokah, line 29:
```suggestion
const mkdirRecursiveAsync = path => makeDir(path)
```

src/lib/fs.js
```diff
@@ -23,4 +25,18 @@ const fileExistsAsync = async filePath => {
...
+
+const rmdirRecursiveAsync = async path => del(path, { force: true })
```
erezrokah, line 31:
```suggestion
const rmdirRecursiveAsync = path => del(path, { force: true })
```

src/utils/proxy.js
```diff
@@ -35,7 +36,7 @@ async function getStatic(pathname, publicFolder) {
   for (const i in alternatives) {
     const p = alternatives[i]
     try {
-      const pathStats = await fs.stat(p)
+      const pathStats = fs.statSync(p)
```
erezrokah, line 39: Can remove the `async` keyword from the function?

tests/utils/siteBuilder.js
erezrokah: Good point

erezrokah (changes requested):
Thanks for the quick follow up @JCMais, add a few minor comments and suggestions.
You're correct about Edge Handlers, they are not available for all accounts yet.

site/fs.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs').promises
+const path = require('path')
+const rimraf = require('rimraf')
+const util = require('util')
+
+const rimrafAsync = util.promisify(rimraf)
+
+const copyDirRecursiveAsync = async (src, dest) => {
+  try {
+    fs.mkdir(dest, { recursive: true })
```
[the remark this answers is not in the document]
JCMais: 🤦‍♂️

src/commands/functions/create.js
JCMais: I kept using sync in there to keep it consistent with the other `fs.*` calls.

src/lib/fs.js
JCMais: weird, eslint is not giving me a warning about this locally.

erezrokah: I just added that rule in https://github.com/netlify/cli/pull/1278 as a follow up to this PR

JCMais: I thought eslint was being crazy or something, it makes sense now. I will rebase this on top of master and fix the lint erros.

src/commands/functions/create.js
JCMais: Let me know if you want me to change these fs calls to async or just keep it the way it is. @erezrokah 

I don't think this is a big issue tho, probably can be done in another PR.

JCMais:
Rebased on top of `master` and addressed all your comments @erezrokah, thanks for the review!

src/commands/functions/create.js
erezrokah: No worries you can keep it `sync`

site/fs.js
```diff
@@ -0,0 +1,49 @@
...
+      if (itemStat.isFile()) {
+        fs.copyFile(srcPath, destPath)
+      } else {
+        await copyDirRecursiveAsync(srcPath, destPath)
+      }
+    })
+  )
+}
+
+const ensureFilePathAsync = async filePath => {
```
ehmicky, line 33: We probably should let consumers pass the `dirname` directly instead.
Like this, functions like `copyDirRecursiveAsync()` will be able to use it.

site/fs.js
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs').promises
+const path = require('path')
+const rimraf = require('rimraf')
+const util = require('util')
+
+const rimrafAsync = util.promisify(rimraf)
+
+const copyDirRecursiveAsync = async (src, dest) => {
```
ehmicky, line 8: We probably should be using a library instead of implementing this.
[`cpy`](https://github.com/sindresorhus/cpy) could work, but alternatives might work too.

site/fs.js
```diff
@@ -0,0 +1,49 @@
...
+
+  await Promise.all(
+    childrenItems.map(async item => {
+      const srcPath = path.join(src, item)
+      const destPath = path.join(dest, item)
+
+      const itemStat = await fs.lstat(srcPath)
+
+      if (itemStat.isFile()) {
+        fs.copyFile(srcPath, destPath)
```
ehmicky, line 25: `await` should be used.

site/fs.js
JCMais: changed to `ensurePathExistsAsync = async path => {`

site/fs.js
JCMais: good catch

site/fs.js
JCMais: I didn't use one as this is a simple to implement and only really used in a single place, `./site/sync.js`.

But yeah, an external dep would be better, as there are not even tests for that implementation I wrote. Let me know if you want me to change to `cpy`.

JCMais: I agree completely, I meant simple for this use case in specific, which is using for syncing files during development, not something that is going to be shipped. I'm going to add `cpy` and use it instead inside `site/`.

JCMais:
By the way, if we look further into this change, it probably makes sense to use graceful-fs directly instead of just fs. As just using fs is not going to handle some scenarios that were being handled before when using `fs-extra` (which uses `graceful-fs` internally).

erezrokah:
@JCMais, this PR brings value in the form of:
1. Doing some refactoring and wrapping `fs` with our own layer (making it easier to make the relevant changes once we drop Node 8 support).
2. Using Node core when possible and falling back to specific dependencies when needed seems to make more sense in our case.

It would be great if you could follow up with the remaining changes.

JCMais:
Thanks @erezrokah, I will do that as soon as I get a chance today.

JCMais:
added `cpy` as a dependency to `./site` and used it instead of the custom implementation.

erezrokah:
> I have not rebased with master yet, will do after the latest changes are reviewed.

Latest changes are good, thanks! Can you follow up with a rebase? I'm re-running the tests to verify

site/fs.js
```diff
@@ -0,0 +1,26 @@
...
+const util = require('util')
+
+const rimrafAsync = util.promisify(rimraf)
+
+const copyDirRecursiveAsync = (src, dest) => cpy(src, dest)
+
+const ensurePathExistsAsync = async path => {
+  try {
+    await fs.mkdir(path, { recursive: true })
+  } catch (_) {
```
ehmicky, line 13: `fs.mkdir()` [does not throw](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback) when `path` does not exist providing `recursive: true` is passed.

ehmicky (approved):

site/fs.js
JCMais: good to know!
I will remove the try/catch and rebase with master.

JCMais:
the branch has been rebased with master

erezrokah:
Sorry for the delay with this @JCMais, going to merge and publish after https://jamstackconf.com/virtual/

JCMais:
no hurries @erezrokah , looking forward to the conf too. 😄

erezrokah (approved):
Thank you for your patience with this @JCMais

Issue #1225, opened 2020-09-10
Community report: CLI behaviour does not match containerized behaviour regarding .gitignore files


PerrySetGo:
### Describe the bug

We have a customer report in our Community forums indicating that the behaviour of the CLI and the behaviour of the web app differ in regards to how .gitignore files are treated.

### Steps to reproduce the behavior

Please see the discussion in this thread:
https://community.netlify.com/t/dont-apply-gitignore-rules-to-the-publish-directory/21244/7

### Expected vs actual behavior

It is our (@PerrySetGo , @depadiernos, @rstavchansky ) expectation that the two should work the same 🤷

erezrokah:
Looks like there are 2 different issues in that thread:
1. regarding `. gitignore` behaviour during deploy (unrelated to the CLI).
2. regarding a difference between how our build system deploys the publish directory vs how the CLI deploys the publish directory. It seems the former deploys everything in the publish directory and the latter doesn't upload `**/node_modules` directories (which was already resolved for the user due to a change in Nuxt behaviour).

erezrokah:
Looks like `node_modules` are not ignored in the buildbot:
https://github.com/netlify/open-api/blob/e72e2eb2d7eedf02bccc5916fc0330022f7f823b/go/porcelain/deploy.go#L692
But are ignored in the CLI:
https://github.com/netlify/js-client/blob/55650048fc7fe6288816c1dc172b27024cbd1586/src/deploy/util.js#L11

It makes sense to ignore those in the CLI in some cases, for example when deploying the project directory via:
`netlify deploy --dir .`

PerrySetGo:
i circled back to the community post and thanked them again for their contribution, fyi.

erezrokah:
Still not published yet, will do that today/tomorrow

PR #1272, opened 2020-09-23
fix(command-deploy): align file filtering with buildbot

erezrokah:
**- Summary**

Related to https://github.com/netlify/js-client/issues/157 and https://github.com/netlify/cli/issues/1227.

Fixes https://github.com/netlify/cli/issues/1225#issuecomment-696957186

The CLI deploy command relies on `js-client` default filter to skip some files when deploying.
This is done here:
https://github.com/netlify/js-client/blob/55650048fc7fe6288816c1dc172b27024cbd1586/src/deploy/util.js#L8 

and trying to emulate the code here:
https://github.com/netlify/open-api/blob/e72e2eb2d7eedf02bccc5916fc0330022f7f823b/go/porcelain/deploy.go#L692

This PR fixes a few discrepancies with our buildbot:
1. Allows publishing a hidden directory (e.g. `.public`) - fixes part of https://github.com/netlify/cli/issues/1227.
2. Allows deploying `node_modules` when inside a publish folder. Fixes https://github.com/netlify/cli/issues/1225#issuecomment-696957186 and matches buildbot behaviour. Only difference is that `node_modules` are ignored if a user tries to publish the current project dir.
3. Splits the regular expression here https://github.com/netlify/js-client/blob/55650048fc7fe6288816c1dc172b27024cbd1586/src/deploy/util.js#L13 into string operations. Also I believe the `\/__MACOSX` part wasn't working as I'm not sure how `basename` can result in a string containing a `/`.

**- Test plan**

Added tests

Align deploy file filtering with buildbot

🐴

### DIFF
--- a/tests/command.deploy.test.js
+++ b/tests/command.deploy.test.js
@@ -7,6 +7,19 @@ const { generateSiteName, createLiveTestSite } = require('./utils/createLiveTest
 
 const siteName = generateSiteName('netlify-test-deploy-')
 
+const validateContent = async ({ siteUrl, path, content, t }) => {
+  let actualContent
+  try {
+    const response = await fetch(`${siteUrl}${path}`)
+    if (response.ok) {
+      actualContent = await response.text()
+    }
+  } catch (e) {
+    // no op
+  }
+  t.is(actualContent, content)
+}
+
 const validateDeploy = async ({ deploy, siteName, content, t }) => {
   t.truthy(deploy.site_name)
   t.truthy(deploy.deploy_url)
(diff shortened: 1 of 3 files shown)


tests/command.deploy.test.js
```diff
@@ -7,18 +7,27 @@ const { generateSiteName, createLiveTestSite } = require('./utils/createLiveTest
 
 const siteName = generateSiteName('netlify-test-deploy-')
 
+const validateContent = async ({ siteUrl, path, content, t }) => {
+  let actualContent
+  try {
+    const response = await fetch(`${siteUrl}${path}`)
+    if (response.ok) {
```
ehmicky, line 14: Should the test always fail when `response.ok` is not `true`? If so, I am wondering if there should be a `t.true(response.ok)`?
Also wondering about whether the test should always fail when `fetch()` or `response.text()` throws. In which case, t.notThrowsAsync() could be used.

ehmicky: Oops, nevermind, I see that `content` is sometimes `undefined` in some of the tests. :+1:

ehmicky (approved):

erezrokah:
Tests are failing since the team we're using to deploy edge handlers on CI doesn't have access to do so anymore.

PR #1274, opened 2020-09-23
fix(command-deploy): handle deploy error when there are no files to deploy

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1227
Dependant on https://github.com/netlify/js-client/pull/161

Handles the error thrown by `js-client` when there are no files to deploy (cancels the created deploy).

**- Test plan**

Will push a test once is https://github.com/netlify/js-client/pull/161 merged and published

Handle deploy error when there are no files to deploy

🐳

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -16,6 +16,7 @@ const { getBuildOptions, runBuild } = require('../lib/build')
 const LinkCommand = require('./link')
 const { NETLIFYDEV, NETLIFYDEVLOG, NETLIFYDEVERR } = require('../utils/logo')
 const { statAsync } = require('../lib/fs')
+const { cancelDeploy } = require('../lib/api')
 const { deployEdgeHandlers } = require('../utils/edge-handlers')
 
 const DEFAULT_DEPLOY_TIMEOUT = 1.2e6
@@ -174,6 +175,7 @@ const runDeploy = async ({
   exit,
 }) => {
   let results
+  let deployId
   try {
     if (deployToProduction) {
       if (isObject(siteData.published_deploy) && siteData.published_deploy.locked) {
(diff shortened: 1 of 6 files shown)


src/utils/edge-handlers.js
```diff
@@ -74,11 +74,7 @@ const deployEdgeHandlers = async ({ site, deployId, api, silent, error, warn })
     } catch (e) {
       const text = `Failed deploying Edge Handlers: ${e.message}`
       stopSpinner({ spinner, text, error: true })
-      try {
-        await api.cancelSiteDeploy({ deploy_id: deployId })
-      } catch (e) {
-        warn(`Failed canceling deploy with id ${deployId}: ${e.message}`)
-      }
+      await cancelDeploy({ api, deployId, warn })
```
erezrokah, line 77: We can throw an error an let the calling function cancel the deploy, but we would need lift the spinner handling too.

src/commands/deploy.js
```diff
@@ -220,6 +222,9 @@ const runDeploy = async ({
       filter: getDeployFilesFilter({ site, deployFolder }),
     })
   } catch (e) {
+    if (deployId) {
```
ehmicky, line 225: [sand] Should this be an additional `try`/`catch` block around `deployEdgeHandlers()` + `api.deploy()`? Like this, there would be no need to check if `deployId` exists.

ehmicky (approved):

src/commands/deploy.js
erezrokah: That would mean we need to re-throw the error after canceling the deploy to reach the other part of the error handling code (we could also refactor that part when we do https://github.com/netlify/js-client/issues/157 to have a dedicated `try/catch` per operation to have better error reporting).

erezrokah: Going to merge and release. We can discuss this in https://github.com/netlify/js-client/issues/157

ehmicky: Makes sense!

Issue #1302, opened 2020-09-30
Bug(command-dev): `.env` vars overriding process.env ones


erezrokah:
**Describe the bug**

When using a `.env` file and setting environment variables at the process level, `.env` ones take precedence
This does not follow https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set

**To Reproduce**

1. Create an empty directory and navigate to it.
2. Add a `.env` file with the content of:
```
TEST=FROM_ENV
```
3. Run `TEST=FROM_COMMAND netlify dev:exec env | grep TEST`

Result:
```
TEST=FROM_ENV
```

Expected:
```
TEST=FROM_COMMAND
```

**Configuration**

```bash
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.78 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.5 - ~/Code/github/netlify/cli/node_modules/.bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.18.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmGlobalPackages:
    netlify-cli: 2.64.1
```

**Expected behavior**

Process level variables should take precedence

**Additional context**

https://github.com/netlify/cli/issues/1301
https://github.com/netlify/cli/pull/924

Issue #1301, opened 2020-09-30
Bug(command-dev): `.env.development` file is not merged with `.env`


erezrokah:
**Describe the bug**

We support sourcing environment variables from `.env` OR `.env.development`:
https://github.com/netlify/cli/blob/25d5aa4fdae19924ac26123b6634b9e1a72a1097/src/utils/env.js#L12

I believe it is the common practice to source from both with `.env.development` taking precedence.

**To Reproduce**

1. Create an empty directory and navigate to it.
2. Add a `.env` file with the content of:
```
TEST=FROM_ENV
TEST_2=FROM_ENV
```
3. Add a `.env.development` with the content of:
```
TEST=FROM_DEV
TEST_3=FROM_DEV
```
4.  run `netlify dev:exec env | grep TEST`

Result:
```
TEST=FROM_DEV
TEST_3=FROM_DEV
```

Expected:
```
TEST_2=FROM_ENV
TEST=FROM_DEV
TEST_3=FROM_DEV
```

**Configuration**

```bash
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.78 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
    Yarn: 1.22.5 - ~/Code/github/netlify/cli/node_modules/.bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.18.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmGlobalPackages:
    netlify-cli: 2.64.1
```

**Expected behaviour**

Both `.env` and `.env.development` should be used with `.env.development` taking precedence. 

**Additional context**

https://github.com/netlify/cli/pull/924

Issue #1303, opened 2020-09-30
Feature: framework specific .env file support


erezrokah:
**Is your feature request related to a problem? Please describe.**

https://github.com/netlify/cli/pull/924 Attempted to align our `.env` support with `create-react-app`.
After doing some thinking on how the CLI is built, we might want to handle `.env` files as a part of the framework detection mechanism, as some frameworks might handle those files differently.

> If we find out that all/most frameworks use same the logic we can close this issue 

**Describe the solution you'd like**

Have Netlify CLI mimic the detected framework `.env` file support and inherit the variables.

**Describe alternatives you've considered**

Using the same logic for all frameworks

**Can you submit a pull request?**

Yes

ercgrat:
The handling of environment variables Next.js and netlify dev doesn't seem to be compatible. When running netlify dev my local .env variables are overriden by the Netlify variables pulled from my production settings (Site settings > Build & deploy > Environment > Environment variables).

This is at least true for the variables that I need to be accessible in the browser with the NEXT_PUBLIC_ prefix ([blog post](https://nextjs.org/blog/next-9-4#new-environment-variables-support)).

For anyone who runs into this, my workaround is to define a separate environment variable only in my local .env file, and add conditionals to code (womp!).

.env
```
NEXT_PUBLIC_PAYPAL_ENV=sandbox
NEXT_PUBLIC_PAYPAL_ENV_DEV=sandbox
```

In my production settings, NEXT_PUBLIC_PAYPAL_ENV=production. The browser value of process.env.NEXT_PUBLIC_PAYPAL_ENV ends up being production when running netlify dev.

workaround:
```
let env = process.env.NEXT_PUBLIC_PAYPAL_ENV_DEV;
if (!env) {
    env = process.env.NEXT_PUBLIC_PAYPAL_ENV;
}
```

erezrokah:
> When running netlify dev my local .env variables are overridden by the Netlify variables pulled from my production settings

This is actually a bug. Do you mind opening a new issue for that?

ercgrat:
Sure thing, created it here: #1386

Soviut:
@erezrokah rather than doing any detection, I think it would be more useful to be able to use `netlify.toml` to explicitly configure which env files are loaded for which environments.

Perhaps some kind of list that specified the priority to load the env files in. For example:

```toml
env_priority = [ ".env", ".env.local" ]
```

This would also mean you could configure it as follows to not inject any `.env` files at all.

```toml
[...]

PR #1305, opened 2020-09-30
fix(command-dev): align .env and .env.development handling

erezrokah:
**- Summary**

Follow up on https://github.com/netlify/cli/pull/924
Fixes https://github.com/netlify/cli/issues/1302
Fixes https://github.com/netlify/cli/issues/1301

This does not:
1. Mimic `create-react-app` as the original PR (see https://github.com/netlify/cli/issues/1303)
2. Adds interpolation see https://github.com/netlify/cli/pull/924#issuecomment-700789859

**- Test plan**

Added some tests

Align .env and .env.development handling with common practices.

🐟


src/commands/dev/exec.js
```diff
@@ -24,14 +23,12 @@ class ExecCommand extends Command {
...
-          path.relative(site.root, envSettings.file)
-        )} file:`,
-        chalk.yellow(Object.keys(envSettings.vars))
+    if (envSettings.vars.length > 0) {
+      this.log(
+        `${NETLIFYDEVLOG} Adding the following env variables from ${envSettings.files.map(f => chalk.blue(f))}:`,
+        chalk.yellow(envSettings.vars.map(([key]) => key))
       )
-      Object.entries(envSettings.vars).forEach(([key, val]) => (process.env[key] = val))
+      envSettings.vars.forEach(([key, val]) => (process.env[key] = val))
```
erezrokah, line 31: Instead of calling `Object.entries` on each usage, the function now returns it.

erezrokah: We can also avoid `process.env[key] = val` and pass an `env` object to `execa`, but I'll do that in another PR

src/commands/dev/exec.js
```diff
@@ -24,14 +23,12 @@ class ExecCommand extends Command {
     }
 
     const envSettings = await getEnvSettings(site.root)
-    if (envSettings.file) {
-      console.log(
-        `${NETLIFYDEVLOG} Overriding the following env variables with ${chalk.blue(
-          path.relative(site.root, envSettings.file)
-        )} file:`,
-        chalk.yellow(Object.keys(envSettings.vars))
+    if (envSettings.vars.length > 0) {
```
erezrokah, line 26: Previous logic had an edge case where it would log if you had a `.env` file with no vars

src/function-builder-detectors/netlify-lambda.js
```diff
@@ -26,11 +26,8 @@ module.exports = async function(projectDir) {
     }
   }
 
-  let env = {}
   const envSettings = await getEnvSettings(projectDir)
-  if (envSettings.file) {
-    env = envSettings.vars
-  }
+  const env = envSettings.vars.reduce((env, [key, value]) => ({ ...env, [key]: value }), {})
```
erezrokah, line 30: It would have been nice to use `Object.fromEntries`....
Also though to also return the `env` object AND the entries, but that's maybe too much?

src/utils/env.js
```diff
@@ -1,27 +1,34 @@
...
-
-  const settings = {}
-
-  try {
-    if ((await statAsync(envFile)).isFile()) settings.file = envFile
-  } catch (err) {
-    // nothing
-  }
-  try {
-    if ((await statAsync(envDevelopmentFile)).isFile()) settings.file = envDevelopmentFile
```
erezrokah, line 17: We could have optimised the previous code actually to return if we manage to read `.env.development` and skip reading `.env`, but that's irrelevant now.

src/commands/dev/index.js
```diff
@@ -102,16 +102,15 @@ const getAddonsUrlsAndAddEnvVariablesToProcessEnv = async ({ api, site, flags })
...
 }
 
-const addDotFileEnvs = async ({ site }) => {
+const addDotFileEnvs = async ({ site, log }) => {
+  // add .env file environment variables
   const envSettings = await getEnvSettings(site.root)
-  if (envSettings.file) {
-    console.log(
-      `${NETLIFYDEVLOG} Overriding the following env variables with ${chalk.blue(
-        path.relative(site.root, envSettings.file)
```
erezrokah, line 110: Changed the function to return just the filenames, so we won't have to trim the full path

tests/command.dev.test.js
```diff
@@ -122,7 +122,7 @@ testMatrix.forEach(({ args }) => {
     })
   })
 
-  test(testName('should override process env vars with ones in .env.development', args), async t => {
+  test(testName('should override .env.development with process env', args), async t => {
     await withSiteBuilder('site-with-override', async builder => {
```
erezrokah, line 126: Looks like the initial intention of this feature was to just override vars set in the UI/netlify.toml:
https://github.com/netlify/cli/pull/736

src/function-builder-detectors/netlify-lambda.js
ehmicky: :+1: 
This ponyfill might help: https://github.com/ungap/from-entries/blob/master/index.js

src/lib/fs.js
```diff
@@ -23,4 +23,13 @@ const fileExistsAsync = async filePath => {
   }
 }
 
-module.exports = { statAsync, readFileAsync, readFileAsyncCatchError, writeFileAsync, fileExistsAsync }
+const isFileAsync = async filePath => {
```
ehmicky, line 26: Should we use `path-type`?
https://github.com/sindresorhus/path-type#isfilepath

src/utils/env.js
```diff
@@ -1,27 +1,34 @@
...
+  )
 
-  return settings
+  const settings = results.filter(Boolean).reduce(
+    ({ files, vars }, { file, env }) => {
+      return { files: [...files, file], vars: { ...env, ...vars } }
+    },
+    { files: [], vars: {} }
+  )
+  return { ...settings, vars: Object.entries(settings.vars) }
```
ehmicky, line 31: [sand] I am wondering whether `Array.map()` would be clearer than `Array.reduce()`? Something like:

```js
const allResults = results.filter(Boolean)
const files = allResults.map(({ file }) => file)
const vars = Object.entries(Object.assign({}, ...allResults.map(({ env }) => env)))
return { files, vars }
```

I have not tried that code, it might be wrong :)
Feel free to dismiss if you think `Array.reduce()` is simpler :+1:

ehmicky (approved):

src/lib/fs.js
erezrokah: Yes, this is much better!

src/utils/env.js
erezrokah: Maybe, I was trying to save that additional iteration 😄
I think you would need to reverse the order of results when mapping to `vars` to make it work as excepted, as files are ordered by precedence.

src/lib/fs.js
erezrokah: Done in https://github.com/netlify/cli/pull/1305/commits/df18aca5268c47dc175cf2361fca54b71c7b7892. Still had to wrap it with `try/catch` since `path-type` throws:
https://github.com/sindresorhus/path-type/blob/8e4e0a24531d639b5b9f1eb8b08c01e9336578e1/index.js#L18

src/utils/env.js
ehmicky: This is less efficient, but it might be clearer. Although that is not clear either... so please feel free to dismiss :)

src/lib/fs.js
ehmicky: This would only throw if there is something wrong with the filesystem, such as permission error. Shouldn't we propagate those errors?

ehmicky (approved):

src/function-builder-detectors/netlify-lambda.js
erezrokah: After making the change and debugging the code realised this code was completely redundant as we're already setting the env vars on the current process here:
https://github.com/netlify/cli/blob/c5cd7112060151ce31292f424e70656385846cf4/src/commands/dev/index.js#L114
before running this code.
Removed in https://github.com/netlify/cli/pull/1305/commits/baf01c7dac994b7b68f780ec954c4fba3c0325bb

src/lib/fs.js
erezrokah: Sure, let me handle that

src/utils/env.js
erezrokah: Done in https://github.com/netlify/cli/pull/1305/commits/6e2f29cd7b7ed2fe7e69643b00542ac99fbd78f6#diff-eac6f78170256c598e70ce4b121d69caL25

src/lib/fs.js
erezrokah: Done in https://github.com/netlify/cli/pull/1305/commits/6e2f29cd7b7ed2fe7e69643b00542ac99fbd78f6

ehmicky (approved):
:tada:

PR #1307, opened 2020-10-01
refactor(command-dev-exec): don't mutate process.env

erezrokah:
**- Summary**

Follow up on https://github.com/netlify/cli/pull/1305 - don't mutate `process.env` when we can pass `env`
 to `execa`

**- Test plan**

Added a test case

Don't mutate `process.env` in `commands/dev/exec.js`

🐈

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2895,6 +2895,11 @@
         "tsutils": "^3.17.1"
       }
     },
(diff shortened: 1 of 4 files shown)


ehmicky (approved):
Yes!
Mutating `process.env` is such a bad idea :p

Issue #1212, opened 2020-09-07
Chore: remove code duplication around opening browsers


erezrokah:
We have some duplicated code that opens browsers:
https://github.com/netlify/cli/blob/b58555ff1298457ef0a270c18ff0b39b99e4ed0e/src/utils/open-browser.js#L17
https://github.com/netlify/cli/blob/b58555ff1298457ef0a270c18ff0b39b99e4ed0e/src/commands/dev/index.js#L156

we should remove the duplicate code and also avoid using `console.log` in:
https://github.com/netlify/cli/blob/b58555ff1298457ef0a270c18ff0b39b99e4ed0e/src/utils/open-browser.js#L7

Related https://github.com/netlify/cli/pull/1211, https://github.com/netlify/cli/pull/1144

PR #1309, opened 2020-10-01
refactor: remove open-browser code duplication

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1212, follow up on https://github.com/netlify/cli/pull/1144

More context - not only we had duplicate code around this, we were using 2 different dependencies to do so. I removed the usage of https://github.com/oclif/cli-ux/blob/515570bbeb40bcd28698b6b7af53cc9bfdf86279/src/open.ts#L12 as it seems like an out of date version of https://github.com/sindresorhus/open/blob/f3748e4f5a04a41e27034aff1eb08e8b86d01c1c/index.js#L34

**- Test plan**

Tested manually the different commands related to opening browsers (`login,dev,deploy,init,open,sites:create --with-ci`).

Remove open browser code duplication

🐕

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -8,7 +8,7 @@ const { serverSettings } = require('../../utils/detect-server')
 const { startFunctionsServer } = require('../../utils/serve-functions')
 const Command = require('../../utils/command')
 const chalk = require('chalk')
-const open = require('open')
+const openBrowser = require('../../utils/open-browser')
 const { NETLIFYDEV, NETLIFYDEVLOG, NETLIFYDEVWARN, NETLIFYDEVERR } = require('../../utils/logo')
 const boxen = require('boxen')
 const { startLiveTunnel } = require('../../utils/live-tunnel')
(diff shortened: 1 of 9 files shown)


src/commands/dev/index.js
```diff
@@ -223,7 +213,9 @@ class DevCommand extends Command {
     const liveTunnelUrl = await handleLiveTunnel({ flags, site, api, settings, log })
     url = liveTunnelUrl || url
 
-    await openBrowser({ devConfig, url, warn })
+    if (devConfig.autoLaunch !== false) {
+      await openBrowser({ url, log, silentBrowserNoneError: true })
```
erezrokah, line 217: In all other cases we print an error when `BROWSER=none` is set as an env variable. When running `netlify dev` is it quite useful to run `BROWSER=none netlify dev` and should not be considered as an error.

src/utils/open-browser.js
```diff
@@ -1,31 +1,32 @@
...
 const isDockerContainer = require('is-docker')
 
-function unableToOpenBrowserMessage(url, err) {
-  // https://github.com/sindresorhus/log-symbols
-  console.log('---------------------------')
-  const errMsg = err ? `\n${err.message}` : ''
-  const msg = `Error: Unable to open browser automatically${errMsg}\n`
-  console.log(`${chalk.redBright(msg)}`)
-  console.log(chalk.greenBright('Please open your browser & open the URL below to login:'))
-  console.log(chalk.whiteBright(url))
```
erezrokah, line 12: The choice of colours here made the message almost unreadable (`whiteBright`).

src/utils/open-browser.js
```diff
@@ -1,31 +1,32 @@
-const cli = require('cli-ux').default
```
erezrokah, line 1: We use `cli-ux` in some other locations so we can remove this

ehmicky (approved):
:tada:

PR #1313, opened 2020-10-03
feat(dev): add trace subcommand to simulate redirects

calavera:
This new subcommand allows checking whether a request will match
a remap rule directly from the cli.
It requires the new traffic mesh agent.

Closes https://github.com/netlify/product/issues/584


erezrokah (approved):
This is very useful, thanks @calavera!

I made the command hidden for now, if this is something that should be available in our docs I can revert.

Also, added a couple of tests

PR #1316, opened 2020-10-04
chore: unhide dev:trace

erezrokah:
This unhides the `dev:trace` command

### DIFF
--- a/docs/commands/dev.md
+++ b/docs/commands/dev.md
@@ -32,6 +32,7 @@ netlify dev
 | Subcommand | description  |
 |:--------------------------- |:-----|
 | [`dev:exec`](/docs/commands/dev.md#devexec) | Exec command  |
+| [`dev:trace`](/docs/commands/dev.md#devtrace) | Trace command  |
 
 
 **Examples**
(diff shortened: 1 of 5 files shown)


erezrokah:
cc @calavera

calavera (approved):
excellent, thanks!

docs/commands/dev.md
```diff
@@ -67,6 +68,36 @@ netlify dev:exec
...
+**Usage**
+
+```bash
+netlify dev:trace
+```
+
+**Flags**
+
+- `debug` (*boolean*) - Print debugging information
+- `httpProxy` (*string*) - Proxy server address to route requests through.
```
calavera, line 88: The flags are not correct, I guess they are copied from somewhere else. These are the flags:

```
- `watch` (*string*) - Path to the publish directory.
- `header` (*string*) - Request header, this flag can be used multiple times. Example: "Host: netlify.test"
- `cookie` (*string*) - Request cookie, this flag can be used multiple times. Example: "nf_jwt=token"
```

calavera (changes requested):

docs/commands/dev.md
erezrokah: Good catch, these are default flags shared by all commands

ehmicky (approved):

erezrokah:
@ehmicky moving to draft until I fix https://github.com/netlify/cli/pull/1316#discussion_r499295834

src/commands/dev/trace.js
```diff
@@ -1,8 +1,11 @@
+const { flags } = require('@oclif/command')
 const Command = require('../../utils/command')
 const { runProcess } = require('../../utils/traffic-mesh')
 
 class TraceCommand extends Command {
   async run() {
+    this.parse(TraceCommand)
```
erezrokah, line 7: This is done to validate the argument and print an error consistent with the other CLI commands (otherwise the error comes directly from the Traffic Mesh agent).

erezrokah:
@calavera flags and arguments are still passed through to the Traffic Mesh agent as is, but known ones are validated by the CLI and will print relevant help/errors.

You should be able to add new flags without this breaking (but not change existing ones without updating the CLI first).

calavera (approved):

Issue #1332, opened 2020-10-06
Bug: running `netlify deploy --build --json` does not return a valid json output


erezrokah:
When `netlify deploy` is invoked with the `build` and `json` flags, the output from `@netlify/build` breaks the json output.
We should silent `@netlify/build` in that case.

We should be able to use https://github.com/netlify/build/blob/134b4baf21c3add93d4e0f531db42147d8d2f432/packages/config/src/bin/flags.js#L96 for that

PR #1333, opened 2020-10-06
fix(command-deploy): pass buffer: true to build if --json or --silent

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1332

**- Test plan**

Added a test case

pass `buffer: true` to Netlify build if `--json` or `--silent`

image


ehmicky (approved):
Looks good :+1: 

Note: the title should probably be renamed to `pass buffer: true`.

erezrokah:
> Note: the title should probably be renamed to `pass buffer: true`.

🤦

Issue #1331, opened 2020-10-06
Bug: addons env variables are not available when running `netlify build`


erezrokah:
When running `netlify build` addons env variables are not set for `@netlify/build`.
We should use a similar approach like https://github.com/netlify/cli/blob/4f376100ff23e41b938bf017bb991896098eff0a/src/commands/dev/exec.js#L19

to set those

### Context

Got this while trying to run `netlify build` locally with https://github.com/nimbella/netlify-plugin-nimbella/blob/beca8fd4b911f939de21ec47a018ba4dfbbe1afa/index.js#L62

ehmicky:
Please note there is an `env` option to `@netlify/build`. It is a plain object with environment variables. Those environment variables will be set in child processes (the build command and Build plugins), but not in the current process (which is good). This option could probably be used for this.

I am not familiar with addons environment variables, do you have more information about those?

erezrokah:
> I am not familiar with addons environment variables, do you have more information about those?

Those can be added to a site when you create an addon via `netlify addons:create`.
For example `netlify addons:create nimbella` will create an env variable `NIMBELLA_LOGIN_TOKEN`.
`netlify addons:create fauna` creates a bunch of those too.

erezrokah:
Closed by mistake

Issue #310, opened 2019-05-02
Use open-api addons endpoints


bcomnes:
We need to swap the use of `netlify/src/addons` to use the open-api definitions 

- `createAddon` -> `createServiceInstance`
- `getAddons` -> `showServiceInstance`
- `deleteAddon` -> `deleteServiceInstance`
- `updateAddon` -> `updateServiceInstance`

See https://open-api.netlify.com/#/default/updateServiceInstance

All the addons command need to be updated with the new names and probably shape: https://github.com/netlify/cli/search?q=src%2Faddons&unscoped_q=src%2Faddons

Then delete this file: https://github.com/netlify/js-client/blob/master/src/addons.js

erezrokah:
We are still referencing the feature as `addons`:
https://docs.netlify.com/partner-add-ons/get-started/#work-with-add-ons

so the current naming makes sense to me, even if the underlying implementation uses `service instances`.
Comment if still relevant

keiko713:
@erezrokah is there any way to somehow make an alias with this, so that it'll keep the name part (or rather, it can be called in both ways) but it'll use the open-api defined method behind the scene?
I just got bitten by this, since current addon code (https://github.com/netlify/js-client/blob/master/src/addons.js) is hard coding `api.netlify.com`.
Although, I guess that means that the functions shape is gonna change so maybe it's not an option. Let me know in that case, maybe we can use xxxServiceInstance way in https://github.com/netlify/netlify-lm-plugin/blob/b9323387bf0cff230739ebef6ebf5953ae2871fe/src/commands/lm/setup.ts#L7 instead for the workaround

erezrokah:
Thanks @keiko713.
Re-opening this.
Clearing up some confusion I had from the initial issue description.
There are 2 issues here:
1. How the internal implementation works - we should update it to use the open-api defined methods.
2. How the commands are called in the CLI - not sure we should update them to mention service instances (but we could alias them).

Looks like:
1. https://github.com/netlify/js-client/blob/960089a289288233bd2a17fcbd8ab4730ca49135/src/addons.js#L8 can be replaced by directly by https://open-api.netlify.com/#operation/createServiceInstance
2. https://github.com/netlify/js-client/blob/960089a289288233bd2a17fcbd8ab4730ca49135/src/addons.js#L50 should be replaced by https://open-api.netlify.com/#operation/deleteServiceInstance but I'm not sure if I should reference the `instanceId` in the `open-api` spec.
[...]

PR #1336, opened 2020-10-06
refactor: cleanup createSiteAddon function

erezrokah:
Related to https://github.com/netlify/cli/issues/310 and https://github.com/netlify/cli/issues/1331

Wanted to fix https://github.com/netlify/cli/issues/1331, but couldn't do it since the addons code doesn't allow getting the env variables easily at the moment:

I going to do some refactoring first so we can have https://github.com/netlify/cli/blob/b88cf6db2028741b93ce271cfaca934c4f812a0d/src/utils/dev.js#L25 return addons envs instead of mutating `process.env`

I removed a bunch of outdated comments from `utils/addons.js` and did some code cleanup.
Note that part of that function is duplicated from:
https://github.com/netlify/cli/blob/b88cf6db2028741b93ce271cfaca934c4f812a0d/src/commands/addons/create.js#L155

I follow up in another PR. 

Also changed some code in `functions/create.js` to use `async/await` instead of promise nesting and added a missing `await`

The code itself is used when running `netlify functions:create` and will add missing addons depending on the function template.
For example if you install a `fauna` template it adds the `fauna` addon

### DIFF
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -353,7 +353,6 @@ async function installAddons(addons = [], fnPath) {
         const addonCreated = await createSiteAddon(api.accessToken, addonName, siteId, siteData, this.log)
         if (addonCreated) {
           if (addonDidInstall) {
-            const { addEnvVariables } = require('../../utils/dev')
             await addEnvVariables(api, site)
             const { confirmPostInstall } = await inquirer.prompt([
               {
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -327,7 +327,7 @@ async function scaffoldFromTemplate(flags, args, functionsDir) {
         spinner.succeed(`installed dependencies for ${name}`)
       }
 
-      installAddons.call(this, addons, path.resolve(functionPath))
+      await installAddons.call(this, addons, path.resolve(functionPath))
       if (onComplete) {
         await addEnvVariables(this.netlify.api, this.netlify.site)
         await onComplete.call(this)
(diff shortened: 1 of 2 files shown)


src/commands/functions/create.js
```diff
@@ -346,36 +346,33 @@ function installAddons(addons = [], fnPath) {
...
-                if (confirmPostInstall) addonDidInstall(fnPath)
-              }
+    const siteData = await api.getSite({ siteId })
+    const arr = addons.map(async ({ addonName, addonDidInstall }) => {
+      this.log(`${NETLIFYDEVLOG} installing addon: ` + chalk.yellow.inverse(addonName))
+      try {
+        const addonCreated = await createSiteAddon(api.accessToken, addonName, siteId, siteData, this.log)
+        if (addonCreated) {
+          if (addonDidInstall) {
+            const { addEnvVariables } = require('../../utils/dev')
```
ehmicky, line 356: Do we have a good reason to dynamically load `require()` here (I think not)?

erezrokah: Definitely not

src/utils/addons.js
```diff
@@ -1,311 +1,42 @@
...
-    // const deleteCmd = chalk.cyan(`\`netlify addons:delete ${addonName}\``)
-    // log(`- To remove this add-on run: ${deleteCmd}`)
-    // log()
     return false
   }
 
-  // const manifest = await getAddonManifest(addonName, accessToken);
-
-  const configValues = rawFlags
-  // if (manifest.config) {
```
ehmicky, line 33: :broom:

src/utils/addons.js
```diff
@@ -1,311 +1,42 @@
-/* eslint no-console: 0 */
 const { getAddons, createAddon } = require('netlify/src/addons')
-// const chalk = require("chalk");
-// const fetch = require("node-fetch");
 
-/** main section - shamelessly adapted from CLI. we can extract and dedupe later. */
-/** but we can DRY things up later. */
```
ehmicky, line 7: :)

ehmicky (approved):

PR #1338, opened 2020-10-06
fix(dev): add missing await

calavera:
Otherwise the process doesn't wait to the server to be running.

### DIFF
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -55,7 +55,7 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
     args.push('--debug')
   }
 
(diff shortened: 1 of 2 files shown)


src/utils/traffic-mesh.js
```diff
@@ -84,7 +84,7 @@ const runProcess = async ({ log, args }) => {
   await installTrafficMesh({ log })
 
   const execPath = path.join(getBinPath(), EXEC_NAME)
-  execa(execPath, args, { stdio: 'inherit' })
+  await execa(execPath, args, { stdio: 'inherit' })
```
erezrokah: This actually breaks `netlify dev --trafficMesh` since it waits for the process to exit.
On a side node, `netlify dev --trafficMesh` is already broken since `runProcess` should return the sub process.

PR #1340, opened 2020-10-07
Lint/format Markdown and HTML

ehmicky:
This runs Prettier on Markdown and HTML files.

This also adds linting to JavaScript embedded in Markdown and HTML files. We don't have any at the moment, but the goal is to more to make it consistent with Netlify Build (https://github.com/netlify/build/pull/1934) so we can eventually abstract the shared linting configuration, and re-use it across our projects.


CHANGELOG.md
```diff
@@ -203,7 +203,7 @@ Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
 ### Merged
 
-- feat(env): add command to list/get/set/import environment variables  [`#1162`](https://github.com/netlify/cli/pull/1162)
+- feat(env): add command to list/get/set/import environment variables [`#1162`](https://github.com/netlify/cli/pull/1162)
```
erezrokah, line 206: Can we ignore the this file and the `README.md` file for now?
Some parts of those files are autogenerated (we should fix this in the generation code).

erezrokah (changes requested):
Thank you so much for this @ehmicky. Great way to align code bases.
Just a small comment

ehmicky:
Fixed! :+1:

erezrokah (approved):

PR #1341, opened 2020-10-07
Lint Ava test files

ehmicky:
Related: https://github.com/netlify/build/pull/1935

This adds [`eslint-plugin-ava`](https://github.com/avajs/eslint-plugin-ava) to lint our Ava test files.

PR #1342, opened 2020-10-07
Lint test files

ehmicky:
Follow up on #1341.

This runs ESLint on `tests/**/*.js`.

### DIFF
--- a/tests/command.dev.test.js
+++ b/tests/command.dev.test.js
@@ -1,3 +1,5 @@
+// Handlers are meant to be async outside tests
+/* eslint-disable require-await */
 const test = require('ava')
 const path = require('path')
 const { withDevServer } = require('./utils/devServer')
@@ -960,3 +962,4 @@ testMatrix.forEach(({ args }) => {
     })
   })
 })
+/* eslint-enable require-await */
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "test": "run-s test:*",
     "ava": "ava --verbose",
     "format": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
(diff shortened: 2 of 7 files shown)


tests/command.dev.test.js
```diff
@@ -185,7 +185,7 @@ testMatrix.forEach(({ args }) => {
         })
         .withFunction({
           path: 'echo.js',
-          handler: async (event, context) => {
+          handler: (event, context) => {
```
erezrokah, line 188: I think functions should return a promise when not using the callback pattern.
We're using `lambda-local` to invoke functions locally - if the tests pass I can only assume it handles it.

erezrokah (approved):

tests/command.dev.test.js
ehmicky: This makes sense, thanks for the additional information/

The related ESLint rule is `require-await`. Generally speaking, this might be a good rule to decrease the number of `async`/`await` in our code, and only make code asynchronous that needs to be.

Is there a way we could still enable this rule (if you find it useful)? For example, one of:
  - Adding `eslint-disable-next-line` comment before each `handler`
  - Adding `eslint-disable` comment for the whole file
  - Adding an `overrides` to disable this rule in all `*.test.js`
  - Relying on `lambda-local` behavior allowing sync functions

What do you think?

erezrokah (approved):

ehmicky:
Merging after we resolve the comment at https://github.com/netlify/cli/pull/1342#discussion_r501060414: what are your thoughts on this @erezrokah ?

tests/command.dev.test.js
erezrokah: I think the best approach will be to disable it for the file with a comment explaining the reason.

ehmicky: Fixed!

erezrokah (approved):

PR #1343, opened 2020-10-07
Add `eslint-plugin-unicorn`

ehmicky:
This adds [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn). `eslint-plugin-unicorn` contains a bunch of awesome linting rules which are geared towards code consistency and preventing bugs. It's pretty opinionated though, and needs some configuration to fit our current styling.

Almost all the code update is due to `eslint --fix`. I have gone through it, but an extra pair of eyes would be useful :)


.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
     "prettier/standard",
+    "plugin:unicorn/recommended",
     "plugin:ava/recommended",
     "plugin:you-dont-need-lodash-underscore/all"
   ],
   "rules": {
     "no-process-exit": 0,
     "object-shorthand": "error",
     "require-await": "error",
+    "unicorn/custom-error-definition": 2,
```
ehmicky, line 16: Not enabled by default in `unicorn/recommended`, but still pretty useful. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/custom-error-definition.md).

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-unused-properties": 2,
```
ehmicky, line 17: Not enabled by default in `unicorn/recommended`, but still pretty useful. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-unused-properties.md)

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/filename-case": [2, { "case": "kebabCase", "ignore": [".*.md"] }],
```
ehmicky, line 18: `@netlify/build` uses `snake_case` instead of `kebab-case`. We might want to harmonize here.
`.*.md` is to allow `README.md` and the like to be uppercase. We could also lowercase them.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/explicit-length-check": [2, { "non-zero": "not-equal" }],
```
ehmicky, line 19: The additional [`non-zero` option](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/explicit-length-check.md#non-zero-comparisons) is useful for code consistency.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-process-exit": 0,
```
ehmicky, line 20: We execute `process.exit()` in many files, which is forbidden by [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-process-exit.md), which recommends throwing an error and running `process.exit()` only in the top-level binary file. This might be a good practice, but this would require too much refactoring at this point.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-null": 0,
```
ehmicky, line 21: That's a contentious one. I personally agree with this rule, but I think this might be too controversial to be useful. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-null.md). Also, this would require lots of refactoring.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/prevent-abbreviations": 0,
```
ehmicky, line 22: The idea behind [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prevent-abbreviations.md) is quite neat. However, in practice, its detection algorithm has lots of false positives in my experiences, which makes it a hassle.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/number-literal-case": 0,
```
ehmicky, line 23: That conflicts with Prettier sometimes. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/number-literal-case.md)

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/expiring-todo-comments": 0,
```
ehmicky, line 24: Another pretty interesting idea. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/expiring-todo-comments.md). However, I think we might want to leave this for the future.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-fn-reference-in-iterator": 0,
```
ehmicky, line 25: I am wondering what you think about this one. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-fn-reference-in-iterator.md).
I left it out here mostly because it would require too much refactoring.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-unreadable-array-destructuring": 0,
```
ehmicky, line 26: This conflicts with the core ESLint `prefer-destructuring` rule.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/import-style": 0,
```
ehmicky, line 27: Another pretty nice idea. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/import-style.md). However, this might require too much refactoring at the very moment. 
Might be worth considering for the future though?

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/prefer-optional-catch-binding": 0,
```
ehmicky, line 28: Optional catch binding is not supported by Node 8.

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/prefer-trim-start-end": 0,
```
ehmicky, line 29: `string.trimEnd()` is not supported by Node 8.

.eslintrc.json
erezrokah: Nice, this can prevent some bugs

.eslintrc.json
```diff
@@ -5,13 +5,29 @@
...
+    "unicorn/no-reduce": 0,
```
ehmicky, line 30: I agree with this one. [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-reduce.md). However, this would require too much refactoring at this point.
I am wondering what your thoughts are on this?

site/scripts/docs.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path')
 const markdownMagic = require('markdown-magic')
 const stripAnsi = require('strip-ansi')
-const generateCommandData = require('./generateCommandData')
+const generateCommandData = require('./generate-command-data')
```
ehmicky, line 4: Use same case as other files.

site/scripts/docs.js
```diff
@@ -25,7 +25,7 @@ const config = {
         md += formatFlags(info.flags)
         md += commandListSubCommandDisplay(info.commands)
         md += commandExamples(info.examples)
-        if (info.commands.length) {
+        if (info.commands.length !== 0) {
```
ehmicky, line 28: Many changes in this PR follow this pattern: this enforces `length` comparison always use the same pattern (`=== 0` or `!== 0`), for consistency.

.eslintrc.json
erezrokah: How does it work for dependencies with only a default export?

src/commands/addons/config.js
```diff
@@ -188,8 +188,8 @@ async function update({ addonName, currentConfig, newConfig, settings, accessTok
   try {
     // TODO update updateAddon to https://open-api.netlify.com/#operation/updateServiceInstance
     updateAddonResponse = await updateAddon(settings, accessToken)
-  } catch (e) {
-    error(e.message)
+  } catch (error_) {
```
ehmicky, line 191: Another autofix that changes lots of files in this PR.
This enforces consistent names for `error`. When there are several `error` in scope, it appends an `_`. Not a big fan of that style, but at least, it's consistent.
See [rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/catch-error-name.md).

.eslintrc.json
erezrokah: `reduce` can be hard to read most of the times. It might be too strict to forbid it completely.

src/commands/functions/create.js
```diff
@@ -86,7 +86,7 @@ async function getNameFromArgs(args, flags, defaultName) {
         message: 'name your function: ',
         default: defaultName,
         type: 'input',
-        validate: val => Boolean(val) && /^[\w\-.]+$/i.test(val),
+        validate: val => Boolean(val) && /^[\w.-]+$/i.test(val),
```
ehmicky, line 89: The backslash is not needed when `-` is the last character in `[...]`.

src/commands/functions/create.js
```diff
@@ -154,9 +154,9 @@ async function pickTemplate() {
   function filterRegistry(registry, input) {
     const temp = registry.map(x => x.name + x.description)
     const filteredTemplates = fuzzy.filter(input, temp)
-    const filteredTemplateNames = filteredTemplates.map(x => (input ? x.string : x))
+    const filteredTemplateNames = new Set(filteredTemplates.map(x => (input ? x.string : x)))
```
ehmicky, line 157: Performance optimization.
When the set is being used only once after creation, I am not sure this is actually faster. But this is faster when the set is in the top-level scope and re-used.
[See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-set-has.md).

.eslintrc.json
erezrokah: Agree. I'm trying to handle those `process.exit` a few at a time

src/commands/sites/create.js
```diff
@@ -70,7 +70,7 @@ class SitesCreateCommand extends Command {
             name: 'name',
             message: 'Site name (optional):',
             filter: val => (val === '' ? undefined : val),
-            validate: input => /^[a-zA-Z0-9-]+$/.test(input) || 'Only alphanumeric characters and hyphens are allowed',
+            validate: input => /^[\dA-Za-z-]+$/.test(input) || 'Only alphanumeric characters and hyphens are allowed',
```
ehmicky, line 73: [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/better-regex.md)

.eslintrc.json
erezrokah: I think it's good for consistency .

src/functions-templates/js/google-analytics/google-analytics.js
```diff
@@ -83,7 +83,7 @@ exports.handler = function(event, context, callback) {
     // allow GET or POST, but only for whitelisted domains
     proxyToGoogleAnalytics(event, done)
   } else {
-    callback(Error('Not found'))
+    callback(new Error('Not found'))
```
ehmicky, line 86: [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/throw-new-error.md)

src/functions-templates/js/oauth-passport/oauth-passport.js
```diff
@@ -17,15 +17,15 @@ app.use(bodyParser.json())
...
+const handleCallback = (req, res) => {
   res.cookie('jwt', req.user.jwt, { httpOnly: true, COOKIE_SECURE }).redirect('/')
 }
 
 app.get(`${ENDPOINT}/auth/github`, passport.authenticate('github', { session: false }))
 app.get(
   `${ENDPOINT}/auth/github/callback`,
   passport.authenticate('github', { failureRedirect: '/', session: false }),
-  handleCallback()
+  handleCallback
```
ehmicky, line 28: The additional function was not needed.

src/lib/http-agent.js
```diff
@@ -21,18 +21,18 @@ class HttpsProxyAgentWithCA extends HttpsProxyAgent {
...
   try {
     proxyUrl = new URL(httpProxy)
-  } catch (e) {
+  } catch (error) {
     log(NETLIFYDEVERR, `${httpProxy} is not a valid URL`)
     exit(1)
   }
 
-  const scheme = proxyUrl.protocol.substr(0, proxyUrl.protocol.length - 1)
+  const scheme = proxyUrl.protocol.slice(0, -1)
```
ehmicky, line 35: [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-string-slice.md)

src/utils/headers.js
```diff
@@ -73,8 +73,8 @@ function parseHeadersFile(filePath) {
       const sepIndex = line.indexOf(':')
       if (sepIndex < 1) throw new Error(`invalid header at line: ${i}\n${lines[i]}\n`)
 
-      const key = line.substr(0, sepIndex).trim()
-      const value = line.substr(sepIndex + 1).trim()
+      const key = line.slice(0, sepIndex).trim()
```
ehmicky, line 76: `sepIndex` is always positive.

src/utils/proxy.js
```diff
@@ -174,7 +174,7 @@ async function serveRedirect(req, res, proxy, match, options) {
     }
 
     if (isExternal(match)) {
-      console.log(`${NETLIFYDEVLOG} Proxying to `, dest.toString())
+      console.log(`${NETLIFYDEVLOG} Proxying to ${dest}`)
```
ehmicky, line 177: `${...}` calls `URL.toString()` under the hood.

src/utils/serve-functions.js
```diff
@@ -114,9 +115,9 @@ function createHandler(dir) {
     // handle proxies without path re-writes (http-servr)
     const cleanPath = request.path.replace(/^\/.netlify\/functions/, '')
 
-    const func = cleanPath.split('/').filter(function(e) {
```
ehmicky, line 117: [See rule documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-array-find.md)

ehmicky:
This a big PR!
The CI tests are failing on Node 8 because `eslint-plugin-unicorn` does not seem to work with Node 8. Looking into it.
Corresponding PR in Netlify Build: https://github.com/netlify/build/pull/1936

.eslintrc.json
ehmicky: I think it might try to detect whether that's the case, although I'm not 100% sure.
One thing where it could be useful in terms of consistency is for Node.js core modules. Some people prefer using the default export, some prefer the named export.

Regardless, this would require too much code change for this single PR, which is why the rule is disabled.

.eslintrc.json
ehmicky: Should we lint against it in the future, and allow individual uses through `eslint-disable` comments?

ehmicky:
`eslint-plugin-unicorn` only works with Node 10. The last version which worked with Node 8 is pretty old, so I solved it by using `.eslintrc.js` instead of `.eslintrc.json` and conditionally enabling this ESLint plugin based on the current Node.js version. Since CI tests are run on multiple Node.js versions, this can be a workaround until we drop Node 8 support. Generally speaking, this allows our linting not be dependent on Node 8 support.

Using a JavaScript also allows comments, so I added some comments above each rule.

ehmicky:
Closed in favor of #1364.

PR #1344, opened 2020-10-07
refactor: use open-api method to manage addons

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/310

**- Test plan**

There are existing tests for almost all `addons` commands. Added a new one for `addons:config` and tested the flow of creating an addon via a function template manually.

refactor: use open-api method to manage addons

image


tests/command.addons.test.js
```diff
@@ -42,19 +40,19 @@ if (process.env.IS_FORK !== 'true') {
...
-    t.is(regex.test(cliResponse), true)
+    t.is(cliResponse.includes('Addon "demo" deleted'), true)
   })
 
   test.after('cleanup', async t => {
     const { execOptions, builder } = t.context
-
     console.log('Performing cleanup')
-    // Run cleanup
-    await callCli(['addons:delete', 'demo', '-f'], execOptions)
```
erezrokah, line 56: We don't need this since we delete the site.

erezrokah:
@ehmicky There is still a lot to be done in `src/commands/functions/create.js` and `src/utils/dev.js` in regards with code cleanup. Don't think we should do it as a part of this PR.
I'm going to tackle `src/utils/dev.js` next.

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+
+const prepareAddonCommand = async ({ context, addonName, validation }) => {
+  const { authenticate, netlify, log, warn, error, exit } = context
+  const { api, site } = netlify
+  const siteId = site.id
+  if (!siteId) {
+    error('No site id found, please run inside a site folder or `netlify link`')
+  }
+
+  await authenticate.call(context)
```
ehmicky, line 93:
```suggestion
  await context.authenticate()
```

Would this be simpler?

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+
+  const [manifest, siteData, addons] = await Promise.all([
```
ehmicky, line 95: Nice parallelization!

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return siteData
+}
+
+const getAddons = async ({ api, siteId, error }) => {
+  let addons
+  try {
+    addons = await api.listServiceInstancesForSite({ siteId })
+  } catch (e) {
+    error(`Failed getting list of addons: ${e.message}`)
```
ehmicky, line 80: Does this throw?
If so, we could directly `return await api...`. (same for the functions above)

src/commands/addons/config.js
```diff
@@ -1,6 +1,4 @@
 const Command = require('../../utils/command')
-const { getAddons, updateAddon } = require('netlify/src/addons')
```
ehmicky, line 2: Nice removal of `require('netlify/src/addons')` :tada:

src/utils/addons/prepare.js
ehmicky, line 80: Should we handle `422` like in https://github.com/netlify/js-client/blob/00b3e939ef133a15db0e9ba8e4e873d748fd442b/src/addons.js#L41?

src/commands/addons/auth.js
```diff
@@ -1,38 +1,18 @@
...
-
-    if (!siteId) {
-      this.log('No site id found, please run inside a site folder or `netlify link`')
-      return false
-    }
-
-    const site = await this.netlify.api.getSite({ siteId })
-    const addons = await getAddons(siteId, accessToken)
-
-    if (typeof addons === 'object' && addons.error) {
```
ehmicky, line 22: This code was removed: is it ok?

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+}
+
+const getCurrentAddon = ({ addons, addonName }) => {
+  const addon = addons.find(addon => addon.service_path === `/.netlify/${addonName}`)
+  return addon
+}
+
+const validateCurrentAddon = ({ addon, validation, addonName, siteData, log, warn, exit }) => {
+  switch (validation) {
```
ehmicky, line 35: Should this validation enum be simplified to a boolean instead?

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+    getAddons({ api, siteId, error }),
+  ])
+
+  let addon
+  if (addonName) {
+    addon = getCurrentAddon({ addons, addonName })
+    validateCurrentAddon({ addon, validation, addonName, siteData, log, warn, exit })
+  }
+
+  return { siteId, manifest, addons, addon, siteData }
```
ehmicky, line 107: Do we need to return `siteId`? The caller could get it from `context.netlify.site.id` directly.

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+  const { api, site } = netlify
+  const siteId = site.id
+  if (!siteId) {
+    error('No site id found, please run inside a site folder or `netlify link`')
+  }
+
+  await authenticate.call(context)
+
+  const [manifest, siteData, addons] = await Promise.all([
+    addonName ? getAddonManifest({ api, addonName, error }) : Promise.resolve(),
```
ehmicky, line 96: Some of the functions calling this function only need one or two of the `manifest`, `siteData` and `addons`. This means they would make unnecessary API calls. Should we optimize to remove those?

We could for example pass an array option to specify which data should be retrieved.

Another way could also be to split this function into three different functions and let the caller do the `await Promise.all([...])`. `context.authenticate()` could be called from within `getAddonManifest()`, etc. so the caller does not need to call it, but it would need to be memoized then.

src/commands/addons/config.js
```diff
@@ -151,52 +128,44 @@ class AddonsConfigCommand extends Command {
...
+      siteId,
+      addon: addonName,
+      instanceId,
+      body: { config: newConfig },
+    })
+    log(`Add-on "${addonName}" successfully updated`)
   } catch (e) {
     error(e.message)
   }
-  if (updateAddonResponse.code === 404) {
```
ehmicky, line 194: Should this `404` log message be kept?

src/commands/functions/create.js
```diff
@@ -336,39 +335,69 @@ async function scaffoldFromTemplate(flags, args, functionsDir) {
...
+    })
+    log(`Add-on "${addonName}" created for ${siteData.name}`)
+    return true
+  } catch (e) {
+    error(e.message)
+  }
+}
+
+async function installAddons(functionAddons = [], fnPath) {
+  if (functionAddons.length > 0) {
```
ehmicky, line 358: [sand] Should we do an early return to decrease identation/nesting?

```js
if (functionAddons.length === 0) {
  return
}
```

src/commands/functions/create.js
```diff
@@ -336,39 +335,69 @@ async function scaffoldFromTemplate(flags, args, functionsDir) {
...
+      log('No site id found, please run inside a site folder or `netlify link`')
       return false
     }
-    this.log(`${NETLIFYDEVLOG} checking Netlify APIs...`)
+    log(`${NETLIFYDEVLOG} checking Netlify APIs...`)
 
-    const siteData = await api.getSite({ siteId })
-    const arr = addons.map(async ({ addonName, addonDidInstall }) => {
-      this.log(`${NETLIFYDEVLOG} installing addon: ` + chalk.yellow.inverse(addonName))
+    const [siteData, siteAddons] = await Promise.all([
```
ehmicky, line 368: Nice parallelization :tada:

tests/command.addons.test.js
```diff
@@ -16,9 +16,8 @@ if (process.env.IS_FORK !== 'true') {
   })
 
   test.serial('netlify addons:list', async t => {
-    const regex = /No addons currently installed/
     const cliResponse = await callCli(['addons:list'], t.context.execOptions)
-    t.is(regex.test(cliResponse), true)
+    t.is(cliResponse.includes('No addons currently installed'), true)
```
ehmicky, line 20:
```suggestion
    t.true(cliResponse.includes('No addons currently installed'))
```

Would this be clearer? (also below)

ehmicky (changes requested):
This is a huge improvement of the addons code quality @erezrokah!

src/utils/addons/prepare.js
```diff
@@ -0,0 +1,110 @@
...
+    log(`- To update this add-on run: ${cmd}`)
+    const deleteCmd = chalk.cyan(`\`netlify addons:delete ${addonName}\``)
+    log(`- To remove this add-on run: ${deleteCmd}`)
+    log()
+    exit(1)
+  }
+}
+
+const getCurrentAddon = ({ addons, addonName }) => {
+  const addon = addons.find(addon => addon.service_path === `/.netlify/${addonName}`)
```
keiko713, line 30: no need to fix this in this PR, but just this check is not really great and we could check this with `addon.slug == addonName` way.
however, sadly realized that we have another mistake in our swagger.yml that `showService` is using service instance as a ref, so we need to fix this before doing that way.
https://github.com/netlify/open-api/blob/6d77ca696837986aa9518d1843bd00cf8e4c387d/swagger.yml#L1576

I'll create a PR

keiko713: ah actually this one is getting addons with `getServices`, which properly use `service` as a ref, so we could use `addon.slug == addonName` way.
anyways this PR is already a big change and it has been working with this way so far, so no need to improve in this PR

keiko713:
thanks Erez for working on this, and Mickael for a thorough review 🎉 really great to close 2 years+ issue (the original issue created 2 years ago :P) 
I'm not really qualified to review the code of this, but looking like a great direction :)

src/utils/addons/prepare.js
erezrokah: `error` prints an error and exists:
https://oclif.io/docs/commands#thiserrormessage-string--error-options-code-string-exit-number-ref-string-suggestions-string

I'm not sure I like this pattern of short circuiting the execution flow, but it does save propagating the error up.

src/commands/addons/config.js
erezrokah: I'm not aware of anywhere we're using that code, so I believe the answer it yes  🎉

src/utils/addons/prepare.js
erezrokah: Oh, didn't realise we're returning the status code of the error:
https://github.com/netlify/micro-api-client/blob/c08716a0d83c4e826bca99ab2a081015c98ebf44/src/index.js#L4
Going to add that back

src/utils/addons/prepare.js
erezrokah: I started with a boolean and had a hard time naming it 🤦
The opposite of `validateExists`  can mean - `don't validate that it exists`, but it our case it means `validate that it doesn't exists`.
If you think a boolean is easier to understand I'm good with changing it

src/utils/addons/prepare.js
erezrokah: Ha, we don't to handle that error code, `js-client` throws an error in that case:
https://github.com/netlify/js-client/blob/00b3e939ef133a15db0e9ba8e4e873d748fd442b/src/methods/response.js#L10

src/commands/addons/auth.js
erezrokah: Since we're using the `js-client` non `ok` status codes will now throw an error and handled as part of the `try/catch`

erezrokah: The old `getAddons` only handled `422`:
https://github.com/netlify/js-client/blob/00b3e939ef133a15db0e9ba8e4e873d748fd442b/src/addons.js#L41

src/utils/addons/prepare.js
erezrokah: I'll remove it. Good point. Only some of the callers use it.

src/utils/addons/prepare.js
ehmicky: Makes sense. :+1:

src/commands/addons/config.js
ehmicky: We should open an issue for this. That's great news!

src/utils/addons/prepare.js
ehmicky: Great. 
It's actually better because it handles non-JSON responses (which sometimes happen due to bugs), and it includes the full response for JSON responses.

src/commands/addons/auth.js
ehmicky: :+1:

src/utils/addons/prepare.js
ehmicky: I think either works :)
A boolean would be simpler, but an enum would more flexible to additional validation cases (but I'm not sure this might happen there).

src/utils/addons/prepare.js
erezrokah: All callers need the call to get `addons`. `list` needs it to display `addons`, but all others require it for validation.
The `manifest` call has a side effect of verifying the user didn't make a typo.
For example `addons:delete non-existing`, would previously print:
```
Add-on non-existing doesn't exist for <site-name>
> Run `netlify addons:create non-existing` to create an instance for this site`
```
Since it was verifying the addon exists in the list of installed addons.

Now it will print:
```
No add-on "non-existing" found. Please double check your add-on name and try again
```

src/commands/addons/config.js
erezrokah: Similar to other `js-client` methods - a 404 status code will throw an error which will be handled in the `try/catch`

src/utils/addons/prepare.js
ehmicky: Oh I see, I did not think of the side effect of `manifest` validation. 
Then it makes all sense :+1:

src/commands/addons/config.js
ehmicky: :+1:

src/commands/addons/config.js
erezrokah: https://github.com/netlify/js-client/issues/173

src/utils/addons/prepare.js
erezrokah: `addon.service_slug` it the way to do it :)

keiko713: ahhh you're right, now I'm looking at where this is called, it's coming from `api.listServiceInstancesForSite`, so it's actually service instance instead of service. cool sounds good to me!

ehmicky (approved):
This is looking great! :+1: 

I have rebased it onto `master` and added one commit to fix some small linting errors due to the new linting rules on `master`.

PR #1360, opened 2020-10-07
Use template strings

ehmicky:
Related to #1343

This fixes the linting errors from `eslint-plugin-unicorn` related to template strings.

### DIFF
--- a/src/utils/proxy.js
+++ b/src/utils/proxy.js
@@ -174,7 +174,7 @@ async function serveRedirect(req, res, proxy, match, options) {
     }
 
     if (isExternal(match)) {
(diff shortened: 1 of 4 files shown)


src/utils/proxy.js
```diff
@@ -174,7 +174,7 @@ async function serveRedirect(req, res, proxy, match, options) {
     }
 
     if (isExternal(match)) {
-      console.log(`${NETLIFYDEVLOG} Proxying to `, dest.toString())
+      console.log(`${NETLIFYDEVLOG} Proxying to ${dest}`)
```
ehmicky, line 177: `dest` is an URL, and `${dest}` automatically calls `dest.toString()`

erezrokah (approved):

PR #1361, opened 2020-10-07
Avoid unnecessary `undefined`

ehmicky:
Related to #1343

This fixes the linting errors from the `eslint-plugin-unicorn` rule [`no-useless-undefined`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-useless-undefined.md).

### DIFF
--- a/src/lib/api.js
+++ b/src/lib/api.js
@@ -20,7 +20,7 @@ const getErrorMessage = async ({ response }) => {
 
 const checkResponse = async ({ response }) => {
   if (!response.ok) {
-    const message = await getErrorMessage({ response }).catch(() => undefined)
+    const message = await getErrorMessage({ response }).catch(() => {})
     const errorPostfix = message && message ? ` and message '${message}'` : ''
     throw new Error(`Request failed with status '${response.status}'${errorPostfix}`)
   }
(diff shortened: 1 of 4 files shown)


src/lib/api.js
```diff
@@ -20,7 +20,7 @@ const getErrorMessage = async ({ response }) => {
 
 const checkResponse = async ({ response }) => {
   if (!response.ok) {
-    const message = await getErrorMessage({ response }).catch(() => undefined)
+    const message = await getErrorMessage({ response }).catch(() => {})
```
erezrokah, line 23: How does this work with https://eslint.org/docs/rules/no-empty-function?
I actually prefer the explicit `undefined` in this case

erezrokah (changes requested):

src/lib/api.js
ehmicky: This is compatible with `no-empty-function`.
Note that this rule catches more cases, such as `return undefined` and `arg = undefined`.

Would you prefer to leave to:
  - leave this rule enabled, and use `=> {}`
  - leave this rule enabled, and use `=> undefined` + `eslint-disable-next-line` (there is no rule option for that specific use case unfortunately) 
  - disable this rule

I am good with any of those solutions, just let me know :)

erezrokah: Let's keep it then

erezrokah (approved):

PR #1355, opened 2020-10-07
Lint RegExps

ehmicky:
Related to #1343

This fixes the linting errors from the `eslint-plugin-unicorn` rule [`better-regex`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/better-regex.md).

PR #1346, opened 2020-10-07
Remove unused variables

ehmicky:
This adds the `no-unused-vars` linting rule to remove unused variables.

PR #1364, opened 2020-10-07
Add `eslint-plugin-unicorn`

ehmicky:
Follow-up on #1343.

This adds [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn). `eslint-plugin-unicorn` contains a bunch of awesome linting rules which are geared towards code consistency and preventing bugs. It's pretty opinionated though, and needs some configuration to fit our current styling.

`eslint-plugin-unicorn` only works with Node 10. The last version which worked with Node 8 is pretty old, so I solved it by using `.eslintrc.js` instead of `.eslintrc.json` and conditionally enabling this ESLint plugin based on the current Node.js version. Since CI tests are run on multiple Node.js versions, this can be a workaround until we drop Node 8 support. Generally speaking, this allows our linting not be dependent on Node 8 support.

Using a JavaScript also allows comments, so I added some comments above each rule.

This PR depends on #1346, #1347, #1348, #1349, #1350, #1351, #1352, #1353, #1354, #1355, #1356, #1357, #1358, #1359, #1360, #1361, #1362, #1363 to be merged first. Until then, the CI tests will fail.


ehmicky:
This is ready for review :+1:

PR #1384, opened 2020-10-12
chore: cleanup format,lint scripts patterns

erezrokah:
This PR does a bit of cleanup to the pattern for linting and formatting files.
It also ignores the build files for our docs site.


package.json
```diff
@@ -42,9 +42,9 @@
     "start": "node ./bin/run",
     "test": "run-s test:*",
     "ava": "ava --verbose",
-    "format": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
+    "format": "prettier --write \"{src,scripts,site/!(dist),tests,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
```
ehmicky, line 45: This includes `site/*/...`, but not `site/...` because of how `**` is interpreted by the underlying globbing library.
For example, `site/fs.js` is not matched anymore.

erezrokah: Good catch. Let me fix this

ehmicky (changes requested):
This is looking good, except for one globbing issue.

package.json
erezrokah: Moved this to `.prettierignore`. The advantage here is that `.prettierignore` is used for editor plugins too.
For `eslint` we already had it in `.eslintignore`

ehmicky (approved):

package.json
ehmicky: Great solution!

PR #1387, opened 2020-10-13
Improve npm scripts

ehmicky:
This PR aims at improving the Netlify CLI npm scripts:
  - Only run ESLint and Prettier on Node 14 in CI, as described in https://github.com/netlify/build/issues/1962 and as implemented in Netlify Build in https://github.com/netlify/build/pull/1969
  - Minor updates (adding some titles) in the GitHub actions `workflow.yml` to make it more similar to [`@netlify/build`'s `workflow.yml`](https://github.com/netlify/build/blob/master/.github/workflows/workflow.yml)
  - Run formatting on `git push`, as described in https://github.com/netlify/build/issues/1950
  - Sort and rename npm scripts
  - Distinguish between `npm test` (local machine) vs `npm test:ci`.
  - Small optimization of the `nyc` arguments to avoid running it twice
  - Add caching to ESLint, and few smaller CLI flags (`--format=codeframe` and `--max-warnings=0`)
  - Make Prettier less verbose (`--loglevel warn`)

This PR includes many unrelated changes that could have been broken down, but because they all modified the npm scripts, it was simpler to group in a single PR.

Please share any thoughts, and we can update it both in this PR and in Netlify Build :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,15 +40,22 @@
   "scripts": {
     "build:site": "cd site && npm ci && npm run build",
     "start": "node ./bin/run",
-    "test": "run-s test:*",
-    "ava": "ava --verbose",
-    "format": "prettier --write \"{src,scripts,site,tests,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:lint": "eslint --fix \"{src,scripts,site,tests,.github}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,tests,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:cli-version": "npm run start -- --version",
-    "test:cli-help": "npm run start -- --help",
-    "test:install-deps": "npm ci --prefix tests/site-cra",
-    "test:ava": "nyc --reporter=lcov ava --verbose && nyc report",
+    "test": "run-s format test:dev",
+    "format": "run-s format:check-fix:*",
+    "format:ci": "run-s format:check:*",
+    "format:check-fix:lint": "./scripts/run-on-error.js format:check:lint format:fix:lint",
+    "format:check:lint": "cross-env-shell eslint $npm_package_scriptsArgs_eslint",
+    "format:fix:lint": "cross-env-shell eslint --fix $npm_package_scriptsArgs_eslint",
+    "format:check-fix:prettier": "./scripts/run-on-error.js format:check:prettier format:fix:prettier",
+    "format:check:prettier": "cross-env-shell prettier --check $npm_package_scriptsArgs_prettier",
+    "format:fix:prettier": "cross-env-shell prettier --write $npm_package_scriptsArgs_prettier",
+    "test:dev": "run-s test:init:* test:dev:*",
+    "test:ci": "run-s test:init:* test:ci:*",
+    "test:init:cli-version": "npm run start -- --version",
+    "test:init:cli-help": "npm run start -- --help",
+    "test:init:install-deps": "npm ci --prefix tests/site-cra",
+    "test:dev:ava": "ava --verbose",
+    "test:ci:ava": "nyc -r lcovonly -r text -r json ava --verbose",
     "docs": "node ./site/scripts/docs.js",
     "watch": "nyc --reporter=lcov ava --watch",
     "prepack": "oclif-dev manifest && npm prune --prod",
@@ -59,6 +66,10 @@
     "release:version": "git push && git push --tags && gh-release",
     "release:publish": "npm publish",
     "postinstall": "node ./scripts/postinstall.js"
+  },
+  "scriptsArgs": {
+    "eslint": "--cache --format=codeframe --max-warnings=0 \"{src,scripts,site,tests,.github}/**/*.{js,md,html}\"",
+    "prettier": "--loglevel=warn \"{src,scripts,site,tests,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\""
   },
   "dependencies": {
     "@netlify/build": "^5.0.0",
@@ -158,6 +169,7 @@
     "auto-changelog": "^2.0.0",
     "ava": "^2.4.0",
     "babel-eslint": "^10.1.0",
+    "cross-env": "^7.0.2",
     "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.0.0",
     "eslint-config-standard": "^14.1.1",
@@ -222,7 +234,8 @@
   },
   "husky": {
     "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
+      "pre-push": "npm run format"
     }
   },
   "standard-version": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -40,15 +40,17 @@
...
-    "test:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:cli-version": "npm run start -- --version",
-    "test:cli-help": "npm run start -- --help",
-    "test:install-deps": "npm ci --prefix tests/site-cra",
-    "test:ava": "nyc --reporter=lcov ava --verbose && nyc report",
+    "test": "npm run format && npm run test:dev",
+    "format": "run-s format:*",
+    "format:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
+    "format:prettier": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
```
ehmicky, line 46: This merges `prettier --check` + `prettier --write`. This is slightly slower, but might avoid duplication in the npm scripts. What do you think?

package.json
```diff
@@ -40,15 +40,17 @@
...
-    "format": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:cli-version": "npm run start -- --version",
-    "test:cli-help": "npm run start -- --help",
-    "test:install-deps": "npm ci --prefix tests/site-cra",
-    "test:ava": "nyc --reporter=lcov ava --verbose && nyc report",
+    "test": "npm run format && npm run test:dev",
+    "format": "run-s format:*",
+    "format:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
```
ehmicky, line 45: `test:lint` is renamed to `format:lint` since both ESLint and Prettier are skipped in Node 8 in CI, and they are both somewhat related.

package.json
```diff
@@ -40,15 +40,17 @@
...
-    "test": "run-s test:*",
-    "ava": "ava --verbose",
-    "format": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:cli-version": "npm run start -- --version",
-    "test:cli-help": "npm run start -- --help",
-    "test:install-deps": "npm ci --prefix tests/site-cra",
-    "test:ava": "nyc --reporter=lcov ava --verbose && nyc report",
+    "test": "npm run format && npm run test:dev",
```
erezrokah, line 43:
```suggestion
    "test": "run-s format test:dev",
```

package.json
```diff
@@ -40,15 +40,17 @@
   "scripts": {
     "build:site": "cd site && npm ci && npm run build",
     "start": "node ./bin/run",
-    "test": "run-s test:*",
-    "ava": "ava --verbose",
-    "format": "prettier --write \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
-    "test:lint": "eslint --fix \"{.github,src,scripts,site,tests}/**/*.{js,md,html}\"",
-    "test:format": "prettier --check \"{src,scripts,site,site/src,tests,.circleci,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\"",
```
erezrokah: `prettier --write` won't fail the script if the formatting is not valid (as opposed to `--check`). That can happen if someone edits the code directly on GitHub or disables commits hooks.

Maybe we can use package json variables to share the configuration?

This is also relevant for `eslint --fix` (that was already an issue with the previous scripts).

ehmicky:
Rebased onto `master`.

package.json
```diff
@@ -222,7 +224,8 @@
   },
   "husky": {
     "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
+      "pre-push": "npm run format"
```
erezrokah, line 228: At the moment the `npm run format` won't fail if `eslint` or `prettier` modify the files, so the committer will have to do another commit and push it.
Do we want those commands to fail and prevent the push? That will save some CI jobs that are bound to fail.

erezrokah (changes requested):
Thank you @ehmicky. This is great and much better than what we have now.

I think we would need to have `format:ci` and `format:dev` as well to address some of my comments.
The pre push hook could use `format:ci`, WDYT?

package.json
ehmicky: I though `prettier --write` exit code was `1` if a file was modified, but you're right, this is not the case.
`eslint --fix` does give an exit code that is `1` on linting errors, but not if the error was autofixed, so you're right on this one too.

I will fix this.

Could you please describe what you mean by `package.json` variables to share the configuration?

package.json
ehmicky: Yes, absolutely, we do. I had wrong assumptions on `eslint` and `prettier` exit codes. I will fix this.

package.json
erezrokah: Can we do something like this:
```json
{
    "name"   : "myapp",
    "prettier" : { "files" : "...." },
    "scripts": {
        "format:write" : "prettier --write $npm_package_prettier_files",
        "format:check" : "prettier --check $npm_package_prettier_files"
    },
    ...
}
```

Haven't tested that one, but it should work

ehmicky: Great solution! Yes, this should work, let's try it.

scripts/run-on-error.js
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env node
+const execa = require('execa')
+const { argv } = require('process')
+
+const [, , npmScript, npmScriptOnError] = argv
```
ehmicky, line 5: This script is only intended for our npm scripts, so I am not sure whether using a CLI input parsing library is needed.

scripts/run-on-error.js
```diff
@@ -0,0 +1,27 @@
...
+
+// Run a npm script. If that script fails, another npm script is run.
+// We use this for example with ESLint and Prettier to be able to fail if
+// anything should be autofixed, while still autofixing it. Those tools do not
+// provide good CLI flags for this.
+const runOnError = async function() {
+  const { failed, exitCode = DEFAULT_ERROR_EXIT_CODE } = await runNpmScript(npmScript)
```
ehmicky, line 12: `exitCode` can be `undefined` when the script has failed due to a standard stream error or a signal termination.

package.json
ehmicky: Fixed.

erezrokah (approved):

erezrokah:
🎉

PR #1388, opened 2020-10-14
test(command-dev-trace): re-enable tests

erezrokah:
This PR reintroduces `netlify dev:trace` command tests with some improvements:
1. Uses the latest version of `gh-release-fetch` that allows passing fetch options. We use it to pass an auth token that allows for much higher rate limits.
2. Pre-installs the traffic mesh agent so tests can run in parallel.


calavera (approved):

ehmicky:
Nice! :tada:

PR #1389, opened 2020-10-15
fix: fix parcel build command being chosen over dev command

### LINKED ISSUE #1252: "netlify dev" is running incorrect command for parcel
**Describe the bug**

`netlify dev` command is running `npm run build` insted of `npm run dev` when parcel bundler is detected.
I noticed that this happens when my `build` command comes before my `dev` command in `package.json`

eg : 

```json
"scripts": {
    "build": "parcel build index.html",
    "dev": "parcel index.html",
}
```
This will trigger `npm run build` when `netlify dev` command is run.

```json
"scripts": {
    "dev": "parcel index.html",
    "build": "parcel build index.html",
}
```
But this will trigger `npm run dev` when `netlify dev` command is run.

**Expected behaviour**

Since parcel has a dev command ( `parcel <filename>`) which creates a live server, `netlify dev` should run `npm run dev` if it is present and `npm run build` only when `dev` command is absent.

monsonjeremy:
**- Summary**

This PR closes #1252. The issue here is that for Parcel the build and development commands do not have anything to sufficiently differentiate them given how the current detector logic works. Parcel's command are as follows:
development: `parcel index.html`
build: `parcel build index.html`

With the current detector logic, both commands will be matched since they both include `parcel`, and whichever occurs first in the scripts will be the one that is used.

The solution to this is to add some sorting of the commands to give preference to commands that exist within the `preferredScriptsArr`.

**- Test plan**

Steps:
1. Create a basic repo:
```sh
mkdir parcel-test
cd parcel-test
npm init
npm install -S parcel
```
2. Add the following scripts to `package.json`
```json
  "scripts": {
    "build": "parcel build index.html",
    "dev": "parcel index.html"
  },
```
3. Run `netlify dev`

fix parcel build command being chosen over dev command


erezrokah (commented):
Thanks @monsonjeremy.
I'm going to do some manual testing on this with other frameworks.
Sorting the scripts to get consistent results makes sense, just want to make sure we don't breaking anything.

monsonjeremy:
@erezrokah Great, let me know if I can help in any way!

erezrokah:
Sorry for the very very late response @monsonjeremy  and thank you for this.
We have some other changes coming to detectors logic via https://github.com/netlify/cli/issues/1495 so I prefer to have it fixed as part of that issue. We'll be testing all frameworks in the process and strive to simplify the detectors code.
We'll keep #1252 in mind while we do so.

PR #1391, opened 2020-10-15
chore: upgrade prettier

erezrokah:
Follow up on https://github.com/netlify/build/pull/1972


ehmicky (approved):

erezrokah:
Looking at why this is failing on windows

PR #1395, opened 2020-10-15
Add `array-callback-return` ESLint rule

ehmicky:
This adds the [`array-callback-return`](https://eslint.org/docs/rules/array-callback-return) ESLint rule. 

This is a useful linting rule, and actually I think applying it in this PR might have found some bugs.

### DIFF
--- a/src/functions-templates/js/google-analytics/google-analytics.js
+++ b/src/functions-templates/js/google-analytics/google-analytics.js
@@ -16,7 +16,9 @@ function whitelistDomain(domain, addWww = true) {
     prefixes.push('https://www.')
     prefixes.push('http://www.')
   }
-  prefixes.forEach((prefix) => originWhitelist.push(prefix + domain))
+  prefixes.forEach((prefix) => {
+    originWhitelist.push(prefix + domain)
+  })
 }
 
 function proxyToGoogleAnalytics(event, done) {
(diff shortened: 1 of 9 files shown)


src/functions-templates/js/google-analytics/google-analytics.js
```diff
@@ -107,8 +109,8 @@ sr: 1440x900
 vp: 945x777
 je: 0
 _u: blabla~
-jid: 
-gjid: 
+jid:
+gjid:
```
ehmicky, line 113: Unrelated change. Not sure why Prettier/ESLint did not pick this.

tests/utils/dev-server.js
```diff
@@ -35,7 +35,7 @@ const startServer = async ({ cwd, env = {}, args = [] }) => {
           close: async () => {
             selfKilled = true
             const pids = await pidtree(ps.pid).catch(() => [])
-            pids.forEach((pid) => () => {
+            pids.forEach((pid) => {
```
ehmicky, line 38: I believe this was a bug. What do you think @erezrokah ?

erezrokah (approved):

tests/utils/dev-server.js
erezrokah: Yes I think so. A bad shameful copy paste probably

PR #1397, opened 2020-10-15
Add `multiline-comment-style` ESLint rule

ehmicky:
This adds the [`multiline-comment-style`](https://eslint.org/docs/rules/multiline-comment-style) ESLint rule.

This uses the `//` style because it is easier to comment/uncomment individual lines. Also, it is the closest to our current style, so this requires the least modification. However, I'm happy with other styles if you'd prefer :+1:


erezrokah (approved):
Big yes to this

PR #1404, opened 2020-10-15
Use `isexe` module

ehmicky:
Some of our code is a full copy/paste of the [`isexe`](https://github.com/isaacs/isexe) dependency. I happened to remember it because I once looked at the code of this library, but our inlining was lacking attribution (this dependency license requires it).

By using the module, we also benefit from:
  - async instead of sync
  - Windows support


src/lib/exec-fetcher.js
```diff
@@ -55,7 +34,7 @@ const isVersionOutdated = async ({ packageName, currentVersion }) => {
 const shouldFetchLatestVersion = async ({ binPath, packageName, execName, execArgs, pattern, log }) => {
   const execPath = path.join(binPath, getExecName({ execName }))
 
-  const exists = await execExist(execPath)
+  const exists = await isExe(execPath)
```
ehmicky, line 37: `isexe` contains the code of `execExist()` not of `isExe()`.
See [source](https://github.com/isaacs/isexe/blob/master/mode.js)

src/lib/exec-fetcher.js
```diff
@@ -25,27 +25,6 @@ const getOptions = () => {
...
-
-  return Boolean(mode & 0o0001 || (mode & 0o0010 && isGroup) || (mode & 0o0100 && isUser))
-}
-
-const execExist = async (binPath) => {
-  try {
-    const stat = await fs.statAsync(binPath)
-    return stat.isFile() && isExe(stat.mode, stat.gid, stat.uid)
-  } catch (error) {
-    if (error.code === 'ENOENT') return false
```
ehmicky, line 44: This `ENOENT` check is the reason we use `ignoreErrors: true`. However, this will ignore all errors, not only `ENOENT`.
Without that option, only `EACCESS` is ignored.

erezrokah (approved):

PR #1406, opened 2020-10-15
Add `no-invalid-this` ESLint rule

ehmicky:
This adds the [`no-invalid-this`](https://eslint.org/docs/rules/no-invalid-this) ESLint rule.

A file is passing `this` around instead of using class methods, which is odd. We should either use class methods there, or pass the `context` object around. This PR chooses the second pattern for the time being, because this generates the smallest diff.

### DIFF
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -28,11 +28,11 @@ class FunctionsCreateCommand extends Command {
   async run() {
     const { flags, args } = this.parse(FunctionsCreateCommand)
     const { config } = this.netlify
-    const functionsDir = ensureFunctionDirExists.call(this, flags, config)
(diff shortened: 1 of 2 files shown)


src/commands/functions/create.js
```diff
@@ -263,8 +263,8 @@ function installDeps(functionPath) {
 }
 
 // no --url flag specified, pick from a provided template
-async function scaffoldFromTemplate(flags, args, functionsDir) {
-  const chosentemplate = await pickTemplate.call(this) // pull the rest of the metadata from the template
+async function scaffoldFromTemplate(context, flags, args, functionsDir) {
+  const chosentemplate = await pickTemplate() // pull the rest of the metadata from the template
```
ehmicky, line 267: Context is not used.

erezrokah (approved):

PR #1419, opened 2020-10-16
Add `default-param-last` ESLint rule

ehmicky:
This adds the [`default-param-last`](https://eslint.org/docs/rules/default-param-last) ESLint rule.

### DIFF
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -358,7 +358,7 @@ async function createFunctionAddon({ api, addons, siteId, addonName, siteData, l
   }
 }
 
-async function installAddons(context, functionAddons = [], fnPath) {
+async function installAddons(context, functionAddons, fnPath) {
   if (functionAddons.length === 0) {
     return
   }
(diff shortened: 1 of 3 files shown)


src/commands/functions/create.js
```diff
@@ -356,7 +356,7 @@ async function createFunctionAddon({ api, addons, siteId, addonName, siteData, l
   }
 }
 
-async function installAddons(functionAddons = [], fnPath) {
+async function installAddons(functionAddons, fnPath) {
```
ehmicky, line 359: `functionAddons` cannot be `undefined` here.

src/utils/proxy.js
```diff
@@ -298,7 +298,7 @@ function initializeProxy(port, distDir, projectDir) {
   return handlers
 }
 
-async function startProxy(settings = {}, addonUrls, configPath, projectDir) {
+async function startProxy(settings, addonUrls, configPath, projectDir) {
```
ehmicky, line 301: `settings` cannot be `undefined` here.

erezrokah (approved):

PR #1421, opened 2020-10-16
Add `func-style` ESLint rule

ehmicky:
This adds the [`func-style`](https://eslint.org/docs/rules/func-style) ESLint rule.

### DIFF
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -70,7 +70,7 @@ module.exports = FunctionsCreateCommand
  */
 
 // prompt for a name if name not supplied
-async function getNameFromArgs(args, flags, defaultName) {
+const getNameFromArgs = async function (args, flags, defaultName) {
   if (flags.name) {
     if (args.name) {
       throw new Error('function name specified in both flag and arg format, pick one')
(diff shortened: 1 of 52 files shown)


src/commands/functions/create.js
```diff
@@ -152,40 +150,43 @@ async function pickTemplate() {
     },
   })
   return chosentemplate
-  function filterRegistry(registry, input) {
```
ehmicky, line 155: Those functions can be safely moved to the top-level scope.

erezrokah (changes requested):
Should we enable https://eslint.org/docs/rules/func-style#allowarrowfunctions?

ehmicky:
That works :+1: 

Should we try to ensure consistency (`const myFunc = function(...){...}` vs `const myFunc = (...) => {...}`). Both styles are almost the same and work, but would it be useful to enforce consistency over one specific style?
If so, which ESLint rule would be the best to enforce this? I have struggled with this specific linting problem in my personal projects, but there might be a solution.

erezrokah:
> That option is always enabled when the rule is set to `expression` (the default),

Cool, didn't notice that.

I would also prefer to choose one of the two styles. Don't think it is a blocker for this PR (especially if we don't have built in support for it).

ehmicky:
Rebased. Ready for another review :+1:

erezrokah (approved):

ehmicky:
This is enforceable with `no-restricted-syntax`. However, enforcing arrow functions would also require writing:

```js
module.exports = {
  onBuild() {...}
}
```

As:

```js
module.exports = {
  onBuild: () => {...}
}
```

Which might be more verbose?

Also, this would not come with autofixing, so I am thinking this specific rule might be more of a nuisance than a help there, and leaving inconsistent style might be more practical. What do you think?

erezrokah:
> Also, this would not come with autofixing, so I am thinking this specific rule might be more of a nuisance than a help there, and leaving inconsistent style might be more practical. What do you think?

Agreed

PR #1423, opened 2020-10-16
Improve the performance of proxy rules parsing

ehmicky:
This improves the performance of proxy rules parsing by running asynchronous operations in parallel.


erezrokah (approved):
🎉

PR #1424, opened 2020-10-16
Improve performance of static proxy

ehmicky:
This improves the performance of the static proxy by checking for file existence in parallel.


erezrokah (approved):
This is great!

PR #1427, opened 2020-10-16
Add `no-shadow` ESLint rule

ehmicky:
This adds the [`no-shadow`](https://eslint.org/docs/rules/no-shadow) ESLint rule.

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -1,4 +1,4 @@
-const { Command, flags } = require('@oclif/command')
+const { Command, flags: flagsLib } = require('@oclif/command')
 const API = require('netlify')
 const merge = require('lodash.merge')
 const { format, inspect } = require('util')
@@ -47,7 +47,7 @@ class BaseCommand extends Command {
     // Get site id & build state
     const state = new StateConfig(cwd)
 
-    const cachedConfig = await this.getConfig(cwd, state, token, argv)
+    const cachedConfig = await this.getConfig(cwd, state, token)
     const { configPath, config, buildDir } = cachedConfig
 
     const { flags } = this.parse(BaseCommand)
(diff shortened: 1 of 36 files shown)


src/utils/command.js
```diff
@@ -91,7 +91,7 @@ class BaseCommand extends Command {
   }
 
   // Find and resolve the Netlify configuration
-  async getConfig(cwd, state, token, argv) {
```
ehmicky, line 94: This is taken from the top-level scope, so it does not need to be passed as argument.

tests/utils/dev-server.js
```diff
@@ -7,12 +7,12 @@ const pidtree = require('pidtree')
 
 // each process gets a starting port based on the pid
 const rng = seedrandom(`${process.pid}`)
-function getRandomPortStart(rng) {
+function getRandomPortStart() {
   const startPort = Math.floor(rng() * 10000) + 10000 // 10000 to avoid collisions with frameworks ports
   return startPort
 }
 
-let currentPort = getRandomPortStart(rng)
```
ehmicky, line 15: This is taken from the top-level scope, so it does not need to be passed as argument.

tests/utils/dev-server.js
```diff
@@ -43,6 +43,7 @@ const startServer = async ({ cwd, env = {}, args = [] }) => {
               }
             })
             ps.kill()
+            // eslint-disable-next-line no-shadow
             await Promise.race([ps.catch(() => {}), new Promise((resolve) => setTimeout(resolve, 1000))])
```
ehmicky, line 47: This conflicts with a ESLint rule which requires promise first callback to be named `resolve`.
The real solution here is not to nest two `Promise`, but that'd be a bigger refactoring.

erezrokah (approved):

PR #1428, opened 2020-10-16
Configure `ava/no-only-test` ESLint rule

ehmicky:
`test.only()` and `test.skip()` are useful in debugging. Enabling those ESLint rules would prevent from using those due to the autofix on file save.


ehmicky:
Not needed anymore because included by #1462.

PR #1433, opened 2020-10-16
Remove unnecessary quotes

ehmicky:
This removes unnecessary quotes, using Prettier.


ehmicky:
Rebased. What do you think @erezrokah ?

erezrokah (approved):

PR #1434, opened 2020-10-16
Use trailing commas

ehmicky:
This uses Prettier to enforce trailing commas.
Trailing commas generate better git diffs. They also prevent forgetting the comma of the previous item when appending a new item to an object or array.

This PR's diff was generated by Prettier.


erezrokah (approved):
YES Please :)

PR #1435, opened 2020-10-16
Add `.editorconfig`

ehmicky:
This adds an [`.editorconfig` file](https://editorconfig.org/).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+max_line_length = 120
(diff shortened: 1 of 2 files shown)


erezrokah (changes requested):
I think we can set `options.editorconfig` to `true` in `.prettierrc.json` and remove the `printWidth` setting from it, WDYT?
See https://prettier.io/docs/en/api.html#prettierresolveconfigfilepath--options

ehmicky:
Good point! I have just removed the redundant Prettier option `printWidth`.
The `editorconfig` option in Prettier is `true` by default, so we don't need to explicitly enable it (I have just tried it locally to make sure).

erezrokah (approved):

PR #1439, opened 2020-10-16
Speedup ESLint and Prettier

ehmicky:
This speeds up and simplifies ESLint and Prettier by using `.gitignore` (which skips more files) instead of `.prettierignore` and `.eslintignore`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,0 @@
-node_modules
-site-cra
-site/dist
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,1 +1,0 @@
-site/dist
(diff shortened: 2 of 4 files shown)


.eslintignore
```diff
@@ -1,3 +0,0 @@
-node_modules
-site-cra
-site/dist
```
ehmicky, line 3: Those are all inside `.gitignore`.

.prettierignore
```diff
@@ -1 +0,0 @@
-site/dist
```
ehmicky, line 1: This is inside `.gitignore`.

erezrokah (approved):
🎉

PR #1440, opened 2020-10-16
Add Codecov test coverage

ehmicky:
This adds Codecov for test coverage.

It's quite useful to see the test coverage of a PR by looking at the PR status. 
This disables PR comments because they can be noisy.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,3 +36,5 @@ jobs:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
           IS_FORK: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == true }}
           NETLIFY_TEST_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Codecov test coverage
+        run: bash scripts/coverage.sh "${{ matrix.os }}" "${{ matrix.node-version }}"
--- a/scripts/coverage.sh
+++ b/scripts/coverage.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Upload test coverage to Codecov
+
+os="$1"
+os="${os/-latest/}"
+
+node="$2"
+node="node_${node//./_}"
+
+# We don't use the -Z flag (which makes CI build fail on upload error) because
+# Codecov fails wait too often.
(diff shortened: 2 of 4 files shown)


scripts/coverage.sh
```diff
@@ -0,0 +1,13 @@
...
+os="$1"
+os="${os/-latest/}"
+
+node="$2"
+node="node_${node//./_}"
+
+# We don't use the -Z flag (which makes CI build fail on upload error) because
+# Codecov fails wait too often.
+curl -s https://codecov.io/bash | \
+  bash -s -- -f coverage/coverage-final.json -F "$os" -F "$node"
```
ehmicky, line 13: We need a Bash script in order to pass the `-F` tags (they do not accept `.` nor `-`).

erezrokah (changes requested):
Cool 😄 Looks like it is failing at the moment:
https://github.com/netlify/cli/pull/1440/checks?check_run_id=1266323266#step:7:33

ehmicky:
This is a [known bug](https://github.com/codecov/codecov-bash/issues/330) in Codecov.
Apparently, the Codecov report are still looking good, so I am assuming the data is still uploaded. See report [here](https://codecov.io/gh/netlify/cli/pull/1440/tree?path=src). From that angle, would it still be useful to merge?

erezrokah:
> would it still be useful to merge?

Sure, go for it

erezrokah (approved):

.github/workflows/main.yml
```diff
@@ -36,3 +36,5 @@ jobs:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
           IS_FORK: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == true }}
           NETLIFY_TEST_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Codecov test coverage
```
erezrokah, line 39: Also, should be do this just for one node version and OS? Any benefit from uploading all results? Does Codecov knows to differentiate?

ehmicky: Yes, it merges them together, and you can even see per-OS results.
That's useful for OS-specific code. For example, if some code only runs on Windows, and another never runs on Windows, we need to run both to get the proper test coverage.

PR #1456, opened 2020-10-21
fix(command-dev-trace): pin traffic mesh agent version

erezrokah:
This PR fixes the `netlify dev:trace` command by pinning the traffic mesh agent to the latest known working version

### DIFF
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -8,6 +8,8 @@ const { shouldFetchLatestVersion, fetchLatestVersion } = require('../lib/exec-fe
 const PACKAGE_NAME = 'traffic-mesh-agent'
 const EXEC_NAME = 'traffic-mesh'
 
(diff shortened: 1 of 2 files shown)


src/utils/traffic-mesh.js
```diff
@@ -18,6 +20,8 @@ const installTrafficMesh = async ({ log }) => {
     execArgs: ['--version'],
     pattern: '\\sv(.+)',
     execName: EXEC_NAME,
+    latestVersion: LATEST_VERSION,
+    log,
```
erezrokah, line 24: The `log` argument was missing. It is used here https://github.com/netlify/cli/pull/1456/files#diff-f8e66f1838601194468460f6e49a5bad17e2a2700afce376413bb7a4a9c8a290R66

ehmicky (approved):

erezrokah:
Just realised one test is still running. Going to wait for it before releasing a new CLI version

PR #1458, opened 2020-10-21
feat(command-dev-local): pass path to bundler plugin CLI to mesh forward

NeoLegends:
**- Summary**

To make resolution more rigid and to avoid bloating the user's PATH with internal tools, mesh-forward now wants the path to the CLI of the edge handler build plugin passed in via its CLI. In this PR we resolve the path to the CLI via `require.resolve` and then pass it on to mesh-forward.

~~Please note the version of the build plugin specified in the dependencies here **does not yet** ship with the CLI, that's why this is a draft (it would be super nice to get this released, but I don't have the permissions to do that).~~

**- Test plan**

I wonder what's the best way forward here.

Ideally both repos (mesh-forward & cli) had separate integration tests against one another, but I'm afraid right now only `cli` as a consumer of mesh-forward actually has them.

I made changes to mesh-forward that will allow unpinning the mesh-forward version again, but eventually we do want to consume it via npm anyway, such that the version again is fixed.

🐞

### DIFF
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -10,6 +10,11 @@ const EXEC_NAME = 'traffic-mesh'
 
 const LATEST_VERSION = 'v0.22.1'
 
+const EDGE_HANDLERS_BUNDLER_CLI_PATH = path.join(
+  path.dirname(require.resolve('@netlify/plugin-edge-handlers')),
+  'cli.js'
+)
+
 const getBinPath = () => getPathInHome([PACKAGE_NAME, 'bin'])
 
 const installTrafficMesh = async ({ log }) => {
@@ -48,6 +53,8 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
     `http://localhost:${frameworkPort}`,
     '--watch',
     publishDir,
+    '--bundler',
+    EDGE_HANDLERS_BUNDLER_CLI_PATH,
     '--log-file',
     getPathInProject(['logs', 'traffic-mesh.log']),
   ]
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -11,7 +11,7 @@ const EXEC_NAME = 'traffic-mesh'
 const LATEST_VERSION = 'v0.22.1'
 
 const EDGE_HANDLERS_BUNDLER_CLI_PATH = path.resolve(
-  path.join(path.dirname(require.resolve('@netlify/plugin-edge-handlers')), 'cli.js')
+  path.join(path.dirname(require.resolve('@netlify/plugin-edge-handlers')), 'cli.js'),
 )
 
 const getBinPath = () => getPathInHome([PACKAGE_NAME, 'bin'])
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -10,9 +10,7 @@ const EXEC_NAME = 'traffic-mesh'
 
 const LATEST_VERSION = 'v0.22.1'
 
-const EDGE_HANDLERS_BUNDLER_CLI_PATH = path.resolve(
-  require.resolve('@netlify/plugin-edge-handlers'), '..', 'cli.js',
-)
+const EDGE_HANDLERS_BUNDLER_CLI_PATH = path.resolve(require.resolve('@netlify/plugin-edge-handlers'), '..', 'cli.js')
 
 const getBinPath = () => getPathInHome([PACKAGE_NAME, 'bin'])
 
--- a/src/utils/traffic-mesh.js
+++ b/src/utils/traffic-mesh.js
@@ -10,9 +10,8 @@ const EXEC_NAME = 'traffic-mesh'
 
 const LATEST_VERSION = 'v0.22.1'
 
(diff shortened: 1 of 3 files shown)


src/utils/traffic-mesh.js
```diff
@@ -48,6 +52,8 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
     `http://localhost:${frameworkPort}`,
     '--watch',
     publishDir,
+    '--bundler',
+    EDGE_HANDLERS_BUNDLER_CLI_PATH,
```
ehmicky, line 56: Just making sure: will this work on Windows? Specifically, inside `cmd.exe`, which does not handle shebangs, i.e. `/path/to/cli.js` does not work, but `node /path/to/cli.js` does.

src/utils/traffic-mesh.js
```diff
@@ -10,6 +10,10 @@ const EXEC_NAME = 'traffic-mesh'
 
 const LATEST_VERSION = 'v0.22.1'
 
+const EDGE_HANDLERS_BUNDLER_CLI_PATH = path.resolve(
+  path.join(path.dirname(require.resolve('@netlify/plugin-edge-handlers')), 'cli.js'),
```
ehmicky, line 14:
```suggestion
  require.resolve('@netlify/plugin-edge-handlers'), 'cli.js'
```

ehmicky, line 14: Note that using `require.resolve()` might make it harder for this repository to switch to ES modules in the future.

I am wondering: the `plugin-build-edge-handlers` binary should be available in `node_modules/.bin`. This can be retrieved with [`npm-run-path`](https://github.com/sindresorhus/npm-run-path) for example, then added to the `$PATH`, so that the binary can be executed using only the binary name (as opposed to full path).

I am not sure this is a practical solution, and it might complicate things more than help. Just raising this as a possible alternative.

src/utils/traffic-mesh.js
NeoLegends: Good catch! Somebody's alert 😉

It will work due to https://github.com/netlify/traffic-mesh/pull/1190

src/utils/traffic-mesh.js
NeoLegends: I'm trying to avoid putting internal binaries on `PATH`, so I'd prefer passing the full path to the binary to mesh-forward.

I do agree with the last part though, `npm-run-path` might be a more future-proof implementation. I'll look into that.

src/utils/traffic-mesh.js
ehmicky: Yes, that works :+1:

src/utils/traffic-mesh.js
ehmicky: > I'm trying to avoid putting internal binaries on PATH, so I'd prefer that we keep passing the full path to the binary to mesh-forward.

Yes, good point. Also, adding `node_modules/.bin` would also bring the other locally installed Node.js binaries to the `PATH`, which might not be a good side effect. The current solution makes sense in that light :+1:

NeoLegends:
CI is still red.

src/utils/traffic-mesh.js
NeoLegends: Let's roll the `npm-run-path` changes in another PR. I'd like to move forward with the implementation and then I'll optimize. If that's okay for you -- of course.

ehmicky: Just to clarify, I actually think using `npm-run-path` might be a bad idea due to:

> Also, adding node_modules/.bin would also bring the other locally installed Node.js binaries to the PATH, which might not be a good side effect. 

So the current solution looks good to me.

ehmicky (approved):

src/utils/traffic-mesh.js
NeoLegends: Ah, I didn't look at its docs yet. I thought it's just a library that was able to resolve a locally installed node binary, such that we could replace `require.resolve(...)` with a solution that worked even for ES modules.

ehmicky: It depends. 
When run in ES modules mode, Node.js [does remove `require.resolve()`](https://nodejs.org/api/esm.html#esm_no_require_resolve). 
However, the experimental `import.meta.resolve()` might be used instead.

NeoLegends: I'm open to anything when that time comes 🙌

src/utils/traffic-mesh.js
```diff
@@ -10,6 +10,8 @@ const EXEC_NAME = 'traffic-mesh'
 
 const LATEST_VERSION = 'v0.22.1'
```
erezrokah, line 11: We would need to update this too right?

NeoLegends: Yes. There's no release of mesh-forward with the new features yet though. I'll post a PR here once that's ready.

NeoLegends:
CI seems red for unrelated reasons?

erezrokah:
> CI seems red for unrelated reasons?

Yes, trying a re-run

NeoLegends:
I see you're not the only ones having to deal with flaky CI 😄

PR #1461, opened 2020-10-22
Expose bundling status to CLI

NeoLegends:
**- Summary**

Based on #1458, this exposes the edge handler bundling status to the CLI.

**- Test plan**

- [x] Release mesh-forward with support for the `--progress` option
- [x] Source mesh-forward from npm
- [x] Add tests bundling edge handlers

Expose edge handler bundler status on CLI

🐘


erezrokah (changes requested):
Thanks @NeoLegends, I know this is a draft, but added some comments.
I can test it a bit more once the new release for traffic mesh it published.

src/utils/traffic-mesh.js
```diff
@@ -92,6 +97,65 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
   }
 }
 
+const forwardMessagesToLog = ({ log, subprocess }) => {
+  let currentId = null
+  let spinner = null
+
+  subprocess.stderr
```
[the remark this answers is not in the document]
NeoLegends: >Can we use something like: [...] To avoid using https://github.com/dominictarr/event-stream?

I can do that.

> Also, should we be reading stdout?

Currently mesh-forward outputs the status events to stderr, to be resilient to the `log-file` option being present or not. If it's not present mesh-forward will output tracing logs to stdout which might cause conflicts if we log progress messages at the same time.

I am looking into creating a separate pipe just for those messages. The `spawn` function has the ability to open multiple pipes, but I'm not yet sure how to consume them in a cross-platform way from Rust.

> We should also change the way we spawn the process as we're using stdio: 'inherit'

Fair.

src/utils/traffic-mesh.js
```diff
@@ -92,6 +97,65 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
+          cb(null, data)
+        } catch (error) {
+          log(NETLIFYDEVERR, 'cannot parse log line as JSON, ignoring')
+          // don't call callback
+        }
+      }),
+    )
+    .on('data', ({ error, id, type }) => {
+      switch (type) {
+        case 'bundle:start':
```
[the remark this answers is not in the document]
NeoLegends: Yes this case is to be expected, which is why the success & error cases return early if the ID doesn't match up anymore.

I was under the impression starting multiple spinners would work, but I agree we should handle this case more explicitly.

src/utils/traffic-mesh.js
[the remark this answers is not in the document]
ehmicky: Could we also just read the stream as is, with `data` events?

src/utils/traffic-mesh.js
```diff
@@ -92,6 +97,67 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
   }
 }
 
+const forwardMessagesToLog = ({ log, subprocess }) => {
+  let currentId = null
+  let spinner = null
+
+  subprocess.stderr
```
ehmicky, line 104: Note: we might want to handle `error` events on this stream.

src/utils/traffic-mesh.js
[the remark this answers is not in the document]
NeoLegends: I don't think we should depend on the data events coming line-by-line as we get them from Rusts `eprintln!(...)`. If the pipe is buffered for some random reason this will break our code.

src/utils/traffic-mesh.js
```diff
@@ -57,6 +59,7 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
     EDGE_HANDLERS_BUNDLER_CLI_PATH,
     '--log-file',
     getPathInProject(['logs', 'traffic-mesh.log']),
+    '--progress',
```
ehmicky, line 62: Quick question: do we need to spawn this is a new process, as opposed to calling it programmatically? Assuming this is a Node.js process. This would remove the need to parse `stderr`, as information could passed directly using callbacks for example.

src/utils/traffic-mesh.js
NeoLegends: Does a pipe error if the child exits unexpectedly? I imagine it'd just return `eof` at some point and that's it.

src/utils/traffic-mesh.js
NeoLegends: This is calling a Rust process.

src/utils/traffic-mesh.js
ehmicky: There a few reason why an `error` event would be triggered on a stream. In that case, we might want to do some cleanup of the spinner for example.

NeoLegends: Sounds fair.

src/utils/traffic-mesh.js
ehmicky: I am wondering if terminal output might be a good channel for IPC. What are your thoughts on some alternatives like using the filesystem (e.g. watching for a file), or using a small TCP server?

NeoLegends: Using pipes for IPC is actually a common choice. You _do_ often set up dedicates pipes for passing the IPC messages though, and don't use stdout / stderr (though I've also seen stdout being used). One prominent example: https://github.com/xi-editor/xi-editor

Pipes have the advantage of being being simple to set up (one syscall), they are automatically connected, they are available on all operating systems, they don't fail and they are automatically cleaned up when you're done with them. Tempfiles and TCP sockets both are either more complex to set up and synchronize or take effort for cleanup. The multiple-connection machine-to-machine IPC case is generally also better served with UDS, but that doesn't exist on Windows other than a super recent insiders build.

Signals might be worth an investigation, but I imagine it's going to be really hard passing actually usable information across them except for "build started", "build complete", and "build failed". This approach can ship detailed error information across process boundaries.

src/utils/traffic-mesh.js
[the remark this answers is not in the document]
NeoLegends: Fixed. We now avoid opening multiple spinners if the old one is still running.

src/utils/traffic-mesh.js
ehmicky: Those are all very good points. :+1:

NeoLegends: I am not sure but I think nodejs also uses pipes if you use their IPC implementation. At least, that would make a lot of sense. 😄

src/utils/traffic-mesh.js
```diff
@@ -92,6 +97,81 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
+const forwardMessagesToLog = ({ log, subprocess }) => {
+  let currentId = null
+  let spinner = null
+
+  const reset = () => {
+    currentId = null
+    spinner = null
+  }
+
+  rl.createInterface(subprocess.stderr)
```
ehmicky, line 109: Should this use [`historySize: 0`](https://nodejs.org/api/readline.html#readline_readline_createinterface_options)?

ehmicky, line 109: Should the `options` object syntax be used instead, since it is the [main one documented](https://nodejs.org/api/readline.html#readline_readline_createinterface_options)?

NeoLegends: Oh, I'm afraid I got things messed up there. I was actually referring to https://www.npmjs.com/package/readline when implementing this. Will take a look and ensure everything is correct!

src/utils/traffic-mesh.js
```diff
@@ -92,6 +97,81 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
+          log(`${NETLIFYDEVLOG} Change any project file to trigger a re-bundle`)
+          reset()
+          break
+
+        default:
+          log(`${NETLIFYDEVWARN} Unknown mesh-forward event '${type}'`)
+          break
+      }
+    })
+    .on('end', () => {
```
ehmicky, line 156: Would the `close` event be [better](https://nodejs.org/api/readline.html#readline_event_close)?

src/utils/traffic-mesh.js
ehmicky: > I am not sure but I think nodejs also uses pipes if you use their IPC implementation. 

Yes, [it does](https://github.com/nodejs/node/blob/master/lib/internal/child_process.js#L985), with [JSON or v8-serialized payloads](https://github.com/nodejs/node/blob/master/lib/internal/child_process/serialization.js) depending on the `serialization` option.

NeoLegends:
Hey @erezrokah there's now a build we can test out next week: https://github.com/netlify/traffic-mesh-agent/releases/tag/v0.24.0-pre.7 This also already includes a packed npm package we can publish to npm eventually after verifying it's good.

NeoLegends:
Status update: edge handlers are run & rerun (after changes) successfully, but tests are failing because we (mesh-forward) seem to be violating the HTTP spec by returning `Content-Encoding: chunked` and a `Content-Length` at the same time. A browser will swallow that, `node-fetch` won't however, so we have to fix that.

In addition, this requires the latest mesh-forward prerelease, so we should be merging #1469 before this one.

tests/utils/site-builder.js
```diff
@@ -26,7 +26,19 @@ const createSiteBuilder = ({ siteName }) => {
     siteName,
     withNetlifyToml: ({ config, pathPrefix = '' }) => {
       const dest = path.join(directory, pathPrefix, 'netlify.toml')
-      const content = toToml(config, { space: 2 })
+      const content = toToml(config, {
+        replace: (_, val) => {
+          // Strip off `.0` from integers that tomlify normally generates
+
+          if (!Number.isFinite(val) || val % 1 !== 0) {
```
ehmicky, line 33:
```suggestion
          if (!Number.isInteger(val)) {
```

tests/utils/site-builder.js
```diff
@@ -50,14 +62,18 @@ const createSiteBuilder = ({ siteName }) => {
       })
       return builder
     },
-    withEdgeHandlers: ({ handlers }) => {
-      const dest = path.join(directory, 'edge-handlers', 'index.js')
+    withEdgeHandlers: ({ name, handlers }) => {
```
ehmicky, line 65:
```suggestion
    withEdgeHandlers: ({ name = 'index.js', handlers }) => {
```

NeoLegends:
Thanks @ehmicky for the suggestions! 🤘🏻

I did fix the spec violation in v0.24.0-pre.10, which is in #1469.

tests/utils/site-builder.js
```diff
@@ -50,14 +62,18 @@ const createSiteBuilder = ({ siteName }) => {
       })
       return builder
     },
-    withEdgeHandlers: ({ handlers }) => {
-      const dest = path.join(directory, 'edge-handlers', 'index.js')
+    withEdgeHandlers: ({ name = 'index.js', handlers }) => {
+      if (!name) {
```
ehmicky, line 66: Those three lines can be removed thanks to the default parameter above.

mraerino:
Whats blocking us from merging this?

ehmicky:
CI tests are currently failing and this PR is still in draft mode.

mraerino:
@NeoLegends can you take a look? we'd like to get this to customers if we can

NeoLegends:
Sorry for the delay. The tests were due to invalid use of the fetch response object. This is fixed now. :)

NeoLegends:
Not sure how those commits slipped into the PR, the diff is much cleaner now.

ehmicky:
Thanks @NeoLegends!
It looks like the test is currently failing on Windows.

NeoLegends:
Let's fire up the Windows PC 😄

ehmicky (approved):

ehmicky:
:tada: Does this look good for you too @erezrokah?

src/utils/traffic-mesh.js
```diff
@@ -64,13 +77,98 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
+          if (currentId !== id) {
+            return
+          }
+
+          stopSpinner({
+            spinner,
+            error: true,
+            text: (error && error.msg) || 'Failed bundling Edge Handlers',
+          })
+          log(`${NETLIFYDEVLOG} Change any project file to trigger a re-bundle`)
```
erezrokah, line 134:
```suggestion
          log(`${NETLIFYDEVLOG} Change any project file to trigger an Edge Handlers re-bundle`)
```

Also a change to any project file would trigger a re-bundle? Or just changes to files under `edge-handlers`

erezrokah (dismissed):
Thanks @NeoLegends and @ehmicky, I added a few more comments, mostly on how logs are printed. It's better to be explicit as these logs are mixed with other CLI logs.

erezrokah:
@calavera and @NeoLegends. This needs a rebase before merging and also a response on:
https://github.com/netlify/cli/pull/1461#discussion_r520499951
https://github.com/netlify/cli/pull/1461#discussion_r520501054
https://github.com/netlify/cli/pull/1461#discussion_r520509505

Not sure why the GitHub actions were not triggered on the latest push. I'm guessing a rebase will solve that

NeoLegends:
I did not find the time to work on the feedback yesterday. Will to today!

NeoLegends:
Hey @erezrokah I think I addressed your feedback. Thank you for that!

The test seems flaky on Windows and I'm investigating why. I might push some temporary commits that allow me to access the logs. Or do you know another way?

NeoLegends:
Tests were green, rerunning to see if it was just a coincidence or if they're stable now.

NeoLegends:
CI seems good.

ehmicky (commented):
Looks good. :+1: 
What do you think @erezrokah?

erezrokah (changes requested):
Thanks @NeoLegends, sorry for taking so long to respond to this.
I tested this PR again.
- Not having an `edge-handlers` directory prints a message that bundling succeeded:
image Can we remove it?
- Can we remove the `an unknown error occured during bundling: `, it seems redundant:
image and there is a typo `occured->occurred`
- In the above screenshot we should be consistent and use `Edge Handlers` instead of `edge handlers`.

src/utils/traffic-mesh.js
```diff
@@ -40,19 +45,27 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
       subprocess.kill(sig, {
         forceKillAfterTimeout: PROXY_EXIT_TIMEOUT,
       })
     })
   })
 
+  // Wait until the first edge handlers bundle is ready
+  //
+  // In case of errors, this delays the startup process until the errors are fixed
+  await firstBundleReady
```
erezrokah, line 67: Is this required? The server can still run even with bundle failure right?

NeoLegends: We won't run any edge handler until the first bundle is done. If the user then accesses the site there is a chance they'll see invalid / unexpected responses if an edge handler is supposed to run on a set of routes but actually doesn't.

This is more of a product decision, however, so therefore, your call. ✌️

erezrokah: > We won't run any edge handler until the first bundle is done. If the user then accesses the site there is a chance they'll see invalid / unexpected responses if an edge handler is supposed to run on a set of routes but actually doesn't.

Will this also happen if the server was started successfully, but at a later stage bundling failed?
Do we fall back to previously bundled handlers or stop running all handlers?

NeoLegends: In that case we'll fall back to previously bundled handlers.

erezrokah: Cool, so in that case I think we should keep the code as is due to the different semantics.

NeoLegends:
> Not having an edge-handlers directory prints a message that bundling succeeded:

Dang, you're right! I think we just need to change the message on this one.

What's actually happening here is that traffic-mesh is done creating its internal representation of the local project, which includes bundling & loading edge handlers, but also the parsing of e. g. redirect rules and header rules.

WDYT about something like "Loading project...", "Done loading project..." and "Failed loading project" instead (overall very bad suggestions, but you get the hint). We'd also need a message conveying to the user which files they can change to trigger a re-bundle (which would be `_headers`, `_redirects`, `netlify.toml` and any edge handlers file).

NeoLegends:
> In the above screenshot we should be consistent and use Edge Handlers instead of edge handlers.

This is an issue in the build plugin and I created https://github.com/netlify/netlify-plugin-edge-handlers/pull/104 to solve this. Would appreciate your review.

src/utils/traffic-mesh.js
```diff
@@ -64,13 +77,100 @@ const startForwardProxy = async ({ port, frameworkPort, functionsPort, publishDi
...
+          }
+          break
+
+        case 'bundle:success':
+          if (currentId !== id) {
+            return
+          }
+
+          firstBundleResolve()
+          stopSpinner({ spinner, error: false, text: 'Done processing' })
```
erezrokah, line 120: Not sure what is the correct message here (or vs and), but we should have better granularity since the log is printed with other CLI logs:
image

NeoLegends:
CI is a bit flaky but otherwise should be good.

@erezrokah The new version of traffic-mesh-agent removed the "unknown error while bundling" text and also uses consistent casing everywhere.

erezrokah (changes requested):
Thanks @NeoLegends. Clearing the spinner is a very good solution.
I'm getting a weird behavior of the log `Processing request remaps, header rules and Edge Handlers` being repeated twice on a single file change (e.g. a single save to `_redirects` or an edge handler).

I would add a GIF, but https://github.com/justinfrankel/licecap/issues/97.

I added a `console.log` here https://github.com/netlify/cli/pull/1461/files#diff-7893aee6262507710747eb1595c225c78567f01ac1957261a9aa41603d7c4938R99 and it confirms that behavior.

Can you confirm that behavior? Is this something local to my machine?

NeoLegends:
Hey @erezrokah, are you on a Mac?

Based on [the documentation of the file watching library](https://docs.rs/notify/4.0.15/notify/trait.Watcher.html#tymethod.new) we're using in mesh-forward, I suspect this is a quirk of the platform. The library suggests debouncing the FS-events by more than 30s to fix the repeated events, but that would certainly drive down our responsiveness to changes by the users too much.

I think we should opt to leave the behavior as-is and see if we can supply a fix in a follow-up. Since we're not duplicating log lines anymore I suspect it won't be too much of a burden if we sometimes process twice.

erezrokah:
> Hey @erezrokah, are you on a Mac?

Yes!


> I think we should opt to leave the behavior as-is and see if we can supply a fix in a follow-up. Since we're not duplicating log lines anymore I suspect it won't be too much of a burden if we sometimes process twice.

Cool, makes sense

erezrokah (approved):
Marked this as `low-risk` since it only impacts a hidden feature

erezrokah:
Released in https://github.com/netlify/cli/releases/tag/v2.69.0

Issue #1437, opened 2020-10-16
Consistent filename case


ehmicky:
The CLI uses a `kebab-case` convention, while `@netlify/build` uses `snake_case`. 
Should we keep it consistent between projects? If so, which one should we use?
I personally think `snake_case` is more idiomatic in Node.js, but I'm fine with any case, as long as we're consistent. What do you think @erezrokah?

Note: both currently lint this using the [`unicorn/filename-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/filename-case.md) ESLint rule, but with different options.

erezrokah:
Let's go with `snake_case`

sarahetter:
Not a priority

PR #1462, opened 2020-10-23
Use `@netlify/eslint-config-node`

ehmicky:
This uses the shared ESLint and Prettier configuration at `@netlify/eslint-config-node` ([link to repository](https://github.com/netlify/eslint-config-node)).

Unfortunately, it is not possible to do the same for `.editorconfig` and `.gitattributes`, which must be copy/pasted instead.


.eslintrc.js
```diff
@@ -1,196 +1,18 @@
...
-    'plugin:eslint-comments/recommended',
-    'plugin:ava/recommended',
-    'plugin:react/recommended',
-    'prettier/react',
-    'plugin:you-dont-need-lodash-underscore/all',
-  ],
-  reportUnusedDisableDirectives: true,
+  extends: '@netlify/eslint-config-node',
   rules: {
-    'require-await': 2,
```
ehmicky, line 18: All those rules and ESLint configuration are now sourced from `@netlify/eslint-config-node`

.eslintrc.js
```diff
@@ -1,196 +1,18 @@
...
-
-    'react/prop-types': 0,
-
-    // Not enabled by default in unicorn/recommended, but still pretty useful
-    'unicorn/custom-error-definition': 2,
-    'unicorn/no-unused-properties': 2,
-    // The additional `non-zero` option is useful for code consistency
-    'unicorn/explicit-length-check': [2, { 'non-zero': 'not-equal' }],
     // TODO: harmonize with filename snake_case in other Netlify Dev projects
     'unicorn/filename-case': [2, { case: 'kebabCase', ignore: ['.*.md'] }],
```
ehmicky, line 7: We currently need to keep this, because of #1437

.eslintrc.js
```diff
@@ -1,196 +1,18 @@
...
   },
   overrides: [
-    {
-      files: '**/*.test.js',
-      rules: {
-        'node/no-unpublished-require': 0,
-        'node/no-missing-require': 0,
-      },
-    },
+    ...overrides,
```
ehmicky, line 10: `extends` does not extend `overrides`, so we need to manually do it.

.eslintrc.js
```diff
@@ -1,196 +1,18 @@
...
     {
```
ehmicky, line 11: Those `overrides` are specific to this project.

.prettierrc.json
```diff
@@ -1,6 +1 @@
-{
-  "semi": false,
-  "singleQuote": true,
-  "proseWrap": "always",
-  "trailingComma": "all"
-}
+"@netlify/eslint-config-node/.prettierrc.json"
```
ehmicky, line 1: Re-use shared Prettier configuration.

We want to use a `.prettierrc.json` and `.eslintrc.js` as opposed to using a `package.json` properties, because I personally found some tools (like VSCode) to not recognize Prettier/ESLint configuration when inside a `package.json` sometimes.

package.json
```diff
@@ -68,8 +68,8 @@
     "postinstall": "node ./scripts/postinstall.js"
   },
   "scriptsArgs": {
-    "eslint": "--ignore-path .gitignore --cache --format=codeframe --max-warnings=0 \"{src,scripts,site,tests,.github}/**/*.{js,md,html}\"",
-    "prettier": "--ignore-path .gitignore --loglevel=warn \"{src,scripts,site,tests,.github}/**/*.{js,md,yml,json,html}\" \"*.{js,yml,json,html}\""
+    "eslint": "--ignore-path .gitignore --cache --format=codeframe --max-warnings=0 \"{src,scripts,site,tests,.github}/**/*.{js,md,html}\" \"*.{js,md,html}\" \".*.{js,md,html}\"",
```
ehmicky, line 71: Lint top-level files too, e.g. `.eslintrc.js` itself.

package.json
```diff
@@ -168,27 +168,12 @@
...
-    "eslint-plugin-html": "^6.1.0",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-markdown": "^1.0.2",
-    "eslint-plugin-node": "^11.0.0",
-    "eslint-plugin-prettier": "^3.1.4",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-react": "^7.20.5",
-    "eslint-plugin-standard": "^4.0.1",
-    "eslint-plugin-unicorn": "^23.0.0",
-    "eslint-plugin-you-dont-need-lodash-underscore": "^6.10.0",
```
ehmicky, line 191: Those are all `dependencies` of `@netlify/eslint-config-node`.

Technically, the "proper" way is to define those as `peerDependencies` of `@netlify/eslint-config-node`, then install those as `devDependencies` of each project that uses `@netlify/eslint-config-node`. However, this would mean each time one ESLint plugin has a new version, we would need to merge Renovate PRs in each project.

Installing those as `dependencies` of `@netlify/eslint-config-node` instead works due to a hack: npm/yarn dependencies deduping makes those available to the top-level `node_modules/*` directory instead of `node_modues/@netlify/eslint-config-node/node_modules/*`. Without this deduping, this would not work because ESLint `require()` those plugins directly in the parent project's root directory.

npm@7 might solve this since `peerDependencies` are now automatically installed by parent projects, which might solve this problem.

erezrokah (approved):

PR #1469, opened 2020-10-26
Install traffic-mesh via npm package

NeoLegends:
**- Summary**

This configures the CLI to load `traffic-mesh` via npm instead of via downloading it manually the first time `netlify dev` is run.

~~The package itself is quite heavy right now (40mb) because it contains the binaries for all operating systems, but we can always improve on that later by shipping a stripped package that downloads the right binaries on installation. For the MVP I wanted to avoid the additional moving parts this brings.~~

**- Test plan**

This should not change any behavior, so the existing tests should continue working.

Install `traffic-mesh` component via npm package.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,5 @@ jobs:
           # We set a flag so we can skip tests that access Netlify API
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
           IS_FORK: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == true }}
-          NETLIFY_TEST_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Codecov test coverage
         run: bash scripts/coverage.sh "${{ matrix.os }}" "${{ matrix.node-version }}"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1638,9 +1638,9 @@
       }
     },
     "@netlify/traffic-mesh-agent": {
-      "version": "0.24.0-pre.10",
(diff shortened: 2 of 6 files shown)


ehmicky:
> The package itself is quite heavy right now (40mb) because it contains the binaries for all operating systems

Can anything be done to remove some bytes, such as removing debugging symbols, etc.?

> but we can always improve on that later by shipping a stripped package that downloads the right binaries on installation.

The main concern with a bigger binary size is that this would add time when installing `netlify-cli`. Downloading on `postinstall` would add even more time, so I think the current solution might be best.

NeoLegends:
> Can anything be done to remove some bytes, such as removing debugging symbols, etc.?

Yeah the binaries aren‘t stripped yet.

NeoLegends:
Follow-up: Stripping binaries removes any useful information from their stack traces (in case of panics), so I'd opt to leave the debug symbols in them for now and only remove those later.

NeoLegends:
Follow up 2: I'm working on leveraging optional dependencies to make npm only download the package you need for your platform.

This gives us the benefit of using npm for all the I/O and has only marginally increased complexity.

See https://github.com/netlify/traffic-mesh/pull/1217

NeoLegends:
Download size per platform is now:

- macOS: 12.8MB
- Linux: 16.6MB
- Windows: 11.2MB

NeoLegends:
Hey @ehmicky @erezrokah, just wanted to ask your status on this PR. I'm happy to move forward with this PR and #1461, but I'm not sure about your schedule and / or whether you're currently blocked on other things. :)

I think I'll have the capacity to finish the integration next week, so up to you, whenever you have time.

erezrokah (approved):
This looks good to me, but let's wait on https://github.com/netlify/cli/pull/1505 before merging.

erezrokah:
`npm ci` failed:
https://github.com/netlify/cli/runs/1341703190?check_suite_focus=true#step:4:4
Going to re-run it.

PR #1472, opened 2020-10-26
Add `plugin:node/recommended`

ehmicky:
This adds `plugin:node/recommended` and few additional rules, like in Netlify Build.

Note: CI tests currently failing. Looking into it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,14 +40,6 @@ module.exports = {
         'node/no-unsupported-features/es-syntax': 0,
       },
     },
-    // Document site's development scripts
-    {
-      files: ['site/*.js', 'site/scripts/**/*.js'],
-      rules: {
-        'node/no-unpublished-require': 0,
-        'node/no-extraneous-require': 0,
-      },
-    },
     // Example functions
     {
       files: ['src/functions-templates/**/*.js'],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,8 +38,6 @@ module.exports = {
       },
       rules: {
         'node/no-unsupported-features/es-syntax': 0,
-        // TODO: enable
-        'node/no-extraneous-import': 0,
       },
     },
     // Document site's development scripts
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,17 +1,61 @@
 const { overrides } = require('@netlify/eslint-config-node')
 
 module.exports = {
-  extends: '@netlify/eslint-config-node',
+  extends: ['@netlify/eslint-config-node', 'plugin:node/recommended'],
   rules: {
+    'node/handle-callback-err': 2,
+    'node/no-new-require': 2,
+    'node/exports-style': 2,
+    'node/file-extension-in-import': 2,
+    'node/no-mixed-requires': 2,
+    // Browser globals should not use `require()`. Non-browser globals should
+    'node/prefer-global/console': 2,
+    'node/prefer-global/text-decoder': 2,
+    'node/prefer-global/text-encoder': 2,
+    // This does not work well in a monorepo
+    'node/shebang': 0,
+
+    // TODO: enable this
+    'no-process-exit': 0,
+    // 'node/no-sync': 2,
+    // 'node/callback-return': 2,
+    // 'node/global-require': 2,
+    // 'node/prefer-global/url-search-params': 2,
+    // 'node/prefer-global/url': 2,
+    // 'node/prefer-global/buffer': [2, 'never'],
+    // 'node/prefer-global/process': [2, 'never'],
+
     // TODO: harmonize with filename snake_case in other Netlify Dev projects
     'unicorn/filename-case': [2, { case: 'kebabCase', ignore: ['.*.md'] }],
   },
   overrides: [
     ...overrides,
+    // Documentation site's browser JavaScript
     {
-      files: 'src/functions-templates/**/*.js',
+      files: ['site/src/**/*.js'],
+      parserOptions: {
+        sourceType: 'module',
+      },
+      rules: {
+        'node/no-unsupported-features/es-syntax': 0,
+        // TODO: enable
+        'node/no-extraneous-import': 0,
+      },
+    },
+    // Document site's development scripts
+    {
+      files: ['site/*.js', 'site/scripts/**/*.js'],
+      rules: {
+        'node/no-unpublished-require': 0,
+        'node/no-extraneous-require': 0,
+      },
+    },
+    // Example functions
+    {
+      files: ['src/functions-templates/**/*.js'],
       rules: {
         'require-await': 0,
+        'node/no-missing-require': 0,
       },
     },
   ],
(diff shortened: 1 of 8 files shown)


.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+    // Browser globals should not use `require()`. Non-browser globals should
+    'node/prefer-global/console': 2,
+    'node/prefer-global/text-decoder': 2,
+    'node/prefer-global/text-encoder': 2,
+    // This does not work well in a monorepo
+    'node/shebang': 0,
+
+    // TODO: enable this
+    'no-process-exit': 0,
+    // 'node/no-sync': 2,
```
ehmicky, line 20: We can add a separate for sync I/O

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+    // 'node/callback-return': 2,
```
ehmicky, line 21: This rule might require a separate PR.

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+    'node/file-extension-in-import': 2,
+    'node/no-mixed-requires': 2,
+    // Browser globals should not use `require()`. Non-browser globals should
+    'node/prefer-global/console': 2,
+    'node/prefer-global/text-decoder': 2,
+    'node/prefer-global/text-encoder': 2,
+    // This does not work well in a monorepo
+    'node/shebang': 0,
+
+    // TODO: enable this
```
ehmicky, line 18: I will add separate PRs for those rules.
Except for `no-process-exit`, since the changes might be too big.

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+    // Documentation site's browser JavaScript
+    {
+      files: ['site/src/**/*.js'],
+      parserOptions: {
+        sourceType: 'module',
+      },
+      rules: {
+        'node/no-unsupported-features/es-syntax': 0,
+        // TODO: enable
+        'node/no-extraneous-import': 0,
```
ehmicky, line 42: `site/src/**/*.js` use many `import` to dependencies that seem to be indirect dependencies, but not direct dependencies: `rebass`, `react`, `styled-system`, `react-router-dom`, `prop-types`, `lodash.sortby`, `@rebass/markdown`.
Should we add those to `package.json`?

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
   overrides: [
     ...overrides,
+    // Documentation site's browser JavaScript
+    {
+      files: ['site/src/**/*.js'],
+      parserOptions: {
+        sourceType: 'module',
+      },
+      rules: {
+        'node/no-unsupported-features/es-syntax': 0,
```
ehmicky, line 40: Site JavaScript uses ES modules

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+        'node/no-extraneous-import': 0,
+      },
+    },
+    // Document site's development scripts
+    {
+      files: ['site/*.js', 'site/scripts/**/*.js'],
+      rules: {
+        'node/no-unpublished-require': 0,
+        'node/no-extraneous-require': 0,
+      },
```
ehmicky, line 51: We need to tell `eslint-plugin-node` that those are development files. This might not be the best way to do this?

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
...
+    },
+    // Example functions
     {
-      files: 'src/functions-templates/**/*.js',
+      files: ['src/functions-templates/**/*.js'],
       rules: {
         'require-await': 0,
+        'node/no-missing-require': 0,
```
ehmicky, line 58: Those files are just examples, so dependencies might be missing.

package.json
```diff
@@ -18,7 +18,8 @@
     "/npm-shrinkwrap.json",
     "/oclif.manifest.json",
     "/scripts/postinstall.js",
-    "/src/**/*.js"
+    "/src/**/*.js",
+    "!/src/**/*.test.js"
```
ehmicky, line 22: We should not publish test files. Is this globbing expression correct?

.eslintrc.js
```diff
@@ -1,17 +1,61 @@
 const { overrides } = require('@netlify/eslint-config-node')
 
 module.exports = {
-  extends: '@netlify/eslint-config-node',
+  extends: ['@netlify/eslint-config-node', 'plugin:node/recommended'],
```
erezrokah, line 4: Can we share these configs via `@netlify/eslint-config-node`? Is this planned for another PR?

package.json
erezrokah: Good catch, that looks correct.

erezrokah (approved):

.eslintrc.js
erezrokah: Fixed in https://github.com/netlify/cli/pull/1472/commits/2e8ebaa77814bb1fea31d26a884c21951663e226

.eslintrc.js
erezrokah: Not sure. This seems ok to me

.eslintrc.js
ehmicky: Yes, the goal of this PR (and upcoming PRs) is to put those in `@netlify/eslint-config-node`.

We could either:
  1. Add those to `@netlify/eslint-config-node`, make a release, then upgrade in CLI/Netlify Build
  2. Or add the rules to CLI/Netlify Build, then do `1.`

I went with `2.` because it allows us to discuss those rules, and see their impact to the repositories before we commit to them in `@netlify/eslint-config-node`. What do you think?

.eslintrc.js
ehmicky: Thanks! :tada:

.eslintrc.js
ehmicky: Improved in d81af33f4f3a888e1e9c93a4d07041a49beaa37f:
  - Use `package.json` `files` so `eslint-plugin-node` can distinguish between files meant for production or development. We don't publish `/site` to npm, so this is only meant for linting/static analysis
  - One `devDependencies` was missing (`strip-ansi`). This only worked because it was an dependency's dependency too.

ehmicky: This seems to fail in CI (although this works locally). Looking into it.

.eslintrc.js
erezrokah: Sounds good to me

.eslintrc.js
ehmicky: This was failing because the `site` was missing `node_modules` in CI. I fixed by adding `npm run build:site` to the CI steps.

erezrokah: Thanks, we can probably only do the `npm ci` part (and not the build) as it is done in the Netlify CI.
Not a big thing though

ehmicky: Good point. Fixed in b19a552e4b029f92d69834d4d58414f1c88a7ab9

erezrokah (approved):

ehmicky:
Looking good now! Ready for another review.

Issue #694, opened 2020-02-03
Following CONTRIBUTING.md instructions for testing results in test errors


gr2m:
Hello there 👋 

I'm the maintainer of [`@octokit/rest`](https://github.com/octokit/rest.js/). I'm looking to test [the upcoming v17 release](https://github.com/octokit/rest.js/issues/1546) with a few dependents to catch errors before you do 😉 I've tried setting up your respository using the instructions at https://github.com/netlify/cli/blob/master/CONTRIBUTING.md, but I see a lot of errors

<details>
<summary>
<code>$ npm test</code>
</summary>

```
 npm test

> netlify-cli@2.31.0 test /Users/gregor/Projects/octokit/rest.js-dependants/netlify-cli
> run-s test:*


> netlify-cli@2.31.0 test:ava /Users/gregor/Projects/octokit/rest.js-dependants/netlify-cli
> nyc --reporter=lcov ava --verbose && nyc report


  ✔ utils › parse-raw-flags › JSONTruthy works with various inputs
creating new site for tests: netlify-test-tftzlmbx
Running command...
CMD: "/Users/gregor/Projects/octokit/rest.js-dependants/netlify-cli/bin/run" sites:create --name="netlify-test-tftzlmbx" --account-slug="netlify-services"
DIR: /Users/gregor/Projects/octokit/rest.js-dependants/netlify-cli/src/tests/dummy-site
  ✔ utils › parse-raw-flags › parseRawFlags works
  ✔ utils › find-root › index › Find root in a git project
  ✔ utils › find-root › index › Find root in a linked folder
  ✔ utils › find-root › index › Find root in a netlify site folder
  ✔ utils › find-root › index › Find root with no indicators
Event name must have single underscore. "testProject:test" contains 0
-----------------------------
Tracking Error:
Incorrect tracking event format: "testProject:test"
-----------------------------
Formatting information: 
eventName must match 'cli:objectName_actionName'
eventName must be camelCased: 'camelCase:camelCase_camelCase'
Example: cli:sites_deploySucceeded
-----------------------------
Event name must have single colon. "testProjecttest_eventName" contains 0
-----------------------------
Tracking Error:
Incorrect tracking event format: "testProjecttest_eventName"
-----------------------------
Formatting information: 
eventName must match 'cli:objectName_actionName'
eventName must be camelCased: 'camelCase:camelCase_camelCase'
Example: cli:sites_deploySucceeded
-----------------------------
-----------------------------
Tracking Error:
[...]

PR #1476, opened 2020-10-27
fix: support index.js function name

ardiewen:
**- Summary**

Per the [Build with Javascript documentation](https://docs.netlify.com/functions/build-with-javascript/), functions ought to support the following naming formats:

- my_functions/functionname.js
- my_functions/functionname/functionname.js
- my_functions/functionname/index.js

Currently, this works perfectly fine on Netlify hosted production environments, but fails when running `netlify dev` via the CLI.

For more details on how to replicate this problem, including a minimal reproduction, please see this [support thread](https://community.netlify.com/t/netlify-dev-function-returns-404-when-using-functions-functionname-index-js-syntax-instead-of-functions-functionname-js/24124/2) I created on the Community Forums.

**- Test plan**

1. I first tested some basic scenarios in a production environment hosted on Netlify to understand the expected behaviour when different naming conventions (and different combinations) are used, for example:

   ```
   .
   |-- my_functions
   |    |    |-- bob
   |    |    |    |-- bob.js
   |    |    |    |-- index.js
   |    |    |-- test
   |    |    |    |-- test.js
   |    |    |-- test2
   |    |    |    |-- index.js
   |    |    |-- test3
   |    |    |    |-- index.js
   |    |    |    |-- test3.js
   |    |    |-- hello-world.js
   ```

   This can be verified here: https://gatsby-cloud-netlify-functions.netlify.app/
   e.g. https://gatsby-cloud-netlify-functions.netlify.app/.netlify/functions/hello-world.js etc...

2. In the above scenarios, everything works as expected with the ambiguous cases being `bob` and `test3` where both `functionname/index.js` and `functionname/functionname.js` files exist. In these cases, the production environment takes the the named function file, if it exists, then the index file, if it exists AND the named function file doesn't exist.

   This behaviour can be confirmed here:
   Uses bob.js: https://gatsby-cloud-netlify-functions.netlify.app/.netlify/functions/bob
   Uses test3.js: https://gatsby-cloud-netlify-functions.netlify.app/.netlify/functions/test3

3. This PR reflects this behaviour in the local `netlify dev` environment.

4. Per [this conversation](https://github.com/netlify/cli/issues/694#issuecomment-716717020), there is a failing test for `should deploy edge handlers when directory exists` but all other tests appear to be passing fine (as far as I can see).

Adds missing support for naming functions using the `my_functions/functionname/index.js` convention per official documentation.

dog tax:
IMG_0786

### DIFF
--- a/src/utils/finders.js
+++ b/src/utils/finders.js
@@ -18,17 +18,15 @@ const findHandler = function (functionPath) {
     return functionPath
   }
 
-  let handlerPath
-  const indexHandlerPath = path.join(functionPath, `index.js`)
   const namedHandlerPath = path.join(functionPath, `${path.basename(functionPath)}.js`)
-
   if (fs.existsSync(namedHandlerPath)) {
-    handlerPath = namedHandlerPath
+    return namedHandlerPath
   }
-  if (fs.existsSync(indexHandlerPath) && !fs.existsSync(namedHandlerPath)) {
-    handlerPath = indexHandlerPath
+
+  const indexHandlerPath = path.join(functionPath, `index.js`)
+  if (fs.existsSync(indexHandlerPath)) {
+    return indexHandlerPath
   }
-  return handlerPath
 }
 
 module.exports = { findModuleDir, findHandler }
--- a/src/utils/finders.js
+++ b/src/utils/finders.js
@@ -18,9 +18,15 @@ const findHandler = function (functionPath) {
     return functionPath
   }
 
-  const handlerPath = path.join(functionPath, `${path.basename(functionPath)}.js`)
-  if (!fs.existsSync(handlerPath)) {
-    return
+  let handlerPath
+  const indexHandlerPath = path.join(functionPath, `index.js`)
+  const namedHandlerPath = path.join(functionPath, `${path.basename(functionPath)}.js`)
+
+  if (fs.existsSync(namedHandlerPath)) {
+    handlerPath = namedHandlerPath
+  }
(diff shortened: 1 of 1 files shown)


src/utils/finders.js
```diff
@@ -18,9 +18,15 @@ const findHandler = function (functionPath) {
...
+  const indexHandlerPath = path.join(functionPath, `index.js`)
+  const namedHandlerPath = path.join(functionPath, `${path.basename(functionPath)}.js`)
+
+  if (fs.existsSync(namedHandlerPath)) {
+    handlerPath = namedHandlerPath
+  }
+  if (fs.existsSync(indexHandlerPath) && !fs.existsSync(namedHandlerPath)) {
+    handlerPath = indexHandlerPath
   }
   return handlerPath
```
erezrokah, line 31:
```suggestion
  const namedHandlerPath = path.join(functionPath, `${path.basename(functionPath)}.js`)
  if (fs.existsSync(namedHandlerPath)) {
    return namedHandlerPath
  }

  const indexHandlerPath = path.join(functionPath, `index.js`)
  if (fs.existsSync(indexHandlerPath)) {
    return indexHandlerPath
  }
```

Could we use something like the suggested code to avoid the second call to `fs.existsSync(namedHandlerPath)`?

erezrokah (changes requested):
Thanks @ardiewen, I was just discussing this internally with a colleague.

Added a small optimization. WDYT?

src/utils/finders.js
ardiewen: looks good to me!

erezrokah (approved):

PR #1478, opened 2020-10-27
Use `zip-it-and-ship-it` to list Functions

### LINKED ISSUE #1477: Use `zip-it-and-ship-it` logic to detect functions
The CLI uses its own logic to detect functions:
https://github.com/netlify/cli/blob/3baac9cd33330064d2c19b1803697fe72800c409/src/utils/get-functions.js#L7

This logic is not aligned with production https://github.com/netlify/cli/pull/1476.

We should re-use the logic from `zip-it-and-ship-it` to do it:
https://github.com/netlify/zip-it-and-ship-it/blob/b211bc7e6cef4c618f04e64782f8bb63856de2a3/src/main.js#L79

ehmicky:
Fixes #1477.

This is currently in draft to get some early feedback on the approach.

I have not tried it and this is a WIP, so this code is probably broken right now.


src/commands/functions/invoke.js
```diff
@@ -178,38 +178,47 @@ const processPayloadFromFlag = function (payloadString) {
 // prompt for a name if name not supplied
 // also used in functions:create
 const getNameFromArgs = async function (functions, args, flags) {
-  // let functionToTrigger = flags.name;
```
ehmicky, line 181: We probably should split this refactoring to its own PR.

src/commands/functions/list.js
```diff
@@ -56,40 +56,43 @@ class FunctionsListCommand extends Command {
...
       this.exit()
     }
 
     if (flags.json) {
-      const jsonData = functionData.map(([functionName, { moduleDir }]) => {
-        const isDeployed = deployedFunctions.map((deployedFunction) => deployedFunction.n).includes(functionName)
-        return {
-          name: functionName,
-          url: `/.netlify/functions/${functionName}`,
-          moduleDir,
```
ehmicky, line 73: This removes `moduleDir`. `moduleDir` is meant to be the site's `package.json` or the function's `package.json`.
Since `package.json` inside Netlify functions is not supported out-of-the-box, I don't think we should expose this to users. This could be a breaking change is anyone is using `--json` and `moduleDir`, i.e. would require a major release. What do you think?
We could put this in a separate PR.

src/commands/functions/list.js
```diff
@@ -56,40 +56,43 @@ class FunctionsListCommand extends Command {
...
-      table.addRow(functionName, `/.netlify/functions/${functionName}`, moduleDir, isDeployed ? 'yes' : 'no')
+    table.setHeading('Name', 'Url', 'deployed')
+    functions.forEach(({ name, localPath }) => {
+      addFunctionRow({ table, deployedFunctions, functionName: name, localPath })
     })
     this.log(table.toString())
   }
 }
 
+const addFunctionRow = function ({ table, deployedFunctions, functionName, localPath }) {
```
ehmicky, line 87: We could put this refactoring in a separate PR.

src/utils/get-functions.js
```diff
@@ -1,34 +1,25 @@
...
-    }
-    return functions
-  },
+// List all Netlify Functions
+const getFunctions = async function (functionsSrcDir) {
+  if (!(await fileExistsAsync(functionsSrcDir))) {
+    return []
+  }
+
+  const functions = await listFunctions(functionsSrcDir)
```
ehmicky, line 11: We need to add the function `name` to `listFunctions()` in `zip-it-and-ship-it` to make this work.

ehmicky: Done in https://github.com/netlify/zip-it-and-ship-it/pull/231

src/commands/functions/list.js
erezrokah: I think we're good here due to https://github.com/netlify/cli/blob/9292288cbb8b5b58aa4420a05c39795ba1fed474/src/commands/functions/list.js#L118

erezrokah:
@ehmicky I fixed the tests and limited this to listing `js` functions for now.
I can't add you as a reviewer since you opened the PR.

erezrokah (approved):

src/commands/functions/invoke.js
ehmicky: Done in https://github.com/netlify/cli/pull/1523

PR #1483, opened 2020-10-27
Add `node/global-require` linting rule

ehmicky:
This adds the [`node/global-require`](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/global-require.md) ESLint rule.

We probably could remove more dynamic requires (currently marked as linting exceptions in this PR), but this would lead to too much refactoring for the moment.

A few dynamic requires had comments "lazily evaluated on purpose", but I don't see any actual good purpose for doing so since those libraries do not perform any logic on `require()` and are fairly small/fast to load.

### DIFF
--- a/src/functions-templates/js/url-shortener/url-shortener.js
+++ b/src/functions-templates/js/url-shortener/url-shortener.js
@@ -1,3 +1,6 @@
+const generateRoute = require('./generate-route')
+const getRoute = require('./get-route')
+
 const handler = (event, context, callback) => {
   const path = event.path.replace(/\.netlify\/functions\/[^/]+/, '')
   const segments = path.split('/').filter(Boolean)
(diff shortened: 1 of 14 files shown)


src/functions-templates/js/url-shortener/url-shortener.js
```diff
@@ -6,10 +9,10 @@ const handler = (event, context, callback) => {
   switch (event.httpMethod) {
     case 'GET':
       // e.g. GET /.netlify/functions/url-shortener
-      return require('./get-route').handler(event, context, callback)
+      return getRoute(event, context, callback)
```
ehmicky, line 12: There was an error here, since the routes are default exports.

erezrokah (approved):

PR #1484, opened 2020-10-27
Add `node/callback-return` ESLint rule

ehmicky:
This adds the [`node/callback-return`](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/callback-return.md) ESLint rule.


src/functions-templates/js/stripe-charge/stripe-charge.js
```diff
@@ -11,7 +11,7 @@ const headers = {
 const handler = function (event, context, callback) {
   // -- We only care to do anything if this is our POST request.
   if (event.httpMethod !== 'POST' || !event.body) {
-    callback(null, {
+    return callback(null, {
```
ehmicky, line 14: It seems the lack of `return` was a bug there.

erezrokah (approved):

Issue #514, opened 2019-09-06
`netlify lm:setup` crashes


ehmicky:
**- Do you want to request a *feature* or report a *bug*?** 
Bug

**- What is the current behavior?**
`netlify lm:setup` crashes.

**- If the current behavior is a bug, please provide the steps to reproduce.**
I ran the following commands in order:
  - `apt-get install git-lfs`
  - `git lfs version`
  - `netlify plugins:install netlify-lm-plugin`
  - `netlify lm:install`
  - `netlify lm:setup`

All commands worked except the last one which shows:

```
Logging into your Netlify account...
Opening https://app.netlify.com/authorize?response_type=ticket&ticket=xxxxxx
---------------------------
Error: Unable to open browser automatically
spawn /home/ether/.local/share/netlify-cli/node_modules/netlify-lm-plugin/node_modules/@netlify/cli-utils/node_modules/cli-ux/lib/xdg-open ENOENT

Please open your browser & open the URL below to login:
https://app.netlify.com/authorize?response_type=ticket&ticket=xxxxxx
---------------------------
TypeError: Cannot read property 'slug' of undefined
    at LmSetup.authenticate (~/.local/share/netlify-cli/node_modules/netlify-lm-plugin/node_modules/@netlify/cli-utils/src/index.js:116:21)
    at async LmSetup.run (~/.local/share/netlify-cli/node_modules/netlify-lm-plugin/lib/commands/lm/setup.js:24:29)
    at async LmSetup._run (~/.local/share/netlify-cli/node_modules/netlify-lm-plugin/node_modules/@netlify/cli-utils/node_modules/@oclif/command/lib/command.js:34:20)
    at async Config.runCommand (~/.nvm/versions/node/v12.10.0/lib/node_modules/netlify-cli/node_modules/@oclif/config/lib/config.js:151:9)
    at async Main.run (~/.nvm/versions/node/v12.10.0/lib/node_modules/netlify-cli/node_modules/@oclif/command/lib/main.js:21:9)
    at async Main._run (~/.nvm/versions/node/v12.10.0/lib/node_modules/netlify-cli/node_modules/@oclif/command/lib/command.js:44:20)
```

**- What is the expected behavior?**
`netlify lm:setup` should not crash

**- Local Environment Information**
Paste the results of `netlify status --verbose` here

```
──────────────────────┐
 Current Netlify User │
──────────────────────┘
Name:  ehmicky
Email: [edited]
Teams: 
[...]

Issue #483, opened 2019-08-20
Merge netlify-lm-plugin into this repo


calavera:
We should merge https://github.com/netlify/netlify-lm-plugin into this repository like we already did with netlify-dev-plugin in #396. It'd be easier to maintain that way.

DavidWells:
Sounds good to me.

Idk it the `ts` will play nicely with the rest of the repo tho

verythorough:
Adding the `needs docs` label in case the new plugin location affects the [Large Media installation docs](https://www.netlify.com/docs/large-media/#enabling-netlify-large-media).

PR #1489, opened 2020-10-29
feat(command-lm): merge netlify-lm-plugin to CLI

keiko713:
**- Summary**

Closes https://github.com/netlify/cli/issues/483
Closes https://github.com/netlify/cli/issues/514
Closes https://github.com/netlify/netlify-lm-plugin/issues/43
Closes https://github.com/netlify/netlify-credential-helper/issues/46

[netlify-lm-plugin](https://github.com/netlify/netlify-lm-plugin) is not maintained well currently, and using some deprecated dependency. This PR is to see if we can merge the plugin into the CLI to make it easy to maintain.
Also, I would love to retire the usage of old getAddons/createAddon, so that it'll better integrated with the development env var like `NETLIFY_API_URL`.

**- Test plan**

Very much WIP.
This is also a great chance to add tests!!

Merge netlify-lm-plugin (netlify lm command) to CLI

image


keiko713:
Currently it is half-working. `lm:info` is running but somehow `lm` is not working:

```
$ bin/run lm              
 ›   Error: command lm not found
    Error: Command failed: netlify lm --help
$ bin/run lm --help
Handle Large Media operations

USAGE
  $ netlify lm

DESCRIPTION
  The lm command will help you manage a large media for the site

EXAMPLE
  netlify lm:info

$ bin/run lm:info  
  ✔ Checking Git version [2.26.0]
  ✔ Checking Git LFS version [2.5.1]
  ✔ Checking Git LFS filters
  ✔ Checking Netlify's Git Credentials version [0.1.9]
```

src/commands/lm/info.js
```diff
@@ -0,0 +1,42 @@
...
+const {
+  GitValidators,
+  checkLFSFilters,
+  checkHelperVersion
+} = require('../../utils/lm/requirements')
+
+class LmInfoCommand extends Command {
+  async run() {
+    const steps = GitValidators
+    steps.push(
```
ehmicky, line 13: This would modify `module.exports.GitValidators`, which means another module importing it would see this mutation.
Instead, using array spread should create a new array, so the original imported value is not modified.

```js
const steps = [
  ...GitValidators,
  {
     title: '...'
  // And so on
]  
```

src/utils/lm/requirements.js
```diff
@@ -0,0 +1,64 @@
...
+      const version = await checkLFSVersion()
+      task.title += chalk.dim(` [${version}]`)
+    }
+  }
+]
+
+module.exports.checkLFSFilters = async function checkLFSFilters() {
+  try {
+    const result = await execa('git', ['config', '--get-regexp', 'filter.lfs'])
+    return Promise.resolve(result.stdout.length > 0)
```
ehmicky, line 25: If a function is `async`, `return Promise.resolve(value)` can be simplified to `return value`.

erezrokah:
Hi @keiko713, how do we want to move forward with this?

keiko713:
thanks for the ping and sorry that I haven't touched this for more than one month...!!! I really would like to get back to this, but at the same time, I'm having hard time finding time for this.
can you please give me, let's say, one more week to see if I can find some time?

erezrokah:
> can you please give me, let's say, one more week to see if I can find some time?

Should we can do that. I can also raise https://github.com/netlify/cli/issues/483 internally to see if we can prioritize it.

src/commands/lm/info.js
keiko713: thanks, updated!

src/utils/lm/requirements.js
keiko713: good to know, updated 👍

keiko713:
current status:
* migrated all the commands
* confirmed that super basic paths (lm:info, lm:install, and lm:setup) with platform: macOS 10.15.7 works

TODOs
* write tests
* more tests with different platform, as well as more edge cases (e.g. when X is not installed)

keiko713:
@erezrokah syncing up with some progress (also no rush, totally fine after 🎄 ):
* I managed to run some test, but still it's been failing with all other platforms than macos
* there is some difficulty with the test in general (will comment more inline)
* I'd love to get some feedback, the tests are failing but the code should be already in a decent shape

Thanks!

tests/command.lm.test.js
```diff
@@ -0,0 +1,53 @@
...
+    const cliResponse = await callCli(['lm:install'], t.context.execOptions)
+    t.true(cliResponse.includes('Checking Git version'))
+    t.true(cliResponse.includes('Checking Git LFS version'))
+    t.true(cliResponse.includes('Checking Git LFS filters'))
+    t.true(cliResponse.includes("Installing Netlify's Git Credential Helper"))
+    t.true(cliResponse.includes("Configuring Git to use Netlify's Git Credential Helper"))
+  })
+
+  test.serial('netlify lm:setup', async (t) => {
+    // Skip the actual setup test as the site needs to be associated with the repo to provision a large media addon.
```
keiko713, line 39: I realized that it's hard to do the integration test with this. The workaround I can think of is to create a repo dedicated to this test. However, there is no good way to link the repo to the site with CLI currently (it needs multiple steps to do that action), so I wasn't able to put that into this.
I'd say that we can keep it as is for now, it's sadly not doing the provisioning but still testing something (actually this test did make me realize that the error message wasn't showing up well, which I was able to tweak).

erezrokah: How about we mock the API call by setting up a mock server and use `NETLIFY_API_URL` to configure it?
I think `lm:setup` has only one API call.

erezrokah:
@keiko713 Thank you so much for following up on this.
I rebased the branch and added a few more commits to do some code clean - please let know what you think. I might follow up with some more next year :).

For the failing tests, I had to run it on my Windows command to get the error, see https://github.com/netlify/netlify-credential-helper/pull/49

That made `lm:install` work on Windows (I updated the code to install from the PR's branch), but `lm:setup` is still failing so I need to look into it.

erezrokah:
cc @keiko713 tests are passing on Windows (Node.js 10 failure is unrelated).

tests/command.lm.test.js
keiko713: oh sorry 🤦  just realized that I dropped the ball here. I like the idea!! I think it's still actually valuable to have this "failure case" test, but I love the idea of mocking the API request and doing the success case test.

the mock success response will look like https://open-api.netlify.com/#operation/createServiceInstance but I don't think we're using any returned data so maybe mocking with empty json response with 201 is fine.

shall I do it, or will you do it?

erezrokah: > shall I do it, or will you do it?

I can do it. I think it will be valuable for other tests we'd like to have

keiko713: thank you!!

src/utils/lm/install.js
```diff
@@ -0,0 +1,262 @@
...
+    task: async () => {
+      await installUnixHelper()
+      await setupUnixPath()
+    },
+  }
+}
+
+const installWithPowershell = async function () {
+  const script = `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+iex (iwr -UseBasicParsing -Uri https://github.com/netlify/netlify-credential-helper/blob/53fbf6e2b82d1ae366d096a24022fa259d19329a/resources/install.ps1)`
```
erezrokah, line 118: I pinned `install.ps1` to a specific commit so we can make breaking changes to it.

The current script installs to `${Home}\.netlify` where newer versions of the CLI use:
https://github.com/netlify/cli/blob/e5c1798cf3f8383990cf3d336b44d8dbce62c2e4/src/lib/settings.js#L16

cc @keiko713 @JGAntunes.

Once we get this merge I'll submit a new PR to update https://github.com/netlify/netlify-credential-helper/blob/53fbf6e2b82d1ae366d096a24022fa259d19329a/resources/install.ps1 then another to the CLI

src/commands/lm/setup.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return false
+}
+
+const provisionService = async function (siteId, api) {
+  const addonName = 'large-media'
+
+  if (!siteId) {
+    return Promise.reject(new Error('No site id found, please run inside a site folder or `netlify link`'))
```
ehmicky, line 32:
```suggestion
    throw new Error('No site id found, please run inside a site folder or `netlify link`')
```

src/commands/lm/setup.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+  try {
+    await api.createServiceInstance({
+      siteId,
+      addon: addonName,
+      body: {},
+    })
+  } catch (error) {
+    // error is JSONHTTPError
+    return Promise.reject(new Error(error.json.error))
```
ehmicky, line 42:
```suggestion
    throw new Error(error.json.error)
```

src/commands/lm/install.js
```diff
@@ -0,0 +1,43 @@
...
+      },
+    })
+
+    try {
+      const installed = await installPlatform(flags.force)
+      if (installed) {
+        printBanner(this, flags.force)
+      }
+    } catch (error) {
+      this.log(error)
```
ehmicky, line 24:
```suggestion
      this.error(error)
```

So the error is printed on `stderr` and the exit code is non `0`.

src/commands/lm/info.js
```diff
@@ -0,0 +1,40 @@
...
+
+    await this.config.runHook('analytics', {
+      eventName: 'command',
+      payload: {
+        command: 'lm:info',
+      },
+    })
+
+    const tasks = new Listr(steps, { concurrent: true, exitOnError: false })
+    tasks.run().catch(() => {})
```
ehmicky, line 32: Should the error be printed instead of ignored?

src/commands/lm/setup.js
```diff
@@ -0,0 +1,115 @@
...
+        command: 'lm:setup',
+      },
+    })
+
+    let helperInstalled = false
+    if (!flags['skip-install']) {
+      try {
+        helperInstalled = await installHelperIfMissing(flags['force-install'])
+      } catch (error_) {
+        this.error(error_.message)
```
ehmicky, line 72:
```suggestion
        this.error(error_.message)
```

To keep the stack trace.

(Also same content as above about default error handling)

src/utils/lm/install.js
```diff
@@ -0,0 +1,262 @@
...
+    task: async () => {
+      await installUnixHelper()
+      await setupUnixPath()
+    },
+  }
+}
+
+const installWithPowershell = async function () {
+  const script = `[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+iex (iwr -UseBasicParsing -Uri https://raw.githubusercontent.com/netlify/netlify-credential-helper/53fbf6e2b82d1ae366d096a24022fa259d19329a/resources/install.ps1)`
```
ehmicky, line 118: Should we inline that file in this repository instead of fetching it over the network?

ehmicky, line 118: Should we inline that file in this repository instead of fetching it over the network?

src/utils/lm/install.js
```diff
@@ -0,0 +1,262 @@
...
+  const temp = await mkdtempAsync(path.join(tempDir, 'netlify-'))
+  const scriptPath = path.join(temp, 'install.ps1')
+
+  await writeFileAsync(scriptPath, script)
+
+  return await execa('powershell', ['-ExecutionPolicy', 'unrestricted', '-File', scriptPath, '-windowstyle', 'hidden'])
+}
+
+const isBinInPath = function () {
+  const envPath = process.env.PATH || ''
```
ehmicky, line 129: Should we use https://github.com/sindresorhus/path-key to make it work on Windows?

src/utils/lm/install.js
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  const initContent = `
+# The next line updates PATH for Netlify's Git Credential Helper.
+if [ -f '${shellInfo.path}' ]; then source '${shellInfo.path}'; fi
+`
+
+  switch (shellInfo.shell) {
+    case 'bash': {
+      const bashPath = `script_link="$( command readlink "$BASH_SOURCE" )" || script_link="$BASH_SOURCE"
```
ehmicky, line 151: Should this be inlined in a sibling Bash file (then `fs.readFile()` used on it)? 
This would make it easier to read, and provide with syntax highlighting.

src/utils/lm/install.js
```diff
@@ -0,0 +1,262 @@
...
+
+const setupWindows = function (skipInstall) {
+  return {
+    title: `Installing Netlify's Git Credential Helper for Windows`,
+    skip: () => skipHelperInstall(skipInstall),
+    task: installWithPowershell,
+  }
+}
+
+const installUnixHelper = async function () {
```
ehmicky, line 84: Most of the code in that file is due to adding some binary to the `PATH`. This requires detecting shells, having shell-specific code, etc.

Is there a potential alternative where we would not need to modify the `PATH`? For example:
  - Calling the binary ourselves (so we can use an absolute/relative file path instead of the `PATH`)
  - Using a Node module + `npx`
  - Using a Node module and globally installing it

Note that those solutions might not make sense here.

src/utils/lm/install.js
erezrokah: Looking at the file:
https://github.com/netlify/netlify-credential-helper/blob/53fbf6e2b82d1ae366d096a24022fa259d19329a/resources/install.ps1
Looks like it would be better to port most of it to JavaScript (at least fetching the release part).

src/utils/lm/install.js
erezrokah: This function doesn't run when the OS is Window. How about we rename it to `isUnixBinPath` to make it more clear?

ehmicky: Sounds good. Sorry to miss this :+1:

src/utils/lm/install.js
ehmicky: Yes, that'd be even better.

Not relying on network requests would also be a win in terms of performance and reliability.

src/utils/lm/install.js
erezrokah: Sure, that makes sense. I'll change it

src/utils/lm/install.js
erezrokah: I believe we install that binary to be used by `git-lfs`. I'll verify that and also check if there's a better alternative

src/utils/lm/install.js
erezrokah: This is done in https://github.com/netlify/cli/pull/1489/commits/f693f79abbb3d0d38313f435ffdebcd542800c17

src/utils/lm/install.js
erezrokah: Done in https://github.com/netlify/cli/pull/1489/commits/f693f79abbb3d0d38313f435ffdebcd542800c17

src/utils/lm/scripts/bash.sh
```diff
@@ -0,0 +1,10 @@
+script_link="$( command readlink "$BASH_SOURCE" )" || script_link="$BASH_SOURCE"
+apparent_sdk_dir="${script_link%/*}"
+if [ "$apparent_sdk_dir" == "$script_link" ]; then
+apparent_sdk_dir=.
+fi
+sdk_dir="$( command cd -P "$apparent_sdk_dir" > /dev/null && command pwd -P )"
+bin_path="$sdk_dir/bin"
+if [[ ":${PATH}:" != *":${bin_path}:"* ]]; then
+export PATH=$bin_path:$PATH
+fi
```
ehmicky, line 10: This file should probably be indented and use some newlines.

src/utils/lm/scripts/bash.sh
```diff
@@ -0,0 +1,10 @@
+script_link="$( command readlink "$BASH_SOURCE" )" || script_link="$BASH_SOURCE"
```
ehmicky, line 1: We should add a shabang to this file.

src/utils/lm/scripts/path.ps1
```diff
@@ -0,0 +1,33 @@
+$ErrorActionPreference = "Stop"
```
ehmicky, line 1: We should add a shabang to this file.

```
#!/usr/bin/env pwsh
```

src/utils/lm/scripts/zsh.sh
```diff
@@ -0,0 +1 @@
+export PATH=${0:A:h}/bin:$PATH
```
ehmicky, line 1: We should also add a shabang there.

src/commands/lm/info.js
erezrokah: Great point!
That suppress a redundant error (which is already printed when a task fails):
image

package.json
```diff
@@ -19,6 +19,7 @@
     "/oclif.manifest.json",
     "/scripts/postinstall.js",
     "/src/**/*.js",
+    "/src/utils/lm/scripts/*.*",
```
ehmicky, line 22: Would this also work?

```
    "/src/utils/lm/scripts/",
```

src/utils/lm/install.js
erezrokah: Per `git` [documentation](https://git-scm.com/docs/gitcredentials#_custom_helpers) looks like this is the way to go:
image

erezrokah:
Corresponding PR in https://github.com/netlify/netlify-credential-helper/pull/51

src/commands/lm/info.js
```diff
@@ -0,0 +1,40 @@
...
+
+    await this.config.runHook('analytics', {
+      eventName: 'command',
+      payload: {
+        command: 'lm:info',
+      },
+    })
+
+    const tasks = new Listr(steps, { concurrent: true, exitOnError: false })
+    await tasks.run().catch(() => {})
```
ehmicky, line 32: [sand] `try`/`catch` could be used here.

src/commands/lm/install.js
```diff
@@ -0,0 +1,43 @@
...
+      },
+    })
+
+    try {
+      const installed = await installPlatform(flags.force)
+      if (installed) {
+        printBanner(this, flags.force)
+      }
+    } catch (error) {
+      this.error(error)
```
ehmicky, line 24: Is `this.error()` the default error handling mechanism when an exception is thrown in a command? If so, the `try`/`catch` block could be removed.

ehmicky (approved):

src/commands/lm/install.js
erezrokah: Nice catch! That is indeed the default behavior

Issue #1386, opened 2020-10-13
Next.js dev environment variables overwritten with production values by netlify dev


ercgrat:
When running `netlify dev`, local root .env variables are overwritten by the Netlify variables pulled from my production settings (Site settings > Build & deploy > Environment > Environment variables). This is at least true for the variables that I need to be accessible in the browser with the NEXT_PUBLIC_ prefix ([blog post](https://nextjs.org/blog/next-9-4#new-environment-variables-support)).

My .env file contains (my file is actually 'env.local' but I reproduced this with .env):
```
NEXT_PUBLIC_PAYPAL_ENV=sandbox
```

My production settings contain: `NEXT_PUBLIC_PAYPAL_ENV=production`. The browser value of `process.env.NEXT_PUBLIC_PAYPAL_ENV` ends up being `production` when running `netlify dev`. I have confirmed that other NEXT_PUBLIC_ variables _not_ set in production are still correct in the browser when running locally, so the problem lies with Netlify.

I have to work around this by creating a new, dev-specific environment variable that takes precedence over the one shared between development and production. This is the code I use in my React component to avoid the issue:
```
let env = process.env.NEXT_PUBLIC_PAYPAL_ENV_DEV;
if (!env) {
    env = process.env.NEXT_PUBLIC_PAYPAL_ENV;
}
```

I don't think it's relevant here but this is my netlify.toml:
```
[build]
publish = "out/"
functions = "functions/"

[dev]
command = "npm run dev"

```

Here's my netlify dev output. What's interesting to me is that the next.js environment variable load does appear later in the sequence, but maybe it chooses not to override existing values.
```
>netlify dev
◈ Netlify Dev ◈
◈ Injected addon env var: FAUNADB_ADMIN_SECRET
◈ Injected addon env var: FAUNADB_SERVER_SECRET
◈ Injected addon env var: FAUNADB_CLIENT_SECRET
◈ Injected build setting env var: NEXT_PUBLIC_PAYPAL_ENV
◈ Injected build setting env var: NODE_ENV
◈ Overriding command with setting derived from netlify.toml [dev] block: npm run dev
◈ Functions server is listening on 60715
◈ Starting Netlify Dev with next

> learn-starter@0.1.0 dev C:\Users\ercgr\Documents\UrbanBasic\creator
> next dev

Loaded env from C:\Users\ercgr\Documents\UrbanBasic\creator\.env.local
ready - started server on http://localhost:3000

[...]

PR #1490, opened 2020-10-29
fix: don't override dot envs with site envs

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1386
Related to https://github.com/netlify/cli/issues/1331

Also fixed some stuff related to addons since during my manual testing I noticed it was broken.

**- Test plan**

Fixed some tests.
I still need to think about adding new tests, as testing https://github.com/netlify/cli/issues/1386 will require creating a live site and then running `netlify dev` or `netlify dev:exec`.

Refactor environment variables handling.

image

### DIFF
--- a/src/commands/dev/exec.js
+++ b/src/commands/dev/exec.js
@@ -1,43 +1,22 @@
-const fromEntries = require('@ungap/from-entries')
-const chalk = require('chalk')
 const execa = require('execa')
 
 const Command = require('../../utils/command')
-const { addEnvVariables } = require('../../utils/dev')
-const { getEnvSettings } = require('../../utils/env')
-const {
-  // NETLIFYDEV,
-  NETLIFYDEVLOG,
-  // NETLIFYDEVWARN,
-  NETLIFYDEVERR,
-} = require('../../utils/logo')
+const { getSiteInformation, addEnvVariables } = require('../../utils/dev')
 
 class ExecCommand extends Command {
   async run() {
-    const { site, api } = this.netlify
-    if (site.id) {
-      // just to show some visual response first
-      this.log(`${NETLIFYDEVLOG} Checking your site's environment variables...`)
-      await addEnvVariables(api, site)
-    } else {
-      this.log(
-        `${NETLIFYDEVERR} No Site ID detected. You probably forgot to run \`netlify link\` or \`netlify init\`. `,
-      )
-    }
-
-    const envSettings = await getEnvSettings({ projectDir: site.root, warn: this.warn })
-    if (envSettings.vars.length !== 0) {
-      this.log(
-        `${NETLIFYDEVLOG} Adding the following env variables from ${envSettings.files.map((file) =>
-          chalk.blue(file),
-        )}:`,
-        chalk.yellow(envSettings.vars.map(([key]) => key)),
-      )
-    }
+    const { log, warn, error, netlify } = this
+    const { site, api } = netlify
+    const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
+      api,
+      site,
+      warn,
+      error,
+    })
+    await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
 
     execa(this.argv[0], this.argv.slice(1), {
       stdio: 'inherit',
-      env: fromEntries(envSettings.vars),
     })
     await this.config.runHook('analytics', {
       eventName: 'command',
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -13,8 +13,7 @@ const wrapAnsi = require('wrap-ansi')
 
 const Command = require('../../utils/command')
 const { serverSettings } = require('../../utils/detect-server')
-const { addEnvVariables } = require('../../utils/dev')
-const { getEnvSettings } = require('../../utils/env')
+const { getSiteInformation, addEnvVariables } = require('../../utils/dev')
 const { startLiveTunnel } = require('../../utils/live-tunnel')
 const { NETLIFYDEV, NETLIFYDEVLOG, NETLIFYDEVWARN, NETLIFYDEVERR } = require('../../utils/logo')
 const openBrowser = require('../../utils/open-browser')
@@ -99,29 +98,7 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
 // 1 minute
 const FRAMEWORK_PORT_TIMEOUT = 6e5
 
-const getAddonsUrlsAndAddEnvVariablesToProcessEnv = async ({ api, site, flags }) => {
-  if (site.id && !flags.offline) {
-    const addonUrls = await addEnvVariables(api, site)
-    return addonUrls
-  }
-  return {}
-}
-
-const addDotFileEnvs = async ({ site, log, warn }) => {
-  // add .env file environment variables
-  const envSettings = await getEnvSettings({ projectDir: site.root, warn })
-  if (envSettings.vars.length !== 0) {
-    log(
-      `${NETLIFYDEVLOG} Adding the following env variables from ${envSettings.files.map((file) => chalk.blue(file))}:`,
-      chalk.yellow(envSettings.vars.map(([key]) => key)),
-    )
-    envSettings.vars.forEach(([key, val]) => {
-      process.env[key] = val
-    })
-  }
-}
-
-const startProxyServer = async ({ flags, settings, site, log, exit, addonUrls }) => {
+const startProxyServer = async ({ flags, settings, site, log, exit, addonsUrls }) => {
   let url
   if (flags.trafficMesh) {
     url = await startForwardProxy({
@@ -137,7 +114,7 @@ const startProxyServer = async ({ flags, settings, site, log, exit, addonUrls })
       exit(1)
     }
   } else {
-    url = await startProxy(settings, addonUrls, site.configPath, site.root)
+    url = await startProxy(settings, addonsUrls, site.configPath, site.root)
     if (!url) {
       log(NETLIFYDEVERR, `Unable to start proxy server on port '${settings.port}'`)
       exit(1)
@@ -202,9 +179,15 @@ class DevCommand extends Command {
       ...flags,
     }
 
-    const addonUrls = await getAddonsUrlsAndAddEnvVariablesToProcessEnv({ api, site, flags })
-    process.env.NETLIFY_DEV = 'true'
-    await addDotFileEnvs({ site, log, warn })
+    const { addonsUrls, teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
+      flags,
+      api,
+      site,
+      warn,
+      error: errorExit,
+    })
+
+    await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
 
     let settings = {}
     try {
@@ -217,7 +200,7 @@ class DevCommand extends Command {
     await startFunctionsServer({ settings, site, log, warn, errorExit, siteInfo: this.netlify.cachedConfig.siteInfo })
     await startFrameworkServer({ settings, log, exit })
 
-    let url = await startProxyServer({ flags, settings, site, log, exit, addonUrls })
+    let url = await startProxyServer({ flags, settings, site, log, exit, addonsUrls })
 
     const liveTunnelUrl = await handleLiveTunnel({ flags, site, api, settings, log })
     url = liveTunnelUrl || url
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -388,20 +388,25 @@ const handleOnComplete = async ({ context, onComplete }) => {
 }
 
 const handleAddonDidInstall = async ({ addonCreated, addonDidInstall, context, fnPath }) => {
-  if (addonCreated && addonDidInstall) {
-    const { confirmPostInstall } = await inquirer.prompt([
-      {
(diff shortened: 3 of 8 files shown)


src/commands/dev/exec.js
```diff
@@ -1,42 +1,22 @@
...
-      // just to show some visual response first
-      this.log(`${NETLIFYDEVLOG} Checking your site's environment variables...`)
-      await addEnvVariables(api, site)
-    } else {
-      this.log(
-        `${NETLIFYDEVERR} No Site ID detected. You probably forgot to run \`netlify link\` or \`netlify init\`. `,
-      )
-    }
-
-    const envSettings = await getEnvSettings({ projectDir: site.root, warn: this.warn })
```
erezrokah, line 27: Reading dotenv files was moved into `getSiteInformation`.
Applying those envs was moved into `addEnvVariables`

src/commands/dev/exec.js
```diff
@@ -1,42 +1,22 @@
...
+      api,
+      site,
+      warn,
+      error,
+    })
+    await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
 
     execa(this.argv[0], this.argv.slice(1), {
       stdio: 'inherit',
-      env: fromEntries(envSettings.vars),
```
erezrokah, line 39: These are applied globally now (which is not nice, but would require more refactoring as we log each env that is applied in a different way)

src/commands/dev/index.js
```diff
@@ -98,29 +98,7 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
 // 1 minute
 const FRAMEWORK_PORT_TIMEOUT = 6e5
 
-const getAddonsUrlsAndAddEnvVariablesToProcessEnv = async ({ api, site, flags }) => {
```
erezrokah, line 101: Moved to `getSiteInformation` and `addEnvVariables`

src/commands/dev/index.js
```diff
@@ -136,7 +114,7 @@ const startProxyServer = async ({ flags, settings, site, log, exit, addonUrls })
       exit(1)
     }
   } else {
-    url = await startProxy(settings, addonUrls, site.configPath, site.root)
```
erezrokah, line 139: `addonUrls` had all addons Urls (not a single addon Urls) so renamed it to `addonsUrls`.

Those are use to redirect local requests to `http://localhost:8888/.netlify/<addon>/*` to the actual addon Url.

src/commands/dev/index.js
```diff
@@ -201,9 +179,15 @@ class DevCommand extends Command {
       ...flags,
     }
 
-    const addonUrls = await getAddonsUrlsAndAddEnvVariablesToProcessEnv({ api, site, flags })
-    process.env.NETLIFY_DEV = 'true'
```
erezrokah, line 205: This is also done in `addEnvVariables`

src/commands/functions/create.js
```diff
@@ -258,10 +258,7 @@ const downloadFromURL = async function (context, flags, args, functionsDir) {
     const { onComplete, addons = [] } = require(fnTemplateFile)
 
     await installAddons(context, addons, path.resolve(fnFolder))
-    if (onComplete) {
-      await addEnvVariables(context.netlify.api, context.netlify.site)
```
erezrokah, line 262: This was done so the `onComplete` handler runs in the same context as `dev`, but missing dotenv envs.
This PR fixes that

src/commands/functions/create.js
```diff
@@ -373,6 +367,42 @@ const createFunctionAddon = async function ({ api, addons, siteId, addonName, si
   }
 }
 
+const injectEnvVariables = async ({ context }) => {
+  const { log, warn, error, netlify } = context
+  const { api, site } = netlify
+  const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
```
erezrokah, line 373: `dotFilesEnv` were not injected before

src/commands/functions/create.js
```diff
@@ -373,6 +367,42 @@ const createFunctionAddon = async function ({ api, addons, siteId, addonName, si
...
+    const { confirmPostInstall } = await inquirer.prompt([
+      {
+        type: 'confirm',
+        name: 'confirmPostInstall',
+        message: `This template has an optional setup script that runs after addon install. This can be helpful for first time users to try out templates. Run the script?`,
+        default: false,
+      },
+    ])
+    if (confirmPostInstall) {
+      await injectEnvVariables({ context })
```
erezrokah, line 400: This was done so the `addonDidInstall` handler runs in the same context as `netlify dev`, but was missing dotenv envs.
This PR fixes that and also optimises to run `injectEnvVariables` only after user confirmation

src/commands/functions/create.js
```diff
@@ -404,20 +434,8 @@ const installAddons = async function (context, functionAddons, fnPath) {
...
-            message: `This template has an optional setup script that runs after addon install. This can be helpful for first time users to try out templates. Run the script?`,
-            default: false,
-          },
-        ])
-        if (confirmPostInstall) {
-          addonDidInstall(fnPath)
-        }
-      }
+
+      await handleAddonDidInstall({ addonCreated, addonDidInstall, context, fnPath })
```
erezrokah, line 438: There is some optimisation to be done here as `handleAddonDidInstall` will read `siteData` and `siteAddons` again internally, but that would require more refactoring.

src/utils/dev.js
```diff
@@ -1,118 +1,128 @@
...
-          `${NETLIFYDEVERR} Error retrieving addons data for site ${chalk.yellow(
-            site.id,
-          )}. Double-check your login status with 'netlify status' or contact support with details of your error.`,
-        )
-        process.exit()
-    }
-  })
-  if (Array.isArray(addons)) {
-    addons.forEach((addon) => {
-      addonUrls[addon.slug] = `${addon.config.site_url}/.netlify/${addon.slug}`
```
erezrokah, line 42: Both `addon.slug` and `addon.config.site_url` are `undefined`.
Not sure if this was broken in https://github.com/netlify/cli/pull/1344. I don't think so since the old code was using https://github.com/netlify/js-client/blob/c4a39c22308b60bce8ef1d6c97553bf8a9da56ab/src/addons.js#L30 which is the same as https://open-api.netlify.com/#operation/listServiceInstancesForSite which we are using now.

Fixed in https://github.com/netlify/cli/pull/1490/files#diff-fc861c54030ca23b6d43638aa92f3f446bca1b187ce3b3fea5b8e6d7d1c0a19cR40

src/utils/dev.js
```diff
@@ -1,118 +1,128 @@
...
-      for (const key in addon.env) {
```
erezrokah, line 43: This is actually a bug as addons env should not override site env.

src/utils/proxy.js
```diff
@@ -52,6 +52,29 @@ const isExternal = function (match) {
...
+  // Get the URL after http://host:port
+  return url.toString().replace(url.origin, '')
+}
+
+const proxyToExternalUrl = function ({ req, res, dest, destURL }) {
+  console.log(`${NETLIFYDEVLOG} Proxying to ${dest}`)
+  const handler = createProxyMiddleware({
+    target: `${dest.protocol}//${dest.host}`,
+    changeOrigin: true,
+    pathRewrite: () => destURL,
```
erezrokah, line 65: The original code was stripping the origin using a regex:
https://github.com/netlify/cli/blob/a3b76061a32f67b24cba557524a91368df081a6d/src/utils/proxy.js#L200

This wasn't need since the origin is already stripped in:
https://github.com/netlify/cli/blob/a3b76061a32f67b24cba557524a91368df081a6d/src/utils/proxy.js#L184

src/utils/proxy.js
```diff
@@ -106,9 +129,9 @@ const serveRedirect = async function ({ req, res, proxy, match, options }) {
   if (isFunction(options.functionsPort, req.url)) {
     return proxy.web(req, res, { target: options.functionsServer })
   }
-  const urlForAddons = getAddonUrl(options.addonUrls, req)
+  const urlForAddons = getAddonUrl(options.addonsUrls, req)
   if (urlForAddons) {
-    return proxy.web(req, res, { target: urlForAddons })
```
erezrokah, line 111: This doesn't work since you need to change the request host to match the certificate of the target.
I believe the intention of this code was to redirect `http://localhost:8888/.netlify/<addon>/*` to `https://<site-url>/.netlify/<addon>/*`

src/commands/functions/create.js
```diff
@@ -373,6 +367,42 @@ const createFunctionAddon = async function ({ api, addons, siteId, addonName, si
...
+    warn,
+    error,
+  })
+  await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
+}
+
+const handleOnComplete = async ({ context, onComplete }) => {
+  if (onComplete) {
+    await injectEnvVariables({ context })
+    await onComplete.call(context)
```
ehmicky, line 385: Passing the `context` as an argument instead of `this` might be simpler. Unfortunately, this would be a breaking change since the template file is supplied by the user, correct?

src/commands/functions/create.js
```diff
@@ -373,6 +367,42 @@ const createFunctionAddon = async function ({ api, addons, siteId, addonName, si
...
+  }
+}
+
+const handleAddonDidInstall = async ({ addonCreated, addonDidInstall, context, fnPath }) => {
+  if (addonCreated && addonDidInstall) {
```
ehmicky, line 390: [sand] Could we use early returns?
`if (!addonCreated || !addonDidInstall) { return }`
`if (!confirmPostInstall) { return }`

src/utils/dev.js
```diff
@@ -1,118 +1,130 @@
...
-        process.env[key] = assignLoudly(process.env[key], addon.env[key], msg)
-      }
-    })
+const { loadDotEnvFiles } = require('./dot-env')
+const { NETLIFYDEVLOG } = require('./logo')
+
+const ERROR_CALL_TO_ACTION =
+  "Double-check your login status with 'netlify status' or contact support with details of your error."
+
+const getSiteData = async ({ api, site, error }) => {
```
ehmicky, line 12: At some point, we should rename this `error()` function passed to all commands. It it confusingly named and clashes with actual error instances. Using a verb like `fail()` might be better.

src/utils/proxy.js
```diff
@@ -52,6 +52,29 @@ const isExternal = function (match) {
   return match.to && match.to.match(/^https?:\/\//)
 }
 
+const stripOrigin = function (url) {
+  // Get the URL after http://host:port
+  return url.toString().replace(url.origin, '')
```
ehmicky, line 57: `url.origin` does not include `username:password`, so this would fail if those are specified. I am not sure this is a possible case though in this instance.
What about `${url.pathname}${url.search}${url.hash}` instead?

src/utils/proxy.js
```diff
@@ -52,6 +52,29 @@ const isExternal = function (match) {
...
+}
+
+const proxyToExternalUrl = function ({ req, res, dest, destURL }) {
+  console.log(`${NETLIFYDEVLOG} Proxying to ${dest}`)
+  const handler = createProxyMiddleware({
+    target: `${dest.protocol}//${dest.host}`,
```
ehmicky, line 63:
```suggestion
    target: dest.origin,
```

src/commands/dev/exec.js
ehmicky: Would it be possible to keep those environment variables into a variable, instead of modifying `process.env`?

src/utils/dev.js
```diff
@@ -1,118 +1,130 @@
...
+    return siteData.build_settings.env
+  }
+  return {}
 }
 
-module.exports = {
-  addEnvVariables,
+const getSiteInformation = async ({ flags = {}, api, site, warn, error }) => {
+  if (site.id && !flags.offline) {
+    const [siteData, accounts, addons, dotFilesEnv] = await Promise.all([
```
ehmicky, line 62: Nice

src/commands/functions/create.js
erezrokah: Correct, see https://github.com/netlify/cli/blob/65c4bd70bc51492b49fbae677526aea881edbd93/src/functions-templates/js/url-shortener/.netlify-function-template.js#L17
We can update the templates in this repo, but this will still break templates installed from URLs.
At some point we can collect analytics on `functions:create` from URLs and see if this can break anything.

src/commands/dev/exec.js
erezrokah: `addEnvVariables` modifies `process.env` but also prints log messages while doing so:
https://github.com/netlify/cli/pull/1490/files#diff-fc861c54030ca23b6d43638aa92f3f446bca1b187ce3b3fea5b8e6d7d1c0a19cR98

We can clone `process.env` in `addEnvVariables`, use the clone instead of `process.env` and return it to `dev:exec`.
That would work for `dev:exec` but we would still need to modify `process.env` in other places due to the way `netlify dev` and `netlify functions:create` command work at the moment.

I would rather be consistent in how we're applying env variables and handle mutating `process.env` (in all locations we're doing it) in another PR. WDYT?

ehmicky: This makes sense :+1: 
Should we create an issue for it?

ehmicky (approved):

src/commands/dev/exec.js
erezrokah: https://github.com/netlify/cli/issues/1517

PR #1494, opened 2020-10-29
Add `promise/always-return` ESLint rule

ehmicky:
This adds the [`promise/always-return`](https://github.com/xjamundx/eslint-plugin-promise/blob/development/docs/rules/always-return.md) ESLint rule.

This PR also rewrites all Function templates that still use callbacks to use `async`/`await` instead. It also refactors and improves several of those templates.


src/functions-templates/js/stripe-charge/package.json
```diff
@@ -16,6 +16,6 @@
   "author": "Netlify",
   "license": "MIT",
   "dependencies": {
-    "stripe": "^6.28.0"
+    "stripe": "^8.119.0"
```
ehmicky, line 19: Added a `stripe` upgrade to this template as well.

src/functions-templates/js/url-shortener/get-route.js
```diff
@@ -1,39 +1,32 @@
...
-        }
-      }
-    } else {
-      return callback(null, {
-        statusCode: 200,
-        body: err,
-      })
+    const {
+      data: { destination },
+    } = body.find(({ data }) => data.code === code)
```
ehmicky, line 22: The original code does not check for whether the route is never found. I am not quite sure if this is a possible outcome, so I kept it as is.

src/functions-templates/js/google-analytics/google-analytics.js
```diff
@@ -39,29 +39,24 @@ const proxyToGoogleAnalytics = function (event, done) {
...
-    },
-    url: GA_ENDPOINT,
-    body: qs,
+  try {
+    const response = await fetch(GA_ENDPOINT, {
+      method: 'POST',
+      headers: { 'Content-Type': 'image/gif' },
+      body: qs,
+    })
+    const { statusCode, statusMessage } = await response.json()
```
erezrokah, line 48: Is `json` correct here? We weren't parsing the response before.
Also should this be `const { status, statusText } = response`?
Should we also check `response.ok`?

src/functions-templates/js/url-shortener/generate-route.js
```diff
@@ -24,26 +25,26 @@ module.exports = function handler(event, context, callback) {
...
+  form.append('expires', '')
 
   // post the new route to the Routes form
-  request.post({ url: rootURL, formData: payload }, function onResponse(err) {
-    const msg = err
-      ? `Post to Routes stash failed: ${err}`
-      : `Route registered. Site deploying to include it. ${rootURL}${code}`
-    console.log(msg)
+  try {
+    await fetch(rootURL, { method: 'POST' })
```
erezrokah, line 36:
```suggestion
    await fetch(rootURL, { method: 'POST', body: form })
```

src/utils/rules-proxy.js
```diff
@@ -86,37 +86,38 @@ const createRewriter = async function ({ distDir, projectDir, jwtSecret, jwtRole
...
+      headers,
+      cookieValues,
+      getHeader: (name) => headers[name.toLowerCase()] || '',
+      getCookie: (key) => cookieValues[key] || '',
+    }
+    const match = matcherFunc.match(matchReq)
+    if (!match) {
+      return
+    }
+
```
erezrokah, line 119:
```suggestion
```

erezrokah (changes requested):
Thanks @ehmicky, this simplifies our code quite a lot.
Added a few comments.

This is also related to https://github.com/netlify/cli/issues/522.

On another note, not sure we should be updating all the function templates in one PR.
I would rather have an issue for it and discuss the necessity of some of those and whether they should be a part of this repo or https://github.com/netlify/functions

src/functions-templates/js/google-analytics/google-analytics.js
ehmicky: Good point, I messed that up. Fixed!

ehmicky:
Even if those were to be moved to `netlify/functions`, they might benefit from the update?
The reason this PR is updating most of them was for this specific ESLint rule, although I did bring some additional refactoring. 
Which way do you suggest for this PR to move forward? I am fine with closing it too if you think this is better. :+1:

erezrokah (approved):

erezrokah:
> I was thinking making those updates was rather low-risk considering those are only templates, and some of them were already somewhat broken.

Good point

PR #1496, opened 2020-10-30
Add `eslint-plugin-import`

ehmicky:
See https://github.com/netlify/build/issues/1955

This adds [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import).

PR #1498, opened 2020-10-30
Add `eslint-plugin-fp`

ehmicky:
This adds [`eslint-plugin-fp`](https://github.com/jfmengels/eslint-plugin-fp).

At the moment, the codebase cannot enable most rules from this ESLint rule, but we can enable them one by one as we go.


ehmicky:
For the failing test in CI, see https://github.com/netlify/cli/pull/1496#issuecomment-719814343

PR #1519, opened 2020-11-03
fix(build): pass team,site, addons env variables to build command

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1331

**- Test plan**

Tested manually with the flow in https://github.com/nimbella/netlify-plugin-nimbella

Inject relevant env variables when running the build command.

image

### DIFF
--- a/src/lib/build.js
+++ b/src/lib/build.js
@@ -1,14 +1,34 @@
 const build = require('@netlify/build')
 
+const { getSiteInformation } = require('../utils/dev')
+
+const getBuildEnv = async ({ context }) => {
+  const { warn, error, netlify } = context
+  const { site, api } = netlify
+  const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
+    api,
+    site,
+    warn,
+    error,
+  })
+  const env = { ...teamEnv, ...addonsEnv, ...siteEnv, ...dotFilesEnv }
+  return env
+}
+
 // We have already resolved the configuration using `@netlify/config`
 // This is stored as `netlify.cachedConfig` and can be passed to
 // `@netlify/build --cachedConfig`.
-const getBuildOptions = ({ netlify, token, flags }) => {
-  const cachedConfig = JSON.stringify(netlify.cachedConfig)
+const getBuildOptions = async ({ context, token, flags }) => {
+  const cachedConfig = JSON.stringify(context.netlify.cachedConfig)
   const { dry, debug } = flags
   // buffer = true will not stream output
   const buffer = flags.json || flags.silent
-  return { cachedConfig, token, dry, debug, mode: 'cli', telemetry: false, buffer }
+
+  const env = await getBuildEnv({
+    context,
+  })
+
+  return { cachedConfig, token, dry, debug, mode: 'cli', telemetry: false, buffer, env }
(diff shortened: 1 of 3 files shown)


src/lib/build.js
```diff
@@ -1,14 +1,34 @@
...
+const getBuildEnv = async ({ context }) => {
+  const { warn, error, netlify } = context
+  const { site, api } = netlify
+  const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
+    api,
+    site,
+    warn,
+    error,
+  })
+  const env = { ...teamEnv, ...addonsEnv, ...siteEnv, ...dotFilesEnv }
```
erezrokah, line 14: @ehmicky  since `@netlify/config` already evaluates some of these, should we move this logic to `@netlify/config`?

erezrokah:
Checking why this is failing on node 8

erezrokah:
Looks like this is another issue with `oclif`. Some of the `oclif` packages we are using are referencing `fs-extra@^9` which only supports node 10.

```
├─┬ @oclif/errors@1.3.3
│ └── fs-extra@9.0.1 
├─┬ @oclif/plugin-not-found@1.2.4
│ └─┬ cli-ux@4.9.3
│   └── fs-extra@7.0.1 
├─┬ @oclif/plugin-plugins@1.9.1
│ └── fs-extra@9.0.1  deduped
└─┬ cli-ux@5.5.0
  └── fs-extra@9.0.1  deduped
```

For example https://github.com/oclif/errors/blob/2233a100fbb8822a15a74e97c5f1ebd32c876a99/package.json#L35. A breaking change was done in https://github.com/oclif/errors/pull/27/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R9, but was published as a minor version https://github.com/oclif/errors/compare/v1.2.2...v1.3.0

erezrokah:
@ehmicky this is ready for review

src/lib/build.js
ehmicky: Yes.
I have filed https://github.com/netlify/build/issues/2036 with additional information. However, this can be done in a follow-up PR.

src/lib/build.js
```diff
@@ -1,14 +1,34 @@
 const build = require('@netlify/build')
 
+const { getSiteInformation } = require('../utils/dev')
+
+const getBuildEnv = async ({ context }) => {
+  const { warn, error, netlify } = context
+  const { site, api } = netlify
+  const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
```
ehmicky, line 8: We should reach a conclusion in https://github.com/netlify/build/issues/1514 before adding `.env` support in local builds. Would it be ok to split this specific feature from this PR?

ehmicky (changes requested):
This is looking good.
Just one comment about `.env` files support.

src/lib/build.js
erezrokah: Done in https://github.com/netlify/cli/pull/1519/commits/1556c5694df980e1b53ffe8eb4fb2592d444ec76

ehmicky (approved):

PR #1521, opened 2020-11-03
Release Tarball to Github Release

mraerino:
**- Summary**

In order to have a single asset that can be downloaded i decided to make a workflow that uses `oclif-dev pack` which is recommended by oclif and actively used by Heroku for their CLI.

Downside: This will include the traffic-mesh agent for linux, because it's built on linux. For anyone who wants to use `netlify dev` i'd recommend to not use those tarballs.

It will try to create a release for the version in the package json, but is fine if there already is one.

**- Test plan**

Testing in CI

Run where the release did not exist: https://github.com/netlify/cli/runs/1349748657
Run where the release already existed: https://github.com/netlify/cli/runs/1349776403
Test release: https://github.com/netlify/cli/releases/tag/v2.67.2.ci-test

Release as tarball to Github

[](https://unsplash.com/photos/3Xd5j9-drDA)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Publish Release Asset
+
+on:
+  push:
+#    tags:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+      - name: Install deps
+        run: npm ci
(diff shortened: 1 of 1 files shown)


fasterthanlime:
Closing, that workflow is being run from another repo, see https://github.com/netlify/pod-the-builder/issues/100#issuecomment-721715158

PR #1522, opened 2020-11-04
Release Tarball to Github Release on a private repository

fasterthanlime:
(follow-up to https://github.com/netlify/cli/pull/1521)

(this closes https://github.com/netlify/pod-the-builder/issues/100)

**- Summary**

In order to have a single asset that can be downloaded i decided to make a workflow that uses `oclif-dev pack` which is recommended by oclif and actively used by Heroku for their CLI.

Downside: This will include the traffic-mesh agent for linux, because it's built on linux. For anyone who wants to use `netlify dev` i'd recommend to not use those tarballs.

It will try to create a release for the version in the package json, but is fine if there already is one.

**- Test plan**

Testing in CI

Run where the release did not exist: (upcoming)
Run where the release already existed: (upcoming)
Test release: (upcoming)

(This should not be mentioned in the changelog, it's for internal use only)

### DIFF
--- a/.github/workflows/private-release.yml
+++ b/.github/workflows/private-release.yml
@@ -1,4 +1,4 @@
-name: Publish Release Asset
+name: Release packed CLI privately
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


fasterthanlime:
Closing, that workflow is being run from another repo, see https://github.com/netlify/pod-the-builder/issues/100#issuecomment-721715158

PR #1524, opened 2020-11-04
feat(dev): set path to geoip location database

calavera:
This will allow testing edge handlers to geoip location capabilities.

That flag was introduced in https://github.com/netlify/traffic-mesh/pull/1199/files

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1638,31 +1638,31 @@
       }
     },
     "@netlify/traffic-mesh-agent": {
-      "version": "0.24.0-pre.23",
-      "resolved": "https://registry.npmjs.org/@netlify/traffic-mesh-agent/-/traffic-mesh-agent-0.24.0-pre.23.tgz",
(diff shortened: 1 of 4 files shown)


erezrokah (changes requested):
This looks good.
Does it need a version bump for the npm traffic mesh agent package?
I'm getting:
```
error: Found argument '--location-db' which wasn't expected, or isn't valid in this context
```
When passing the argument

calavera:
That's weird, I assumed that the most recent version had that option already, I'll double check.

package-lock.json
```diff
@@ -10111,7 +10111,7 @@
         },
         "pify": {
           "version": "2.3.0",
-          "resolved": "https://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
+          "resolved": "http://registry.npmjs.org/pify/-/pify-2.3.0.tgz",
```
erezrokah, line 10114: Going to merge this once I figure out these `https->http` changes

erezrokah (changes requested):
~~@calavera I'm still getting the same error with version `0.24.0`~~
~~Maybe it's because I'm submitting a wrong db?~~
Argument name should be `geo`

PR #1530, opened 2020-11-06
feat: default to XDG spec for global config

### LINKED ISSUE #526: Respect XDG Base Directory Specification
Netlify CLI should use `$XDG_CONFIG_HOME/netlify`(or `~/.config/netlify` if `$XDG_CONFIG_HOME` is not set) instead of `~/.netlify ` to store config files.
This keeps users' home directories clean of dotfiles. It is being widely adopted by Linux/BSD applications.

See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html for more info.

evans:
**- Summary**

Thank you for making tools with slick dx(love the branch previews)! I was looking into #526 as a first issue 🎉  , which requests the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for the global netlify configuration. In other words use to `~/.config/netlify ...` instead of `~/.netlify/config.json`.

This PR suggests using `~/.netlify/config.json` if it exists. Otherwise configure `configstore` to place the config in `~/.config/netlify/config.json` or `$XDG_CONFIG_HOME/netlify/config.json`. I'm not fully happy with this plan, so I'm open to suggestions.

If `config.json` is retrieved in more places than the netlify cli([source](https://github.com/netlify/cli/blob/c704e22269a806b99adef9633fb1e473b5bbc8f4/src/utils/global-config.js#L13-L17)), then I'm happy to add that to this PR if possible! Also if the `~/.netlify` folder is used for more files, I recommend closing this PR and discussing the feasibility on #526 . I also worry about documentation, testing, and code style. Please let me know if there's something I missed.

Closes #526

<details><summary>My understanding of netlify configurations(not necessary to look at)</summary>
To frame the problem, there are three different configurations that I've come across: `netlify.toml`, `.netlify/config.json`, and `.netlify/state.json`. As I understand, their functions are:

* [`netlify.toml`](https://docs.netlify.com/configure-builds/file-based-configuration/): build settings(ex: use `npm run` vs `jekyll build`), deploy settings, and env variables
* `~/.netlify/config.json`: global netlify state, such as userId and auth tokens
* `.netlify/state.json`: local netlify state, such as siteId

Their path resolution: 

* `netlify.toml`: provided through flag or [source](https://github.com/netlify/build/blob/0b965ae2b1a1cc57f4724b1e18947f3aed6de891/packages/config/src/path.js#L16-L27)
* `~/.netlify/config.json`: set to `{os.homedir()}/.netlify/config.json` using configstore's [`configPath`](https://github.com/yeoman/configstore#configpath) [source](https://github.com/netlify/cli/blob/c704e22269a806b99adef9633fb1e473b5bbc8f4/src/utils/global-config.js#L13-L17)
* `.netlify/state.json`: searched for up from `cwd` like [node module resolution](https://www.typescriptlang.org/docs/handbook/module-resolution.html#how-nodejs-resolves-modules)
([source](https://github.com/netlify/cli/blob/c704e22269a806b99adef9633fb1e473b5bbc8f4/src/utils/state-config.js#L22))
</details>

**- Test plan**

I can add a test that uses [`mock-fs`](https://www.npmjs.com/package/mock-fs), though we may run into issues with mock-fs not playing nice with tests using node's `fs` (https://github.com/tschaub/mock-fs/issues/103) and I vaguely remember having difficulty across os's.

https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/lib/exec-fetcher.test.js#L16-L18

I've run locally, though I don't consider that worth much for maintainability, so I'd love suggestions here!

feat: default to XDG spec for global config when `~/.netlify/config.json` doesn't exist.

<details><summary>༼ つ ◕_◕ ༽つ</summary>

</details>

### DIFF
--- a/src/utils/global-config.js
+++ b/src/utils/global-config.js
@@ -1,7 +1,11 @@
+const fs = require('fs')
+
 const Configstore = require('configstore')
 const { v4: uuidv4 } = require('uuid')
 
 const { getPathInHome } = require('../lib/settings')
+
+const CONFIG_NAME = 'netlify'
 
 const globalConfigDefaults = {
   /* disable stats from being sent to Netlify */
@@ -10,8 +14,16 @@ const globalConfigDefaults = {
   cliId: uuidv4(),
 }
 
+const legacyConfigPath = getPathInHome(['config.json'])
+const legacyConfigExists = fs.existsSync(legacyConfigPath)
+
+// Configuration location can be:
+// - ~/.netlify/config
+// - $XDG_CONFIG_HOME/netlify/config.json
+// - ~/.config/netlify/config.json
 const globalConfigOptions = {
-  configPath: getPathInHome(['config.json']),
+  configPath: legacyConfigExists ? legacyConfigPath : undefined,
+  globalConfigPath: true,
 }
 
-module.exports = new Configstore(null, globalConfigDefaults, globalConfigOptions)
+module.exports = new Configstore(CONFIG_NAME, globalConfigDefaults, globalConfigOptions)


src/utils/global-config.js
```diff
@@ -1,17 +1,29 @@
...
+const CONFIG_NAME = 'netlify'
+
 const globalConfigDefaults = {
   /* disable stats from being sent to Netlify */
   telemetryDisabled: false,
   /* cliId */
   cliId: uuidv4(),
 }
 
+const legacyConfigPath = getPathInHome(['config.json'])
```
evans, line 17: We could look for the folder instead in case there are other files in `~/.netlify`, which would render this PR moot

ehmicky:
Hi @evans,

Thanks a lot for this feature!
XDG does sound like a good idea.

However, this not apply to Windows users. For them, I believe using a library like [`env-paths`](https://github.com/sindresorhus/env-paths#pathsconfig) would be more proper. This uses XDG for Linux but use more OS-idiomatic locations for macOS and Windows. On a typical machine this would be:
  - `/home/evans/.config/netlify/config.json` for Linux
  - `/Users/evans/Library/netlify/config.json` for macOS
  - `C:\Users\evans\AppData\Roaming\netlify\Config\config.json` for Windows

This can be easily handled with the `configPath` option of `configstore`. What do you think?

erezrokah:
To add to @ehmicky comment, would it make sense to change the file `exports` to a function?
By doing that we could:
1. Have better error handling.
2. Avoid calling `fs.existsSync` (which can fail) when the file is required.
3. Make it easier to have backward compatibility - for example we could read the old config, provide it as defaults and then remove the old config file. With this approach we can be sure the config file is always located in a single place.

evans:
@ehmicky @erezrokah I have some embarrassing news. I've started a new job and won't have the time to devote to open source outside of that scope, so I won't be pushing this PR forward.

In terms of your comments, I agree with the decision to make the retrieval of the config a function and `env-path` sounds like a great move too! As an macOS user, I love the thought of putting it inside `Library` 🎉  I imagine some warning or messaging would be useful to the user in case they're expecting the config to stay put, for example if they symlink their configuration from somewhere else.

I wish that I could make the time/brainspace to make the changes. If you have bandwidth, feel free to take the branch over. If not, feel free to close the PR. In any case, thank you for your reviews and work on the Netlify CLI! I'm rooting for you and look forward to the developer wins you bring 💪

erezrokah:
> I've started a new job and won't have the time to devote to open source outside of that scope, so I won't be pushing this PR forward.

Congrats on the new job and thank you for following up. Since the PR and the issue make sense we'll try to push it forward.

ehmicky:
@evans Thanks for stirring us in the right direction with this, we'll take it from there. Congratulations on the new job!

erezrokah:
Closing this PR for https://github.com/netlify/cli/pull/1725. Thanks again @evans for pushing this

PR #1531, opened 2020-11-07
docs: include npm install for site subdirectory

evans:
**- Summary**

I got unresolved dependencies errors during the pre-push hooks, since I hadn't installed the dependencies in the `site` subdirectory. This PR adds instructions to install those missing deps to the contributing guide. If you think there's a better place or alternative, happy to change the PR or you're welcome to take over!

Screen Shot 2020-11-06 at 8 33 20 PM

**- Test plan**

I've tested locally, though I wish there was a way to test/maintain the contributor onboarding experience. If there's something I'm not aware of, I'm down to look at the implementation.

docs: add instructions to install npm dependencies in the site subdirectory

<details><summary>༼ つ ◕_◕ ༽つ🤳</summary>

</details>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ First fork and clone the repository. If you're not sure how to do this, please w
 Run:
 
 ```bash
-npm install
+npm install && npm run site:build:install
 ```
 
 Tests are run with:


ehmicky (changes requested):
Thanks @evans for your contribution!

There is a `npm run site:build` task that could be used instead:

https://github.com/netlify/cli/blob/060fde2a999bdb9e0cbc5d166ff358bd614228e9/package.json#L68

evans:
Thanks @ehmicky!

I thought about adding something similar to the readme's [development section](https://github.com/netlify/cli#development), though that seems to be aimed at how to get up and running quickly

CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ First fork and clone the repository. If you're not sure how to do this, please w
 Run:
 
 ```bash
-npm install
+npm install && npm run site:build
```
ehmicky, line 25: We just add a discussion with @erezrokah and were thinking `npm run site:build:install` might actually make more sense, since the linting errors are only due to uninstalled dependencies. What do you think?

evans: That makes sense 👍  I'll make the update

Originally I was hesitant to use `npm ci` when reading into `site:build:install`. Now I think it's a reasonable decision over `npm install`, since it'll stay out of the way, installing stable dependencies. I'm assuming there's more value to smoothing the starting path than staying compatibility with the state of the user's `site/node_modules`, since I'm guessing they're coming to the contributing getting started as a landing page or while debugging. Also a vast majority of contributions are probably made to the main project, so I figure it's a non-issue and changes made to `site` will be done by someone familiar with the setup.

ehmicky (approved):

ehmicky:
Thanks @evans!

PR #1551, opened 2020-11-11
Stop using lodash standalone packages

nikoladev:
**- Summary**

[Lodash recommends to not use the separate per method packages](https://lodash.com/per-method-packages). This will also decrease the size of the bundle as well ensuring end-users download less dependencies.

**- Test plan**

Only 1 test failed after this change, but it seems unrelated:

```
  1 test failed

  tests › command.deploy › should deploy edge handlers when directory exists

  /[...]/netlify-cli/node_modules/execa/lib/error.js:56

  Rejected promise returned by test. Reason:

  Error {
    command: '/[...]/netlify-cli/bin/run deploy --json',
    exitCode: 2,
    failed: true,
    isCanceled: false,
    killed: false,
    signal: undefined,
    signalDescription: undefined,
    stderr: ` ›   Warning: ␊
     ›   {␊
     ›      "oclif": {␊
     ›        "exit": 2␊
     ›      }␊
     ›   }␊
     ›␊
     ›   Error: `,
    stdout: '',
    timedOut: false,
    message: 'Command failed with exit code 2: /[...]/netlify-cli/bin/run deploy --json',
  }

  makeError (node_modules/execa/lib/error.js:56:11)
  handlePromise (node_modules/execa/index.js:114:26)
  callCli (tests/utils/call-cli.js:6:22)
  tests/command.deploy.test.js:122:24
  withSiteBuilder (tests/utils/site-builder.js:139:12)
  tests/command.deploy.test.js:94:7
```

Switch to using lodash modules from the main package instead of the separate module packages. More info: https://lodash.com/per-method-packages

image


erezrokah:
> [Lodash recommends to not use the separate per method packages](https://lodash.com/per-method-packages). This will also decrease the size of the bundle as well ensuring end-users download less dependencies.

Thanks @nikoladev, since we're not bundling the CLI code I don't think that is relevant for us.
Do you have a package size comparison of `before/after` this change so we can measure the impact?

> The failed test is unrelated

nikoladev:
> Do you have a package size comparison of before/after this change so we can measure the impact?

How can I do a package size comparison?

For end-users the effective size should decrease because they almost certainly already have lodash in their node_modules since so many packages depend on it (at least if you install it locally per project. Which I do)

erezrokah:
> How can I do a package size comparison?

You can check the size of `node_modules` before/after the change.

nikoladev:
This is the size of `node_modules` in the netlify-cli repo.

Before PR:
```
➜ du -sk node_modules
402888  node_modules
```

After PR:
```
➜ du -sk node_modules
402020  node_modules
```

I also tried installing it as a dependency but I couldn't get it to work because I either couldn't install the new version through my git branch because of a cache error, or because when using `npm link` to install it and running `du -sk node_modules` I get `0` for the size.

erezrokah (approved):

erezrokah:
Thanks @nikoladev, did a rebase with another required change.

Issue #1553, opened 2020-11-12
Bug(command-dev): Netlify CLI should warn when a user has a background but the team doesn't supports it


erezrokah:
**Describe the bug**

Background functions are only available for specific plans. 
Netlify CLI should print a warning when a user has a function with a name ending in `-background` when the user runs `netlify dev`

**To Reproduce**

1. Set up a site on a free plan with a function named `hello-background.js`
2. Run `netlify dev`
3. See that no warning is printed.

**Expected behavior**

A warning should be printed explaining what can be done to fix the issue.

Issue #958, opened 2020-06-25
Avoid calling `getSite` twice


ehmicky:
When the configuration is being resolved, `@netlify/config` is receiving the `siteId` from `@netlify/cli`. It retrieves the site information (such as UI build settings) [using a `getSite` API call](https://github.com/netlify/build/blob/a5d09eac3b3e81063c34fa1ba9ba59cea499b0f0/packages/config/src/api/site_info.js#L6). This is returned as a `siteInfo` object (which is only `{}` if neither the `siteId` nor API `token` was available, and `{ id: siteId }` if only the `siteId` was available).

However, there are about 20 places in `@netlify/cli` that call the `getSite` endpoint. Instead, those should simply re-use the return value from `@netlify/config`.

sarahetter:
Closing as stale. Please re-open if needed

PR #1554, opened 2020-11-12
fix(command-dev): warn when background functions are not supported

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1553
Also related to https://github.com/netlify/cli/issues/958

**- Test plan**

Tested locally with an account that supports background functions, and another one that doesn't

See example warning:
image

Warn when background functions are not supported when `netlify dev` is invoked

image

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -169,7 +169,7 @@ class DevCommand extends Command {
     this.log(`${NETLIFYDEV}`)
     const { error: errorExit, log, warn, exit } = this
     const { flags } = this.parse(DevCommand)
-    const { api, site, config } = this.netlify
+    const { api, site, config, siteInfo } = this.netlify
     config.dev = { ...config.dev }
     config.build = { ...config.build }
     const devConfig = {
@@ -180,12 +180,13 @@ class DevCommand extends Command {
       ...flags,
     }
 
-    const { addonsUrls, teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
+    const { addonsUrls, teamEnv, addonsEnv, siteEnv, dotFilesEnv, siteUrl, capabilities } = await getSiteInformation({
       flags,
       api,
       site,
       warn,
       error: errorExit,
+      siteInfo,
     })
 
     await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
@@ -198,7 +199,15 @@ class DevCommand extends Command {
       exit(1)
     }
 
-    await startFunctionsServer({ settings, site, log, warn, errorExit, siteInfo: this.netlify.cachedConfig.siteInfo })
+    await startFunctionsServer({
+      settings,
+      site,
+      log,
+      warn,
+      errorExit,
+      siteUrl,
+      capabilities,
+    })
     await startFrameworkServer({ settings, log, exit })
 
     let url = await startProxyServer({ flags, settings, site, log, exit, addonsUrls })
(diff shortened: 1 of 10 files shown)


src/commands/dev/index.js
```diff
@@ -198,7 +199,15 @@ class DevCommand extends Command {
       exit(1)
     }
 
-    await startFunctionsServer({ settings, site, log, warn, errorExit, siteInfo: this.netlify.cachedConfig.siteInfo })
```
erezrokah, line 201: `siteInfo` was passed only for the sake of getting `ssl_url`

src/lib/account.js
```diff
@@ -0,0 +1,18 @@
+const dotProp = require('dot-prop')
+
+const supportsBooleanCapability = (account, capability) => {
```
erezrokah, line 3: A non boolean capability is build minuets (for example)

src/utils/dev.js
```diff
@@ -2,28 +2,26 @@ const process = require('process')
...
-const getSiteData = async ({ api, site, failAndExit }) => {
-  try {
-    const siteData = await api.getSite({ siteId: site.id })
-    return siteData
-  } catch (error) {
-    failAndExit(
-      `Failed retrieving site data for site ${chalk.yellow(site.id)}: ${error.message}. ${ERROR_CALL_TO_ACTION}`,
-    )
+const validateSiteInfo = ({ site, siteInfo, failAndExit }) => {
+  if (isEmpty(siteInfo)) {
```
erezrokah, line 16: An empty object is returned on error from `@netlify/config`

src/utils/dev.js
```diff
@@ -38,50 +36,64 @@ const getAddons = async ({ api, site, failAndExit }) => {
...
-    return siteData.build_settings.env
+const getSiteEnv = ({ siteInfo }) => {
+  if (siteInfo.build_settings && siteInfo.build_settings.env) {
+    return siteInfo.build_settings.env
   }
   return {}
 }
 
-const getSiteInformation = async ({ flags = {}, api, site, warn, error: failAndExit }) => {
+const getSiteInformation = async ({ flags = {}, api, site, warn, error: failAndExit, siteInfo }) => {
```
erezrokah, line 68: `siteInfo` is passed from `@netlify/config` to avoid calling `getSite` twice

src/utils/serve-functions.js
```diff
@@ -339,9 +354,8 @@ const createFormSubmissionHandler = function (siteInfo) {
           ...fields,
           ...Object.entries(files).reduce((prev, [name, { url }]) => ({ ...prev, [name]: url }), {}),
         }).map(([key, val]) => ({ title: capitalize(key), name: key, value: val })),
-        site_url: siteInfo.ssl_url,
+        site_url: siteUrl,
       },
-      site: siteInfo,
```
erezrokah, line 344: Passing `siteInfo` to the form submission function handler is redundant.

src/utils/serve-functions.js
```diff
@@ -410,48 +424,54 @@ const getBuildFunction = ({ functionBuilder, log }) => {
   }
 }
 
-const startFunctionsServer = async ({ settings, site, log, warn, errorExit, siteInfo }) => {
```
erezrokah, line 413: Broke this down into multiple functions

src/lib/account.js
ehmicky: Should this be added as a code comment?

ehmicky (approved):

Issue #1569, opened 2020-11-18
fix(command-deploy): Netlify CLI should print a relevant error message when background functions deploy fails due to unsupported plan


erezrokah:
**Describe the bug**

Related to https://github.com/netlify/cli/issues/1553.

Deploying a background function to a site under an account that doesn't support it generates a nasty error:
![image](https://user-images.githubusercontent.com/26760571/99533566-c882dd80-29ae-11eb-8bfe-492fd701e5c2.png)

We should provide a better error message

**To Reproduce**

1. Setup a Netlify site and add a function ending with `-background.js`
2. Run `netlify deploy`
3. See verbose error  

**Configuration**

```bash
  System:
    OS: macOS 11.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 58.27 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.15.0 - ~/.nvm/versions/node/v14.15.0/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v14.15.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmGlobalPackages:
    netlify-cli: 2.68.5
```

**Expected behavior**

A relevant error message and call to action should be displayed.

PR #1570, opened 2020-11-18
fix(command-deploy): print relevant log error when background functions are not supported

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1569

**- Test plan**

See error message:
image

Print relevant error message when failing to deploy a background function to an account that doesn't support it

🐼

### DIFF
--- a/src/lib/log.js
+++ b/src/lib/log.js
@@ -1,0 +1,23 @@
+const chalk = require('chalk')
+const dotProp = require('dot-prop')
+
+const RED_BACKGROUND = chalk.red('-background')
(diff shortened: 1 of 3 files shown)


src/lib/log.js
```diff
@@ -0,0 +1,23 @@
+const chalk = require('chalk')
```
erezrokah, line 1: I'm planning on moving more log messages to a centralized location.
This will add to consistency and we can even lint those as some point.

ehmicky (approved):

PR #1607, opened 2020-11-30
refactor(command-init): remove duplicate call to getSite

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/issues/958

**- Test plan**

Tested manually

Use `siteInfo` from `@netlify/config` instead of calling `api.getSite`

🐱

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -135,8 +135,7 @@ git remote add origin https://github.com/YourUserName/RepoName.git
       this.error(repo.error)
     }
 
-    // TODO: remove siteInfo.id === undefined once https://github.com/netlify/build/pull/2081 is published
-    if (isEmpty(siteInfo) || siteInfo.id === undefined) {
+    if (isEmpty(siteInfo)) {
       const NEW_SITE = '+  Create & configure a new site'
       const EXISTING_SITE = '⇄  Connect this directory to an existing Netlify site'
 
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -2,6 +2,7 @@ const { flags: flagsLib } = require('@oclif/command')
 const chalk = require('chalk')
 const inquirer = require('inquirer')
 const get = require('lodash/get')
+const isEmpty = require('lodash/isEmpty')
 
 const Command = require('../utils/command')
 const getRepoData = require('../utils/get-repo-data')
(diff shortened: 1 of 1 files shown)


src/commands/init.js
```diff
@@ -32,56 +32,41 @@ class InitCommand extends Command {
...
-    if (siteId && siteData && get(siteData, 'build_settings.repo_url') && !flags.force) {
-      const repoUrl = get(siteData, 'build_settings.repo_url')
+    const repoUrl = get(siteInfo, 'build_settings.repo_url')
+    if (repoUrl && !flags.force) {
       this.log()
       this.log(`This site has been initialized`)
       this.log()
-      this.log(`Site Name:  ${chalk.cyan(siteData.name)}`)
-      this.log(`Site Url:   ${chalk.cyan(siteData.ssl_url || siteData.url)}`)
-      if (repoUrl) {
```
erezrokah, line 55: This condition was redundant since it was check already here:
https://github.com/netlify/cli/pull/1607/files#diff-d9c0f8c5c3ab192e3bbc073115e4f12d530bcad9db684563da982b0767c5a94fL48

and since `repoUrl === get(siteData, 'build_settings.repo_url')`

src/commands/init.js
```diff
@@ -32,56 +32,41 @@ class InitCommand extends Command {
...
-    } catch (error) {
-      // silent api error
-      // TODO handle expected errors
-      // Throw unexpected ones
-    }
-
     // Add .netlify to .gitignore file
     await ensureNetlifyIgnore(site.root)
 
-    if (siteId && siteData && get(siteData, 'build_settings.repo_url') && !flags.force) {
```
erezrokah, line 48: It's enough to check for `get(siteInfo, 'build_settings.repo_url')` since it will falsy if `siteId` and `siteData` are falsy

erezrokah:
Added `low-risk`. This shouldn't change any behavior

ehmicky (approved):

src/commands/init.js
```diff
@@ -150,7 +135,8 @@ git remote add origin https://github.com/YourUserName/RepoName.git
       this.error(repo.error)
     }
 
-    if (!siteData) {
+    // TODO: remove siteInfo.id === undefined once https://github.com/netlify/build/pull/2081 is published
```
ehmicky, line 138: `@netlify/config@2.4.1` was published with that change.

erezrokah: Done in https://github.com/netlify/cli/pull/1607/commits/f13d566f663f8bcfcf6ae7727544a96bd7051733

Issue #1611, opened 2020-12-01
`netlify init` throws a 403 "GitHub error" on `Create & configure a new site`


lindsaylevine:
**Describe the bug**

<img width="735" alt="Screen Shot 2020-12-01 at 1 13 41 AM" src="https://user-images.githubusercontent.com/57350178/100705967-21bb1b80-3376-11eb-865c-bce378108e2e.png">

`netlify logout` and `netlify login` does not make a difference. `rm -rf .netlify` also does not make a difference.

The site actually does get created though, despite the 403s.

cc @erezrokah 

**To Reproduce**

1. Clone https://github.com/lindsaylevine/blender-resources
2. Navigate to the cloned repository
3. Run `netlify init` and auth through Github
4. Choose create & configure a new site
5. See output with 403

**Configuration**

[build]
  command   = "npm run build"
  functions = "out_functions"
  publish   = "out_publish"

<img width="435" alt="Screen Shot 2020-12-01 at 1 41 02 AM" src="https://user-images.githubusercontent.com/57350178/100706039-4911e880-3376-11eb-9f8d-b42abefd8f68.png">

**Expected behavior**

No 403 which to be honest made me think the site creation actually failed for awhile (😅)

** Additional **

Also happened with another project:

<img width="1200" alt="Screen Shot 2020-11-28 at 8 21 54 PM" src="https://user-images.githubusercontent.com/57350178/100706483-19afab80-3377-11eb-96b8-f97767a97f5c.png">

erezrokah:
Thanks for opening this @lindsaylevine, and sorry for the late reply.
I can't be 100% sure this is a CLI issue since we're missing some error handling and logs in that part of the code.
I've created a PR to address that https://github.com/netlify/cli/pull/1619.

From the second screenshot it looks like that deleting `~/.netlify` (the one in the home directory and not the repo) doesn't help with this.

When authentication with Netlify the CLI receives the token from the UI app, so it looks like the CLI is getting a token with wrong permissions.

For the second screenshot, does the error happens after approving the authentication in the opened browser window or before?

lindsaylevine:
no worries at all!!!!!!!

for the second screenshot, the error happens *after* authing in the browser

lindsaylevine:
*and* _after_ the site is created, which successfully gets created, weirdly

erezrokah:
The CLI first creates the site, then tries to add a deploy key on GitHub - which I believe this is the part that's failing.
[...]

PR #1614, opened 2020-12-01
refactor(command-init): extract code into functions

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/issues/1611

**- Test plan**

Tested manually the different flows

Extract init command into separate functions for code clarity

🧸 

Marking as `low-risk` as this shouldn't impact existing behavior

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -1,7 +1,7 @@
 const { flags: flagsLib } = require('@oclif/command')
 const chalk = require('chalk')
+const dotProp = require('dot-prop')
 const inquirer = require('inquirer')
-const get = require('lodash/get')
 const isEmpty = require('lodash/isEmpty')
 
 const Command = require('../utils/command')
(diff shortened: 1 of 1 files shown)


src/commands/init.js
```diff
@@ -14,206 +14,241 @@ const { track } = require('../utils/telemetry')
...
+class InitCommand extends Command {
+  async run() {
+    const { flags } = this.parse(InitCommand)
+    const { site, state } = this.netlify
+    let { siteInfo } = this.netlify
+
+    // Check logged in status
+    await this.authenticate()
+
+    await reportAnalytics({ context: this, flags })
```
ehmicky, line 221: Should we directly pass the `context` property that each function is actually using instead of the full object? This would make it clearer what the function actually needs, and might remove the need for a context object altogether (or minimize it).

For example, `reportAnalytics()` only needs the `config`. And some others only need the `log` helper function.

src/commands/init.js
```diff
@@ -14,206 +14,241 @@ const { track } = require('../utils/telemetry')
...
-        this.exit()
-      }
+  context.exit()
+}
 
-      // Throw github remote error
-      this.error(repo.error)
-    }
+const handleNoGitRemoteAndExit = async ({ context, error, state }) => {
+  context.log()
```
ehmicky, line 99: Unrelated to this PR, but I was wondering: do you know why we are passing the `log` function around instead of importing it in the top-level scope?

ehmicky (approved):
:crossed_swords:

src/commands/init.js
erezrokah: Good point. Let me fix that.

src/commands/init.js
erezrokah: Not sure what you mean by this. We use the `log` method from the `Command` class - we even override it:
https://github.com/netlify/cli/blob/6ea7ab428ce88c9ce4ea682eff8799777672efae/src/utils/command.js#L137

src/commands/init.js
erezrokah: Fixed in https://github.com/netlify/cli/pull/1614/commits/2fd92a455bc81f22f5609a2a46357aad1d27585e

src/commands/init.js
ehmicky: Yes, exactly.
We should extract this `log` class method into a normal function instead. Like this, we don't need to pass the `context` object around just due to `log()`. What do you think?

ehmicky (approved):

src/commands/init.js
erezrokah: Opened https://github.com/netlify/cli/issues/1615 to continue the discussion there

Issue #1568, opened 2020-11-18
Announcement: Dropping support for Node.js 8


erezrokah:
As Node.js 8 is out of the [maintenance window](https://nodejs.org/en/about/releases/) we should drop support for it and require Node.js 10.
This will allow us to upgrade [a bunch of packages](https://github.com/netlify/cli/blob/d09fa45874839ee4bb2a69ef9aa7e2249a68ddbe/renovate.json5#L13) and drop others we are using due to missing featured in Node.js 8.

### What to expect from this change
- It will be done behind a major version bump
- `package.json` [`engines` field](https://github.com/netlify/cli/blob/d09fa45874839ee4bb2a69ef9aa7e2249a68ddbe/package.json#L14) will be updated accordingly

### Potential breakage
- If you're installing the latest version of the CLI in CI that is running Node.js 8 (e.g. running `npm install -g netlify-cli` inside a GitHub action), that CI job will probably error out (the nature of the error depends on the package manager you're using and the settings passed to it). We recommend to either use a specific version range or update the CI Node.js version to  Node.js 10 or greater.
- If you're installing the CLI locally in a Node.js 8 environment you should get a relevant warning/error message saying the CLI is meant to be used on Node.js 10 or greater.

**If you're already running Node.js 10 or greater no change is needed on your part**

ehmicky:
Once the CLI drops support for Node 8, should we also drop support for Node 8 in the `js-client` and `zip-it-and-ship-it`?

The biggest backward compatibility problem we've got are production builds which set a custom Node version. Those builds require the build command and some plugins to use that version. However, `zip-it-and-ship-it` is used as a core command, i.e. it is not impacted by this. `js-client` is not impacted either.

erezrokah:
Sounds good to me as long as we follow `semver`. I think some customers are using `js-client` directly

PR #1664, opened 2020-12-16
feat!: drop Node.js 8 support

erezrokah:
Fixes https://github.com/netlify/cli/issues/1568

Not to be merged yet, but opened for visibility

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "Bret Comnes <bcomnes@gmail.com> (https://bret.io)"
   ],
   "engines": {
-    "node": ">=8.17.0"
+    "node": ">=10"
   },
   "files": [
     "/bin",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "Bret Comnes <bcomnes@gmail.com> (https://bret.io)"
   ],
   "engines": {
-    "node": ">=10"
+    "node": ">=10.18.0"
   },
   "files": [
     "/bin",
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,32 +3,4 @@
   ignorePresets: [':prHourlyLimit2'],
   semanticCommits: true,
   masterIssue: true,
-  packageRules: [
-    {
-      // Those cannot be upgraded to a major version until we drop support for Node 8
-      packageNames: [
-        '@octokit/rest',
-        'ava',
-        'boxen',
-        'chalk',
-        'concordance',
-        'cosmiconfig',
-        'del',
-        'dot-prop',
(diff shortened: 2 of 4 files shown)


renovate.json5
```diff
@@ -9,29 +9,5 @@
...
-        'dot-prop',
-        'execa',
-        'find-up',
-        'globby',
-        'locate-path',
-        'log-symbols',
-        'npm-packlist',
-        'ora',
-        'wrap-ansi',
-        'update-notifier',
```
ehmicky, line 30: It would be nice to follow up with upgrading those dependencies as well.

ehmicky (approved):

renovate.json5
erezrokah: Yes, I'm counting on renovate to open those PRs so I can examine the changelogs for breaking changes (there might ones not related to the Node.js upgrade).

package.json
```diff
@@ -11,7 +11,7 @@
     "Bret Comnes <bcomnes@gmail.com> (https://bret.io)"
   ],
   "engines": {
-    "node": ">=8.17.0"
+    "node": ">=10"
```
ehmicky, line 14: Should we specific a specific minor and patch version?

erezrokah: From a quick check most of the deps we haven't updated don't require it:
https://github.com/sindresorhus/path-exists/blob/82516fa81d83e75477e517a4ba6c2c8637198521/package.json#L13

ehmicky: Unless I'm mistaken, when the minor and patch version are omitted, `engines.node` understands it as "at least the latest 10.*.* available", as opposed to `>=10.0.0` which might be what we mean?

ehmicky: I think `engines.node >= 10` would be interpreted by `npm` and `yarn` as `>= 10.0.0`, so it's probably to match what our dependencies are doing :+1:

erezrokah: Looks like there is just 1 package that doesn't follow that pattern:
https://github.com/concordancejs/concordance/blob/2bf27e08a2512a586febad3f4eb40548c998fcaa/package.json#L11

We could specify `>=10.18.0`, WDYT?

ehmicky: From [the data](https://github.com/netlify/engineering/issues/50#issuecomment-704451333) of Node.js in production (not CLI) from 2 months ago, 93% of users were using `>=10.0.0` and 90% were using `>=10.18.0`. This probably will increase a lot in April when Node 10 is being officially deprecated. 
Based on this, this seems ok to me :+1:

PR #1674, opened 2020-12-20
feat: add warning message if running Node.js < 10

erezrokah:
Related to https://github.com/netlify/cli/issues/1568

Example output:
image

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -25,6 +25,9 @@ const { NETLIFY_AUTH_TOKEN, NETLIFY_API_URL } = process.env
 // Todo setup client for multiple environments
 const CLIENT_ID = 'd6f37de6614df7ae58664cfca524744d73807a377f5ee71f1a254f78412e3750'
 
+// 'api' command uses JSON output by default
+const isDefaultJson = () => argv._[0] === 'api' && argv.list !== true
+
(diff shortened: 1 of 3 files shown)


src/utils/command.js
```diff
@@ -23,6 +25,19 @@ const { NETLIFY_AUTH_TOKEN, NETLIFY_API_URL } = process.env
 // Todo setup client for multiple environments
 const CLIENT_ID = 'd6f37de6614df7ae58664cfca524744d73807a377f5ee71f1a254f78412e3750'
 
+// 'api' command uses JSON output by default
+const isDefaultJson = () => argv._[0] === 'api' && argv.list !== true
```
erezrokah, line 29: We override the `log` method and don't print messages if the `silent` or `json` flags are passed.
For the `api` command I needed to add a special case since it uses `json` format by default and not flag is passed (this is ugly, but a better fix would require a bigger refactor).

The main problem with our current approach is that we don't use `oclif` flags, but we parse the arguments using `minimist`:
https://github.com/netlify/cli/blob/4171bc6d26c057b6c986a6277498ba1ca0b58fc6/src/utils/command.js#L9.
That prevents use from adding a `json` flag to the `api` command with a default value of `true`.

Benaiah (approved):

PR #1675, opened 2020-12-20
refactor(command-functions-invoke): use async/await

erezrokah:
A small refactoring to use `async/await` when invoking functions

### DIFF
--- a/src/commands/functions/invoke.js
+++ b/src/commands/functions/invoke.js
@@ -38,7 +38,6 @@ class FunctionsInvokeCommand extends Command {
       flags.functions || (config.dev && config.dev.functions) || (config.build && config.build.functions)
     if (typeof functionsDir === 'undefined') {
       this.error('functions directory is undefined, did you forget to set it in netlify.toml?')
-      process.exit(1)
     }
 
     if (!flags.port)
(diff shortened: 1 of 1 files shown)


src/commands/functions/invoke.js
```diff
@@ -38,7 +38,6 @@ class FunctionsInvokeCommand extends Command {
       flags.functions || (config.dev && config.dev.functions) || (config.build && config.build.functions)
     if (typeof functionsDir === 'undefined') {
       this.error('functions directory is undefined, did you forget to set it in netlify.toml?')
-      process.exit(1)
```
erezrokah, line 41: `error` already exists the process

src/commands/functions/invoke.js
```diff
@@ -122,26 +121,20 @@ class FunctionsInvokeCommand extends Command {
...
-      method: 'post',
-      headers,
-      body: JSON.stringify(body),
-    })
-      .then((response) => {
-        let data
-        data = response.text()
-        try {
-          // data = response.json();
-          data = JSON.parse(data)
```
erezrokah, line 136: `JSON.parse(data)` was always failing as `response.text()` returns a promise.
Regardless a successful parse would log an object which might log something like:
`{ message: [ 1, { test: [Array] } ] }`

Benaiah (approved):

Issue #473, opened 2019-03-29
functions NODE_ENV === production


kentcdodds:
Ok, so here's a bit of an issue that I think will confuse a lot of people: `netlify dev` actually runs as if everything's in production (NODE_ENV=production), BUT it runs your `develop` script rather than your `start` script.

This is problematic because I expect to be able to run everything in development (like NODE_ENV=development) because the script is called `dev` and it runs my `develop` script. So it's confusing that my functions would run with NODE_ENV=production.

Does that make sense? I think this could probably be resolved if `netlify dev` just basically replaced `netlify-lambda` and ran the `netlify-lambda serve` script for me. What would be best is if `netlify dev` could find the script in my `package.json` that runs `netlify-lambda serve` and just run that, because my scripts looks like this:

```
"functions:build": "netlify-lambda build netlify/functions",
"functions:dev": "dotenv -e ./netlify/.env netlify-lambda serve netlify/functions",
```

If `netlify dev` just found the script that runs `netlify-lambda serve` and ran that, then I think this issue would be resolved as would netlify/netlify-dev-plugin#61 and netlify/netlify-dev-plugin#60.

tysonmatanich:
When I running `netlify dev`, `process.env.NODE_ENV` is always `undefined`. I would expect `netlify dev` to set it to `'development'`. I would also expect other [Netlify build environment variables](https://docs.netlify.com/configure-builds/environment-variables/) to get set such as `CONTEXT`.

eduardoboucas:
Hi! I wanted to share a couple of updates on the items mentioned in this issue.

> I would expect netlify dev to set it to 'development'. I would also expect other Netlify build environment variables to get set such as CONTEXT.

@tysonmatanich This should be available in the latest version of CLI.

> Even with the following,
>
> ```
> [build.environment]
>   NODE_ENV="development"
> ```

@jonathan-mothership variables declared in `build.environment` are now picked up by `netlify dev`.

---

As for making `NODE_ENV=development` by default, it's something we're looking into and we'll update this issue as soon as there's any progress there.

Thanks all!

erezrokah:
Adding more context to clear up confusion.
1. `netlify-lambda` uses Webpack under the hood. Webpack inlines environments variables so `netlify dev` cannot override them. As a workaround you can use a [custom Webpack config](https://github.com/netlify/netlify-lambda#webpack-configuration):
```js
const webpack = require('webpack');

module.exports = {
  plugins: [
[...]

PR #1685, opened 2020-12-29
Set CONTEXT when running Netlify Dev

jon-sully:
**- Summary**

In the last few months the addition of Build Plugins (particularly https://github.com/bencao/netlify-plugin-inline-functions-env) combined with the [deploy context env var](https://docs.netlify.com/configure-builds/environment-variables/#build-metadata) have made it possible for Functions to feel environment-agnostic a la:

```js
const contextualEnvVar = (v) => {
  const currentContext = process.env.CONTEXT
  const formattedContext = currentContext.replace('-', '_').toUpperCase()
  return process.env[`${formattedContext}_${v}`]
}

exports.handler = async _ => {
  return {
    statusCode: 200,
    body: JSON.stringify({
      status: contextualEnvVar('API_KEY')
    })
  }
}

```

I'll be writing up a post on this soon, but the premise is that I can set `PRODUCTION_API_KEY`, `DEPLOY_PREVIEW_API_KEY`, and `DEV_API_KEY` all in the Netlify Admin UI then my function will correctly get the key value for each environment using `contextualEnvVar('API_KEY')` as shown above.

The ultimate result is that my Function will run with the correct ENV vars for each environment it's in, I can (securely) set ENV var _values_ via the Netlify Admin UI for all four environments (dev, deploy-preview, branch-preview, and production), and I can _easily_ override the dev key locally by just adding the key to my local `.env` file (adding `DEV_API_KEY` in this case).

This is really neat. Using Stripe as the canonical example, this structure would let me have my Stripe Prod and Testing keys configured safely in the Netlify Admin UI and feel good knowing that my local dev, deploy-previews, and production environments would be using the correct keys all the time.

This PR just adds `CONTEXT` = `dev` to the process thread when running netlify dev locally. Currently no `CONTEXT` is set and I think this just adds parity to Netlify's other environments where `CONTEXT` _is_ set.

Add CONTEXT=dev to environment variables when running netlify dev

**- What else**

I currently have this working great on a micro testing site, Tenv (test env)! Here are a couple of `cURL`s to show:

`$ curl https://deploy-preview-2--tenv.netlify.app/.netlify/functions/output-env-var`

`$ curl https://tenv.netlify.app/.netlify/functions/output-env-var`

The site can be inspected as needed too. [tenv.netlify.app](https://tenv.netlify.app)

I believe this addition / workflow would also satisfy a [few](https://community.netlify.com/t/environment-variable-per-branch-for-lambda-functions/5009/14) different [community](https://community.netlify.com/t/netlify-functions-and-env-variables-from-netlify-toml/4404/12) requests.

image

### DIFF
--- a/src/utils/dev.js
+++ b/src/utils/dev.js
@@ -163,6 +163,7 @@ const addEnvVariables = ({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv }) => {
   })
 
   process.env.NETLIFY_DEV = 'true'
+  process.env.CONTEXT = 'dev'
 }
 
 module.exports = {


jon-sully:
I'm not sure why some of the CI builds failed 😕 A couple of the Windows builds on the first pass and now one of the ubuntu builds

erezrokah:
Thanks @jon-sully, the CI failure seems unrelated. This is also related to https://github.com/netlify/cli/issues/473#issuecomment-715522228.
Please bare with us while we discuss this internally to figure out what are the best built in env variables the CLI should inject.

erezrokah (approved):
Thanks @jon-sully  this is good to go and sorry for the delay in approving it.

jon-sully:
Yay! Thanks!

PR #1694, opened 2021-01-05
test(command-dev): log dev server output on timeout

erezrokah:
This PR reports the dev server output in our tests when we fail to spawn it.


tests/utils/dev-server.js
```diff
@@ -63,6 +66,9 @@ const startServer = async ({ cwd, env = {}, args = [] }) => {
     })
     ps.catch((error) => !selfKilled && reject(error))
   })
+
+  const { timeout, ...server } = await Promise.race([serverPromise, getTimeoutPromise(SERVER_START_TIMEOUT)])
```
ehmicky, line 70: [sand] Do you think `p-timeout` might help with this pattern?
https://github.com/sindresorhus/p-timeout

tests/utils/dev-server.js
```diff
@@ -37,10 +37,13 @@ const startServer = async ({ cwd, env = {}, args = [] }) => {
...
     env: { BROWSER: 'none', ...env },
   })
-  return new Promise((resolve, reject) => {
+  let output = ''
+  const serverPromise = new Promise((resolve, reject) => {
     let selfKilled = false
-    ps.stdout.on('data', (data) => {
-      if (data.toString().includes('Server now ready on')) {
+    ps.stdout.on('data', (dataBuffer) => {
+      const data = dataBuffer.toString()
```
ehmicky, line 44: I believe using the `encoding: 'utf-8'` option of `execa` should provide with `data` string chunks instead directly. Would this work?

ehmicky (approved):

tests/utils/dev-server.js
erezrokah: Sure, done in https://github.com/netlify/cli/pull/1694/commits/a0d81587763b7673c23de5b294f02cb7a873b63b

tests/utils/dev-server.js
erezrokah: Done in https://github.com/netlify/cli/pull/1694/commits/12b4af197cbfc48ec946ccbf15dc05da571f8aa9

PR #1713, opened 2021-01-11
test(command-dev): limit concurrency

erezrokah:
This is an attempt to make our CI more stable by limiting the concurrency of the servers we spawn during tests.

### DIFF
--- a/tests/command.dev.test.js
+++ b/tests/command.dev.test.js
@@ -4,8 +4,8 @@ const http = require('http')
 const path = require('path')
 const process = require('process')
 
-const { Semaphore } = require('async-mutex')
-const test = require('ava')
+// eslint-disable-next-line ava/use-test
+const avaTest = require('ava')
 const dotProp = require('dot-prop')
 const FormData = require('form-data')
 const jwt = require('jsonwebtoken')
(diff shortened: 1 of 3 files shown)


erezrokah:
@ehmicky this doesn't solve all our CI issues, but looks like it doesn't impact performance:
https://github.com/netlify/cli/actions/runs/477184868
https://github.com/netlify/cli/actions/runs/477774268

tests/command.dev.test.js
```diff
@@ -14,6 +15,9 @@ const { withDevServer } = require('./utils/dev-server')
 const { startExternalServer } = require('./utils/external-server')
 const { withSiteBuilder } = require('./utils/site-builder')
 
+const CONCURRENCY = 1
+const semaphore = process.env.CI === 'true' ? new Semaphore(CONCURRENCY) : null
```
ehmicky, line 19: Would the following work too?

```js
const testFunc = process.env.CI === 'true' ? test.serial.bind(test) : test

testFunct('test title', (t) => { ... })
```

erezrokah: You're right, I was playing with different values or `CONCURRENCY` (`4->2->1`).
We can remove the dependency and enforce `serial`.

erezrokah: I chose to do it a bit different. It required disabling a lint rule, but makes the spec consistent with other ones.

ehmicky (approved):

PR #1722, opened 2021-01-12
Use environment variables from @netlify/config

eduardoboucas:
**- Summary**

This PR is the CLI counterpart of https://github.com/netlify/build/pull/2040. It introduces the updated environment variables object coming from `@netlify/build` to various commands.

Copying across this list created by @ehmicky in the build PR. I'll use it to track the progress on the various commands.

  - `netlify build`:
     - [x] Do not pass [the `env` option](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/lib/build.js#L31) to `@netlify/build` anymore
     - [x] This means [`getBuildEnv()`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/lib/build.js#L5) can be removed.
  - [`netlify env:list`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/commands/env/list.js#L26) and [`netlify env:get`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/commands/env/get.js#L23):
     - [x] Use the `env` return value from `@netlify/config` instead of `config.build.environment`.
       We might want to select only specific environment variables categories, and merge them.
  - `netlify dev`, `netlify dev:exec`, `netlify functions:create`:
     - [x] Use the `env` return value from `@netlify/config`. Note: this will also add support to `build.environment`.
     - [x] The `dev.environment` property in `netlify.toml` should be merged to them: `{ ...env.all, ...dev.environment }`.
     - [x] Do not use [`getSiteInformation()`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/utils/dev.js#L62) anymore, except for the [`addonsUrl`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/utils/dev.js#L42) and [`dotenv`](https://github.com/netlify/cli/blob/master/src/utils/dot-env.js) logic
     - [x] Adjust logs in [`addEnvVariables()`](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/utils/dev.js#L97) to dissociate between netlify.toml and UI site environment variables
     - Pass [`--offline` CLI flag](https://github.com/netlify/cli/blob/e031d8eb3bc2b56e85fc19e5fc395440acd30158/src/commands/dev/index.js#L254) to `@netlify/config`. `netlify dev:exec` and `netlify functions:create` might not have this flag yet, but `netlify dev` does.

**- Test plan**

Added new tests that assert the capture of environment variables from `[build.environment]`. Rest of the test suite unchanged.

Mimic the environment variables used in production

5f510bbc10345b91b189bf45dde83ddb

### DIFF
--- a/src/commands/dev/exec.js
+++ b/src/commands/dev/exec.js
@@ -1,20 +1,13 @@
 const execa = require('execa')
 
 const Command = require('../../utils/command')
-const { getSiteInformation, addEnvVariables } = require('../../utils/dev')
+const { injectEnvVariables } = require('../../utils/dev')
 
 class ExecCommand extends Command {
   async run() {
-    const { log, warn, error, netlify } = this
-    const { site, api, siteInfo } = netlify
-    const { teamEnv, addonsEnv, siteEnv, dotFilesEnv } = await getSiteInformation({
-      api,
-      site,
-      warn,
-      error,
-      siteInfo,
-    })
-    await addEnvVariables({ log, teamEnv, addonsEnv, siteEnv, dotFilesEnv })
+    const { log, warn, netlify } = this
+    const { site } = netlify
+    await injectEnvVariables({ env: this.netlify.cachedConfig.env, log, site, warn })
 
     execa(this.argv[0], this.argv.slice(1), {
       stdio: 'inherit',
--- a/src/commands/dev/exec.js
+++ b/src/commands/dev/exec.js
@@ -6,8 +6,8 @@ const { injectEnvVariables } = require('../../utils/dev')
 class ExecCommand extends Command {
   async run() {
     const { log, warn, netlify } = this
-    const { site } = netlify
-    await injectEnvVariables({ env: this.netlify.cachedConfig.env, log, site, warn })
+    const { cachedConfig, site } = netlify
+    await injectEnvVariables({ env: cachedConfig.env, log, site, warn })
 
     execa(this.argv[0], this.argv.slice(1), {
       stdio: 'inherit',
--- a/src/commands/env/list.js
+++ b/src/commands/env/list.js
@@ -1,3 +1,4 @@
+const fromEntries = require('@ungap/from-entries')
 const AsciiTable = require('ascii-table')
 const isEmpty = require('lodash/isEmpty')
 
@@ -22,17 +23,12 @@ class EnvListCommand extends Command {
     })
 
     const siteData = await api.getSite({ siteId })
-    const environment = Object.entries(cachedConfig.env).reduce((env, [key, variable]) => {
-      // Omitting general variables to reduce noise.
-      if (variable.sources[0] === 'general') {
-        return env
-      }
-
-      return {
-        ...env,
-        [key]: variable.value,
-      }
-    }, {})
+    const environment = fromEntries(
+      Object.entries(cachedConfig.env)
+        // Omitting general variables to reduce noise.
+        .filter(([, variable]) => variable.sources[0] !== 'general')
+        .map(([key, variable]) => [key, variable.value]),
+    )
 
     // Return json response for piping commands
     if (flags.json) {
--- a/src/commands/env/list.js
+++ b/src/commands/env/list.js
@@ -6,7 +6,7 @@ const Command = require('../../utils/command')
 class EnvListCommand extends Command {
   async run() {
     const { flags } = this.parse(EnvListCommand)
-    const { api, site, config } = this.netlify
+    const { api, site, cachedConfig } = this.netlify
     const siteId = site.id
 
     if (!siteId) {
@@ -22,9 +22,17 @@ class EnvListCommand extends Command {
     })
 
     const siteData = await api.getSite({ siteId })
-    const {
-      build: { environment = {} },
-    } = config
+    const environment = Object.entries(cachedConfig.env).reduce((env, [key, variable]) => {
+      // Omitting general variables to reduce noise.
+      if (variable.sources[0] === 'general') {
+        return env
+      }
+
+      return {
+        ...env,
+        [key]: variable.value,
+      }
+    }, {})
 
     // Return json response for piping commands
     if (flags.json) {
--- a/src/lib/build.js
+++ b/src/lib/build.js
@@ -1,35 +1,23 @@
 const build = require('@netlify/build')
 
-const { getSiteInformation } = require('../utils/dev')
-
-const getBuildEnv = async ({ context }) => {
-  const { warn, error, netlify } = context
-  const { site, api, siteInfo } = netlify
-  const { teamEnv, addonsEnv, siteEnv } = await getSiteInformation({
-    api,
-    site,
-    warn,
-    error,
-    siteInfo,
-  })
-  const env = { ...teamEnv, ...addonsEnv, ...siteEnv }
-  return env
-}
-
 // We have already resolved the configuration using `@netlify/config`
 // This is stored as `netlify.cachedConfig` and can be passed to
 // `@netlify/build --cachedConfig`.
-const getBuildOptions = async ({ context, token, flags }) => {
-  const cachedConfig = JSON.stringify(context.netlify.cachedConfig)
+const getBuildOptions = ({ context, token, flags }) => {
+  const { cachedConfig } = context.netlify
+  const serializedConfig = JSON.stringify(cachedConfig)
   const { dry, debug } = flags
   // buffer = true will not stream output
   const buffer = flags.json || flags.silent
+  const env = Object.entries(cachedConfig.env).reduce(
+    (obj, [key, variable]) => ({
+      ...obj,
+      [key]: variable.value,
+    }),
+    {},
+  )
 
-  const env = await getBuildEnv({
-    context,
-  })
-
-  return { cachedConfig, token, dry, debug, mode: 'cli', telemetry: false, buffer, env }
+  return { cachedConfig: serializedConfig, token, dry, debug, mode: 'cli', telemetry: false, buffer, env }
 }
 
 const runBuild = async (options) => {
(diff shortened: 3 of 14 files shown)


src/commands/env/list.js
```diff
@@ -22,9 +22,17 @@ class EnvListCommand extends Command {
...
+      // Omitting general variables to reduce noise.
+      if (variable.sources[0] === 'general') {
+        return env
+      }
+
+      return {
+        ...env,
+        [key]: variable.value,
+      }
+    }, {})
```
erezrokah, line 35: [sand] Could we use `const fromEntries = require('@ungap/from-entries')` for this?

```suggestion
    const environment = fromEntries(
      Object.entries(cachedConfig.env)
        .filter(([, variable]) => variable.sources[0] !== 'general')
        .map(([key, variable]) => [key, variable.value]),
    )
```

src/commands/dev/exec.js
```diff
@@ -1,20 +1,13 @@
 const execa = require('execa')
 
 const Command = require('../../utils/command')
-const { getSiteInformation, addEnvVariables } = require('../../utils/dev')
+const { injectEnvVariables } = require('../../utils/dev')
```
erezrokah, line 4: I like this rename `addEnvVariables-> injectEnvVariables`!

src/lib/build.js
```diff
@@ -1,35 +1,23 @@
...
 // `@netlify/build --cachedConfig`.
-const getBuildOptions = async ({ context, token, flags }) => {
-  const cachedConfig = JSON.stringify(context.netlify.cachedConfig)
+const getBuildOptions = ({ context, token, flags }) => {
+  const { cachedConfig } = context.netlify
+  const serializedConfig = JSON.stringify(cachedConfig)
   const { dry, debug } = flags
   // buffer = true will not stream output
   const buffer = flags.json || flags.silent
+  const env = Object.entries(cachedConfig.env).reduce(
```
erezrokah, line 12: Do we need to pass `env` as a build option? Won't `@netlify/build` use the one in `cachedConfig`?
cc @ehmicky.

If we need it could we [sand] use:
```js
const env = fromEntries(Object.entries(cachedConfig.env).map(([key, variable]) => [key, variable.value]))
```

src/utils/dev.js
```diff
@@ -96,77 +124,61 @@ const getSiteInformation = async ({ flags = {}, api, site, warn, error: failAndE
...
-      logIgnoredEnvVar({ key, log, parentSource: environment.get(key).source, source })
-    } else {
-      environment.set(key, { source, value: siteEnv[key] })
+// Takes a set of environment variables in the format provided by @netlify/config, augments it with variables from both
+// dot-env files and the process itself, and injects into `process.env`.
+const injectEnvVariables = async ({ env, log, site, warn }) => {
+  const environment = new Map(Object.entries(env))
+  const dotEnvFiles = await loadDotEnvFiles({ projectDir: site.root, warn })
+
+  for (const { file, env: fileEnv } of dotEnvFiles.reverse()) {
```
erezrokah, line 139: I think `utils/dev.js` is the only consumer for `loadDotEnvFiles`.
If that's the case should we return the result in `loadDotEnvFiles` in the correct order instead of reversing it?

src/lib/build.js
```diff
@@ -1,35 +1,23 @@
 const build = require('@netlify/build')
 
-const { getSiteInformation } = require('../utils/dev')
```
erezrokah, line 3: 🧹

src/utils/dev.js
```diff
@@ -96,77 +124,61 @@ const getSiteInformation = async ({ flags = {}, api, site, warn, error: failAndE
...
+    for (const key in fileEnv) {
+      const newSourceName = `${file} file`
+      const sources = environment.has(key) ? [newSourceName, ...environment.get(key).sources] : [newSourceName]
```
erezrokah, line 142: [dust] we can also keep the order of `loadDotEnvFiles` and append the new source.
I'm not sure if that would be better.

src/utils/dev.js
```diff
@@ -9,6 +9,34 @@ const { supportsBackgroundFunctions } = require('../lib/account')
 const { loadDotEnvFiles } = require('./dot-env')
 const { NETLIFYDEVLOG } = require('./logo')
 
+// Possible sources of environment variables. For the purpose of printing log messages only. Order does not matter.
+const ENV_VAR_SOURCES = {
```
erezrokah, line 13: Nice!

erezrokah (changes requested):
Thanks @eduardoboucas. Looks good and I love all the code removal.
It's great the CLI doesn't have to resolve those env variables.
Added a few comments/questions.

src/commands/env/list.js
eduardoboucas: Sure! I'll update it.

src/lib/build.js
eduardoboucas: I don't think we need it. @ehmicky's list had an item for removing it, which I somehow missed. I'll remove it. Thanks!

src/utils/dev.js
eduardoboucas: Good point. I'll update.

eduardoboucas:
@erezrokah I addressed your comments in github.com/netlify/cli/pull/1722/commits/320d0746f10e2d4d4721af754a8679e0078edbfb, except for https://github.com/netlify/cli/pull/1722#discussion_r556512390 (removing this is causing the tests to fail, I'm investigating ⏳ )

Note that in https://github.com/netlify/cli/pull/1722/commits/320d0746f10e2d4d4721af754a8679e0078edbfb#diff-98719cf70cb4c9ae11108e4f1dc3b629caf2e0f429a9ad4dee054596f8f69950L24-R22 I removed the filtering for `process.env` variables, because we're already doing that in `injectEnvVariables`. If we don't remove this, variables in `.env` that also exist in `process.env` would never reach `injectEnvVariables` and therefore wouldn't get logged.

As a result of this, I had to remove [the corresponding test](https://github.com/netlify/cli/pull/1722/commits/320d0746f10e2d4d4721af754a8679e0078edbfb#diff-4c84114a2d104f8a46df989ffa1958e178abf3c0ada412a7470ca6518af6698bL84-L97). I'll try to add a similar test in a more appropriate place, to assert that `process.env` is still taking precedence over `.env` files.

erezrokah:
> Note that in [320d074#diff-98719cf70cb4c9ae11108e4f1dc3b629caf2e0f429a9ad4dee054596f8f69950L24-R22](https://github.com/netlify/cli/commit/320d0746f10e2d4d4721af754a8679e0078edbfb#diff-98719cf70cb4c9ae11108e4f1dc3b629caf2e0f429a9ad4dee054596f8f69950L24-R22) I removed the filtering for `process.env` variables, because we're already doing that in `injectEnvVariables`. If we don't remove this, variables in `.env` that also exist in `process.env` would never reach `injectEnvVariables` and therefore wouldn't get logged.
> 
> As a result of this, I had to remove [the corresponding test](https://github.com/netlify/cli/pull/1722/commits/320d0746f10e2d4d4721af754a8679e0078edbfb#diff-4c84114a2d104f8a46df989ffa1958e178abf3c0ada412a7470ca6518af6698bL84-L97). I'll try to add a similar test in a more appropriate place, to assert that `process.env` is still taking precedence over `.env` files.

Great catch! Removing the filtering logic makes sense.

src/utils/dev.js
ehmicky: :art:

src/commands/env/list.js
```diff
@@ -22,9 +23,12 @@ class EnvListCommand extends Command {
     })
 
     const siteData = await api.getSite({ siteId })
-    const {
-      build: { environment = {} },
-    } = config
+    const environment = fromEntries(
+      Object.entries(cachedConfig.env)
+        // Omitting general variables to reduce noise.
```
ehmicky, line 28: :+1:

ehmicky (approved):
This is looking great, thanks @eduardoboucas!

Everything is looking neat :v: 

The CI tests are failing, but I'm not quite sure this is related to this PR considering different tests are failing, and not for every OS and Node.js version... @erezrokah What do you think?

erezrokah (approved):

erezrokah:
> The CI tests are failing, but I'm not quite sure this is related to this PR considering different tests are failing, and not for every OS and Node.js version... @erezrokah What do you think?

Yes, everything looks good. Looks the deploy command tests are especially flaky, Seems unrelated. I'm working to improve those too.

eduardoboucas:
Contrary to our internal discussion, I ended up including the addition of the `--offline` flag in this PR, via https://github.com/netlify/cli/pull/1722/commits/1c49d959615781e55c453ef06278873b8d2b0c9b. I hope it's okay, as since it's literally a one-line change and it made my testing process a bit easier.

@erezrokah @ehmicky if you're okay with that, and once the tests finish running, what's our plan for releasing? Do we release `@netlify/build` and `@netlify/config`, and then merge this PR and release CLI?

Thanks! 🙌🏻

erezrokah:
> @erezrokah @ehmicky if you're okay with that, and once the tests finish running, what's our plan for releasing? Do we release `@netlify/build` and `@netlify/config`, and then merge this PR and release CLI?

I'm good with it. The release process you described is 💯

erezrokah (approved):

ehmicky:
@eduardoboucas and I will pair to release this and the related Netlify Build PR tomorrow.

ehmicky (approved):
@eduardoboucas This should be ready for merge, after `@netlify/build` and `@netlify/config` are updated to the latest versions in `package.json`.

PR #1727, opened 2021-01-13
test: add CLI timeout

erezrokah:
This PR adds a timeout of 30 seconds for CLI invocations in our tests.

### DIFF
--- a/tests/utils/call-cli.js
+++ b/tests/utils/call-cli.js
@@ -2,8 +2,15 @@ const execa = require('execa')
 
 const cliPath = require('./cli-path')
 
+const CLI_TIMEOUT = 3e5
+
 const callCli = async function (args, execOptions) {
-  const { stdout } = await execa(cliPath, args, { windowsHide: true, windowsVerbatimArguments: true, ...execOptions })
+  const { stdout } = await execa(cliPath, args, {
+    windowsHide: true,
+    windowsVerbatimArguments: true,
+    timeout: CLI_TIMEOUT,
+    ...execOptions,
+  })
   return stdout
 }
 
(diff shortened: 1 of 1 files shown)


tests/utils/call-cli.js
```diff
@@ -2,8 +2,15 @@ const execa = require('execa')
 
 const cliPath = require('./cli-path')
 
+const CLI_TIMEOUT = 3e5
+
 const callCli = async function (args, execOptions) {
-  const { stdout } = await execa(cliPath, args, { windowsHide: true, windowsVerbatimArguments: true, ...execOptions })
+  const { stdout } = await execa(cliPath, args, {
+    windowsHide: true,
```
ehmicky, line 9: This is already [the default value](https://github.com/sindresorhus/execa#windowshide).
Note: `windowsHide` is a tricky option which has [pros and cons](https://github.com/sindresorhus/execa/issues/433). We probably want to keep the default value here and let Execa decides what the best value for it is.

tests/utils/call-cli.js
```diff
@@ -2,8 +2,15 @@ const execa = require('execa')
...
+    windowsVerbatimArguments: true,
```
ehmicky, line 10: Is this needed? [The default behavior](https://github.com/sindresorhus/execa#windowsverbatimarguments) might be better at handling Windows.

ehmicky (approved):
Looks good.
I added two comments which are about options already present in the current code, not changed by this PR.

tests/utils/call-cli.js
erezrokah: Removed in https://github.com/netlify/cli/pull/1727/commits/130ee481e20c00e3d4c45f9fd3ceea59414b28e7

tests/utils/call-cli.js
erezrokah: Don't think it is. Removed in https://github.com/netlify/cli/pull/1727/commits/130ee481e20c00e3d4c45f9fd3ceea59414b28e7

Issue #1703, opened 2021-01-07
Test: fix the create react app test


erezrokah:
Related to https://github.com/netlify/cli/issues/404
See https://github.com/netlify/cli/runs/1664340078?check_suite_focus=true#step:7:370

This test has been randomly failing in our CI.

We should fix it.

erezrokah:
Closed via https://github.com/netlify/cli/pull/1795

PR #1733, opened 2021-01-14
test: use got instead of node-fetch

erezrokah:
Replaces the usage of `node-fetch` to `got` in our tests.
This allows:
1. Built in retries
2. Built in timeouts

Related to https://github.com/netlify/cli/issues/1703

### DIFF
--- a/tests/command.deploy.test.js
+++ b/tests/command.deploy.test.js
@@ -1,7 +1,6 @@
 const process = require('process')
 
 const test = require('ava')
-const fetch = require('node-fetch')
 const omit = require('omit.js').default
 
 const { supportsEdgeHandlers } = require('../src/lib/account')
@@ -9,22 +8,25 @@ const { getToken } = require('../src/utils/command')
 
 const callCli = require('./utils/call-cli')
 const { generateSiteName, createLiveTestSite } = require('./utils/create-live-test-site')
+const got = require('./utils/got')
 const { withSiteBuilder } = require('./utils/site-builder')
 
 const SITE_NAME = generateSiteName('netlify-test-deploy-')
 
 const validateContent = async ({ siteUrl, path, content, t }) => {
-  const response = await fetch(`${siteUrl}${path}`)
-  if (content === undefined) {
-    t.is(response.status, 404)
-    return
+  try {
+    const { body } = await got(`${siteUrl}${path}`)
+    t.is(body, content)
+  } catch (error) {
+    const {
+      response: { statusCode, statusMessage, body },
+    } = error
+    if (content === undefined) {
+      t.is(statusCode, 404)
+      return
+    }
+    throw new Error(`Failed getting content: ${statusCode} - ${statusMessage} - ${body}`)
   }
-  const actualContent = await response.text()
-  if (!response.ok) {
-    throw new Error(`Failed getting content: ${response.status} - ${response.statusText} - ${actualContent}`)
-  }
-
-  t.is(actualContent, content)
 }
 
 const validateDeploy = async ({ deploy, siteName, content, t }) => {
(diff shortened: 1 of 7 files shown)


tests/command.deploy.test.js
```diff
@@ -1,30 +1,32 @@
...
 const { withSiteBuilder } = require('./utils/site-builder')
 
 const SITE_NAME = generateSiteName('netlify-test-deploy-')
 
 const validateContent = async ({ siteUrl, path, content, t }) => {
-  const response = await fetch(`${siteUrl}${path}`)
-  if (content === undefined) {
-    t.is(response.status, 404)
-    return
+  try {
```
erezrokah, line 17: Some tests are expecting 404 so we need to catch that error

tests/command.deploy.test.js
```diff
@@ -330,12 +333,7 @@ if (process.env.IS_FORK !== 'true') {
     })
   })
 
-  test.after('cleanup', async (t) => {
-    const { siteId } = t.context
-    console.log(`deleting test site "${SITE_NAME}". ${siteId}`)
-    await callCli(['sites:delete', siteId, '--force'])
-  })
-  test('should exit with error when deploying an empty directory', async (t) => {
+  test.serial('should exit with error when deploying an empty directory', async (t) => {
```
erezrokah, line 336: This test was missing a `.serial`

tests/cra.test.js
```diff
@@ -31,86 +32,83 @@ test.after(async (t) => {
 
 test('homepage', async (t) => {
   const { url } = t.context.server
-  const response = await fetch(`${url}/`).then((res) => res.text())
+  const response = await got(`${url}/`).text()
 
-  t.regex(response, /Web site created using create-react-app/)
+  t.true(response.includes('Web site created using create-react-app'))
```
erezrokah, line 37: I don't really mind if we use `t.regex` or `string.includes` as long as we are consistent

tests/cra.test.js
```diff
@@ -31,86 +32,83 @@ test.after(async (t) => {
...
-  t.is(response.headers.get('location'), `${url}/otherthing.html`)
-  t.is(await response.text(), 'Redirecting to /otherthing.html')
+  t.is(statusCode, 301)
+  t.is(headers.location, `/otherthing.html`)
+  t.is(body, 'Redirecting to /otherthing.html')
 })
 
 test('normal rewrite', async (t) => {
   const { url } = t.context.server
+  // TODO: replace with got like the rest of the tests
```
erezrokah, line 72: This is the only test that fails when switching to `got` - it throws with a status code of 404.
I've tested with `fetch` and `curl` and it passes.

I'm not sure what is the issue (yet) but don't think this should block the PR.

tests/cra.test.js
```diff
@@ -31,86 +32,83 @@ test.after(async (t) => {
...
 test('functions rewrite echo without body', async (t) => {
   const { url, host, port } = t.context.server
-  const response = await fetch(`${url}/api/echo?ding=dong`).then((res) => res.json())
+  const response = await got(`${url}/api/echo?ding=dong`).json()
 
   t.is(response.body, undefined)
   t.deepEqual(response.headers, {
-    accept: '*/*',
-    'accept-encoding': 'gzip,deflate',
+    accept: 'application/json',
```
erezrokah, line 106: Some of these values can be take from the `got` request object.
I find it easier to have the expected values in plain strings.

tests/utils/got.js
```diff
@@ -0,0 +1,16 @@
+const got = require('got')
+
+const TIMEOUT = 1e4
+
+// Default got retry status code with the addition of 403
+// eslint-disable-next-line no-magic-numbers
+const STATUS_CODE = [403, 408, 413, 429, 500, 502, 503, 504, 521, 522, 524]
```
erezrokah, line 7: Some of the deploy tests are occasionally failing with 403:
https://github.com/netlify/cli/runs/1696253114?check_suite_focus=true#step:7:366

I'm not sure yet if this is an issue with our tests because of https://github.com/netlify/cli/pull/1733/files#r557434973.

I would rather to stabilize the tests first then look into it.

ehmicky (approved):

Issue #1665, opened 2020-12-16
chore: add release flow via PRs


erezrokah:
We should switch to https://github.com/google-github-actions/release-please-action like our [other repos](https://github.com/netlify/zip-it-and-ship-it/pull/266) to align the way we publish packages.

erezrokah:
What's blocking this at the moment is that we generate `docs` and commit them back to the repo on release:
https://github.com/netlify/cli/blob/9e7819b38e53ef99f0c44c1c60b0d62acaf5ec9d/package.json#L64

We should probably not commit autogenerated files to the repo and we're already generating them using Netlify:
https://github.com/netlify/cli/blob/9e7819b38e53ef99f0c44c1c60b0d62acaf5ec9d/netlify.toml#L3

Issue #304, opened 2019-04-24
Port CLI docs site to new static site tool


DavidWells:
Future enhancement for cli.netlify.com

`x0` is cool but won't scale to fit the needs of the ref site. So we'll need to put the site into more robust static site tool.

Something like gatsby / react static / phenomic would probably be a better fit.

Stretch goal: Use netlify marketing component library 🚀

danez:
Duplicate of #5516

PR #1745, opened 2021-01-15
chore: verify docs using a GitHub action

erezrokah:
Related to https://github.com/netlify/cli/issues/1665.

Solves the blocker for adding a release PR workflow by verifying docs are synced using a GitHub action.
Makes sure a change to a command has the relevant doc change (instead of pushing the docs to the default branch during publishing).

Useful until we do https://github.com/netlify/cli/issues/304


eduardoboucas (approved):
👏🏻

PR #1747, opened 2021-01-15
chore: update PR labeler

erezrokah:
Uses a matrix in the PR labeler action and updates the action to use the new repo path (we transferred it from `erezrokah` to `netlify`)


eduardoboucas (approved):
😍

PR #1752, opened 2021-01-15
feat: warn when project has netlify dir

erezrokah:
Related to https://github.com/netlify/product/issues/257#issuecomment-746446846 and https://github.com/netlify/build/pull/2175

See example log:
image

Not sure if we should avoid duplicating the logic by having `@netlify/conifg` or `@netlify/build` return it.
Since it's a temporary message I think we can live with it.

### DIFF
--- a/src/lib/deprecations.js
+++ b/src/lib/deprecations.js
@@ -1,0 +1,66 @@
+const path = require('path')
+const process = require('process')
+
+const pathExists = require('path-exists')
+const semverLessThan = require('semver/functions/lt')
+
+const { NETLIFYDEVWARN } = require('../utils/logo')
+
+const warnOnOldNodeVersion = ({ log, chalk }) => {
+  if (semverLessThan(process.version, '10.0.0')) {
+    log(
+      `${NETLIFYDEVWARN} ${chalk.bold('Netlify CLI')} will require ${chalk.magenta.bold(
+        'Node.js 10',
+      )} or greater soon. Please update your Node.js version.`,
+    )
+  }
+}
+
+const NETLIFY_DIR = 'netlify'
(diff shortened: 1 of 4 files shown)


src/lib/deprecations.js
```diff
@@ -0,0 +1,66 @@
+const path = require('path')
+const process = require('process')
+
+const pathExists = require('path-exists')
+const semverLessThan = require('semver/functions/lt')
+
+const { NETLIFYDEVWARN } = require('../utils/logo')
+
+const warnOnOldNodeVersion = ({ log, chalk }) => {
```
erezrokah, line 9: Going to remove this in another PR

src/lib/deprecations.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const NETLIFY_DIR = 'netlify'
+const DEFAULT_FUNCTIONS_SRC = path.join(NETLIFY_DIR, 'functions')
+
+const formatDirectory = (chalk, directory) => chalk.redBright(directory)
+
+const getCommunityLink = (chalk) =>
+  chalk.magenta(
+    'community.netlify.com/t/upcoming-change-netlify-functions-as-zero-config-default-folder-for-deploying-netlify-functions/28789',
```
eduardoboucas, line 26: Are we intentionally leaving out `https://`?

ehmicky (approved):
Yes, you're right, I should have been added this to `@netlify/config` instead of `@netlify/build`. 
Fortunately, this is only temporary as you mentioned.

src/lib/deprecations.js
erezrokah: The purpose was to keep the link shorter (and the browser redirects to `https`).
But you're right, better to add it.

PR #1756, opened 2021-01-15
test: generate only json coverage file in CI

erezrokah:
Removes `verbose` flag from `ava` and only keep the JSON coverage file in CI.

I'm hoping this will help with our CI jobs getting stuck during tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "test:init:cli-help": "npm run start -- --help",
     "test:init:install-deps": "npm ci --prefix tests/site-cra",
     "test:dev:ava": "ava --verbose",
-    "test:ci:ava": "nyc -r lcovonly -r text -r json ava --verbose",
+    "test:ci:ava": "nyc -r json ava",
     "docs": "node ./site/scripts/docs.js",
     "watch": "nyc --reporter=lcov ava --watch",
     "prepack": "oclif-dev manifest && npm prune --prod",


ehmicky (changes requested):
Istanbul [reporters](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-reports/README.md) are called after [instrumentation](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-instrument/README.md), so I don't think adding/removing reporters to nyc should impact the unit tests?

When it comes to the verbose mode, this also only impacts reporting, so I would also not think it should impact the unit tests?

Please let me know if I got something wrong.

erezrokah:
> Istanbul [reporters](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-reports/README.md) are called after [instrumentation](https://github.com/istanbuljs/istanbuljs/blob/master/packages/istanbul-lib-instrument/README.md), so I don't think adding/removing reporters to nyc should impact the unit tests?
> 
> When it comes to the verbose mode, this also only impacts reporting, so I would also not think it should impact the unit tests?
> 
> Please let me know if I got something wrong.

Sorry for not being clear with my explanation. When our tests get stuck we don't have [any access to the logs](https://github.com/netlify/cli/runs/1709102228?check_suite_focus=true). This seems like an issue with GitHub actions not piping the logs (I also tried downloading the the logs zip file and the tests logs are missing from the zip).
I'm hoping that by removing the `verbose` flag I could get more insights into this (if this is a buffering issue).
As for the Istanbul reporter, I'm trying to get the CI to only do the tasks it needs to do. Since we only use the JSON file here:
https://github.com/netlify/cli/blob/967ecdaa2db5bca363611c2282a044ce01d3b1ed/.github/workflows/main.yml#L50

TLDR: I still don't know what causes the CI tests to get stuck, so I'm trying helplessly to get more information :)

ehmicky (approved):
This makes sense :+1:

PR #1758, opened 2021-01-16
Fix links to api docs in `netlify api --list`

russelldavis:
**- Summary**
The anchor portion of the links had an extra slash which meant the page wouldn't scroll to the right section.

Fixed links to api docs in `netlify api --list`


ehmicky (approved):
Thanks @russelldavis!

PR #1765, opened 2021-01-18
Compute slug for site name suggestions when user does not have one

### LINKED ISSUE #1730: Suggested site names computed with a potentially null value
**Describe the bug**

When suggesting names for new sites, CLI is relying on [the `slug` property of the signed-in user](https://github.com/netlify/cli/blob/master/src/commands/sites/create.js#L55-L60), which is not guaranteed to exist. When it doesn't exist, the string `null` is used to compute the name suggestions.

**To Reproduce**

1. Create a Netlify account using the "sign in with email" option
2. Run `ntl login` to authenticate the CLI
3. Run `ntl init` in an empty directory

**CLI Output**

```
Choose a unique site name (e.g. the-great-null-site.netlify.app) or leave it blank for a random name. You can update the site name later.
```

eduardoboucas:
**- Summary**

When CLI offers suggestions for a new site name (e.g. as part of `ntl init`), it uses the user's `slug` property, which isn't guaranteed to exist. This PR addresses that by computing a slug in those cases, using the `full_name` property.

Closes #1730 

**- Test plan**

- Verified that the behaviour for a user that does have the `slug` property remains unchanged
- Verified that the behaviour for a user that does not have the `slug` property now includes a slug

fix: compute slug for site name suggestions when user does not have one

3-toed-sloth-TRR

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13145,7 +13145,20 @@
     "p-event": {
       "version": "4.2.0",
       "resolved": "https://registry.npmjs.org/p-event/-/p-event-4.2.0.tgz",
-      "integrity": "sha512-KXatOjCRXXkSePPb1Nbi0p0m+gQAwdlbhi4wQKJPI1HsMQS9g+Sqp2o+QHziPr7eYJyOZet836KoHEVM1mwOrQ=="
+      "integrity": "sha512-KXatOjCRXXkSePPb1Nbi0p0m+gQAwdlbhi4wQKJPI1HsMQS9g+Sqp2o+QHziPr7eYJyOZet836KoHEVM1mwOrQ==",
+      "requires": {
+        "p-timeout": "^3.1.0"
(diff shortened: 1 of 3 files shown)


erezrokah (approved):

eduardoboucas:
@erezrokah I've added https://github.com/netlify/cli/pull/1765/commits/dfdb8cca51386e81717e46ba6fe4050349590ff7, which computes the slug using `email` in case `full_name` isn't defined. If you're okay with that, I'm going to merge and release.

erezrokah:
> @erezrokah I've added [dfdb8cc](https://github.com/netlify/cli/commit/dfdb8cca51386e81717e46ba6fe4050349590ff7), which computes the slug using `email` in case `full_name` isn't defined. If you're okay with that, I'm going to merge and release.

I'm good with!

PR #1768, opened 2021-01-18
test(command-build): don't run on forked repos

erezrokah:
Build tests now require a valid auth token which is not available to forked PRs.
Fixes CI failures in https://github.com/netlify/cli/pull/1758

### DIFF
--- a/tests/command.build.test.js
+++ b/tests/command.build.test.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const process = require('process')
 
 const test = require('ava')
 const execa = require('execa')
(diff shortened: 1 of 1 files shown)


ehmicky (approved):

eduardoboucas:
What does this mean for external contributors that submit a PR? Are we able to run the tests "on our side" before merging?

erezrokah:
> What does this mean for external contributors that submit a PR? Are we able to run the tests "on our side" before merging?

We can run those tests locally and they'll run on the default branch post merge.
Contributors _should also_ run those tests locally, but we don't have a good way (yet) to enforce tests that require GitHub secrets.
I'm planning on adding a mock API server as a part of https://github.com/netlify/cli/pull/1489#discussion_r548092573 so maybe we could fallback to a mock API on forks?

eduardoboucas:
> maybe we could fallback to a mock API on forks?

Sounds good. Until we get there, perhaps there's something we can do internally to ensure we don't merge a PR from a forked repo without running the test suite locally, because going forward the PR status will be misleading and there's nothing stopping us from merging a PR that actually breaks the build.

erezrokah:
> Sounds good. Until we get there, perhaps there's something we can do internally to ensure we don't merge a PR from a forked repo without running the test suite locally, because going forward the PR status will be misleading and there's nothing stopping us from merging a PR that actually breaks the build.

Good point. Let's continue this discussion in https://github.com/netlify/cli/issues/1769

PR #1776, opened 2021-01-19
Improve robustness of netlify-lambda function-builder-detector

### LINKED ISSUE #1726: Netlify Function Builder Detector for netlify-lambda can detect wrong functions source folder
**Describe the bug**

Having an NPM script which uses `netlify-lambda` and options that come **after** the functions source folder will incorrectly detect the option as the function source.

**To Reproduce**

1. Set up a new Netlify repo with `netlify-lambda` as a dependency
2. Add an NPM script, e.g. `functions:build` that runs netlify `netlify-lambda build` and passes an argument **after** specifying the functions source folder, e.g. `netlify-lambda build functions/src -c ./functions/webpack.config.js`
3. Run `netlify dev`, which will pick up the `functions:build` script
4. Notice that Netlify Dev logs "building functions from directory **./functions/webpack.config.js**"
5. Changes in the function source won't be picked up either, only changes to `./functions/webpack.config.js`

**Expected behavior**

The functions source directory should be able to be detected regardless of which spot any options passed to `netlify-lambda` are passed

**Additional context**

As `netlify-lambda build` allows any options to be placed either before or after the functions source folder itself, that should also work when used together with Netlify Dev.

The problem should be quite easily fixed by modifying the regex used here 
https://github.com/netlify/cli/blob/305f8f4a92e030384dc58ae9d572a611880a7b7d/src/function-builder-detectors/netlify-lambda.js#L16

Happy to take a look at the regex and open a PR, just give me a shout

Robin-Hoodie:
**- Summary**

This PR fixes a few edge cases where the `netlify-lambda build` command would be detected, but would not run correctly

- The functions folder can be detected, regardless of the placement of any options. This fixes #1726
- Add a test for multiple `nelify-lambda` commands (though this was already working, this was not tested)
- Disable detection if functions folder is not specified, warn the user in this case
- Disable detection if multiple function folders were specified, warn the user in this code

**- Test plan**

You can view the added tests in the commits

Improve detection of invalid `netlify-lambda build` commands & detect functions folder regardless of options placement

My 2 cats when they were kittens:

image


src/function-builder-detectors/netlify-lambda.js
```diff
@@ -13,12 +14,20 @@ const detectNetlifyLambda = async function ({ dependencies, devDependencies, scr
...
 
-    const match = script.match(/netlify-lambda build.* (\S+)\s*$/)
-    if (match) {
-      const [, src] = match
-      settings.src = src
-      settings.npmScript = key
-      break
+    if (script.match(/netlify-lambda\s+build/)) {
+      // E.g. ["netlify-lambda", "build", "functions/folder"]
+      const match = minimist(script.split(' '))
```
erezrokah, line 19: This isn't whitespace safe (e.g. a functions folder with a whitespace will break this), but the current approach isn't either and this works for more use case.
🚀

erezrokah (approved):
Thanks for the fix @Robin-Hoodie. This looks good and sorry for the delay reviewing it

PR #1779, opened 2021-01-20
test: upgrade ava to v3

erezrokah:
Since we dropped Node.js 10 we should be able to use `ava v3`


package.json
```diff
@@ -193,8 +193,7 @@
     "failFast": false,
     "failWithoutAssertions": false,
     "tap": false,
-    "compileEnhancements": false,
```
erezrokah, line 196: Those are disabled by default now

ehmicky (approved):

Issue #1773, opened 2021-01-19
Use `@netlify/framework-info` when creating a site


ehmicky:
When creating a site (`netlify init` or `netlify sites:create`), we set the build command, publish directory and functions directory based on framework detection in the UI.

The [CLI does it too](https://github.com/netlify/cli/blob/425e2c6eeb6a3b589370a78407b18c53f4b850bc/src/utils/init/utils.js#L19), but not based on framework detection. We should add this, using `@netlify/framework-info`.

Related: https://github.com/netlify/cli/issues/1495 (for `@netlify/framework-info` + Netlify Dev)

PR #1784, opened 2021-01-20
feat(command-init): recommend framework specific settings on site creation

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1773

**- Test plan**

See flow when running from a `create-react-app` site:
image

recommend framework specific settings on site creation

image

### DIFF
--- a/src/utils/init/utils.js
+++ b/src/utils/init/utils.js
@@ -1,5 +1,6 @@
 const path = require('path')
 
+const { listFrameworks } = require('@netlify/framework-info')
 const chalk = require('chalk')
 const cleanDeep = require('clean-deep')
 const inquirer = require('inquirer')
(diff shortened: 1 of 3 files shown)


src/utils/init/utils.js
```diff
@@ -16,8 +17,26 @@ const normalizeDir = ({ siteRoot, dir, defaultValue }) => {
...
+  return {}
+}
+
+const getDefaultSettings = async ({ siteRoot, config }) => {
+  const { frameworkBuildCommand, frameworkBuildDir } = await getFrameworkDefaults({ siteRoot })
+  const {
+    command: defaultBuildCmd = frameworkBuildCommand,
+    publish: defaultBuildDir = frameworkBuildDir,
+    functions: defaultFunctionsDir,
+  } = config.build
```
ehmicky, line 39: Is it common for `config.build.command` or `config.build.publish` to be defined?
If so, would it make sense to compute the default value only when needed (for performance reason due to the time required to run `listFrameworks()`?

ehmicky (approved):

src/utils/init/utils.js
erezrokah: As always another great observation.
The call is redundant when running `npm init` and linking to an existing that has both of those settings configured in the UI or there is a `netlify.toml` with both of those configured.

I'm not sure how often those happen with a new repo, but I think we can keep the call to `listFrameworks` for 2 reasons:
1. Code simplicity
2. We'll need to run `listFrameworks` always soon to recommend the Next.js plugin

PR #1785, opened 2021-01-20
fix(dev): fix role based access tests

calavera:
Remove the public directory because the dev server doesn't know about it.


erezrokah (approved):
🚀

PR #1790, opened 2021-01-21
refactor: clean up netlify-lambda detector's code

erezrokah:
This is a follow up on https://github.com/netlify/cli/pull/1776 to make the detector's code more readable.

I didn't require it as a part of the review since it wasn't an issue with the PR (that made the code much better btw), but an issue with our codebase.


eduardoboucas (approved):
LGTM! 🚢

PR #1791, opened 2021-01-21
fix: pass NETLIFY_API_URL override to @netlify/config

eduardoboucas:
**- Summary**

The `NETLIFY_API_URL` is currently not being passed to `@netlify/config`, which means that the default API URL will always be used when making any config-related API call.

Additionally, it stops any API calls from being made as part of the `logout` (by unsetting the access token). This addresses an issue where an invalid token is used on every command (including `logout`), preventing the user from clearing it.

⚠️ **Requires https://github.com/netlify/build/pull/2190.**

fix: pass NETLIFY_API_URL override to @netlify/config

6118325_450_450_81393_0_fill_0_7f8cf6e0aefef4c047621d965a072e57


erezrokah (approved):
This looks good to me pending a release of `@netlify/config` 🚀

PR #1795, opened 2021-01-21
fix: make ntl dev wait for local server to be ready

### LINKED ISSUE #404: 'socket hang up' error thrown randomly

Bug

At points during development the following error occurs which stops the entire Netlify Dev stack:

```
.../node_modules/netlify-redirector/lib/redirects.js:116
      throw ex;
      ^

Error: socket hang up
    at createHangUpError (_http_client.js:323:15)
    at Socket.socketCloseListener (_http_client.js:364:25)
    at Socket.emit (events.js:194:15)
    at TCP._handle.close (net.js:600:12)
```

Looks like it was also reported on the community forum back in April - https://community.netlify.com/t/netlify-dev-regularly-crashes/551

I haven't been able see any common pattern for what causes this. 

Should not crash. 

**- Please mention your node.js, and operating system version.**

- Node: `v10.15.1`
- Using `yarn`
- Netlify CLI `2.12.0`

eduardoboucas:
**- Summary**

When running `ntl dev`, we wait for the port of the local server to open before launching the browser session. However, there is sometimes a delay between the port being open and the server actually being ready to receive requests. This means that the page we open may be blank and unable to recover without a refresh.

I'm able to reproduce this consistently with a fresh Gatsby site.

The solution proposed in this PR is to poll the local server with `HEAD` requests. Once one is completed successfully, we consider the server to be ready and only then launch the browser session.

Closes #404

fix: make ntl dev wait for local server to be ready

samoyed-outdoors-gallery-8-min

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -5,6 +5,8 @@ const process = require('process')
 const { flags: flagsLib } = require('@oclif/command')
 const boxen = require('boxen')
 const chalk = require('chalk')
+const fetch = require('node-fetch')
+const waitFor = require('p-wait-for')
 const StaticServer = require('static-server')
 const stripAnsiCc = require('strip-ansi-control-characters')
 const waitPort = require('wait-port')
@@ -20,6 +22,13 @@ const openBrowser = require('../../utils/open-browser')
 const { startProxy } = require('../../utils/proxy')
 const { startFunctionsServer } = require('../../utils/serve-functions')
 const { startForwardProxy } = require('../../utils/traffic-mesh')
+
+// 1 second
+const SERVER_POLL_INTERVAL = 1e3
+// 150 milliseconds
+const SERVER_POLL_REQUEST_TIMEOUT = 150
+// 30 seconds
+const SERVER_POLL_TIMEOUT = 3e4
 
 const startFrameworkServer = async function ({ settings, log, exit }) {
   if (settings.noCmd) {
@@ -82,7 +91,30 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
   })
 
   try {
-    const open = await waitPort({ port: settings.frameworkPort, output: 'silent', timeout: FRAMEWORK_PORT_TIMEOUT })
+    const open = await waitPort({
+      port: settings.frameworkPort,
+      output: 'silent',
+      timeout: FRAMEWORK_PORT_TIMEOUT,
+    })
+
+    const waitForServerToRespond = async () => {
+      try {
+        await fetch(`http://localhost:${settings.frameworkPort}`, {
+          method: 'HEAD',
+          timeout: SERVER_POLL_REQUEST_TIMEOUT,
+        })
+      } catch (_) {
+        return false
+      }
+
+      return true
+    }
+
+    await waitFor(waitForServerToRespond, {
+      interval: SERVER_POLL_INTERVAL,
+      timeout: SERVER_POLL_TIMEOUT,
+    })
+
     if (!open) {
       throw new Error(`Timed out waiting for port '${settings.frameworkPort}' to be open`)
     }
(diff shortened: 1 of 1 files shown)


src/commands/dev/index.js
```diff
@@ -82,7 +91,30 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
...
+          timeout: SERVER_POLL_REQUEST_TIMEOUT,
+        })
+      } catch (_) {
+        return false
+      }
+
+      return true
+    }
+
+    await waitFor(waitForServerToRespond, {
```
erezrokah, line 113: Can we skip this `waitFor` if `open` is `false`?

src/commands/dev/index.js
```diff
@@ -82,7 +91,30 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
...
+      port: settings.frameworkPort,
+      output: 'silent',
+      timeout: FRAMEWORK_PORT_TIMEOUT,
+    })
+
+    const waitForServerToRespond = async () => {
+      try {
+        await fetch(`http://localhost:${settings.frameworkPort}`, {
+          method: 'HEAD',
+          timeout: SERVER_POLL_REQUEST_TIMEOUT,
```
erezrokah, line 104: I'm not sure the `fetch` API allows passing a timeout.

We can consider using https://github.com/sindresorhus/got but that's out of scope for this PR.

erezrokah (changes requested):
Hi @eduardoboucas - I like the approach in this PR!
I have 2 questions/suggestions:
1. Can we verify a few other framework servers accept `HEAD` requests?
2. If `waitFor` times out and `open` is `true`, how about we print a warning instead of failing?

src/commands/dev/index.js
eduardoboucas: > Hi @eduardoboucas - I like the approach in this PR!
> I have 2 questions/suggestions:
> 
> 1. Can we verify a few other framework servers accept `HEAD` requests?
> 2. If `waitFor` times out and `open` is `true`, how about we print a warning instead of failing?

Both great suggestions, thanks @erezrokah!

1. I tested a few, happy to test a few more. Would you feel more comfortable if we used a `GET` request instead?
2. I was actually experimenting with this locally. Since you brought it up, we should definitely do it.

eduardoboucas:
> Hi @eduardoboucas - I like the approach in this PR!
> I have 2 questions/suggestions:
> 
> 1. Can we verify a few other framework servers accept `HEAD` requests?
> 2. If `waitFor` times out and `open` is `true`, how about we print a warning instead of failing?

Both great suggestions, thanks @erezrokah!

1. I tested a few, happy to test a few more. Would you feel more comfortable if we used a `GET` request instead?
2. I was actually experimenting with this locally. Since you brought it up, we should definitely do it.

src/commands/dev/index.js
eduardoboucas: Absolutely.

src/commands/dev/index.js
eduardoboucas: I think it exists in the version of `node-fetch` we're using, but it was [removed in version 3](https://github.com/node-fetch/node-fetch/blob/cbd4d895767e33491219e40013dd8daa5c7ac024/docs/v3-UPGRADE-GUIDE.md#the-timeout-option-was-removed). Definitely something we need to take into account if/when upgrading.

eduardoboucas:
I've addressed your suggestions in https://github.com/netlify/cli/pull/1795/commits/f85789365390c40ffe285d0aed5389c656eb9764. I also reduced the overall timeout significantly, so that we don't wait for too long if the server doesn't respond to the requests.

src/commands/dev/index.js
erezrokah: Oh that's very helpful to know thanks!

erezrokah (approved):

erezrokah:
> I've addressed your suggestions in [f857893](https://github.com/netlify/cli/commit/f85789365390c40ffe285d0aed5389c656eb9764). I also reduced the overall timeout significantly, so that we don't wait for too long if the server doesn't respond to the requests.

This looks great!

erezrokah:
Also closes https://github.com/netlify/cli/issues/1703

PR #1804, opened 2021-01-25
feat: remove warning for default functions directory

eduardoboucas:
**- Summary**

As part of https://github.com/netlify/build/issues/2016, we're updating the default functions directory to be `netlify/functions`, and therefore the current warning message is no longer relevant.

**- Test plan**

N/A

feat: remove warning for default functions directory

cute-baby-animals-31


eduardoboucas:
@erezrokah At first I thought it'd be a good idea to keep the deprecations file (even though we don't have any deprecation messages currently), to serve as a skeleton for future cases, but I'm not so sure now. What do you think?

erezrokah:
> @erezrokah At first I thought it'd be a good idea to keep the deprecations file (even though we don't have any deprecation messages currently), to serve as a skeleton for future cases, but I'm not so sure now. What do you think?

I'd say remove it - we can always add it back based on history.

eduardoboucas:
> I'd say remove it - we can always add it back based on history.

https://github.com/netlify/cli/pull/1804/commits/86a3b9d44174fffd23f24d7c660a7db0cd0a49a5 ✅

erezrokah (approved):
🚀

PR #1810, opened 2021-01-26
fix: recover from unauthorized error in @netlify/config call

eduardoboucas:
**- Summary**

Since #1722, CLI is making an API call for every command, even for the ones that don't require API data (e.g. `login` and `logout`).

As a consequence, users might get locked in their accounts if they have an invalid/revoked token. If they try to login or logout, CLI will still try to use their existing token and fail before the command they're trying to execute has a change to succeed.

This PR adds a temporary solution that re-tries a call to `@netlify/config` _without_  a token if a call with the existing token fails.

It also adds a message to the `status` command checking for an unauthorised call to `api.listAccountsForUser` and `api.getCurrentUser` (currently it only exists for `api.getSite`. We should abstract this behaviour, so that any time we retrieve any user data with an invalid token, we're able to display a helpful error message (and possibly log the user out automatically?).

fix: recover from unauthorized error in @netlify/config call

maxresdefault

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -127,7 +127,19 @@ class BaseCommand extends Command {
         offline: argv.offline,
       })
     } catch (error) {
-      const message = error.type === 'userError' ? error.message : error.stack
+      const isUserError = error.type === 'userError'
+
+      // If we're failing due to an error thrown by us, it might be because the token we're using is invalid.
+      // To account for that, we try to retrieve the config again, this time without a token, to avoid making
+      // any API calls.
+      //
+      // @todo Replace this with a mechanism for calling `resolveConfig` with more granularity (i.e. having
+      // the option to say that we don't need API data.)
+      if (isUserError && token) {
+        return this.getConfig(cwd, state)
+      }
+
+      const message = isUserError ? error.message : error.stack
       console.error(message)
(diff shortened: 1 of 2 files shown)


src/utils/command.js
```diff
@@ -127,7 +127,19 @@ class BaseCommand extends Command {
...
     } catch (error) {
-      const message = error.type === 'userError' ? error.message : error.stack
+      const isUserError = error.type === 'userError'
+
+      // If we're failing due to an error thrown by us, it might be because the token we're using is invalid.
+      // To account for that, we try to retrieve the config again, this time without a token, to avoid making
+      // any API calls.
+      //
+      // @todo Replace this with a mechanism for calling `resolveConfig` with more granularity (i.e. having
+      // the option to say that we don't need API data.)
```
ehmicky, line 137: Would the `offline` flag of `@netlify/config` work here?

eduardoboucas: Yes, that works too. Done in https://github.com/netlify/cli/pull/1810/commits/1c8cc6d1ffec91a27d47199d37022b551d35895d.

Thanks @ehmicky! 👊🏻

erezrokah (approved):

Issue #1707, opened 2021-01-08
on `ntl login` exit new sign up gracefully


tzmanics:
**Describe the bug**

when running `ntl login` if a user is signing up (in any form: third-party, email, etc.) there is a disconnect from the cli and the command is just left hanging (i.e. the command doesn't exit).

<img width="837" alt="Screen Shot 2021-01-08 at 6 02 01 PM" src="https://user-images.githubusercontent.com/3611928/104073447-b3227680-51db-11eb-950b-6c2e9244f52c.png">

**To Reproduce**

1. Run `ntl login`
2. Choose 'Sign in' from the opened website

**Expected behavior**

Quick fix: we could exit the command with a message like,
> "Process timed out. If you do not currently have a Netlify account, please create one at https://app.netlify.com/signup, then run `netlify login` again."

In the long run, it would be great to figure out a process to make the signup process work in the cli.

**CLI Output**

see image above

jlengstorf:
maybe we could check for `~/.netlify/config.json` and, if none is found, ask if they have an account?

the "no account" flow might look like this:

```
$ Do you have a Netlify account? [Y/n] no

Before we can log you in, you'll need to set up an account!
Opening https://app.netlify.com/signup...

Please run `netlify login` again once you've finished setting up your account.

$ # at this point the command has exited
```

erezrokah:
Thank you for opening the issue @tzmanics and for the suggestion @jlengstorf!

Could we try making the flow work when creating a new account?
e.g. have a UI state to redirect to the CLI authorization page once the account was created?

Regardless, we probably need to set an explicit  timeout here:
https://github.com/netlify/cli/blob/078a28bfc81e8a2630886e80f3c8528e3b9f4e70/src/utils/command.js#L242
as the default is 1 hour:
https://github.com/netlify/js-client/blob/05981e404f6067db743a6cf199da2e7742576284/src/index.js#L62

PR #1833, opened 2021-02-02
fix(command-login): reduce auth timeout, display spinner

erezrokah:
**- Summary**

Partially addresses https://github.com/netlify/cli/issues/1707 pending the long term solution in that issue.

**- Test plan**

Run `netlify logout` then `netlify login` and don't follow through with authentication

Timeout authorization after 5 minutes instead of 1 hour

image

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -50,7 +50,7 @@ const getToken = async (tokenFromFlag) => {
   return [null, 'not found']
 }
 
-// 5 Minuets
+// 5 Minutes
 const TOKEN_TIMEOUT = 3e5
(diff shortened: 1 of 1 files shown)


src/utils/command.js
```diff
@@ -49,6 +50,34 @@ const getToken = async (tokenFromFlag) => {
   return [null, 'not found']
 }
 
+// 5 Minuets
```
eduardoboucas, line 53:
```suggestion
// 5 Minutes
```

eduardoboucas (approved):

Issue #1832, opened 2021-02-02
Feature: recommend/auto-install build plugins when initializing a site


erezrokah:
**Is your feature request related to a problem? Please describe.**

When creating a new site via `ntl init` we should recommend/auto-install relevant plugins for that site

**Describe the solution you'd like**

`ntl init` should display a list of plugins the user can select from based on the detected framework

**Describe alternatives you've considered**

Users can install plugins manually 

**Additional context**

This is supported by `framework-info` (https://github.com/netlify/framework-info/issues/109)

Blocked by https://github.com/netlify/netlify-plugin-nextjs/issues/93

**Can you submit a pull request?**

Yes

PR #1836, opened 2021-02-02
feat(command-init): recommend build plugins

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1832 pending https://github.com/netlify/framework-info/issues/123 and https://github.com/netlify/netlify-plugin-nextjs/issues/93

**- Test plan**

image

Going to do some more testing once the blocking issues are completed

Recommend plugins during `ntl init` command and let the user choose if to install them

image

### TODO

- [x] Get site Node.js runtime version and pass to `framework-info`
- [x] Maybe use plugin names instead of package names?

### DIFF
--- a/src/utils/init/frameworks.js
+++ b/src/utils/init/frameworks.js
@@ -1,0 +1,23 @@
+const { listFrameworks } = require('@netlify/framework-info')
+
+const getFrameworkInfo = async ({ siteRoot, nodeVersion }) => {
+  const frameworks = await listFrameworks({ projectDir: siteRoot, nodeVersion })
+  if (frameworks.length !== 0) {
+    const [
+      {
+        name,
+        build: { directory, commands },
+        plugins,
+      },
+    ] = frameworks
+    return {
+      frameworkName: name,
+      frameworkBuildCommand: commands[0],
+      frameworkBuildDir: directory,
+      frameworkPlugins: plugins,
+    }
+  }
+  return {}
+}
+
+module.exports = { getFrameworkInfo }
--- a/src/utils/init/frameworks.js
+++ b/src/utils/init/frameworks.js
@@ -2,6 +2,7 @@ const { listFrameworks } = require('@netlify/framework-info')
 
 const getFrameworkInfo = async ({ siteRoot, nodeVersion }) => {
   const frameworks = await listFrameworks({ projectDir: siteRoot, nodeVersion })
+  // several frameworks can be detected - first one has highest priority
   if (frameworks.length !== 0) {
     const [
       {
--- a/src/utils/init/utils.js
+++ b/src/utils/init/utils.js
@@ -18,16 +18,22 @@ const normalizeDir = ({ siteRoot, dir, defaultValue }) => {
   return relativeDir || defaultValue
 }
 
-const getFrameworkDefaults = async ({ siteRoot }) => {
+const getFrameworkInfo = async ({ siteRoot }) => {
   const frameworks = await listFrameworks({ projectDir: siteRoot })
   if (frameworks.length !== 0) {
     const [
       {
+        name,
         build: { directory, commands },
         plugins,
       },
     ] = frameworks
-    return { frameworkBuildCommand: commands[0], frameworkBuildDir: directory, frameworkPlugins: plugins }
+    return {
+      frameworkTitle: name,
+      frameworkBuildCommand: commands[0],
+      frameworkBuildDir: directory,
+      frameworkPlugins: plugins,
+    }
   }
   return {}
 }
@@ -35,8 +41,7 @@ const getFrameworkDefaults = async ({ siteRoot }) => {
 const isPluginInstalled = (configPlugins, plugin) =>
   configPlugins.some(({ package: configPlugin }) => configPlugin === plugin)
 
-const getDefaultSettings = async ({ siteRoot, config }) => {
-  const { frameworkBuildCommand, frameworkBuildDir, frameworkPlugins } = await getFrameworkDefaults({ siteRoot })
+const getDefaultSettings = ({ siteRoot, config, frameworkPlugins, frameworkBuildCommand, frameworkBuildDir }) => {
   const recommendedPlugins = frameworkPlugins.filter((plugin) => !isPluginInstalled(config.plugins, plugin))
   const {
     command: defaultBuildCmd = frameworkBuildCommand,
@@ -52,7 +57,13 @@ const getDefaultSettings = async ({ siteRoot, config }) => {
   }
 }
 
-const getPromptInputs = ({ defaultBuildCmd, defaultBuildDir, defaultFunctionsDir, recommendedPlugins }) => {
+const getPromptInputs = ({
+  defaultBuildCmd,
+  defaultBuildDir,
+  defaultFunctionsDir,
+  recommendedPlugins,
+  frameworkTitle,
+}) => {
   const inputs = [
     {
       type: 'input',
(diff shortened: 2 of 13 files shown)


verythorough:
This is cool, @erezrokah! Is it possible to add an explanatory line before the question prompt?

erezrokah:
> This is cool, @erezrokah! Is it possible to add an explanatory line before the question prompt?

Great idea! Of course we can

verythorough:
> > This is cool, @erezrokah! Is it possible to add an explanatory line before the question prompt?
> 
> Great idea! Of course we can

🎉 @rstavchansky Can you work with @erezrokah to develop an appropriate message? Since this functionality will be used for multiple frameworks/plugins, I'm guessing this will need to be a more general message, like, "Based on the technology used in your site, we recommend the following plugin(s):"

rstavchansky:
After some [discussion in Slack](https://netlify.slack.com/archives/CBC2U9MMG/p1612543838002600) to understand the mode of detection, I'm making this recommendation:

**Single recommended plugin flow**
Seems like this is a {detected framework} site. 
Recommended Build Plugin: {list single plugin}
Install {single plugin name}?

**Multiple recommended plugins flow**
Seems like this is a {detected framework} site. 
Recommended Build Plugins: {list multiple plugins}
Which plugins to install?

One thing to consider - Is it possible to add a link to the plugin's readme so people can see what they're installing? (I supposed you’d have to get that link from the [plugins.json](https://github.com/netlify/plugins/blob/master/site/plugins.json) file, though, unless you could predict it programmatically.)

A note on capitalization from the [Style Guide](https://netlify-handbook.netlify.app/devex/docs-style-guide/?refresh=login#build-plugins%2C-plugins): Build Plugins is capitalized since it's a formal feature name. plugins is not.

erezrokah:
> One thing to consider - Is it possible to add a link to the plugin's readme so people can see what they're installing? (I supposed you’d have to get that link from the [plugins.json](https://github.com/netlify/plugins/blob/master/site/plugins.json) file, though, unless you could predict it programmatically.)

Sure we can add the link

erezrokah:
### Update

- Single plugin:
image

- Multiple plugins:
image


This is pending a decision if to use plugin package names or plugins names.
I also opened https://github.com/netlify/framework-info/issues/133 so we can show a pretty title of the framework.
Also I'm not sure where we can fit a README link (unless we shorten them, I think @eduardoboucas had that idea a while back).

verythorough:
> This is pending a decision if to use plugin package names or plugins names.

If the human-readable names are available, I think those are best.

erezrokah:
Latest version:
## Single plugin
image

## Multiple plugins
### Pre-selection
image

### Post-selection
image

tests/command.init.test.js
```diff
@@ -0,0 +1,283 @@
...
+    {
+      build_settings: {
+        functions_dir: functions,
+      },
+    },
+  ])
+}
+
+test('netlify init existing site', async (t) => {
+  const initQuestions = [
```
erezrokah, line 67: Figured it's time for us to be able to test the CLI input prompts. If you think of a better way to do this please let me know.

src/utils/init/frameworks.js
```diff
@@ -0,0 +1,23 @@
+const { listFrameworks } = require('@netlify/framework-info')
+
+const getFrameworkInfo = async ({ siteRoot, nodeVersion }) => {
+  const frameworks = await listFrameworks({ projectDir: siteRoot, nodeVersion })
+  if (frameworks.length !== 0) {
+    const [
```
JGAntunes, line 6: Are we only considering the first element of the array on purpose (i.e. the `framework-info` lib is only returning one framework for the time being)?

tests/command.init.test.js
JGAntunes: This is great! 🎉

JGAntunes (approved):
👍 🚀

src/utils/init/frameworks.js
erezrokah: It might detect a few frameworks - the first one has higher priority based on the order in https://github.com/netlify/framework-info/blob/32c58bb8de2f86874a61bdc7abab059c4c638f6c/src/frameworks/main.js#L4.

JGAntunes: I wouldn't mind having a comment here just highlighting that (we pick the first as its the highest priority). Not critical though, happy to move forward if you feel like it doesn't make sense @erezrokah 👍

erezrokah: That makes sense. I'll add it

erezrokah: Done in https://github.com/netlify/cli/pull/1836/commits/e1a410e10e1fc4390dc7c58ae8b786ec5caf9a54

erezrokah:
FYI, this is pending the release of https://github.com/netlify/netlify-plugin-nextjs/pull/94 (it was merged but not released).

iKristy:
Just a suggestion, and please show me the door if you think this is crazy or stupid, but wondering if some iconography (utilizing emojis) might help to grok this (since the message is a bit longer than the other one-liner prompts), as well kind of indent the text under the first line.

So... something like this, maybe?

Seems like this is a {detected framework} site. 
❇️ Recommended Build Plugin: {list single plugin}
➡️ Install {single plugin name}?

Again, totally a non-blocking suggestion! Feel free to ignore.

lindsaylevine:
> FYI, this is pending the release of [netlify/netlify-plugin-nextjs#94](https://github.com/netlify/netlify-plugin-nextjs/pull/94) (it was merged but not released). cc @lindsaylevine

released to npm, pending this PR to plugins https://github.com/netlify/plugins/pull/222

erezrokah:
> Again, totally a non-blocking suggestion! Feel free to ignore.

We can do that - that's a great suggestion thank you.

erezrokah:
New version per @iKristy suggestion:

### Single plugin
image

### Multiple plugins
image

JGAntunes (approved):

erezrokah:
@JGAntunes I added another commit to only install plugins to the UI (via API call) and not add them to `netlify.toml`.
When adding to `netlify.toml` the user should also add them to `package.json` and it's probably better to avoid editing `package.json` files.

Also, installing to the UI is the recommended way.

JGAntunes (approved):

JGAntunes:
Good call @erezrokah 👍 thanks for the explanation.

src/utils/init/utils.js
```diff
@@ -67,9 +64,72 @@ const getBuildSettings = async ({ siteRoot, config }) => {
...
+
+  const prefix = `Seems like this is a ${formatTitle(frameworkName)} site.${EOL}❇️  `
+  if (recommendedPlugins.length === 1) {
+    const { name } = getPluginInfo(pluginsList, recommendedPlugins[0])
+    return [
+      ...inputs,
+      {
+        type: 'confirm',
+        name: 'installSinglePlugin',
+        message: `${prefix}Recommended Build Plugin: ${formatTitle(`${name} plugin`)}${EOL}➡️  Install ${name} plugin?`,
```
verythorough, line 83:
```suggestion
        message: `${prefix}We're going to install this Build Plugin: ${formatTitle(`${name}`)}${EOL}➡️  OK to install?`,
```

src/utils/init/utils.js
```diff
@@ -67,9 +64,72 @@ const getBuildSettings = async ({ siteRoot, config }) => {
...
+  return [
+    ...inputs,
+    {
+      type: 'checkbox',
+      name: 'plugins',
+      message: `${prefix}Recommended Build Plugins: ${infos
+        .map(({ name }) => `${name} plugin`)
+        .map(formatTitle)
+        .join(', ')}${EOL}➡️  Which plugins to install?`,
+      choices: infos.map(({ name, package }) => ({ name: `${name} plugin`, value: package })),
```
verythorough, line 99:
```suggestion
      message: `${prefix}We're going to install these plugins: ${infos
        .map(({ name }) => `${name}`)
        .map(formatTitle)
        .join(', ')}${EOL}➡️  OK to install?`,
      choices: infos.map(({ name, package }) => ({ name: `${name}`, value: package })),
```

verythorough (changes requested):
@rstavchansky and I were chatting last week about how we might consider changing the language so that it _feels_ more like an auto-install, even though the behavior is the same. I've taken a stab at some wording for this, and would love feedback.

If we go this route, it might also be worthwhile to change the internal variables to replace `recommended` with something like `autoInstall`, particularly because it leaves room for us to later add a tier of "recommended" plugins, which are not quite so encouraged as the auto-install ones. (Though I suppose there's still room to do something like "suggested" for that lower tier, so we aren't totally boxed in if we stick with "recommended".)

verythorough (approved):
Thanks for your patience with this PR! 😛

PR #1777, opened 2021-01-20
fix(headers): add support for consistent * wildcard

ctjlewis:
### Summary
The CLI's current behavior is to only allow the asterisk wildcard `*` at the end of a rule. This PR aims to add support for wildcard which matches Netlify's deploy logic, which does not seem to have the same restriction.
 
1. See https://github.com/netlify/next-on-netlify/issues/151
2. Closes #1148

### Test plan

Tests were updated in `headers.test.js`. More robust testing should likely be added before merging.

### Changelog

```none
Previously, (*) wildcards had to appear at the end of a rule path, as in /path/to/dir/*. The CLI now supports the asterisk wildcard at any part of the rule, i.e. /test/*/test/*/test.
```

PR #1855, opened 2021-02-08
fix(command-dev): handle headers with colon in value

erezrokah:
Follow up on https://github.com/netlify/cli/pull/1777 to fix header values such as `https://www.example.com`

### DIFF
--- a/src/utils/headers.js
+++ b/src/utils/headers.js
@@ -67,6 +67,8 @@ const headersForPath = function (rules, path) {
   return pathObject
 }
 
+const HEADER_SEPARATOR = ':'
+
 const parseHeadersFile = function (filePath) {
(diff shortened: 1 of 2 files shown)


src/utils/headers.js
```diff
@@ -94,9 +96,9 @@ const parseHeadersFile = function (filePath) {
       throw new Error('path should come before headers')
     }
 
-    if (line.includes(':')) {
-      const [key = '', value = ''] = line.split(':', 2)
```
erezrokah, line 98: This is a port of our Ruby code. TIL `limit` in Ruby's split doesn't mean the same thing as `limit` in JavaScript's split.

eduardoboucas (approved):

PR #1860, opened 2021-02-09
test: allow configuring test account slug

erezrokah:
This is useful when the user belongs to multiple teams and the first one is not the one we want to test against.


JGAntunes (approved):
yolo

Issue #1663, opened 2020-12-16
Also update version numbers in package.json when updating dependencies


nikoladev:
**Is your feature request related to a problem? Please describe.**

The [latest update](https://github.com/netlify/cli/compare/v2.69.9...v2.69.10) has no changes for end users. Only the lockfile was updated, but this means nothing for end users unless the dependencies are also updated in `package.json`.

**Describe the solution you'd like**

To ensure that users benefit from updates to dependencies we should also update the dependency versions in `package.json`.

**Additional context**

> One key detail about `package-lock.json` is that it cannot be published

[Source](https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json)

**Can you submit a pull request?**

No. I've never used update bots.

nikoladev:
After a quick search it seems like `bumpVersion` might be the config option for this: https://docs.renovatebot.com/configuration-options/#bumpversion

PR #1876, opened 2021-02-14
chore(deps): update renovate to use  rangeStrategy bump

erezrokah:
Fixes https://github.com/netlify/cli/issues/1663

Testing this configuration so we don't get releases with just a package lock file change

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,6 +3,7 @@
   ignorePresets: [':prHourlyLimit2'],
   semanticCommits: true,
   dependencyDashboard: true,
+  rangeStrategy: 'bump',
   packageRules: [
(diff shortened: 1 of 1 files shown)


JGAntunes (approved):

JGAntunes:
If this ends up working accordingly it would be a good addition to the rest of our packages IMO 👍

erezrokah:
> If this ends up working accordingly it would be a good addition to the rest of our packages IMO

Yep, just testing it a bit more before I change https://github.com/netlify/renovate-config/blob/9b7e2be6bd9bef2fc8546dd136891cf1ce8c6ed9/default.json#L11

Issue #1873, opened 2021-02-12
Environment variables from the `context.production` toml config are being injected during netlify dev


askrzypczak:
**Describe the bug**

My environment variables from the `context.production` toml config are being injected during netlify dev

**Configuration**
```
[build]
  base = ""
  functions = "functions"
  command = "/bin/bash ./netlify-build.sh"
  environment = { GRIDSOME_DEPLOY_MODE = "development", NODE_VERSION="12", AWS_LAMBDA_JS_RUNTIME="nodejs12.x" }

[context.staging]
  environment = { GRIDSOME_DEPLOY_MODE = "staging", NODE_ENV="production" }

[context.feature-branch]
  environment = { GRIDSOME_DEPLOY_MODE = "feature-branch", NODE_ENV="production" }
  
[context.production]
  environment = { GRIDSOME_DEPLOY_MODE = "production", NODE_ENV="production" }
```

in my main.js i have the following: `console.debug("deploy mode: ", process.env.GRIDSOME_DEPLOY_MODE)`

(some background: gridsome includes any env var which is prefixed by 'GRIDSOME" in the client env, which is why you can console.log it)

the output is `deploy mode: production` Additionally, the webpack config files that I am using are listening to NODE_ENV and are packing for production instead of dev.

 - Did you run your build through the UI or the CLI?
 I used the CLI
- If using the CLI, which flags did you use?
no flags, just `netlify dev`
- If using the CLI, please enter the following command in a terminal and copy/paste its output:
```  System:
    OS: Windows 10 10.0.19042
    CPU: (6) x64 Intel(R) Core(TM) i5-8400 CPU @ 2.80GHz
    Memory: 7.07 GB / 15.92 GB
  Binaries:
    Node: 12.13.1 - C:\Program Files\nodejs\node.EXE    
    Yarn: 1.22.10 - C:\Projects\virtualship\node_modules\.bin\yarn.CMD
    npm: 6.12.1 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    netlify-cli: ^3.7.1 => 3.7.1
```

I just updated from `netlify-cli@2.70` to `netlify-cli@3.7.1`. 
Before the update the `GRIDSOME_DEPLOY_MODE` var was always undefined if you did not have it set in a `.env` (the `build.environment` was being ignored which was OK with me)

[...]

PR #1878, opened 2021-02-14
fix(command-dev): pass dev context to '@netlify/config'

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1873

**- Test plan**

Added a couple of tests

Pass `context: dev` to `@netlify/config` when running dev related commands. This should tell `@netlify/config` to resolve settings from the correct `context`.

image

### DIFF
--- a/tests/command.dev.test.js
+++ b/tests/command.dev.test.js
@@ -256,6 +256,58 @@ testMatrix.forEach(({ args }) => {
     })
   })
 
+  test(testName('[context.dev.environment] should override [build.environment]', args), async (t) => {
+    await withSiteBuilder('site-with-build-environment', async (builder) => {
+      builder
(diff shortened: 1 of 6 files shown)


tests/command.dev.test.js
```diff
@@ -333,7 +385,7 @@ testMatrix.forEach(({ args }) => {
 
       await builder.buildAsync()
 
-      await withDevServer({ cwd: builder.directory, env: { CONTEXT: 'production' }, args }, async (server) => {
+      await withDevServer({ cwd: builder.directory, args }, async (server) => {
```
erezrokah, line 388: I think we should allow overriding it via process level env, but can go either way

eduardoboucas (approved):
🎉

PR #1879, opened 2021-02-14
test(command-lm): add uninstall command

erezrokah:
Follow up to https://github.com/netlify/cli/pull/1489.

Adds a hidden command to undo the changes made by `ntl lm:install` (to be used by our tests).

### DIFF
--- a/src/utils/lm/install.js
+++ b/src/utils/lm/install.js
@@ -270,10 +270,7 @@ const shellVariables = function () {
   }
 }
 
-const uninstall = async function () {
-  await rmdirRecursiveAsync(getHelperPath())
-  await removeConfig(GIT_CONFIG, getGitConfigContent(getGitConfigPath()))
-
+const cleanupShell = async function () {
   try {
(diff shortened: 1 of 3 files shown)


src/utils/lm/install.js
```diff
@@ -181,6 +192,13 @@ const getCurrentCredentials = async () => {
   }
 }
 
+// Git expects the config path to always use / even on Windows
+const getGitConfigContent = (gitConfigPath) => `
+# This next lines include Netlify's Git Credential Helper configuration in your Git configuration.
+[include]
+  path = ${path.posix.normalize(gitConfigPath)}
```
ehmicky, line 199: This might not work as expected.
`path.posix.normalize()` actually returns backslashes on Windows (which is confusing).

Is the intent to only use forward slashes? If so, I have personally used this unmaintained package: https://github.com/jonschlinkert/unixify (we use it in `zip-it-and-ship-it`). If you find a better package, I'd be curious to learn about it too :)

ehmicky (approved):

PR #1887, opened 2021-02-15
fix(command-lm-install): fix gitconfig path in windows

erezrokah:
Follow up on https://github.com/netlify/cli/pull/1879#discussion_r576247092

@ehmicky we can't use `unixify` as it replaces drive names


ehmicky (commented):
Just to make sure: doesn't git support backward slashes in the config?

I'm relying on https://github.com/netlify/netlify-lm-plugin/blob/f129b0ccb90fefd2226cf1f468797f12708a2bf4/src/install.ts#L260 and also looking at my Windows VM `gitconfig` having Windows paths normalized.

Also https://stackoverflow.com/questions/28520565/what-is-the-syntax-for-specifying-windows-paths-in-gitconfig

ehmicky (approved):

PR #1901, opened 2021-02-17
fix: reduce package size

ehmicky:
This removes `src/functions-templates/**/node_modules` from the package published to npm.

Before: `305.8 KB` (packed), `1.3 MB` (unpacked)
After: `160.5 KB` (packed), `599.2 KB` (unpacked)


erezrokah:
This is 🚀  I'm curious to know how those ended up being installed on a dev machine

erezrokah (approved):

PR #1906, opened 2021-02-19
feat: rename trafficMesh flag to edgeHandlers

tiffanosaurus:
**- Summary**
The `--trafficMesh` flag is not a super clear term to users, let's go for `--edgeHandlers` which is more to the point. Deprecate the current flag in favor of a. new flag and add a warning to users that use the old one.

image

**- Test plan**

Choose your poison:
- `npm run test` (follow Contributing.md for instructions)
- `netlify dev --edgeHandlers`
- `netlify dev -eh`

Rename `netlify dev` flag `--trafficMesh` (short: `-t`) to `--edgeHandlers` (short: `-eh`) and update description to `activates the edge handlers runtime`.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -136,7 +136,7 @@ const FRAMEWORK_PORT_TIMEOUT = 6e5
 
 const startProxyServer = async ({ flags, settings, site, log, exit, addonsUrls }) => {
   let url
-  if (flags.trafficMesh) {
+  if (flags.edgeHandlers || flags.trafficMesh) {
     url = await startForwardProxy({
       port: settings.port,
       frameworkPort: settings.frameworkPort,
@@ -221,6 +221,12 @@ class DevCommand extends Command {
       ...(config.build.publish && { publish: config.build.publish }),
       ...config.dev,
       ...flags,
+    }
+
+    if (flags.trafficMesh) {
+      warn(
+        '--trafficMesh and -t are deprecated and will be removed in the near future. Please use --edgeHandlers or -e instead.',
+      )
     }
 
     await injectEnvVariables({ env: this.netlify.cachedConfig.env, log, site, warn })
@@ -311,10 +317,15 @@ DevCommand.flags = {
     char: 'l',
     description: 'start a public live session',
   }),
+  edgeHandlers: flagsLib.boolean({
+    char: 'e',
+    hidden: true,
+    description: 'activates the Edge Handlers runtime',
+  }),
   trafficMesh: flagsLib.boolean({
     char: 't',
     hidden: true,
-    description: 'uses Traffic Mesh for proxying requests',
+    description: '(DEPRECATED: use --edgeHandlers or -e instead) uses Traffic Mesh for proxying requests',
   }),
   locationDb: flagsLib.string({
     description: 'specify the path to a local GeoIP location database in MMDB format',
--- a/tests/command.dev.test.js
+++ b/tests/command.dev.test.js
@@ -32,7 +32,7 @@ const testMatrix = [
   { args: [] },
 
   // some tests are still failing with this enabled
-  // { args: ['--trafficMesh'] }
+  // { args: ['--edgeHandlers'] }
 ]
 
 const testName = (title, args) => (args.length <= 0 ? title : `${title} - ${args.join(' ')}`)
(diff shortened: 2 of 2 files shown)


tiffanosaurus:
I have this as `fix:` but maybe `chore:` is better? It is a breaking change even if it's hidden

erquhart:
Thanks @tiffanosaurus!

cc/ @erezrokah I believe this is okay if `--trafficMesh` is considered beta, but defer to you on the decision.

src/commands/dev/index.js
```diff
@@ -311,10 +311,10 @@ DevCommand.flags = {
     char: 'l',
     description: 'start a public live session',
   }),
-  trafficMesh: flagsLib.boolean({
-    char: 't',
+  edgeHandlers: flagsLib.boolean({
+    char: 'eh',
     hidden: true,
-    description: 'uses Traffic Mesh for proxying requests',
+    description: 'activates the edge handlers runtime',
```
erezrokah, line 317: I believe this should be `Edge Handlers` per our style guide cc @rstavchansky

erezrokah (changes requested):
Thanks @tiffanosaurus is there any issue we can discuss this?
To my knowledge the `trafficMesh` flag does more than add [Edge Handlers](https://www.netlify.com/products/edge/edge-handlers/) support (which allows running custom logic at the edge).
Should we should call it [`edge`](https://www.netlify.com/products/edge/) or `local-edge` or something similar.

Also, could we keep the old flag and add a warning message to users instructing them to switch to the new flag so they can self serve?
We can drop the old flag completely after a couple of weeks with the new one.

As for the PR type:
1. `chore:` is a change that isn't visible to users (doesn't even require a new release)
2. `fix:` is a bug fix (suggests a patch version)
3. `feat:` is a user facing change (suggests a minor version) - so I believe this should be a `feat:`

To declare a breaking change you can use `feat!:`, but this isn't once since:
1. The flag is hidden (like you mentioned).
2. This is considered Beta (again like you mentioned).

tiffanosaurus:
Thanks @erezrokah! The issue and discussion around how to handle the flag can be found here: https://github.com/netlify/traffic-mesh/issues/1517

We can definitively add a new flag instead, we figured since it was beta, we could be more aggressive with the approach and replace it.

src/commands/dev/index.js
```diff
@@ -311,10 +311,15 @@ DevCommand.flags = {
...
+  edgeHandlers: flagsLib.boolean({
+    char: 'eh',
+    hidden: true,
+    description: 'activates the Edge Handlers runtime',
+  }),
   trafficMesh: flagsLib.boolean({
     char: 't',
     hidden: true,
-    description: 'uses Traffic Mesh for proxying requests',
+    description: '(DEPRECATED: use --edgeHandlers or -eh instead) uses Traffic Mesh for proxying requests',
```
tiffanosaurus, line 322: Let me know if a different copy would be better here!

src/commands/dev/index.js
```diff
@@ -223,6 +223,12 @@ class DevCommand extends Command {
       ...flags,
     }
 
+    if (flags.trafficMesh) {
+      warn(
+        '--trafficMesh and -t are deprecated and will be removed in the near future. Please use --edgeHandlers or -eh instead.',
```
tiffanosaurus, line 228: Let me know if a different copy here would be better!

tiffanosaurus:
Updated the PR with the new flag and a deprecation notice for the `--trafficMesh`.

This might have more changes as discussed in the issue since this not only enables edge handlers, but also the new redirect engine and people might want to enable it without wanting to use edge handlers.

src/commands/dev/index.js
```diff
@@ -311,10 +317,15 @@ DevCommand.flags = {
     char: 'l',
     description: 'start a public live session',
   }),
+  edgeHandlers: flagsLib.boolean({
+    char: 'eh',
```
mraerino, line 321: it's common in the unix world to only have single-letter flags following a single-dash parameter so that you can combine multiple single letters in one param, e.g. `tar -xzf` which is short for `tar -x -z -f`. this kinda breaks with this convention and might leave users confused if there are isolated `-e` or `-h` (help?) flags?

multi-letter flags should always use double-dashes as a prefix imho

erezrokah:
Thank you for making these changes @tiffanosaurus. Can we finalize the discussion [here](https://github.com/netlify/traffic-mesh/issues/1517) before merging this? Also great point by @mraerino about the single letter convention.

tiffanosaurus:
Yep of course and will update for the letter convention

rstavchansky:
Heads up that the `--trafficMesh flag` is currently publicly documented. https://docs.netlify.com/routing/edge-handlers/configure-and-deploy/#local-development
@jacklyn-net for a docs update

erezrokah (approved):

erezrokah:
@tiffanosaurus when you merge this can you keep the commit message as `feat: rename trafficMesh flag to edgeHandlers (#1906)` and clear the commit description from any intermediate messages? 
That will ensure an automated version bump as a minor version.

Issue #1913, opened 2021-02-24
Integrate LaunchDarkly


ehmicky:
LaunchDarkly is integrated with most of our platform, but not this repository.
We should load LaunchDarkly flags in CLI commands.

Background: https://github.com/netlify/build/pull/2308#discussion_r582147957

charliegroll:
Food for thought: I know I've seen an issue floating around in the past with regards to privacy/network calls the cli makes. Just want to make sure we're cognizant of this and get our thoughts on paper. Do we disable LD in such cases and only use defaults?

PR #1914, opened 2021-02-25
feat: add environment variable for esbuild

eduardoboucas:
**- Summary**

With the rollout of esbuild (https://github.com/netlify/zip-it-and-ship-it/pull/339 and https://github.com/netlify/build/pull/2308), `zip-it-and-ship-it` will no longer look for a `NETLIFY_EXPERIMENTAL_ESBUILD` environment variable. This means that people who want to bundle locally using esbuild will lose their ability to opt-in.

Until we have #1913, there's no feature flag mechanism extended to the CLI, so this PR reuses the same environment variable that `zip-it-and-ship-it` was using. When it's detected, the correct feature flag (i.e. `buildbot_esbuild`) is passed along to Netlify Build.

feat: add environment variable for esbuild

shutterstock_132484832

### DIFF
--- a/src/lib/build.js
+++ b/src/lib/build.js
@@ -1,3 +1,5 @@
+const { env } = require('process')
+
 const build = require('@netlify/build')
 
 // We have already resolved the configuration using `@netlify/config`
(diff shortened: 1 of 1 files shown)


erezrokah (approved):
🚀

PR #1942, opened 2021-03-01
chore: use default rangeStrategy

erezrokah:
Use the default renovate strategy to avoid restricting ranges for non Netlify packages.

For Netlify packages we still use `bump` from https://github.com/netlify/renovate-config/blob/e7e66c5e8ec56923e0acfcb8b335564e86e7ff2d/default.json#L23

This should close all of theses PRs 
image

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -3,7 +3,6 @@
   ignorePresets: [':prHourlyLimit2'],
   semanticCommits: true,
   dependencyDashboard: true,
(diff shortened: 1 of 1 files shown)


erezrokah:
@JGAntunes tests are failing since we have some tests that are hitting the Netlify API and we've hit the rate limit

eduardoboucas (approved):
🤖 🤖 🤖 🤖 🤖 🤖

Issue #1279, opened 2020-09-24
Invalid Content-Type for Netlify Dev forms request


philiplindberg:
**Describe the bug**

When running Netlify Functions locally via `netlify dev` and making a `POST` request to a function endpoint, I'm getting the following console error:

```
Invalid Content-Type for Netlify Dev forms request
```

**To Reproduce**

1. Generate a new netlify function using `netlify functions:create hello-world`
2. Pick `[hello-world]` template
3. Serve newly created function locally with `netlify dev`
4. Make request to local function endpoint (using Postman, curl etc...)
   * Set method to `POST`
   * Add header: `Content-Type: applications/json`
5. Response will contain `Invalid Content-Type for Netlify Dev forms request` error

**Configuration**

Using `netlify-cli@2.64.1`.

netlilfy.toml:
```
[build]
  functions = "functions"
```

envinfo:
```
System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 183.36 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    Yarn: 1.22.5 - ~/.nvm/versions/node/v12.18.2/bin/yarn
    npm: 6.14.5 - ~/.nvm/versions/node/v12.18.2/bin/npm
```

**Expected behavior**

No error.

Xaconi:
+1

masonhipp:
Also getting this error

sandman18:
+1

jasperchess:
Also getting this error.

I've reverted back to using `netlify-lambda serve` for the time being

Dimotro:
I'm still experiencing exactly this error as described on:
netlify-cli/2.68.0 linux-x64 node-v14.15.0

ksnyde:
This is actually quite shocking. POST is critical to almost any RESTful API and from what I'm seeing I can send a request via GET, PUT, DELETE, etc. and it all works as expected. Using POST however returns the `Invalid Content-Type for Netlify Dev forms request` message to the node server console.

To this point, my testing has been within the `netlify dev` sandbox. Will try in a post-deploy environment soon too but for testing purposes it really needs to work in both environments anyway.

ksnyde:
I can validate that `POST` _does_ work once deployed but `netlify dev` is broken

[...]

PR #1949, opened 2021-03-01
fix(command-dev): call middleware next on invalid content-type

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/1279

**- Test plan**

Added a test case

Call the `next` function when an invalid request was sent to the functions form submission handler to avoid blocking forever.

image

### DIFF
--- a/src/utils/detect-server.js
+++ b/src/utils/detect-server.js
@@ -176,7 +176,7 @@ const serverSettings = async (devConfig, flags, projectDir, log) => {
   settings.jwtRolePath = devConfig.jwtRolePath || 'app_metadata.authorization.roles'
   settings.functions = devConfig.functions || settings.functions
   if (settings.functions) {
-    settings.functionsPort = await getPort({ port: settings.functionsPort || 0 })
+    settings.functionsPort = await getPort({ port: devConfig.functionsPort || 0 })
   }
 
   return settings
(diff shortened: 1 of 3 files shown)


src/utils/detect-server.js
```diff
@@ -176,7 +176,7 @@ const serverSettings = async (devConfig, flags, projectDir, log) => {
   settings.jwtRolePath = devConfig.jwtRolePath || 'app_metadata.authorization.roles'
   settings.functions = devConfig.functions || settings.functions
   if (settings.functions) {
-    settings.functionsPort = await getPort({ port: settings.functionsPort || 0 })
+    settings.functionsPort = await getPort({ port: devConfig.functionsPort || 0 })
```
erezrokah, line 179: `settings.functionsPort` is always `undefined` since `settings` is returned from the framework detectors.
`devConfig.functionsPort` can be overridden in the `dev` block of `netlify.toml`

src/utils/serve-functions.js
```diff
@@ -305,10 +305,12 @@ const createFormSubmissionHandler = function ({ siteUrl }) {
           })
         })
       } catch (error) {
-        return console.error(error)
+        warn(error)
```
erezrokah, line 308: Both these error cases didn't call `next` so the server hanged instead of returning `404`

eduardoboucas (approved):

Issue #1957, opened 2021-03-04
CLI can't make a functions folder more than one level deep


cfjedimaster:
**Describe the bug**

I specified `netlify/functions` as my functions folder in `netlify.toml`. When the CLI tried to create it, it failed:

```
ray:~/projects/tempfornetlify$ ntl functions:create test1
◈ functions folder /home/ray/projects/tempfornetlify/netlify/functions specified in netlify.toml but folder not found, creating it...
    Error: ENOENT: no such file or directory, mkdir '/home/ray/projects/tempfornetlify/netlify/functions'
    Code: ENOENT
```

I assumed because it was multiple levels deep. I created `netlify` as a subdirectory, ran it again and it made the `functions` folder.

**To Reproduce**

Specify a functions folder that is 2+ levels deep and does not exist.

**Expected behavior**

It should intelligently create any folders "along the path" down to the end.

erezrokah:
Closed via https://github.com/netlify/cli/pull/1996

PR #1996, opened 2021-03-15
feat: add functions directory creation when non-existent in netlify.toml

### LINKED ISSUE #1822: Prompt the user for function directory on function:create

Currently, if no directory is specified for functions, you get the error below. The user is required to go to the docs to learn how to configure the setting in netlify.toml.
```
mysite % netlify functions:create
◈ No functions folder specified in netlify.toml
```

```
mysite % netlify functions:create
◈ functions folder not specified in netlify.toml
? Enter the path where your functions should live: netlify/functions
◈ netlify/functions folder not found, creating it...
? Pick a template 
...
```

- Hint text -  `netlify/functions`
- Once specified, auto-populate the path into `netlify.toml`

**Alternate solution**
Since we're actively doing the work to make `netlify/functions` the default functions directory, we could create it automatically if it doesn't exist and no path is specified in `netlify.toml`.

**Additional context**

This will also help with https://github.com/netlify/netlify-react-ui/issues/6943

## Success Metric
https://github.com/netlify/data/issues/1061

tiffanosaurus:
**- Summary**

Currently, if no directory is specified for functions in the user's `netlify.toml`, you get an error. The user is required to go to the docs to learn how to configure the setting in netlify.toml. This PR creates a directory for the user with a default to `netlify/functions`

Also updates instances of `folder` to `directory` in the functions command.

Closes #1822, #1957 

**- Test plan**

Run `netlify functions:create`

With a functions directory already existing:
image

Without a functions directory:
image

### DIFF
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -211,9 +211,7 @@ const ensureFunctionDirExists = async function (context) {
     log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or UI settings`)
 
     if (!siteId) {
-      context.error(
-        `${NETLIFYDEVERR} No site id found, please run inside a site directory or \`netlify link\``,
-      )
+      context.error(`${NETLIFYDEVERR} No site id found, please run inside a site directory or \`netlify link\``)
     }
 
     const { functionsDir } = await inquirer.prompt([
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -225,18 +225,18 @@ const ensureFunctionDirExists = async function () {
   }
 
   if (!functionsDirHolder) {
-    log(`${NETLIFYDEVLOG} functions folder not specified in netlify.toml or app settings`)
+    log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or UI settings`)
     const { functionsDir } = await inquirer.prompt([
       {
         type: 'input',
         name: 'functionsDir',
-        message: 'Enter the path where your functions should live:',
+        message: 'Enter the path, relative to your site’s base directory in your repository, where your functions should live:',
         default: 'netlify/functions',
       },
     ])
 
     try {
-      log(`${NETLIFYDEVLOG} updating app build settings with ${chalk.magenta.inverse(functionsDirHolder)}`)
+      log(`${NETLIFYDEVLOG} updating site settings with ${chalk.magenta.inverse(functionsDirHolder)}`)
       await api.updateSite({
         siteId,
         body: {
@@ -246,7 +246,7 @@ const ensureFunctionDirExists = async function () {
         },
       })
       log(
-        `${NETLIFYDEVLOG} functions folder ${chalk.magenta.inverse(functionsDirHolder)} updated in app build settings`,
+        `${NETLIFYDEVLOG} functions directory ${chalk.magenta.inverse(functionsDirHolder)} updated in site settings`,
       )
     } catch (error) {
       error('Error updating site settings')
@@ -257,12 +257,12 @@ const ensureFunctionDirExists = async function () {
 
   if (!fs.existsSync(functionsDirHolder)) {
     log(
-      `${NETLIFYDEVLOG} functions folder ${chalk.magenta.inverse(
+      `${NETLIFYDEVLOG} functions directory ${chalk.magenta.inverse(
         functionsDirHolder,
       )} does not exist yet, creating it...`,
     )
     fs.mkdirSync(functionsDirHolder, { recursive: true })
-    log(`${NETLIFYDEVLOG} functions folder ${chalk.magenta.inverse(functionsDirHolder)} created`)
+    log(`${NETLIFYDEVLOG} functions directory ${chalk.magenta.inverse(functionsDirHolder)} created`)
   }
 
   return functionsDirHolder
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -211,7 +211,9 @@ const ensureFunctionDirExists = async function (context) {
     log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or UI settings`)
 
     if (!siteId) {
-      context.log(`${NETLIFYDEVERR} Unable to find a site id. Please make sure you are in a directory that is linked to a site with Netlify.`)
+      context.log(
+        `${NETLIFYDEVERR} Unable to find a site id. Please make sure you are in a directory that is linked to a site with Netlify.`,
+      )
       process.exit(1)
     }
 
--- a/src/commands/functions/create.js
+++ b/src/commands/functions/create.js
@@ -33,8 +33,7 @@ const templatesDir = path.resolve(__dirname, '../../functions-templates')
 class FunctionsCreateCommand extends Command {
   async run() {
     const { flags, args } = this.parse(FunctionsCreateCommand)
-    const { config } = this.netlify
-    const functionsDir = await ensureFunctionDirExists(this, flags, config)
+    const functionsDir = await ensureFunctionDirExists()
 
     /* either download from URL or scaffold from template */
     const mainFunc = flags.url ? downloadFromURL : scaffoldFromTemplate
@@ -197,11 +196,36 @@ const pickTemplate = async function () {
 
 const DEFAULT_PRIORITY = 999
 
-/* get functions dir (and make it if necessary) */
-const ensureFunctionDirExists = async function (context, flags, config) {
+/**
+ * Get functions dir (and make it if necessary)
+ * @this FunctionsCreateCommand
+ * @returns {string | never} - functions directory or throws an error
+ */
+const ensureFunctionDirExists = async function () {
+  const { api, config, site } = this.netlify
+  const { log } = this
   let functionsDirHolder = config.functionsDirectory
+  let siteData = null
+  const siteId = site.id
+
+  // try to find functions_dir in site settings if not in user's netlify config
   if (!functionsDirHolder) {
-    context.log(`${NETLIFYDEVLOG} functions folder not specified in netlify.toml or UI settings`)
+    try {
+      siteData = await api.getSite({ siteId })
+    } catch (error) {
+      if (error.status === 404) {
+        error('Site not found')
+      } else {
+        error(error.message)
+      }
+    }
+
+    // https://open-api.netlify.com/#operation/getSite
+    functionsDirHolder = siteData?.build_settings?.functions_dir
+  }
+
+  if (!functionsDirHolder) {
+    log(`${NETLIFYDEVLOG} functions folder not specified in netlify.toml or app settings`)
     const { functionsDir } = await inquirer.prompt([
       {
         type: 'input',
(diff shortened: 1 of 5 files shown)


src/commands/functions/create.js
```diff
@@ -197,22 +197,32 @@ const pickTemplate = async function () {
...
+    const { functionsDir } = await inquirer.prompt([
+      {
+        type: 'input',
+        name: 'functionsDir',
+        message: 'Enter the path where your functions should live: netlify/functions:',
+        default: 'netlify/functions',
+      },
+    ])
+
+    functionsDirHolder = functionsDir
```
erezrokah, line 213: Should we update the user's site settings with the value retrieved from the user?
If the user choses `my_netlify_functions` as the functions directory how will our build system know about it?

src/commands/functions/create.js
```diff
@@ -197,22 +197,32 @@ const pickTemplate = async function () {
...
-    process.exit(1)
+const ensureFunctionDirExists = async function (context, flags, config) {
+  let functionsDirHolder = config.build && config.build.functions
+  if (!functionsDirHolder) {
+    context.log(`${NETLIFYDEVLOG} functions folder not specified in netlify.toml or UI settings`)
+    const { functionsDir } = await inquirer.prompt([
+      {
+        type: 'input',
+        name: 'functionsDir',
+        message: 'Enter the path where your functions should live: netlify/functions:',
```
erezrokah, line 208:
```suggestion
        message: 'Enter the path where your functions should live:',
```

Per you screenshot it seems we don't need to print `netlify/functions` twice.

src/commands/functions/create.js
```diff
@@ -197,22 +197,32 @@ const pickTemplate = async function () {
...
+    functionsDirHolder = functionsDir
   }
-  if (!fs.existsSync(functionsDir)) {
+
+  if (!fs.existsSync(functionsDirHolder)) {
     context.log(
       `${NETLIFYDEVLOG} functions folder ${chalk.magenta.inverse(
-        functionsDir,
+        functionsDirHolder,
       )} specified in netlify.toml but folder not found, creating it...`,
```
erezrokah, line 220: It seems that if the functions directory is not configured we first print `functions folder not specified in netlify.toml` and later on print `<functions-folder> specified in netlify.toml but folder not found`.
Should we change the latter message in that case?

erezrokah (changes requested):
Thanks @tiffanosaurus. not sure if this is ready yet, but I added some comments.

Also I think this fixes https://github.com/netlify/cli/issues/1957 too

src/commands/functions/create.js
tiffanosaurus: `netlify/functions` and `functions` directories should be detected by bitballoon and added as the site's `functions_dir`, but to ensure it's in the CLI too, I'm going to update the `netlify.toml` to have the new functions directory.

erezrokah: > I'm going to update the `netlify.toml` to have the new functions directory.

Please consider that some sites don't have a `netlify.toml`. In that case it might be better to use an `api` call to update the site settings.

If the functions directory is the default one `netlify/functions` that won't be necessary as you said.

src/commands/functions/create.js
tiffanosaurus: Yeah cc @klavavej, I think we can put `Creating functions folder `${functionsDirHolder}...`

src/commands/functions/create.js
tiffanosaurus: Ahh, got it! I'll do that then, thanks 😊

tiffanosaurus: Seems like it might make sense after this to add a `functions:set` command that updates the functions directory

tiffanosaurus: nvm we already have that :P

tiffanosaurus:
Updated to call the API, adding tests now. I don't see existing tests for testing user input so I might need to add an inquirer mock library.

erezrokah:
> I don't see existing tests for testing user input so I might add an inquirer mock library.

We've introduced such tests in https://github.com/netlify/cli/pull/1836/files#diff-d9344dcdb5e545ca0abd67faea6550ebcab2739e7d489055daa28621f79c2281R150.

Using a mock is a bit challenging since we launch the CLI as a separate process in our tests.

src/commands/functions/create.js
```diff
@@ -196,23 +195,76 @@ const pickTemplate = async function () {
...
+ */
+const ensureFunctionDirExists = async function () {
+  const { api, config, site } = this.netlify
+  const { log } = this
+  let functionsDirHolder = config.build && config.build.functions
+  let siteData = null
+  const siteId = site.id
+
+  // try to find functions_dir in site settings if not in user's netlify config
+  if (!functionsDirHolder) {
```
erezrokah, line 211: I don't think this part is required as we already retrieve site information when resolving the configuration in the `Command` base class.

src/commands/functions/create.js
```diff
@@ -196,23 +195,76 @@ const pickTemplate = async function () {
...
+        name: 'functionsDir',
+        message: 'Enter the path where your functions should live:',
+        default: 'netlify/functions',
+      },
+    ])
+
+    try {
+      log(`${NETLIFYDEVLOG} updating app build settings with ${chalk.magenta.inverse(functionsDirHolder)}`)
+      await api.updateSite({
+        siteId,
```
erezrokah, line 240: `siteId` is not guaranteed to exist, for example when a user hasn't linked the directory to a site yet.
Should we check that use case and exit with a relevant error?

erezrokah (changes requested):
Thanks @tiffanosaurus looking great.
Two smaller comments. Please let me know what you think

src/commands/functions/create.js
```diff
@@ -196,23 +195,76 @@ const pickTemplate = async function () {
...
+    }
+
+    // https://open-api.netlify.com/#operation/getSite
+    functionsDirHolder = siteData?.build_settings?.functions_dir
   }
-  if (!fs.existsSync(functionsDir)) {
-    context.log(
+
+  if (!functionsDirHolder) {
+    log(`${NETLIFYDEVLOG} functions folder not specified in netlify.toml or app settings`)
```
klavavej, line 227: [pebble] Per the [writing style guide](https://netlify-handbook.netlify.app/devex/docs-style-guide/?refresh=login#directory%2C-folder), we generally prefer the term `directory` over `folder`. Another reason to use `directory` here is to match how the corresponding UI input is labeled.


```suggestion
    log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or app settings`)
```

src/commands/functions/create.js
```diff
@@ -196,23 +195,76 @@ const pickTemplate = async function () {
...
+      {
+        type: 'input',
+        name: 'functionsDir',
+        message: 'Enter the path where your functions should live:',
+        default: 'netlify/functions',
+      },
+    ])
+
+    try {
+      log(`${NETLIFYDEVLOG} updating app build settings with ${chalk.magenta.inverse(functionsDirHolder)}`)
```
klavavej, line 238: [pebble] I recommend changing `app build settings` to `site settings` to better reflect where the Functions directory setting is in the information architecture of the app UI.

```suggestion
      log(`${NETLIFYDEVLOG} updating site settings with ${chalk.magenta.inverse(functionsDirHolder)}`)
```

klavavej (approved):
Thanks for adding me as a reviewer @tiffanosaurus!

Left a few pebble/sand comments, but nothing blocking so I'm approving. 

One note that doesn't fit in the diff:
[pebble] I'm recommending changing several instances of the term "folder" to "directory" to adhere to the writing style guide and to match the language we use in the UI. There's one instance outside of the diff that I can't suggest changes on that I think should be similarly changed for consistency. This is on line 52

> description: 'name of your new function file inside your functions folder',

src/commands/functions/create.js
tiffanosaurus: Removed the one where we retrieve site info thanks!

src/commands/functions/create.js
```diff
@@ -196,30 +195,78 @@ const pickTemplate = async function () {
...
+  let functionsDirHolder = config.build && config.build.functions
+
+  if (!functionsDirHolder) {
+    log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or UI settings`)
+
+    if (!siteId) {
+      context.log(
+        `${NETLIFYDEVERR} Unable to find a site id. Please make sure you are in a directory that is linked to a site with Netlify.`,
+      )
+      process.exit(1)
```
tiffanosaurus, line 216: Wasn't totally sure if I should throw an error or use `NETLIFYDEVERR`

src/commands/functions/create.js
```diff
@@ -196,30 +195,78 @@ const pickTemplate = async function () {
...
+  const siteId = site.id
+  const { log } = context
+  let functionsDirHolder = config.build && config.build.functions
+
+  if (!functionsDirHolder) {
+    log(`${NETLIFYDEVLOG} functions directory not specified in netlify.toml or UI settings`)
+
+    if (!siteId) {
+      context.log(
+        `${NETLIFYDEVERR} Unable to find a site id. Please make sure you are in a directory that is linked to a site with Netlify.`,
```
tiffanosaurus, line 214: @klavavej This is new copy, please let me know what you think!

src/commands/functions/create.js
tiffanosaurus: I think we should update it if the user is explicitly entering it, it might be confusing for them to see the functions directory as "not set" if they set it as the default.

tests/command.functions.test.js
```diff
@@ -28,4 +45,70 @@ test('should return function response when invoked', async (t) => {
     })
   })
 })
+
+test('should create a new function directory when none is found', async (t) => {
+  const { binPath } = t.context
+
+  const initQuestions = [
```
erezrokah, line 52: Should we rename this variable to `createFunctionsQuestions`?

erezrokah (changes requested):
Thanks @tiffanosaurus again for following up.

I made a few comments on the code and also some suggested improvements to the tests.

I think we might need to change the flow for creating the directory since we support putting functions inside `netlify/functions`.

src/commands/functions/create.js
klavavej: Thanks for the ping @tiffanosaurus !

[pebble] We already have a few variants of this error message in the CLI, I'm reluctant to introduce a new variant so I suggest using one of the existing ones. 

Here's a suggestion to change this to [the most common similar error message](https://github.com/netlify/cli/search?q=%22No+site+id+found%2C+please+run+inside+a+site+folder+or+%60netlify+link%60%22)

```suggestion
        `${NETLIFYDEVERR} No site id found, please run inside a site folder or `netlify link``,
```

Other similar error messages include 

> Could not find the site ID. Please run netlify link.

> Folder is not linked to a Netlify site. Run 'netlify link' to link it

But those are used far less.

klavavej (approved):
Thanks for committing my suggestions @tiffanosaurus. A couple pebbles still but nothing blocking so I'm approving. 

I didn't see any edit in response to the following comment of mine from my initial review, I'm not sure if that's because you skipped it on accident or if you're intentionally not making this change for some reason. So, I wanted to highlight this note again. Please let me know what you think

   - One note that doesn't fit in the diff:   
   [pebble] I'm recommending changing several instances of the term "folder" to "directory" to adhere to the writing style guide and to match the language we use in the UI. There's one instance outside of the diff that I can't suggest changes on that I think should be similarly changed for consistency. This is on line 52

      > description: 'name of your new function file inside your functions folder',

src/commands/functions/create.js
```diff
@@ -196,30 +195,78 @@ const pickTemplate = async function () {
...
+      )
+      process.exit(1)
+    }
+
+    const { functionsDir } = await inquirer.prompt([
```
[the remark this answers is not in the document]
erezrokah: Cool, that makes sense

Issue #1891, opened 2021-02-16
chore: enable lint rules in `.eslintrc.js` and fix errors


erezrokah:
We have a bunch of `eslint` rules disabled in out configuration https://github.com/netlify/cli/blob/f9ec20633cd2aa22f25f19f4112b46f80562d69a/.eslintrc.js#L8

We should enable those one by one and fix related errors. This is an envelope issue for related PRs.

### TODO
- [ ] `class-methods-use-this`
- [ ] `complexity`
- [ ] `max-depth`
- [ ] `max-lines`
- [ ] `max-lines-per-function`
- [ ] `max-nested-callbacks`
- [ ] `max-statements`
- [ ] `no-param-reassign`
- [ ] `no-process-exit`
- [ ] `fp/no-class`
- [ ] `fp/no-delete`
- [ ] `fp/no-get-set`
- [ ] `fp/no-let`
- [ ] `fp/no-loops`
- [ ] `fp/no-mutating-assign`
- [ ] `fp/no-mutating-methods`
- [ ] `fp/no-mutation`
- [ ] `fp/no-proxy`
- [ ] `fp/no-this`
- [ ] `import/max-dependencies`
- [ ] `node/no-sync`
- [x] `promise/catch-or-return`
- [x] `promise/no-callback-in-promise`
- [x] `promise/no-nesting`
- [x] `promise/no-return-wrap`
- [x] `promise/prefer-await-to-callbacks`
- [x] `promise/prefer-await-to-then`
- [x] `you-dont-need-lodash-underscore/get`
- [ ] `unicorn/prefer-spread`
- [ ] `unicorn/consistent-destructuring`
- [x] `unicorn/no-array-push-push`
- [ ] `unicorn/filename-case`

ehmicky:
Related: https://github.com/netlify/cli/issues/1437

tinfoil-knight:
@erezrokah Can I pick up the first one?

ehmicky:
@tinfoil-knight Please feel free to pick any of these, including `class-methods-use-this`. Thanks a lot for your help! :heart:

tinfoil-knight:
I'm picking up `fp/no-loops` while `class-methods-use-this` is in process.

PR #2007, opened 2021-03-16
refactor: enable and fix promise rules

erezrokah:
Related to https://github.com/netlify/cli/issues/1891

It seems `prefer-await-to-callbacks` relies too much on argument names.

### DIFF
--- a/src/functions-templates/js/using-middleware/using-middleware.js
+++ b/src/functions-templates/js/using-middleware/using-middleware.js
@@ -2,16 +2,16 @@ const middy = require('middy')
 const { jsonBodyParser, validator, httpErrorHandler, httpHeaderNormalizer } = require('middy/middlewares')
 
 /* Normal lambda code */
-const businessLogic = (event, context, callback) => {
+const businessLogic = async (event) => {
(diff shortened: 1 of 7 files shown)


src/functions-templates/js/using-middleware/using-middleware.js
```diff
@@ -2,16 +2,16 @@ const middy = require('middy')
 const { jsonBodyParser, validator, httpErrorHandler, httpHeaderNormalizer } = require('middy/middlewares')
 
 /* Normal lambda code */
-const businessLogic = (event, context, callback) => {
+const businessLogic = async (event) => {
```
erezrokah, line 5: See https://github.com/middyjs/middy/tree/0.x#using-asyncawait

tests/utils/dev-server.js
```diff
@@ -65,6 +65,7 @@ const startServer = async ({ cwd, env = {}, args = [] }) => {
         })
       }
     })
+    // eslint-disable-next-line promise/prefer-await-to-callbacks
```
erezrokah, line 68: I'm not sure how to handle this error.

eduardoboucas (approved):

Issue #2013, opened 2021-03-16
event.multiValueQueryStringParameters in redirects drops values


jlengstorf:
**Describe the bug**

In production, passing two query parameters with the same key results in an array of values in `event.multiValueQueryStringParameters`. This works with redirects as well.

In Netlify Dev, this works if you go to the function directly, but _does not_ work if a redirect is used.

**To Reproduce**

1. deploy https://github.com/jlengstorf/fn-payload-test
2. pass multiple query params with the same key to any path (all redirected to the function) e.g. https://hungry-golick-f8f51d.netlify.app/?test=one&test=two
3. see the multi-value query params array (it's correct)
4. run `ntl dev`
5. do the same thing
6. see the multi-value query params drop values

Issue #1605, opened 2020-11-30
Redirect rewrites inconsistent between netlify dev and production


parkan:
**Describe the bug**

Referencing [this netlify community post](https://community.netlify.com/t/querystringparameters-not-working-with-redirect-on-production/3828/15)

Redirects of the form:

```
/echo/:msg /.netlify/functions/echo?msg=:msg 200
```

will populate the queryString parameter `msg` in netlify dev, however they will not in production. While I prefer the netlify dev behavior, I was advised to open a bug here to change it for consistency.

**To Reproduce**

Set a redirect that uses parameter matching and rewrites them into the target URL. The value will be copied to the target query parameter.

**Configuration**
N/A

**Expected behavior**

Nothing should happen (value in target URL should be ignored)

**Additional context**

Of course, I would vastly prefer that one of these two things happened:
- The current netlify dev behavior worked in production
- (even better) any redirects containing path parameters populated the `pathParameters` property of the [APIGatewayEvent](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L72) as they do on AWS API Gateway

However, I was advised by the platform team that this probably won't happen, so I will need to add the path expressions to each function, which is brittle and not DRY, but nonetheless the CLI behavior should match production.

My guess is that whatever netlify dev uses as the proxy server does this automagically.

robbiehudson:
Thanks for raising this bug @parkan. This just bit us when we deployed a new function to production. Very unexpected that the dev behaviour doesn't match production (and of course, trying to debug this was hard!).

parkan:
@robbiehudson this definitely took a second to track down, I thought I was losing my mind! this is how I ended up tackling the issue (using pathParameters rather than queryParameters), using the excellent lambda-middleware: https://gist.github.com/parkan/f0b95f343e18aaadd1bb93b4ca67806d

robbiehudson:
@parkan this is great, thank you (and much better than my `String.split` hack that I used to get it working).

PR #2017, opened 2021-03-17
fix(command-dev): fix query params handling in redirects

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2013
Fixes https://github.com/netlify/cli/issues/1605

Also see https://answers.netlify.com/t/querystringparameters-not-working-with-redirect-on-production/3828 and https://github.com/netlify/proxy/issues/449

This PR aligns the way query parameters redirects work in `netlify dev` with our production environment.
Given the following `_redirects` file:
```
/api/* /.netlify/functions/test?foo=1&foo=2 200
/foo / 302
/bar /?a=1&a=2&b=3 302
/test id=:id /?param=:id
```

The expected behavior based on the repo https://github.com/erezrokah/fn-payload-test deploy to https://erez-redirects-query.netlify.app/

1. Rewrites to functions always strip the redirect rule query params in favor of passing through the request query params.
2. If a non function redirect rule doesn't have query params in the rule the request params are passed through.
3. If a non function redirect rule has query params those supersede the request params.
4. If a non function redirect rule has param matching syntax those supersede the request params (we can think of this as a specific version of 3).

Before this change the CLI code kind of tried to merge both the request params and the redirect rule params, overriding multiple values of the same param with the last value (since it was doing `set` instead of `append`).

**- Test plan**

See new test

image

### DIFF
--- a/tests/utils/site-builder.js
+++ b/tests/utils/site-builder.js
@@ -77,7 +77,7 @@ const createSiteBuilder = ({ siteName }) => {
       const dest = path.join(directory, pathPrefix, '_redirects')
       tasks.push(async () => {
         const content = redirects
-          .map(({ from, to, status, condition = '' }) => `${from} ${to} ${status} ${condition}`)
+          .map(({ from, to, status, condition = '' }) => [from, to, status, condition].filter(Boolean).join(' '))
           .join(os.EOL)
         await ensureDir(path.dirname(dest))
         await fs.writeFileAsync(dest, content)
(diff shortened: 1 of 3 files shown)


tests/utils/site-builder.js
```diff
@@ -77,7 +77,7 @@ const createSiteBuilder = ({ siteName }) => {
       const dest = path.join(directory, pathPrefix, '_redirects')
       tasks.push(async () => {
         const content = redirects
-          .map(({ from, to, status, condition = '' }) => `${from} ${to} ${status} ${condition}`)
+          .map(({ from, to, status, condition = '' }) => [from, to, status, condition].filter(Boolean).join(' '))
```
erezrokah, line 80: This is so we don't have to pass `status` to the site builder

JGAntunes (approved):

PR #2028, opened 2021-03-20
feat: add support for shell completion

MunifTanjim:
**- Summary**

Add support for shell completion for `bash`, `fish` and `zsh`

**Commands**:

- `netlify completion --shell bash|fish|zsh`
- `netlify completion:generate --shell bash|fish|zsh`
- `netlify completion:generate:alias --shell bash|fish`

**- Test plan**

**BASH**

image

image

**FISH**

image

image

**ZSH**

image

image

image


erezrokah (changes requested):
Hi @MunifTanjim this is cool thanks.
Can you explain:
1. How is this plugin different from https://github.com/oclif/plugin-autocomplete (other than `fish` support)?
2. How does the plugin work?
3. Under `help` this shows up as `completion plugin` with not much information.
4. How can we add this to https://cli.netlify.com/? The reference doc site is autogenerated using a bunch of scripts here https://github.com/netlify/cli/tree/14b1b7751fa259c78fc12e4f1207ca765ac7b483/site/scripts

MunifTanjim:
> How is this plugin different from https://github.com/oclif/plugin-autocomplete (other than fish support)?

### No Modification to Dotfiles (e.g. bashrc, zshrc etc.)

`@oclif/plugin-autocomplete` requires you to modify your `.bashrc`, `.zshrc` or equivalent files to enable completion for a specific cli. That means `netlify` cli will need a line in your config, another cl will need another line in your config. It shows you instruction for how to do that:

**`netlify autocomplete zsh`**:

```
Building the autocomplete cache... done

Setup Instructions for NETLIFY CLI Autocomplete ---

1) Add the autocomplete env var to your zsh profile and source it
$ printf "eval $(netlify autocomplete:script zsh)" >> ~/.zshrc; source ~/.zshrc

NOTE: After sourcing, you can run `$ compaudit -D` to ensure no permissions conflicts are present

2) Test it out, e.g.:
$ netlify <TAB>                 # Command completion
$ netlify command --<TAB>       # Flag completion

Enjoy!
```

On the other hand, `oclif-plugin-completion` will just tell you where to put the generated script so that you shell can pick it up automatically, no need to modify dotfiles.

**`netlify completion --shell zsh`**:

```
Running the following command will generate the completion script for zsh shell:

  $ netlify completion:generate --shell=zsh > _netlify

You need to put that "_netlify" file in one of the directories present in "$FPATH" variable:

  $ echo $FPATH

Usually this should work by automatically find an appropriate directory for you:

  $ netlify completion:generate --shell=zsh | tee "$(echo ${FPATH} | tr ':' '\n' | grep site-functions | head -n1)/_netlify"

For more info, visit: https://www.npmjs.com/package/oclif-plugin-completion#zsh

Enjoy!
```

I think no modification to dotfiles will be preferable to developers as many of them keeps a copy of their dotfiles in git (https://github.com/topics/dotfiles). And having dedicated lines for enabling completion for different cli applications is noisy. Also, not having those lines gives you a "tiny" performance boost (in terms of shell loading time).

### More Advanced Features

`@oclif/plugin-autocomplete` will only give you the basic completion features (i.e. command completion + long flag completion). `oclif-plugin-completion` supports more features (short flags, enum options) on all shells and even more (short flag stacking, alternative long and short syntax etc.) on supporting shells. Here's the feature matrix: https://github.com/MunifTanjim/oclif-plugin-completion/blob/3d2e17b0df/README.md#feature-support-matrix

> How does the plugin work?

It generates a shell completion script based on how you've configured your commands, flags, args etc. and prints it to console. It will not create or modify any files automatically on your system, instead it will show you instructions about where to create the file with that completion script.

> Under help this shows up as completion plugin with not much information.

Running `netlify completion`  will show you detailed information about how to enable completion.

> How can we add this to https://cli.netlify.com/? 

I'll take a look if this can be generated automatically.

MunifTanjim:
Some examples of other CLI tools that behave the same way `oclif-plugin-completion` does:

- `cargo`: `rustup completions zsh cargo`
- `deno`: `deno completions zsh`
- `gh`: `gh completion --shell zsh`
- `chezmoi`: `chezmoi completion zsh`
- `starship`: `starship completions zsh`
- `volta`: `volta completions zsh`

All of these just prints the script to console, instead of creating/modifying files on your system itself. And these are just the ones that I use, there should be plenty more.

erezrokah:
@MunifTanjim I fixed a small issue with the docs in https://github.com/netlify/cli/pull/2028/commits/5917842b7295e504e465f6adc8d98104c72dbca0.
I also noticed this doesn't work with `ntl` (only with `netlify`) but I think this is more of an `oclif` limitation.

MunifTanjim:
> I also noticed this doesn't work with `ntl` (only with `netlify`) but I think this is more of an `oclif` limitation.

It's really easy to implement it in `oclif-plugin-completion` for `fish` and `zsh`. I'm not sure about `bash` yet.

I'll be publishing the support for `fish` and `zsh` shortly.

erezrokah:
@MunifTanjim Thanks again for all the follow ups. I added some code to mark this command as `(Beta)` which is something we do with new commands until we get some user feedback (in case we want to do breaking changes).

I've tried the new version of the plugin with `zsh` and still can't get `ntl` to work.

I'm running `netlify completion:generate --shell=zsh | tee "$(echo ${FPATH} | tr ':' '\n' | grep site-functions | head -n1)/_netlify"` and the content of `_netlify` starts with `#compdef netlify ntl`.

What am I missing?

MunifTanjim:
> I've tried the new version of the plugin with `zsh` and still can't get `ntl` to work.
> 
> I'm running `netlify completion:generate --shell=zsh | tee "$(echo ${FPATH} | tr ':' '\n' | grep site-functions | head -n1)/_netlify"` and the content of `_netlify` starts with `#compdef netlify ntl`.
> 
> What am I missing?

@erezrokah I think your shell is using the old completion cache file. Quoting the [zsh Doc: Completion System](http://zsh.sourceforge.net/Doc/Release/Completion-System.html):

> To speed up the running of compinit, it can be made to produce a dumped configuration that will be read in on future invocations; this is the default, but can be turned off by calling compinit with the option -D. The dumped file is .zcompdump in the same directory as the startup files (i.e. $ZDOTDIR or $HOME); alternatively, an explicit file name can be given by ‘compinit -d dumpfile’. The next invocation of compinit will read the dumped file instead of performing a full initialization.
>
> If the number of completion files changes, compinit will recognise this and produce a new dump file. However, if the name of a function or the arguments in the first line of a #compdef function (as described below) change, it is easiest to delete the dump file by hand so that compinit will re-create it the next time it is run.

So, you need to remove the old cache:

```sh
rm "${ZDOTDIR:-$HOME}/.zcompdump"
```

And it should work.

erezrokah:
> So, you need to remove the old cache:
> 
> ```shell
> rm "${ZDOTDIR:-$HOME}/.zcompdump"
> ```
> 
> And open a new shell, it should work.

That works 🎉

erezrokah (approved):
This is 🚀  I'll merge and release after we're done with https://github.com/netlify/cli/pull/2059

Issue #2029, opened 2021-03-20
Netlify CLI keeps adding overzealous and incorrect rule to .gitignore


SleeplessByte:
**Describe the bug**

In my `.gitignore` I have the following intentional line:

```gitignore
/.netlify
```

Linking a new site / running CLI in several ways will **incorrectly add**:

```gitignore
# Local Netlify folder
.netlify
```

In order to make my monorepos work again I have to add:

```gitignore
!sites/**/.netlify
```

**To Reproduce**

1. Create a new git repo
2. Add `.gitignore` with `/.netlify`
3. Run `netlify deploy`
4. It has added the line to `.gitignore`

**Configuration**

```bash
npx envinfo --system --binaries --npmPackages netlify-cli --npmGlobalPackages netlify-cli


  System:
    OS: Windows 10 10.0.19042
    CPU: (12) x64 Intel(R) Core(TM) i7-6800K CPU @ 3.40GHz
    Memory: 5.80 GB / 31.91 GB
  Binaries:
    Node: 12.18.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - ~\Documents\GitHub\delftsolutions\@introcloud\app\node_modules\.bin\yarn.CMD
    npm: 6.14.5 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    netlify-cli: ^3.10.12 => 3.10.12

```

**Expected behavior**

It does not add `.netlify` to `.gitignore`.

SleeplessByte:
The biggest issue with this is that it will silently fail on changes / new files. You won't notice until your build starts failing somewhere else (for example, because there is no `state.json` available. It's counter-intuitive and frankly, I hate that it's doing something to my `.gitignore` _without telling me_.

erezrokah:
Hi @SleeplessByte, linking https://github.com/netlify/cli/issues/506 for context.
I agree this is not an ideal experience.

To clarify - you only want to ignore `.netlify` at the root level? Any reason you want to commit `state.json`? 
Please note that the `.netlify` folder can also contain temporary files generated during `netlify build` and other information that should not be committed to the repo.

SleeplessByte:
You got it.

We have our own build and then only use netlify deploy, so those other files never get generated.

state.json holds the siteId and is a far better tool to configure ci builds for a folder then trying to do it via environment variables.

SleeplessByte:
The linked issue has very similar concerns (and remains unresolved).

I see the following options:

[...]

PR #2031, opened 2021-03-21
fix(command-link): don't modify gitignore if a pattern includes .netlify

erezrokah:
**- Summary**

Partially fixes https://github.com/netlify/cli/issues/2029 (I'll open other separates issue per https://github.com/netlify/cli/issues/2029#issuecomment-803581401)

**- Test plan**

On a linked repo change your `.gitignore` `.netlify` pattern to `/.netlify` and re-run `ntl link`.

image

### DIFF
--- a/src/utils/gitignore.js
+++ b/src/utils/gitignore.js
@@ -92,7 +92,7 @@ const ensureNetlifyIgnore = async function (dir) {
     // ignore
   }
   /* Not ignoring .netlify folder. Add to .gitignore */
-  if (!ignorePatterns || !ignorePatterns.patterns.includes('.netlify')) {
+  if (!ignorePatterns || !ignorePatterns.patterns.some((pattern) => pattern.includes('.netlify'))) {
     const newContents = `${gitIgnoreContents}\n${ignoreContent}`
     await writeFileAsync(gitIgnorePath, newContents, 'utf8')
   }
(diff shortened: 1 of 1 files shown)


src/utils/gitignore.js
```diff
@@ -92,7 +92,7 @@ const ensureNetlifyIgnore = async function (dir) {
     // ignore
   }
   /* Not ignoring .netlify folder. Add to .gitignore */
-  if (!ignorePatterns || !ignorePatterns.patterns.includes('.netlify')) {
+  if (!ignorePatterns || !ignorePatterns.patterns.some((pattern) => pattern.includes('.netlify'))) {
```
eduardoboucas, line 95: Do you think there's a risk that someone uses sub-directories like `mysite.netlify.app`, which we'd be targeting with this new logic and incorrectly skipping the `.gitignore` modification? It's probably an edge case, but I wonder if a regular expression like `(^|\/)\.netlify($|\/)` would be more accurate?

erezrokah: Great point. Done in https://github.com/netlify/cli/pull/2031/commits/38497c1640b31f87f7380551b07ea5791e2386ce

eduardoboucas: Sorry, I just realised that this might not account for Windows paths? Perhaps `(^|\/|\\)\.netlify($|\/|\\)` is better?

erezrokah: Nice one, done in https://github.com/netlify/cli/pull/2031/commits/cff8f941ba917fc8ba3c514d71d5e18d2af6bd1d

eduardoboucas (approved):
🚀

PR #2094, opened 2021-04-07
feat: serve bundled functions in ntl dev

eduardoboucas:
**- Summary**

This PR uses `zip-it-and-ship-it` to bundle functions before serving them with `netlify dev`.

**- Test plan**

⏳ 

cute-lion-cub


src/function-builder-detectors/zisi.js
```diff
@@ -0,0 +1,67 @@
...
+
+const getTargetDirectory = async ({ log }) => {
+  const targetDirectory = path.resolve(getPathInProject(['functions-serve']))
+
+  try {
+    await makeDir(targetDirectory)
+  } catch (error) {
+    log(`${NETLIFYDEVERR} Could not create directory: ${targetDirectory}`)
+
+    process.exit(1)
```
eduardoboucas, line 51: Not sure whether to `exit` or let the error bubble up?

src/utils/serve-functions.js
```diff
@@ -458,12 +461,24 @@ const startServer = async ({ server, settings, log, errorExit }) => {
...
 }
 
-const startFunctionsServer = async ({ settings, site, log, warn, errorExit, siteUrl, capabilities }) => {
+const startFunctionsServer = async ({ config, settings, site, log, warn, errorExit, siteUrl, capabilities }) => {
   // serve functions from zip-it-and-ship-it
   // env variables relies on `url`, careful moving this code
   if (settings.functions) {
-    await setupFunctionsBuilder({ site, log, warn })
-    const server = await getFunctionsServer({ dir: settings.functions, siteUrl, capabilities, warn })
+    const { target: functionsDirectory } = await setupFunctionsBuilder({
```
eduardoboucas, line 468: The functions builder API now has the ability to return a `target` property, which defines the new functions directory.

src/function-builder-detectors/zisi.js
erezrokah: To exit with an error we use `oclif` `error` utility, see https://github.com/netlify/cli/blob/f881b668de15bf5a9ec6988e5ce543caa7e22df8/src/commands/functions/create.js#L214

Regardless, I think this can be a warning as `netlify dev` will be partially useful?

eduardoboucas: We won't be able to serve the functions if we can't write to the directory, though.

erezrokah: We currently warn on function builder errors, so I think we can remove the `try/catch` here and let the CLI handle it:
https://github.com/netlify/cli/blob/f5b26d2295f46975194fc4c984e7695727b02029/src/utils/serve-functions.js#L409

eduardoboucas: Changed in https://github.com/netlify/cli/pull/2094/commits/78f4ec0c541d8d21d7c3f2d42596c08aff87ef95 to use the `errorExit` property already being passed to the functions server. What do you think?

eduardoboucas: Oh, I think we commented at the same time! 😅 

Just checking that you've seen https://github.com/netlify/cli/commit/78f4ec0c541d8d21d7c3f2d42596c08aff87ef95.

src/function-builder-detectors/zisi.js
```diff
@@ -0,0 +1,65 @@
...
+  try {
+    await makeDir(targetDirectory)
+  } catch (error) {
+    errorExit(`${NETLIFYDEVERR} Could not create directory: ${targetDirectory}`)
+  }
+
+  return targetDirectory
+}
+
+module.exports = async function handler({ config, errorExit, functionsDirectory: sourceDirectory }) {
```
erezrokah, line 54: At the moment we evaluate detecters based on the order of `readdirSync`:
https://github.com/netlify/cli/blob/f5b26d2295f46975194fc4c984e7695727b02029/src/utils/detect-functions-builder.js#L6

I don't think it is guaranteed to have a consistent order.

1. Should we sort the files by name?
2. Should we add a condition to not use `zip-it-and-ship-it` builder when `netlify-lambda` is used?
3. Should we add a condition to use `zip-it-and-ship-it` only when someone is using `esbuild` (I think that would minimize the risk for the PR and we can later on make this behavior the default).

src/utils/serve-functions.js
```diff
@@ -417,32 +425,43 @@ const getBuildFunction = ({ functionBuilder, log }) =>
...
+    functionsDirectory,
+    log,
+    projectRoot: site.root,
+  })
 
-    const debouncedBuild = debounce(getBuildFunction({ functionBuilder, log }), 300, {
-      leading: true,
-      trailing: true,
-    })
+  if (!functionBuilder) {
```
erezrokah, line 437: I like this early return!

erezrokah (changes requested):
Hi @eduardoboucas, this looks great pending one code comment.

What do you think about adding a test case for this?
It will be nice to test es6 modules and TypeScript functions as part of the CLI tests.

src/function-builder-detectors/zisi.js
eduardoboucas: Great points.

I have done 1 and 3 in https://github.com/netlify/cli/pull/2094/commits/1b2278b755c89a57965d5f7dc2d730d366079183.

I don't think we need 2, because `netlify-lambda` will be evaluated before `zisi`, due to the alphabetical order, and if it returns a truthy value, we stop looking for the next handler (https://github.com/netlify/cli/blob/feat/serve-bundled-functions/src/utils/detect-functions-builder.js#L18).

eduardoboucas:
> Hi @eduardoboucas, this looks great pending one code comment.
> 
> What do you think about adding a test case for this?
> It will be nice to test es6 modules and TypeScript functions as part of the CLI tests.

Thanks @erezrokah! I've addressed the comment and also added tests in https://github.com/netlify/cli/pull/2094/commits/88f1f47bf5e68589a8cecb626a464e89b1732884.

Let me know what you think!

erezrokah (approved):
🚀

src/utils/serve-functions.js
```diff
@@ -417,32 +425,43 @@ const getBuildFunction = ({ functionBuilder, log }) =>
...
-    const functionWatcher = chokidar.watch(functionBuilder.src)
-    functionWatcher.on('ready', () => {
-      functionWatcher.on('add', debouncedBuild)
-      functionWatcher.on('change', debouncedBuild)
-      functionWatcher.on('unlink', debouncedBuild)
-    })
-  }
+  log(`${NETLIFYDEVLOG} Function builder ${chalk.yellow(functionBuilder.builderName)} detected${npmScriptString}.`)
+  warn(
+    `${NETLIFYDEVWARN} This is a beta feature, please give us feedback on how to improve at https://github.com/netlify/cli/`,
```
eduardoboucas, line 447: @erezrokah Do we still want to display this message for people using the ZISI function builder?

erezrokah: Great catch - I forgot about it. Let's remove it all together

eduardoboucas: Done in https://github.com/netlify/cli/pull/2094/commits/4ad71a4960407fc2397ac485c07acb331245b490.

erezrokah (approved):

Issue #455, opened 2019-04-17
Feature request: support secure HTTPS local dev environnement


borisschapira:
Right now, I'm developping locally with a compiled Jekyll with a self-signe cert. To use Netlify Dev, I would need it to support secure environment and the ability to either produce it's own self-signed cert, or be able to pass mine as argument.

jhull:
Wondering the update on this...it's February 2020...

RaeesBhatti:
`http://localhost` is already treated as [Secure Context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) by [major browsers](https://www.chromestatus.com/feature/6269417340010496), so, you don't really need an `https://` address.
Or are you using a custom hostname for local development? In which case I would suggest using `localhost` instead.
Please elaborate on your problem if this doesn't work for you!

borisschapira:
Use case: several web projects and each of them has its own context (cookies, service workers, local storage…). If you expose all of them on http://localhost, you mix the contexts. You'd rather want independant self-signed certs on development domains to be able to work on several contexts without interferences.

erezrokah:
> Use case: several web projects and each of them has its own context (cookies, service workers, local storage…). If you expose all of them on http://localhost, you mix the contexts.

Thanks @borisschapira, can this issue be resolved by using a different port per context?

borisschapira:
@erezrokah Yes, it could, with [the appropriate certificate](https://letsencrypt.org/docs/certificates-for-localhost/).

erezrokah:
@borisschapira, I think was not clear with my suggestion.
Won't `http://localhost:8000` and `http://localhost:9000` have different context (cookies, service workers, local storage…)?
So to resolve your use case you could run each project on a different port via the `--port` option?

probablyup:
I think this is needed. Gatsby CLI for example supports running an HTTPS environment via the "-S" flag which uses some library to automatically provision the needed self-signed certificates (I think it's called "devcert".)

`localhost` isn't enough in a lot of cases to bypass the need for HTTPS, for example some browser APIs won't work without being in an explicit SSL context.

erezrokah:
> for example some browser APIs won't work without being in an explicit SSL context.

Thank you for the additional information @probablyup, can you share those APIs and cases? That would make it easier to verify those work when the CLI is running in HTTPS mode.

probablyup:
[...]

Issue #2128, opened 2021-04-14
chore: switch to a known validator library (e.g. ajv)


erezrokah:
We're doing a lot of user input validation in the codebase.
See example https://github.com/netlify/cli/blob/da75eb1304e604f6dcf68a6854b9f5f209364588/src/utils/detect-server.js#L160

I think it would be much easier to understand the code if we use a schema and a known validator.

Issue #2127, opened 2021-04-14
bug(command-dev): `netlify.toml` dev block should use snake case


erezrokah:
**Describe the bug**

`netlify.toml` dev block uses camel case which doesn't match other configurations which use snake case.
We should switch to snake case configuration properties and deprecate the camel case ones.

**To Reproduce**

See https://cli.netlify.com/netlify-dev#netlifytoml-dev-block and https://docs.netlify.com/configure-builds/file-based-configuration/

**Expected behavior**

`netlify.toml` dev block should use snake case

cc @rstavchansky

PR #2132, opened 2021-04-14
feat(command-dev): support secure https development

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/455

Doesn't do anything fancy like generating certificates on the fly, but still unblocks users by letting them bring their own certification and key file.

**- Test plan**

Added a test with self signed cert and key (validity until 2048)

image


src/utils/proxy.js
```diff
@@ -378,15 +367,36 @@ const startProxy = async function (settings, addonsUrls, configPath, projectDir)
     }
 
     proxy.web(req, res, options)
+  }
+
+  return onRequest
+}
+
+const createSecureServer = ({ onRequest, cert, key }) => https.createServer({ cert, key }, onRequest)
```
ehmicky, line 375: [sand] Would it be simpler to inline that function?

docs/netlify-dev.md
```diff
@@ -118,6 +118,10 @@ Netlify Dev is meant to work with zero config for the majority of users, by usin
   jwtSecret = "secret" # The secret used to verify tokens for JWT based redirects
   jwtRolePath = "app_metadata.authorization.roles" # Object path we should look for role values for JWT based redirects
   autoLaunch = true # a Boolean value that determines if Netlify Dev launches the local server address in your browser
+  # to start an https server instead of an http one, configure a certificate and key files
+  [dev.https]
+    certFile = "cert.pem" # path to the certificate file
```
erezrokah, line 123: This is for consistency.

See https://github.com/netlify/cli/issues/2127

src/utils/proxy.js
```diff
@@ -378,15 +367,36 @@ const startProxy = async function (settings, addonsUrls, configPath, projectDir)
...
   })
 
+  const onRequest = getRequestListener({ proxy, rewriter, settings, addonsUrls, functionsServer })
+  const server = settings.https ? createSecureServer({ ...settings.https, onRequest }) : http.createServer(onRequest)
   server.on('upgrade', function onUpgrade(req, socket, head) {
     proxy.ws(req, socket, head)
   })
 
   return new Promise((resolve) => {
     server.listen({ port: settings.port }, () => {
```
ehmicky, line 397: I am not sure whether this is something to fix separately, but we should promisify this with `await promisify(server.listen.bind(server))({ port: settings. port })`. This would also allow handle errors (e.g. port errors).

src/utils/proxy.js
```diff
@@ -378,15 +367,36 @@ const startProxy = async function (settings, addonsUrls, configPath, projectDir)
...
-      resolve(`http://localhost:${settings.port}`)
+      const scheme = settings.https ? 'https' : 'http'
+      resolve(`${scheme}://localhost:${settings.port}`)
```
ehmicky, line 399: Can `settings.port` be `0` (i.e. "use any port")? If so, we might consider using `server.address().port` instead to get the resolved port.

eduardoboucas (approved):

src/utils/proxy.js
```diff
@@ -328,20 +329,8 @@ const initializeProxy = function (port, distDir, projectDir) {
...
-    distDir: settings.dist,
-    jwtSecret: settings.jwtSecret,
-    jwtRoleClaim: settings.jwtRolePath,
-    configPath,
-    projectDir,
-  })
-
-  const server = http.createServer(async function onRequest(req, res) {
+const getRequestListener = ({ proxy, rewriter, settings, addonsUrls, functionsServer }) => {
+  const onRequest = async (req, res) => {
```
ehmicky, line 333: [sand] What are your thoughts on merging both functions to a single one, then using `onRequest.bind(undefined, { proxy, ... })`?

ehmicky (approved):

docs/netlify-dev.md
```diff
@@ -337,6 +341,21 @@ After you have installed an add-on, it will be visible with the `netlify addons:
 
 For now, it is important to include instructions to create addons for each member of your team, as there is no way to specify addons inside of `netlify.toml`. We are working on this.
 
+### Running an https server for local development
+
+By default, `netlify dev` starts an http server. If you require an https one, you can configure a certificate and key file to be used by `netlify dev`.
```
rstavchansky, line 346: minor suggestions:

```suggestion
By default, `netlify dev` starts an HTTP server. If you require HTTPS, you can configure a certificate and key file for use by `netlify dev`.
```

src/utils/proxy.js
erezrokah: It cannot be `0`. We capture it before in:
https://github.com/netlify/cli/blob/41ec0d9a4626efa16efd0e2e292d84c66fd57d3d/src/utils/detect-server.js#L170

We do that to alert the user in case they configured a custom port, but we cannot capture it.

docs/netlify-dev.md
```diff
@@ -337,6 +341,21 @@ After you have installed an add-on, it will be visible with the `netlify addons:
...
+
+```toml
+[dev]
+[dev.https]
+certFile = "cert.pem"
+keyFile = "key.pem"
+```
+
+**Important -** if you're using a self signed certificate you might need to configure your browser to accept it when running on `localhost`.
+For Chrome you can do it by visiting the following URL in your browser `chrome://flags/#allow-insecure-localhost`.
```
rstavchansky, line 357: Minor suggestions:

```suggestion
**Important -** If you're using a self-signed certificate, you might need to configure your browser to accept it when running on `localhost`.
To enable this setting for Chrome, visit `chrome://flags/#allow-insecure-localhost` in your browser.
```

rstavchansky (approved):
Left some minor docs feedback. Looks good!

src/utils/detect-server.js
```diff
@@ -7,9 +7,40 @@ const fuzzy = require('fuzzy')
...
 const inquirer = require('inquirer')
 const inquirerAutocompletePrompt = require('inquirer-autocomplete-prompt')
+const isPlainObject = require('is-plain-obj')
+
+const { readFileAsyncCatchError } = require('../lib/fs')
 
 const { NETLIFYDEVLOG, NETLIFYDEVWARN } = require('./logo')
 
+const readHttpsSettings = async (options) => {
+  if (!isPlainObject(options)) {
```
erezrokah, line 17: See https://github.com/netlify/cli/issues/2128

src/utils/proxy.js
erezrokah: Done in https://github.com/netlify/cli/pull/2132/commits/e763e9c33e23f3cdbee88b4f827d5265a372210e

src/utils/proxy.js
erezrokah: Done https://github.com/netlify/cli/pull/2132/commits/c323df52977c3798da2e93e3b969eb675e08405b

src/utils/proxy.js
erezrokah: Done in https://github.com/netlify/cli/pull/2132/commits/ec5c40a204502cb11960aa7ee15387ce60f6cc18

As discussed we can't use `promisify` as errors are emitted via `server.on('error')`

Issue #909, opened 2020-05-19
Support reloading/watching of source files outside the functions folder with "netlify dev"


oles:
**- Do you want to request a _feature_ or report a _bug_?**
Feature, though could be interpreted as a bug.

It is maybe not smart to watch the whole root folder by default.
Perhaps not many people have the same usecase of shared utilities for their functions.

Therefore perhaps it is better to add a `--watch` flag to your `netlify dev` command? 
For example, `netlify dev --port 9000 --watch <file or folder path>`

This works when deployed, by the way, thanks to your excellent "zip-it-and-ship-it" :heart: 

**- What is the current behavior?**
While developing some Netlify Functions, the reload/watching functionality works as expected, as long as the files are in the folder specified as your "functions" folder.

If you, for example, create an utility folder next to your "functions" folder (not inside it), and `require` some shared utilities from it, `netlify dev` will not emit the `◈ /<path>/functions/<function>.js modified, reloading...` when you edit the utilities.

You have to restart `netlify dev` for the changes in your utilities to be picked up.

**- If the current behavior is a bug, please provide the steps to reproduce.**
1. clone https://github.com/oles/netlify-functions-missing-reload-proof-of-concept
2. run `netlify dev`
3. open http://localhost:8888/.netlify/functions/test-function and/or 
http://localhost:8888/.netlify/functions/nested-test-function and see the properly capitalized text.
4. change `utilities/index.js` by changing the `capitalize` function (remove the `.toLowerCase()`, for example)
5. reload one of the opened URLs, and see that the output is still the properly capitalized text.
6. relaunch `netlify dev`
7. reload one of the opened URLs, and see now that the output is in all uppercase.

**- What is the expected behavior?**
For the first reload in step 5 to show the text in all uppercase.

**- Local Environment Information**
netlify-cli/2.51.0 linux-x64 node-v10.20.1

Also wanted to thank you for both the great support you've provided me, and the quality of this product. Been a pleasure to develop our current project with it!

caesarsol:
+1!
Customizing watch folders is absolutely needed for a modern web development experience IMHO.

Better off, since often some files are shared between backend and frontend, it would be cool if only the files imported in the netlify-functions could be watched, so that the backend gets restarted as less times as possible and no manual tracking is necessary.

erezrokah:
[...]

PR #2143, opened 2021-04-15
feat(command-dev): watch all functions files

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/909

~~Still in draft since I need to test it more~~

I believe we would need to add similar code to the functions builders to make this work for `esbuild`.

Another way to implement this is via configuration (e.g. flags/`netlify.toml` or both)

**- Test plan**

Added a test case

🐼

### DIFF
--- a/src/utils/get-functions.js
+++ b/src/utils/get-functions.js
@@ -1,4 +1,6 @@
-const { listFunctions } = require('@netlify/zip-it-and-ship-it')
+const path = require('path')
+
+const { listFunctions, listFunctionsFiles } = require('@netlify/zip-it-and-ship-it')
 
 const { fileExistsAsync } = require('../lib/fs')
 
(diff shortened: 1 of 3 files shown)


src/utils/get-functions.js
```diff
@@ -24,4 +26,29 @@ const getFunctions = async (functionsSrcDir) => {
   return functionsWithProps
 }
 
-module.exports = { getFunctions }
+const getWatchDirs = (functionsSrcDir, functions) => {
+  const localFilesDirs = functions
```
erezrokah, line 30: This can cause duplicate events if someone imports from both `<repo>/utils` and `<repo>/utils/common`.

Not sure how to handle it

src/utils/get-functions.js
```diff
@@ -24,4 +26,29 @@ const getFunctions = async (functionsSrcDir) => {
...
+    .filter(({ srcFile }) => !srcFile.startsWith(functionsSrcDir) && !srcFile.includes('node_modules'))
```
eduardoboucas, line 31: Why do we need this `includes('node_modules')`?

erezrokah: When calling `listFunctionsFiles` `zisi` returns all the files required by the function (including ones from `node_modules`).

We only want to watch user code.

We pass it also as a flag to `chokidar` too (in case there are `node_modules` under the functions directory): https://github.com/netlify/cli/blob/a2799075df2b61e1653ead3f764ea09b8c5b8e4d/src/utils/serve-functions.js#L185

so I think it will work without it, but I don't see a reason to pass directories we know we don't want to watch, just for them to be ignored later on.

eduardoboucas: Makes sense! Thanks for clarifying.

src/utils/get-functions.js
eduardoboucas: Is this still the case, or did the debounced events fix this?

src/utils/get-functions.test.js
```diff
@@ -76,3 +76,36 @@ test('should mark background functions based on filenames', async (t) => {
     ])
   })
 })
+
+test('should return additional watch dirs when functions requires a file outside function dir', async (t) => {
+  await withSiteBuilder('site-without-functions', async (builder) => {
```
eduardoboucas, line 81: [sand] Should we use a different site name, since it does have a function?

eduardoboucas (approved):

src/utils/get-functions.js
erezrokah: We'll still get duplicate events, but we'll handle them once

src/utils/get-functions.test.js
erezrokah: Yes, good point I'll update it

erezrokah: Done in https://github.com/netlify/cli/pull/2143/commits/25be62d9ec0cfdcca080d677c16c5b572041abb1

okwme:
does this add the ability to watch directories outside of the functions dir?
is there a flag?

PR #2150, opened 2021-04-18
chore: add package size benchmark

eduardoboucas:
**- Summary**

Adds a benchmark workflow to track the evolution of the package size. It works by running `npm install --production` and calculating the size of the generated `node_modules` directory. It makes use of https://github.com/netlify/delta-action to report the results and compare against the baseline.

**- Test plan**

N/A

fat_animals_00

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: npm install --production
       - name: Get size
-        run: du -sk node_modules > .delta.packageSize && echo "b (Package size)" >> .delta.packageSize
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
```
erezrokah, line 18:
```suggestion
          node-version: 15.x
```

erezrokah (approved):
🚀

PR #2169, opened 2021-04-19
feat: reuse existing tabs in `netlify dev` command

### LINKED ISSUE #786: Feature request: Re-use tab when starting `netlify dev`
**Problem:** Every time `netlify dev` is run, a new browser tab is opened. This results in a LOTS of open tabs in your browser.

**Desired Solution:** Re-use open tab if one already exists.

**Possible Implementation:** 

Create react app checks if there is already an open browser tab at localhost:3000 and reuses it if there is already one open. Otherwise it will open a new tab. 

^ This is a smooth developer experience and should help alleviate some issues folks are having with auto open in general.

This logic is MIT and can be found here: https://github.com/facebook/create-react-app/blob/589b41aecaa10d410713794f66a648bf3a72fb62/packages/react-dev-utils/openBrowser.js#L70-L71

eduardoboucas:
**- Summary**

Uses [CRA's `openBrowser` logic](https://github.com/facebook/create-react-app/blob/7e4949a20fc828577fb7626a3262832422f3ae3b/packages/react-dev-utils/openBrowser.js#L67) to reuse existing tabs in `netlify dev` when possible, as opposed to opening a new one each time.

Closes #786.

**- Test plan**

⏳ 

🐨


erezrokah (commented):
Adding https://www.npmjs.com/package/react-dev-utils for context.

We should be able to import it from that package (as long as the package size is not too big)

src/utils/open-browser-cra/index.js
```diff
@@ -0,0 +1,146 @@
...
+      'Brave Browser',
+      'Vivaldi',
+      'Chromium',
+    ]
+
+    for (const chromiumBrowser of supportedChromiumBrowsers) {
+      try {
+        // Try our best to reuse existing tab
+        // on OSX Chromium-based browser with AppleScript
+        execSync(`ps cax | grep "${chromiumBrowser}"`)
```
ehmicky, line 88: The code seems to be inconsistent about how processes are being spawned: it uses both `cross-spawn` and `childProcess.execSync()`, while our code uses `execa`. 

`execa` uses `cross-spawn` and adds a few improvements on top of it, while `cross-spawn` uses `childProcess.spawn()`, and adds a few cross-platform improvements on top of it. `childProcess.execSync()` uses `childProcess.spawn()` under the hood.

It seems to me it might be worth it to use `execa` and `async`/`await` instead?

src/utils/open-browser-cra/index.js
```diff
@@ -0,0 +1,146 @@
...
+    return false
+  }
+}
+
+/**
+ * Reads the BROWSER environment variable and decides what to do with it. Returns
+ * true if it opened a browser or ran a node.js script, otherwise false.
+ */
+const openBrowser = function (url) {
+  const { action, value, args } = getBrowserEnv()
```
ehmicky, line 131: It seems to me it might be possible to merge `getBrowserEnv()` and `openBrowser()` and remove the intermediary `Actions` enum. I.e. replace the `case` below by the conditions tested in `getBrowserEnv()` directly.

package.json
```diff
@@ -93,6 +93,7 @@
     "@ungap/from-entries": "^0.2.1",
     "ansi-styles": "^5.0.0",
     "ascii-table": "0.0.9",
+    "better-opn": "^3.0.0",
```
erezrokah, line 96: We'll need to use `v2` since `v3` requires Node.js 12

eduardoboucas: Good catch! Done in https://github.com/netlify/cli/pull/2169/commits/07383eb749d05cf9ee220e009fbeb9cb96d94f2e.

erezrokah (approved):

PR #2188, opened 2021-04-21
chore(ci): switch to cms shell on windows

erezrokah:
**- Summary**

A shot in the dark trying to prevent our Windows CI from hanging. The default shell for windows is PowerShell.

I couldn't reproduce this on my Windows VM and also getting nowhere with this https://github.community/t/github-actions-times-out-with-no-logs/156707/2

**- Test plan**

Merge to `main` and see if this makes a difference

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@ name: Testing for CLI
 
 on:
   push:
-    branches: [main, chore/use_cmd_windows_ci]
+    branches: [main]
   pull_request:
     branches: [main]
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,22 +2,28 @@ name: Testing for CLI
 
 on:
   push:
-    branches: [main]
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -8,16 +8,22 @@ on:
...
+    runs-on: ${{ matrix.config.os }}
     timeout-minutes: 30
+    defaults:
+      run:
+        shell: ${{ matrix.config.shell }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        config:
+          - { os: macOS-latest, shell: bash }
```
erezrokah, line 19: Oddly enough `bash` is the default for macOS (though Apple already changed the default to `zsh`).
`zsh` is not even installed with the MacOS [image](https://github.com/actions/virtual-environments/blob/a91dbe89493b1661224afeff1d898e8e9a87e3e2/images/macos/macos-10.15-Readme.md).

eduardoboucas (approved):
🏆  Best usage of cute animal picture ever.

PR #2195, opened 2021-04-22
fix: don't infinite redirect during file shadowing

### LINKED ISSUE #1595: Standard 404 rule causes infinite redirect when running static Hugo site with dev command
**Describe the bug**
The following has been present, unchanged, in our `_redirects` file on our live site without any issues for months if not years:
```
/* /404.html 404
```
However, it seems that this redirect command causes the website served by `netlify dev` to enter an infinite redirect loop when trying to navigate to any page, including `/`. This renders the entire site inaccessible. Removing this line from `_redirects` causes the site to work as normal (although 404s then simply redirect to the home page instead of displaying 404.html). 

My reading of the docs suggests that this should be a valid rule and that this would be expected behavior only if `!` was used, which it is not in this case. 

**To Reproduce**
Unfortunately I cannot share our GitHub repository as it is private, however this is the full text of our `_redirects` file:
[_redirects.txt](https://github.com/netlify/cli/files/5599195/_redirects.txt) (`.txt` added so Github would allow the upload)
The issue happens when the last line is un-commented. 

You can see these rules in action on our live site at [jaquishbiomedical.com](https://www.jaquishbiomedical.com/)

**Configuration**

- If possible, please copy/paste below your `netlify.toml`:
<details>
<summary>Full netlify.toml</summary>

```
# Default build command.
[build]
publish = "public"
command = "hugo --gc --minify --baseURL $URL --environment production"
HUGO_VERSION = "0.78.2"
HUGO_ENV = "production"
HUGO_ENABLEGITINFO = "true"


##############################################################################
# Master branch context: all deploys from the Master branch set in your site's
# deploy contexts will inherit these settings.
##############################################################################

[context.master]
command = "hugo --gc --minify --enableGitInfo --baseURL $URL --environment production"

[context.master.environment]
HUGO_ENV = "production"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"


##############################################################################
# Staging branch context: all deploys from the staging branch set in your site's
# deploy contexts will inherit these settings.
##############################################################################

[context.staging]
command = "hugo --gc --minify --enableGitInfo --baseURL $URL --environment development"

[context.staging.environment]
HUGO_ENV = "development"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"

##############################################################################
# Deploy Preview context: all deploys generated from a pull/merge request will
# inherit these settings.
##############################################################################

[context.deploy-preview]
command = "hugo --gc --buildFuture --baseURL $DEPLOY_PRIME_URL --environment development"

[context.deploy-preview.environment]
HUGO_ENV = "development"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"

##############################################################################
# Devstore Deploy Context: 
# 
##############################################################################

[context.task1314c]
command = "hugo --gc --minify --enableGitInfo --baseURL $URL --environment devstore --config config.toml,devstoreconfig.toml"

[context.task1314c.environment]
HUGO_ENV = "devstore"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"


##############################################################################
# Branch Deploy context: all deploys that are not from a pull/merge request or
# from the Production branch will inherit these settings.
##############################################################################

[context.branch-deploy]
command = "hugo --gc --baseURL $DEPLOY_PRIME_URL --environment development"

[context.branch-deploy.environment]
HUGO_ENV = "development"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"


##############################################################################
# devsite.jaquishbiomedical.com just a site to always have the devstore products running
##############################################################################

[context.devstore]
command = "hugo --gc --minify --enableGitInfo --baseURL $URL --config config.toml,devstoreconfig.toml --environment devstore"

[context.devstore.environment]
HUGO_ENV = "development"
HUGO_VERSION = "0.77.0"
HUGO_ENABLEGITINFO = "true"

##############################################################################
# Develop site context: must manually set branch
##############################################################################

[context.task1607a-save-fb-tracking-data-for-shopify]
command = "hugo --gc --minify --enableGitInfo --baseURL $URL --environment development"

[context.task1607a-save-fb-tracking-data-for-shopify.environment]
HUGO_ENV = "development"
HUGO_VERSION = "0.78.2"
HUGO_ENABLEGITINFO = "true"

##############################################################################
# Plugins
##############################################################################
#[[plugins]]
#package = "netlify-plugin-cypress"
#    [plugins.inputs]
#    record = true

##############################################################################
# Headers
# For info about Netlify/TOML header formatting see: https://docs.netlify.com/routing/headers/#syntax-for-the-headers-file
##############################################################################
[[headers]]
  for = "/*"
  [headers.values]
    Access-Control-Allow-Origin = "*"
    Strict-Transport-Security = '''
    max-age=63072000,
    includeSubDomains, 
    preload'''
  
# Proxy Caching
# The default Netlify cache-control header, are `max-age=0, must-revalidate, public` which means all content should be cached but also re-validated.
# `s-max-age` Takes precedence over `max-age` or the `Expires` header, but it only applies to shared caches (e.g., proxies) and is ignored by a private cache. This means we’re using custom headers on Netlify to tell Cloudflare (our CDN) how long to cache assets. We’ll then control how long browsers cache using CDN settings.
# 604800 seconds translates to 7 days, so during normal operation Cloudflare will cache images, CSS and JavaScript files for 7 days.
# Even if Cloudflare has cached our .css or .js we are using adding a hash to the filenames which causes any chance to be cache busting. Hugo has a cache-busting hash baked in called fingerprinting.  Old versions and new versions will be fetched and cached separately due to this new filename with a hash. This makes it safe to aggressively cache across the CDN and in browsers.
# For more info: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#defining_optimal_cache-control_policy 

# ******** Non-hashed static assets ********
# These we're ok caching for 7 days on a local machine and 7 days on Edge. (see above)
[[headers]]
  for = "/img-gen/*"
  [headers.values]
    cache-control = '''
    public, 
    s-max-age=604800, 
    max-age=604800'''

[[headers]]
  for = "/images*"
  [headers.values]
    cache-control = '''
    public, 
    s-max-age=604800, 
    max-age=604800'''

[[headers]]
  for = "/fonts/*"
  [headers.values]
    cache-control = '''
    public, 
    s-max-age=604800, 
    max-age=604800'''

[[headers]]
  for = "/favicon/*"
  [headers.values]
    cache-control = '''
    public, 
    s-max-age=604800, 
    max-age=604800'''

# ******** Hashed assets ********
# These we're ok to cache for 1 year on a local machine and 1 year on Edge (see above)

[[headers]]
  for = "/*.css"
  [headers.values]
    cache-control = '''
    public, 
    s-max-age=31536000, 
    max-age=31536000'''
  
[[headers]]
  for = "/*.js"
  [headers.values]
    cache-Control = '''
    public, 
    s-max-age=31536000, 
    max-age=31536000'''
```

</details>

```
System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 7.40 GB / 15.96 GB
  Binaries:
    Node: 12.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 6.14.8 - C:\Program Files\nodejs\npm.CMD
```

**Expected behavior**
I expect for the local version of the website served by `netlify dev` to be accessible and useable just like it is when the redirect rule is removed.

**CLI Output**
n/a (Site launches successfully) 

**Additional context**
The error causes the only thing visible when trying to visit the site to be the error page showing "too many redirects" error on both Chrome and Firefox.

erezrokah:
**- Summary**

Fixes #1595 

The cause for the issue is a bit of an edge case and involves passing the `-d` flag to `hugo`.
When you pass `-d` to `hugo` not only it watches and serves files, but it also outputs static files to a directory.
For example, the result of running `hugo server -d out` for the test site will create a directory with the following content:
```
out
├── _redirects
├── categories
│   ├── index.html
│   └── index.xml
├── index.html
├── index.xml
├── sitemap.xml
└── tags
    ├── index.html
    └── index.xml
```

The scenario in the bug has a `/* /404.html 404` redirect rule which means each path is matched by the rule.
Due to file shadowing, the CLI first checks for an existing static file. For example when you access `/` the CLI will check for an existing `/index.html` file (which exists) and then rewrite directly to the static file.

When you try to access `/index.html` on a Hugo site it redirects you with a `301` status code to `/` (you can try this by accessing `http://localhost:1313/index.html`). 

The redirect to `/` will be matched again by the CLI which will again try to retrieve the static file, hence the infinite redirect.

The solution is to not match anymore redirects when this happens.

> This doesn't happen without the `-d` flag as the CLI will not find a static file hence won't try to retrieve `/index.html`

**- Test plan**

Added a test case for `hugo`


tests/hugo-site/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "serve": "$(npm bin)/hugo server -d out"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "hugo-bin": "^0.70.2"
```
erezrokah, line 13: This downloads releases directly from https://github.com/gohugoio/hugo/releases so the performance impact should be ok

.gitignore
```diff
@@ -29,3 +29,6 @@ oclif.manifest.json
 .netlify
 
 .eslintcache
+
+tests/hugo-site/resources
```
erezrokah, line 33: This is here so it can get ignored by eslint and prettier

ehmicky (approved):

PR #2197, opened 2021-04-22
fix: delete the directory from which functions will be served

### LINKED ISSUE #2190: TypeScript function cache is not cleaned
**Describe the bug**

If you create a TypeScript function, then delete it, its built .js files still live on within `.netlify/functions-serve` for subsequent runs.

**To Reproduce**

1. Create a new TypeScript function (hello world is fine)
2. Run `netlify dev`, note that `.netlify/functions-serve/funcname` now exists
3. Delete the function file
4. Run `netlify dev` again. The existing function is still in `.netlify/functions-serve/funcname` and is evaluated and run as part of `netlify dev` still.

**Configuration**

Default settings (no netlify.toml), CLI 3.21.2

**Expected behavior**

Deleting a function that previously had a build error, the build error should stop being reported on later runs of `netlify dev`

eduardoboucas:
**- Summary**

Delete the functions serving directory when the ZISI function builder is initialised.

Closes #2190.

**- Test plan**

N/A

🐦

### DIFF
--- a/src/function-builder-detectors/zisi.js
+++ b/src/function-builder-detectors/zisi.js
@@ -1,24 +1,16 @@
 const path = require('path')
 
-const { zipFunction, zipFunctions } = require('@netlify/zip-it-and-ship-it')
+const { zipFunctions } = require('@netlify/zip-it-and-ship-it')
+const del = require('del')
 const makeDir = require('make-dir')
 
 const { getPathInProject } = require('../lib/settings')
 const { getFunctions } = require('../utils/get-functions')
 const { NETLIFYDEVERR } = require('../utils/logo')
 
-const bundleFunctions = ({ config, sourceDirectory, targetDirectory, updatedPath }) => {
-  // If `updatedPath` is truthy, it means we're running the build command due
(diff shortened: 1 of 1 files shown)


src/function-builder-detectors/zisi.js
```diff
@@ -64,6 +56,14 @@ module.exports = async function handler({ config, errorExit, functionsDirectory:
 
   const targetDirectory = await getTargetDirectory({ errorExit })
 
+  // Emptying the directory from which functions will be served, to clear any
+  // deleted functions from a previous run.
+  try {
+    await del(targetDirectory)
```
ehmicky, line 62: Should `force: true` be used? (in case `process.cwd()` is not a parent of `targetDirectory`)

ehmicky (approved):

src/function-builder-detectors/zisi.js
eduardoboucas: Good point, thanks! Done in https://github.com/netlify/cli/pull/2197/commits/e50efb2d243459527f7b292d1f405571ef3f1ca7.

Issue #1846, opened 2021-02-07
chore: enforce Netlify style guide in CLI


erezrokah:
Per this comment https://github.com/netlify/cli/pull/1836#issuecomment-774165013, we should find a way to enforce our style guide in the CLI output.

Some work has been done to extract log messages to a centralized location:
https://github.com/netlify/cli/blob/33fc7fbc068a825c14730047162900ca93737044/src/lib/log.js

We should add any new message to that file and move existing ones.
Having all log messages in a centralized location will make it easier to lint them per the style guide.

PR #2200, opened 2021-04-22
chore(ci): add vale

erezrokah:
**- Summary**

Initial work towards https://github.com/netlify/cli/issues/1846. Only lints docs for now (not CLI logs).

This creates a lot of annotations so I might just enable it for changes files only, but I wanted to get some visibility into the warnings.

**- Test plan**

See GitHub annotations.

🐼

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -2,7 +2,7 @@ name: Lint Docs
 
 on:
   push:
-    branches: [main, chore/add_vale]
+    branches: [main]
   pull_request:
     branches: [main]
 
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -1,0 +1,22 @@
+name: Lint Docs
+
+on:
+  push:
+    branches: [main, chore/add_vale]
+  pull_request:
+    branches: [main]
+
+jobs:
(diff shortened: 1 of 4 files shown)


eduardoboucas:
This looks great! Would it be possible to check the string messages we output in the CLI, or will the action freak out when it sees code?

erezrokah:
> Would it be possible to check the string messages we output in the CLI, or will the action freak out when it sees code?

I think we will need to extract those into a known format https://github.com/errata-ai/vale#functionality

.github/workflows/vale.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+
+      - name: Vale
+        uses: errata-ai/vale-action@master
+        with:
+          styles: https://vale-library.netlify.app/styles.zip
+          files: '["docs", "src", "README.md", "CODE_OF_CONDUCT.md", "CONTRIBUTING.md"]'
+          onlyAnnotateModifiedLines: true
```
erezrokah, line 21: I'll remove this once we fix all the current warnings

ehmicky (approved):

PR #2204, opened 2021-04-23
fix(deps): pin memoize-one to 5.1.1

waynehoover:
This fixes https://github.com/netlify/cli/issues/2202 (and https://github.com/alexreardon/memoize-one/issues/80).

Since currently we're using `^5.1.1` format, `5.2.0` is used for the CLI and it is causing the error. By pinning the memoize-one version to `5.1.1`, the error will be gone.
We can look into for upgrading `5.2.0` or not later on.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "netlify-cli",
-      "version": "3.21.4",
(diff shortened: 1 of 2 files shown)


keiko713 (approved):
test passed 🎉

PR #2218, opened 2021-04-24
fix(deps): upgrading memoize-one to 5.2.1

alexreardon:
Context: https://github.com/netlify/cli/issues/2202

`memoize-one@5.2.0` introduced an unintentional breaking change to the CommonJS bundle. `5.2.0` has been deprecated, and `5.2.1` removes the named import to revert the breaking change to the CommonJS bundle.

This PR simply bumps you back to `5.2.1` and moves away from the named import which is no longer supported (as `5.2.0` has been deprecated)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,7 +75,7 @@
         "lodash": "^4.17.20",
         "log-symbols": "^4.0.0",
         "make-dir": "^3.0.0",
-        "memoize-one": "^5.2.0",
+        "memoize-one": "^5.2.1",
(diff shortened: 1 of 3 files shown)


alexreardon:
Feel free to throw away this PR and do the bump yourselves :)

JGAntunes:
Thanks @alexreardon, this should take care of half of it and https://github.com/netlify/build/pull/2653 will address the rest of the problem 👍 I'll go ahead and merge this.

JGAntunes (approved):

Issue #2216, opened 2021-04-23
TypeError [ERR_INVALID_ARG_TYPE] if publish direct already exists


unclejustin:
**Describe the bug**

When running `ntl dev` in a project that already contains the publish directory the following error is thrown:

```sh
TypeError [ERR_INVALID_ARG_TYPE]: The "url" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at Url.parse (url.js:160:3)
    at Object.urlParse [as parse] (url.js:155:13)
    at Object.common.setupOutgoing (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/node_modules/http-proxy/lib/http-proxy/common.js:87:12)
    at Array.stream (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/node_modules/http-proxy/lib/http-proxy/passes/web-incoming.js:127:14)
    at ProxyServer.<anonymous> (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/node_modules/http-proxy/lib/http-proxy/index.js:81:21)
    at Object.web (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/src/utils/proxy.js:331:20)
    at serveRedirect (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/src/utils/proxy.js:114:28)
    at ProxyServer.<anonymous> (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/src/utils/proxy.js:305:14)
    at ProxyServer.emit (/Users/username/.asdf/installs/nodejs/14.15.4/.npm/lib/node_modules/netlify-cli/node_modules/eventemitter3/index.js:183:35)
```

**To Reproduce**

1. Open project
2. Run `ntl build`
3. Run `ntl dev` 
4. Try to navigate to dev URL
5. Errors here
6. Delete the designated publish directory i.e. "dist"
7. Run `ntl dev` again
8. Try to navigate to dev URL
9. This time it will work

**Configuration**

```toml
[build]
  command = "npm run build"
  functions = "functions"
  publish = "dist"

[[redirects]]
  from = "/api/*"
  to = "/.netlify/functions/:splat"
  status = 200

[[redirects]]
  from = "/*"
[...]

PR #2224, opened 2021-04-25
fix(command-dev): handle undefined location

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2216.
Probably caused by https://github.com/netlify/cli/pull/2195

I couldn't reproduce, the fix is based on the issue stacktrace.

Still it would be great to get a reproduction so I can understand the cause.

**- Test plan**

Waiting on a reproduction for the issue.

🐼

### DIFF
--- a/src/utils/proxy.js
+++ b/src/utils/proxy.js
@@ -300,7 +300,7 @@ const initializeProxy = function (port, distDir, projectDir) {
       }
     }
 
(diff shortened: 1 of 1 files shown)


eduardoboucas:
I'd also love to understand the underlying cause, especially given this comment: https://github.com/netlify/cli/issues/2216#issuecomment-826140379

eduardoboucas (approved):

Issue #2227, opened 2021-04-25
Error with @netlify/functions typescript support - Cannot find module 'exports'


jaredh159:
Problem with `netlify-cli`. I created a extremely simple [reproduction repository.](https://github.com/jaredh159/netlify-cli-bug-repro)

I'm using the new `@netlify/functions` first-class typescript support feature. _Love it (except for this problem), by the way, thanks!_

The problem is, I'm importing a dependency, the official AWS S3 client `@aws-sdk/client-s3`, and this is causing `netlify dev` to break.  Here's the relevant code from the repro:

```ts
import { Handler } from '@netlify/functions';
import { S3 } from '@aws-sdk/client-s3';

const handler: Handler = async (event) => {
  // Omitted: actually doing something with the S3 Client
  // vvvv--- but this line causes `netlify dev` to break
  const client = new S3({});

  return {
    statusCode: 200,
    body: `Hello world`,
  };
};

export { handler };

```

The error message reported by `netlify-cli` is "**`Cannot find module 'exports'`**" (full error message below).

NOTE: the function _builds just fine on netlify and works_, it's just `netlify-cli` chokes on it.

If I comment out the usage of the `S3` imported function, the problem goes away (I presume because of esbuild tree-shaking or something).

I also made the recommended settings of `esModuleInterop` and `isolatedModules` in the `tsconfig.json` file.

Again, there's a super simple reproduction here:

https://github.com/jaredh159/netlify-cli-bug-repro

Just clone, npm install, and run `npm run dev` to see the issue.  Thanks in advance for any help, and again, really love the first-class typescript support! 🙏

```txt
◈ Function builder zip-it-and-ship-it building functions from directory /netlify-cli-bug-repo/netlify/functions
◈ Function builder zip-it-and-ship-it finished building functions from directory /netlify-cli-bug-repo/netlify/functions
    Error: In file "/netlify-cli-bug-repo/.netlify/functions-serve/test/test.js"
    Cannot find module 'exports'
    Require stack:
    - /netlify-cli-bug-repo/node_modules/@netlify/zip-it-and-ship-it/src/node_dependencies/resolve.js
    - /netlify-cli-bug-repo/node_modules/@netlify/zip-it-and-ship-it/src/node_dependencies/index.js
    - /netlify-cli-bug-repo/node_modules/@netlify/zip-it-and-ship-it/src/main.js
    - /netlify-cli-bug-repo/node_modules/netlify/src/deploy/hash_fns.js
[...]

PR #2228, opened 2021-04-25
fix: pass functions config to zisi when listing functions files

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2227

Still in draft as there are issues to fix before merging (I'll comment on the PR).

cc @eduardoboucas 

**- Test plan**

TBD

🐒

### DIFF
--- a/src/utils/get-functions.js
+++ b/src/utils/get-functions.js
@@ -34,13 +34,13 @@ const getWatchDirs = (functionsSrcDir, functions) => {
   return [functionsSrcDir, ...new Set(localFilesDirs)]
 }
 
-const getFunctionsAndWatchDirs = async (functionsSrcDir) => {
+const getFunctionsAndWatchDirs = async (functionsSrcDir, functionsConfig) => {
   if (!(await fileExistsAsync(functionsSrcDir))) {
     return { functions: [], watchDirs: [functionsSrcDir] }
   }
 
   // get all functions files so we know which directories to watch
-  const functions = await listFunctionsFiles(functionsSrcDir)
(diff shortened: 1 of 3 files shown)


src/utils/get-functions.js
```diff
@@ -34,13 +34,13 @@ const getWatchDirs = (functionsSrcDir, functions) => {
...
 
-const getFunctionsAndWatchDirs = async (functionsSrcDir) => {
+const getFunctionsAndWatchDirs = async (functionsSrcDir, functionsConfig) => {
   if (!(await fileExistsAsync(functionsSrcDir))) {
     return { functions: [], watchDirs: [functionsSrcDir] }
   }
 
   // get all functions files so we know which directories to watch
-  const functions = await listFunctionsFiles(functionsSrcDir)
+  const functions = await listFunctionsFiles(functionsSrcDir, { config: functionsConfig })
```
erezrokah, line 43: Since we're listing functions files to get the watch directories, we need to pass the bundler type.

One thing I noticed is that if the bundler is `esbuild` we return the bundled function files.
e.g.

```
[
  {
    srcFile: '<repo>/.netlify/functions-serve/test/test.js',
    name: 'test',
    mainFile: '<repo>/.netlify/functions-serve/test/test.js',
    runtime: 'js',
    extension: '.js'
  }
]
```

We will need `zisi` to return the original files required by the function to make this work. Also since `functionsSrcDir` is also set to `.netlify/functions-serve` by the function builder, we'll need to handle that too.

erezrokah:
Closing in favor of https://github.com/netlify/cli/pull/2236

PR #2236, opened 2021-04-26
feat: improve function builder's file watching service

eduardoboucas:
**- Summary**

With this PR, the ZISI function builder takes advantage of the `inputs` property added in version 3.8.0 of `@netlify/zip-it-and-ship-it`, so that it's capable of bundling only the functions that have been affected by a change event, as opposed to building every function each time.

**- Test plan**

No new tests added.


src/function-builder-detectors/zisi.js
```diff
@@ -1,18 +1,142 @@
...
 const { NETLIFYDEVERR } = require('../utils/logo')
 
-const bundleFunctions = async ({ config, sourceDirectory, targetDirectory }) => {
-  // @todo Build only the functions affected by the path that has changed.
-  await zipFunctions(sourceDirectory, targetDirectory, {
+const ZIP_CONCURRENCY = 5
+
+const zipFunctionsAndUpdateCache = async ({ cache, functions, sourceDirectory, targetDirectory, zipOptions }) => {
+  if (functions !== undefined) {
+    await pFilter(
```
eduardoboucas, line 16: Using `p-filter` just for its capability of handling concurrency. We're effectively mimicking what `zipFunctions` does, but limiting it to a subset of functions. Hence the `ZIP_CONCURRENCY` constant using the same default value as `zip-it-and-ship-it`.

src/utils/serve-functions.js
```diff
@@ -204,10 +206,8 @@ const createHandler = async function ({ dir, capabilities, omitFileChangesLog, w
     leading: false,
     trailing: true,
   })
-  watcher
-    .on('change', debouncedOnChange)
-    .on('unlink', debouncedOnUnlink)
-    .on('add', async (path) => {
```
eduardoboucas, line 210: I was seeing duplicate event callbacks by not debouncing the `add` event.

src/utils/serve-functions.js
```diff
@@ -430,28 +435,37 @@ const getFunctionsServer = async function ({ dir, omitFileChangesLog, siteUrl, c
...
 
 const getBuildFunction = ({ functionBuilder, log }) =>
-  async function build(updatedPath) {
-    log(
-      `${NETLIFYDEVLOG} Function builder ${chalk.yellow(functionBuilder.builderName)} ${chalk.magenta(
-        'building',
-      )} functions from directory ${chalk.yellow(functionBuilder.src)}`,
-    )
+  async function build(updatedPath, eventType) {
+    const relativeFunctionsDir = relative(cwd(), functionBuilder.src)
```
eduardoboucas, line 439: This is a bit outside of the scope of this PR, and I'm happy to break it out if it makes it easier to review, but I thought we could show the function path relative to the functions directory as opposed to showing the full path, which is a bit heavy on the output.

src/utils/serve-functions.js
```diff
@@ -484,9 +498,9 @@ const setupFunctionsBuilder = async ({ config, errorExit, functionsDirectory, lo
 
   const functionWatcher = chokidar.watch(functionBuilder.src)
   functionWatcher.on('ready', () => {
-    functionWatcher.on('add', debouncedBuild)
-    functionWatcher.on('change', debouncedBuild)
-    functionWatcher.on('unlink', debouncedBuild)
+    functionWatcher.on('add', (path) => debouncedBuild(path, 'add'))
```
eduardoboucas, line 501: The build function now receives the path of the updated file _and_ the type of event.

eduardoboucas:
It feels like we should be adding some tests around this functionality. @erezrokah what do you think?

erezrokah:
> It feels like we should be adding some tests around this functionality. @erezrokah what do you think?

:100

src/utils/serve-functions.js
```diff
@@ -153,16 +155,12 @@ const buildClientContext = function (headers) {
   }
 }
 
-const logBeforeAction = ({ omitLog, path, action }) => {
```
eduardoboucas, line 156: The `omitLog` property was introduced to avoid duplicate logging messages when using the ZISI function builder, but it's no longer necessary since the ZISI function builder now bypasses the base file watchers and thus removing duplicate events.

tests/utils/site-builder.js
```diff
@@ -130,11 +130,21 @@ const createSiteBuilder = ({ siteName }) => {
...
+      })
+      return builder
+    },
     buildAsync: async () => {
       for (const task of tasks) {
         // eslint-disable-next-line no-await-in-loop
         await task()
       }
+
+      tasks = []
```
eduardoboucas, line 146: By emptying the tasks array at the end of `buildAsync`, we're able to run new batches of tasks. For example, we can use `builder.withFunction` to create/update a function after the site has been built and served, which is useful when testing `netlify dev`.

tests/utils/site-builder.js
```diff
@@ -130,11 +130,21 @@ const createSiteBuilder = ({ siteName }) => {
       })
       return builder
     },
+    withoutFile: ({ path: filePath }) => {
```
eduardoboucas, line 133: Allows us to remove a file that has been previously created.

erezrokah (approved):

PR #2239, opened 2021-04-26
fix(command-dev): use custom functions timeout if present

erezrokah:
**- Summary**

The CLI synchronous functions timeout was hardcoded to 10 seconds.
This PR retrieves the timeout based on the site information and uses it, if present.

**- Test plan**

Added a test

🐼 

Fixes https://github.com/netlify/pod-workflow/issues/198 and https://answers.netlify.com/t/increase-function-timeout/26361/12

### DIFF
--- a/tests/command.dev.test.js
+++ b/tests/command.dev.test.js
@@ -15,6 +15,7 @@ const { copyFileAsync } = require('../src/lib/fs')
 const { withDevServer } = require('./utils/dev-server')
 const { startExternalServer } = require('./utils/external-server')
 const got = require('./utils/got')
+const { withMockApi } = require('./utils/mock-api')
 const { withSiteBuilder } = require('./utils/site-builder')
 
 const gotCatch404 = async (url, options) => {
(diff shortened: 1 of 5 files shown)


tests/command.dev.test.js
```diff
@@ -1633,6 +1634,67 @@ export const handler = async function () {
...
+
+        await withMockApi(routes, async ({ apiUrl }) => {
+          await withDevServer(
+            {
+              cwd: builder.directory,
+              args,
+              env: {
+                NETLIFY_API_URL: apiUrl,
+                NETLIFY_SITE_ID: 'site_id',
+                NETLIFY_AUTH_TOKEN: '',
```
erezrokah, line 1687: We need to set `NETLIFY_AUTH_TOKEN: ''` since our CI sets it for all tests.
Having an `NETLIFY_AUTH_TOKEN` makes `@netlify/config` thinks we want to access a "real" API:
https://github.com/netlify/build/blob/b442daf2efa2a31057e8628309903f6ec3976945/packages/config/src/options/main.js#L40

erezrokah:
Still looking into the test failure (it passes locally)

tests/command.dev.test.js
erezrokah: A better way to do it is https://github.com/netlify/cli/pull/2239/files#diff-e175b551bfa3b766a5f2559e0c3e037415c7034a04776cf76162ccbdcd0370aaR34

eduardoboucas (approved):

PR #2258, opened 2021-04-27
feat: memoize build function

eduardoboucas:
**- Summary**

Use memoization instead of debouncing the build function.

**- Test plan**

Reinstates the TypeScript tests.

cat-sample_1313

### DIFF
--- a/src/utils/serve-functions.js
+++ b/src/utils/serve-functions.js
@@ -328,6 +328,8 @@ const createFormSubmissionHandler = function ({ siteUrl, warn }) {
     const fakeRequest = new Readable({
       read() {
         this.push(req.body)
+        // TODO: remove when https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1057 is fixed
+        // eslint-disable-next-line unicorn/no-array-push-push
         this.push(null)
       },
     })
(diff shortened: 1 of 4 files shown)


src/utils/serve-functions.js
```diff
@@ -515,21 +515,35 @@ const setupFunctionsBuilder = async ({ config, errorExit, functionsDirectory, lo
...
+    const key = `${eventType}@${path}`
+
+    if (!cache.has(key)) {
+      cache.set(
+        key,
+        buildFunction(path, eventType).finally(() => cache.delete(key)),
+      )
+    }
+
+    return cache.get(key)
```
erezrokah, line 534: I couldn't find where the return value is used

src/utils/serve-functions.js
```diff
@@ -515,21 +515,35 @@ const setupFunctionsBuilder = async ({ config, errorExit, functionsDirectory, lo
...
-    leading: true,
-    trailing: true,
-  })
+  const cache = new Map()
+
+  const buildFunction = getBuildFunction({ functionBuilder, log })
+
+  // `memoizedBuild` will avoid consecutive calls to the build function with
+  // the same arguments until the previous invokation has finished running.
+  const memoizedBuild = (path, eventType) => {
```
erezrokah, line 524: If the user edits the same file multiple times only the first triggers a build?
Do we want to run the last one too?

erezrokah (changes requested):
Thanks @eduardoboucas, I think we might need a combination of debounce and memoize

src/utils/serve-functions.js
eduardoboucas: It's not used currently, but `debouncedBuild` preserved the return value of the build function and I thought `memoizedBuild` should respect that.

src/utils/serve-functions.js
```diff
@@ -515,21 +515,35 @@ const setupFunctionsBuilder = async ({ config, errorExit, functionsDirectory, lo
...
+    const key = `${eventType}@${path}`
+
+    if (!cache.has(key)) {
+      cache.set(
+        key,
+        buildFunction(path, eventType).finally(() => cache.delete(key)),
```
ehmicky, line 530: Would using an async-aware memoization library like [`moize`](https://github.com/planttheidea/moize) work here?
https://github.com/planttheidea/moize#ispromise

eduardoboucas:
In https://github.com/netlify/cli/pull/2258/commits/b9f99f49130755c5d431bda31a3a00a16b10f1de I've updated the `memoizeBuild` to ensure any updates performed while the function builder is running are executed afterwards, whilst at the same time being able to discard any duplicate filesystem events in the same way that `debounce` used to do.

The implementation is non-trivial, but hopefully the comments I've added help clarify the behaviour.

> If the user edits the same file multiple times only the first triggers a build?
> Do we want to run the last one too?

@erezrokah This should now be addressed.

> Would using an async-aware memoization library like moize work here?

I think this is a special case where we want a time-based memoization with a debounce interval. I'm not aware of any library that does exactly what we need, but I'm happy to explore alternatives.

erezrokah (approved):

ehmicky (approved):

erezrokah (approved):

PR #1704, opened 2021-01-07
Feat: use framework info

erezrokah:
### Summary

Fixes https://github.com/netlify/cli/issues/1495
Fixes https://github.com/netlify/cli/issues/1252

- [x] Verify this works with https://github.com/netlify/cli/pull/2312

Related to https://github.com/netlify/cli/pull/843

### Changes in behavior

Changes from #843

- `framework = "#custom"`. 
  - Current behavior - an error is shown if `command` and `targetPort` are not configured. 
  - New behavior - same as current.
- `framework` is not `#custom` and user has specified `command` and `targetPort`. 
  - Current behavior - an error is shown to set `framework = "#custom"`. 
  - New behavior - everything works as the CLI has all the information it needs to run. This allows running `ntl dev --command "echo hello" --targetPort 8000` which was not supported before (adding `--framework "#custom"` didn't work as well).
 - `dir` flag is specified.
    - Current behavior - we don't run framework detection, use a static server and show an error if `targetPort` or `command` are specified.
    - New behavior - we don't run framework detection and use a static server. 
      - If `command` is configured we print a warning and suggest to use `targetPort` for a custom setup.
      - If `targetPort` is configured we print a warning and suggest to use `staticServerPort` to configure the static server port.

Other changes
- `framework = "create-react-app"` (or any other specific framework)
  - Current behavior - run the `create-react-app` detector, [if it fails report an error](https://github.com/netlify/cli/blob/8152bb838177ee7c2666c6cfad8b9f0523fbe293/src/utils/detect-server.js#L135), otherwise use it
  - New behavior - always use the specific framework settings

- ~~Added a new `--framework` flag. You can only configure `framework` using the `dev` block at the moment. That doesn't make sense since we already have `command` and `targetPort` flags.~~ Done in https://github.com/netlify/cli/pull/2818

### Test plan

~~Testing this manually at the moment.~~
~~Will think of a good way to test the integration without having a performance hit.~~

~~Tested only the happy path, so others are probably broken.~~

~~Replaced the unit tests with integration tests~~

Tests were added in https://github.com/netlify/cli/pull/2821, I've updated them to reflect the new behavior

**- A picture of a cute animal (not mandatory but encouraged)**
image

PR #2259, opened 2021-04-27
fix(command-dev): Update detect-server.js

markbello:
This was a very confusing error message. "args" in this case is an implementation detail, and the error message didn't provide any insight as to what might have gone wrong.

In my case, it was using `craco` to override my base `create-react-app` config, which meant a change to the start script from `react-scripts-start` to `craco start`. I didn't know how to use a custom framework until after writing a craco-specific detector in my local install of the CLI

**- Summary**

**- Test plan**


erezrokah (approved):
Thanks @markbello this is much better.

FYI, we're in the process of refactoring the detectors code to make it easier to support and extend https://github.com/netlify/cli/pull/1704

Issue #2209, opened 2021-04-23
Test: improve CLI CI times


erezrokah:
Related to https://github.com/netlify/cli/issues/2202

The CLI tests can take up to 20 minuets sometimes (mostly on Windows).

This means the turnaround for an important bug fix can take around 40 minuets (tests for the fix PR and tests for the release PR).

We should investigate reducing the time it takes to run the tests.

**Can you submit a pull request?**

Yes

erezrokah:
One way to improve the CLI CI times is to use a mock API instead of hitting the API directly.

I think it's valuable to test the integration with the live API, but maybe we should move those tests to https://github.com/netlify/js-client

JGAntunes:
Related to https://github.com/netlify/engineering/issues/318.

erezrokah:
Closing this as we've done several improvements related to this:
1. Use a mock API in all relevant tests except deploy tests
2. We don't re-run the tests on release PRs
3. We don't re-run the tests before publishing

If we can think of more actions items on this, we can open individual tasks

erezrokah:
Re-opening as we need another integration on this. We should first benchmark the tests, but possible improvements:
1. Move to CircleCI and use stronger machines (GitHub actions Windows machines are especially slow).
2. Run the deploy tests in parallel (create multiple sites, instead [just one site](https://github.com/netlify/cli/blob/68fd36200899595180d3a29a24dc7550cfe56e2f/tests/command.deploy.test.js#L46)).
3. Large Media tests are also [serial](https://github.com/netlify/cli/blob/68fd36200899595180d3a29a24dc7550cfe56e2f/tests/command.lm.test.js#L45). I don't think we can do anything about that as Large Media installs globally on the machine.

PR #2315, opened 2021-05-06
test(windows): increase port limit and reduce time wait delay

erezrokah:
**- Summary**
Related to https://github.com/netlify/cli/issues/2209

Another attempt to improve Windows tests.
See https://docs.microsoft.com/en-us/biztalk/technical-guides/settings-that-can-be-modified-to-improve-network-performance#adjust-the-maxuserport-and-tcptimedwaitdelay-settings and https://www.ibm.com/docs/en/records-manager/8.5.0?topic=performance-configuring-maxuserport-maximum-user-ports

Default port limit is between 1024 to 5000 and default time wait delay is 120

**- Test plan**

Existing test

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,12 @@ jobs:
       fail-fast: false
 
     steps:
+      # This improves Windows network performance, we need this since we open many ports in our tests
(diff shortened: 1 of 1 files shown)


JGAntunes (approved):
windows

Issue #2325, opened 2021-05-06
bug(install): wrap chalk require in postinstall script


erezrokah:
**Describe the bug**

Related to https://github.com/netlify/cli/issues/2280 and https://github.com/netlify/cli/issues/2266

The error seems to be coming from the `postinstall` script with certain `npm` versions.

We should be able to handle this error by using a dynamic require and don't use `chalk` if we fail to require it.

**To Reproduce**

N/A

**Expected behavior**

If we can't require `chalk` we can still print the logs without it

PR #2330, opened 2021-05-09
fix(postinstall): wrap chalk require

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2325

Hopefully this fixes the installations issues (and not just move the error to when a user invokes the CLI).

**- Test plan**

N/A

🐒

### DIFF
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -1,30 +1,33 @@
 const id = (message) => message
 
-const postInstall = () => {
-  let chalk
+const format = (message, styles) => {
+  let func = id
   try {
     // this fails sometimes on outdated npm versions
     // eslint-disable-next-line node/global-require
-    chalk = require('chalk')
-  } catch (_) {
-    chalk = {
-      greenBright: { bold: { underline: id } },
-      cyanBright: { bold: id },
-    }
-  }
+    func = require('chalk')
+    styles.forEach((style) => {
+      func = func[style]
+    })
+  } catch (_) {}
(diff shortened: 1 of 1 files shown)


scripts/postinstall.js
```diff
@@ -1,15 +1,31 @@
...
-console.log('')
+const postInstall = () => {
+  let chalk
+  try {
+    // this fails sometimes on outdated npm versions
+    // eslint-disable-next-line node/global-require
+    chalk = require('chalk')
+  } catch (_) {
+    chalk = {
+      greenBright: { bold: { underline: id } },
```
erezrokah, line 11: Not sure if there's a better way to do this

eduardoboucas: If you don't want to be defining all the possible permutations of the chalk options, perhaps we could use something like:

```js
const format = (message, styles) => {
  let func = i => i
  
  try {
    func = require('chalk')
    
    styles.forEach(style => {
      func = func[style]
    })
  } catch (_) {}
  
  
  return func(message)
}
```

And then:

```js
// Instead of
chalk.greenBright.bold.underline(`Success! Netlify CLI has been installed!`)

// We'd do
format(`Success! Netlify CLI has been installed!`, ['greenBright', 'bold', 'underline'])
```

eduardoboucas (approved):
Sorry for the late review! Left a minor comment for an alternative implementation, but what you wrote works just as well, so I'm happy to go with whatever you prefer!

PR #2348, opened 2021-05-10
fix(command-dev): add DEPLOY_PRIME_URL

Xenonym:
**- Summary**

The `DEPLOY_PRIME_URL` environment variable is [available in the build environment](https://docs.netlify.com/configure-builds/environment-variables/#deploy-urls-and-metadata), but not when running `dev`.

**- Test plan**

1. Use `process.env.DEPLOY_PRIME_URL` in your code (eg. in a Netlify function).
2. `process.env.DEPLOY_PRIME_URL` should no longer be `undefined`.

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -274,6 +274,7 @@ class DevCommand extends Command {
 
     process.env.URL = url
     process.env.DEPLOY_URL = url
+    process.env.DEPLOY_PRIME_URL = url
 
     printBanner({ url, log })
   }


Xenonym:
@erezrokah oops, you are right 😅 Turns out I already had `bencao/netlify-plugin-inline-functions-env` installed on my site, which made `DEPLOY_URL` and `DEPLOY_PRIME_URL` available.

While removing `DEPLOY_URL` would bring `dev` into parity with the actual build environment, it would make local testing for me more difficult - some of my functions are providing `DEPLOY_PRIME_URL` as a callback to an external API. If I use `URL`, it breaks deploy previews as the external API would redirect from the deploy preview to the production site, which was not expecting that callback. 

1. Could `DEPLOY_URL` / `DEPLOY_PRIME_URL` be provided at function runtime? Alternatively, would there be a way to simulate build plugins, so that `bencao/netlify-plugin-inline-functions-env` can run?
2. It's not entirely clear what environemnt variables _are_ available to functions, if at all? [`bencao/netlify-plugin-inline-functions-env`](https://github.com/bencao/netlify-plugin-inline-functions-env) seems to suggest only environment variables in the Netlify UI would be available to functions, but `URL` already seems to be an exception.
3. While I _could_ update this PR to remove `DEPLOY_URL` instead, that seems like a completely different change. Should I close this PR and submit a new one instead?

erezrokah:
Hi @Xenonym, I think we can derive a few issues from this PR and your use case:

1. Document which built-in environment variables are exposed to functions
2. Consider exposing more built-in variables (like `DEPLOY_URL` to functions)
3. Align the CLI with whatever our production environment is doing

I'll raise the first 2 internally.
Once we handle 1, we should be able to do 3.

Closing the PR for now. I don't think we need to open a new one. We'll do that once we know what's missing in the CLI

Issue #2302, opened 2021-05-04
Support base configuration in `init` command


erezrokah:
**Is your feature request related to a problem? Please describe.**

The `netlify init` command only prompts for publish dir, build command and functions directory.
We should also add support for the base directory.

**Describe the solution you'd like**

Prompting the user for a base directory, default is undefined

**Describe alternatives you've considered**

Letting the user configure it manually via the UI, after the first build fails.

**Additional context**

N/A

**Can you submit a pull request?**

Yes

- [x] https://github.com/netlify/cli/issues/2380

erezrokah:
Related to https://github.com/netlify/cli/issues/433

verythorough:
Note that there's some trickiness here with the `netlify.toml` file:
- If the `init` flow is pulling settings from the toml file, it should look for the file in the `base` directory, so the base directory prompt should probably be first, before the build command & publish directory. (It will also be first in the UI.)
- The end of the `init` flow asks whether the user would like to save the settings to a `netlify.toml` file. If the `base` directory is set, the file should be saved there, and it should _not_ include the `base` setting. (The `base` setting _could_ be set in a root toml file, but this won't work if there's more than one sub-project in the repo.)

It occurs to me that the first point is also relevant for framework heuristics, which should also be run in the `base` directory.

erezrokah:
Great points @verythorough!

We need to consider that the user can run `netlify init` from within the repository root or from within the base directory itself.

Looks like running from within the base directory will work as we search the config file from that directory and up:
https://github.com/netlify/build/blob/6ed8ab9f55690981d8f6c1216ebb1b5bc0efd8c4/packages/config/src/path.js#L16
cc @ehmicky can you confirm?

If the user runs `netlify init` from the repository root, we'll need to pass the `base` to `netlify/config` so the correct file will be resolved.
That is a bit tricky as at the moment the first thing we do in the CLI is to resolve the configuration, so we would need to call `netlify/config` a second time after getting the `base` directory from the user.

ehmicky:
Yes, we are searching the config file from that directory and up :+1:

verythorough:
> We need to consider that the user can run `netlify init` from within the repository root or from within the base directory itself.

Oh, I didn't think about that!

So I did some experimenting with that workflow, using a basic monorepo testing repo I had:
https://github.com/verythorough/monorepo-test

1. Ran `ntl init` in `/projects/project1` (no config file in that folder), while there was no config file in the root:
[...]

PR #2378, opened 2021-05-13
test(mock-api): use object comparison

erezrokah:
**- Summary**

A bit of refactoring before opening a PR for https://github.com/netlify/cli/issues/2302 to make the upcoming PR easier to review.

This PR changes the mock API to use object comparison instead of string comparison when matching request body to allow comparing the same object with different properties order.

**- Test plan**

Exiting tests

🐐

### DIFF
--- a/tests/utils/mock-api.js
+++ b/tests/utils/mock-api.js
@@ -1,5 +1,6 @@
 const bodyParser = require('body-parser')
 const express = require('express')
+const equal = require('fast-deep-equal')
 
 const addRequest = (requests, request) => {
   requests.push({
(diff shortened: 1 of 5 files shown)


tests/utils/mock-api.js
```diff
@@ -19,7 +20,7 @@ const startMockApi = ({ routes }) => {
   routes.forEach(({ method = 'get', path, response = {}, status = 200, requestBody }) => {
     app[method.toLowerCase()](`/api/v1/${path}`, function onRequest(req, res) {
       // validate request body
-      if (requestBody !== undefined && requestBody !== JSON.stringify(req.body)) {
+      if (requestBody !== undefined && !equal(requestBody, req.body)) {
```
ehmicky, line 23: Would [`util.isDeepStrictEqual()`](https://nodejs.org/api/util.html#util_util_isdeepstrictequal_val1_val2) work? This is supported now that we've dropped Node 8 from the CLI.

erezrokah: Yes! I thought about using `assert.deepStrictEqual` but didn't want a method that throws. Didn't know about `util.isDeepStrictEqual` I'll give it a try

erezrokah: Done in https://github.com/netlify/cli/pull/2378/commits/3937df4a05dc65878db32a515c6c140f1fd03d2d

ehmicky (approved):

PR #2382, opened 2021-05-13
fix(command-init): use main instead of master

rockorager:
**- Summary**
Change `git push -u origin master` to `git push -u origin main` to match github default branch name

**- Test plan**
None, only changed one word

Not today, folks!


erezrokah (approved):
Thanks @rockorager, nice catch!

PR #2394, opened 2021-05-14
chore(deps): remove dependency random-item

Nixinova:
## Summary

Removes dependency [`random-item`](https://www.npmjs.com/package/random-item).
It is only used once and is very simple so there is no loss of readability when it is inlined.
Removes an item from the renovate blocked list.

## Test plan

This uses the exact same code that the dependency uses: `arr[Math.floor(Math.random() * arr.length)]`
https://github.com/sindresorhus/random-item/blob/main/index.js#L6
https://github.com/Nixinova/fork_netlify-cli/blob/8c53236586512923ce2379ec4f4219998c6087e3/src/commands/deploy.js#L632

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -103,7 +103,6 @@
         "path-type": "^4.0.0",
         "prettyjson": "^1.2.1",
         "pump": "^3.0.0",
-        "random-item": "^3.0.0",
         "raw-body": "^2.4.1",
         "resolve": "^1.12.0",
         "safe-join": "^0.1.3",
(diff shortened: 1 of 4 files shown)


erezrokah (changes requested):
Hi @Nixinova, thank you for the PR.

Can you please explain what is the problem the PR is solving?

By using a well maintained dependency instead of inlining we can have the benefit of of upstream bug fixes without the maintenance cost (even 1 line of code can have bugs).

Nixinova:
Well it doesn't have bugs and this removes one item from the install process and from renovate's major version backlog.

erezrokah:
Hi @Nixinova,

> Well it doesn't have bugs

It might, see https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Modulo_bias and this can be improved via https://nodejs.org/api/crypto.html#crypto_crypto_randomint_min_max_callback

> this removes one item from the install process

It doesn't really removes it as the code is now part of the CLI codebase (and without any proper license/credit to the author, unit tests or type definitions). Also inlining the code means that if another dependency uses it - it will be duplicated.

> and from renovate's major version backlog

True, but we prefer to clear it when we drop support for older Node.js versions.

> Updates to the package are just to change the min required Node version.

See https://github.com/sindresorhus/random-item/pull/2 and https://github.com/sindresorhus/random-item/pull/4 for non min Node.js version changes.

Closing the PR.

Issue #2380, opened 2021-05-13
feat(command-init): prompt for base directory if cwd is not repository root


erezrokah:
**Is your feature request related to a problem? Please describe.**

Related to https://github.com/netlify/cli/issues/2302#issuecomment-837388313

This is an issue to describe the first iteration in that comment.

**Describe the solution you'd like**

When running `ntl init` from a subdirectory it should prompt the user to set the `base` directory to `cwd`

**Describe alternatives you've considered**

Making users fix the UI settings post `ntl init`

**Additional context**

Somehow related to https://github.com/netlify/cli/issues/433

**Can you submit a pull request?**

Yes

PR #2401, opened 2021-05-16
feat(command-init): prompt for base

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2380

**- Test plan**

I'll add tests soon

image

### DIFF
--- a/src/utils/init/utils.js
+++ b/src/utils/init/utils.js
@@ -1,5 +1,6 @@
 const { EOL } = require('os')
 const path = require('path')
+const process = require('process')
 
 const chalk = require('chalk')
 const cleanDeep = require('clean-deep')
@@ -12,16 +13,29 @@ const { getFrameworkInfo } = require('./frameworks')
 const { detectNodeVersion } = require('./node-version')
 const { getPluginsList, getPluginInfo, getRecommendPlugins, getPluginsToInstall, getUIPlugins } = require('./plugins')
 
-const normalizeDir = ({ siteRoot, dir, defaultValue }) => {
+const normalizeDir = ({ baseDirectory, dir, defaultValue }) => {
   if (dir === undefined) {
     return defaultValue
   }
 
-  const relativeDir = path.relative(siteRoot, dir)
+  const relativeDir = path.relative(baseDirectory, dir)
   return relativeDir || defaultValue
 }
 
-const getDefaultSettings = ({ siteRoot, config, frameworkPlugins, frameworkBuildCommand, frameworkBuildDir }) => {
+const getDefaultBase = ({ repositoryRoot, baseDirectory }) => {
+  if (baseDirectory !== repositoryRoot && baseDirectory.startsWith(repositoryRoot)) {
+    return path.relative(repositoryRoot, baseDirectory)
+  }
+}
+
+const getDefaultSettings = ({
+  repositoryRoot,
+  config,
+  baseDirectory,
+  frameworkPlugins,
+  frameworkBuildCommand,
+  frameworkBuildDir,
+}) => {
   const recommendedPlugins = getRecommendPlugins(frameworkPlugins, config)
   const {
     command: defaultBuildCmd = frameworkBuildCommand,
@@ -30,9 +44,10 @@ const getDefaultSettings = ({ siteRoot, config, frameworkPlugins, frameworkBuild
   } = config.build
 
   return {
+    defaultBaseDir: getDefaultBase({ repositoryRoot, baseDirectory }),
     defaultBuildCmd,
-    defaultBuildDir: normalizeDir({ siteRoot, dir: defaultBuildDir, defaultValue: '.' }),
-    defaultFunctionsDir: normalizeDir({ siteRoot, dir: defaultFunctionsDir, defaultValue: 'netlify/functions' }),
+    defaultBuildDir: normalizeDir({ baseDirectory, dir: defaultBuildDir, defaultValue: '.' }),
+    defaultFunctionsDir: normalizeDir({ baseDirectory, dir: defaultFunctionsDir, defaultValue: 'netlify/functions' }),
     recommendedPlugins,
   }
 }
@@ -41,10 +56,17 @@ const getPromptInputs = async ({
   defaultBuildCmd,
   defaultBuildDir,
   defaultFunctionsDir,
+  defaultBaseDir,
   recommendedPlugins,
   frameworkName,
 }) => {
   const inputs = [
+    defaultBaseDir !== undefined && {
+      type: 'input',
+      name: 'baseDir',
+      message: 'Base directory (e.g. projects/frontend):',
+      default: defaultBaseDir,
+    },
     {
       type: 'input',
       name: 'buildCmd',
@@ -64,7 +86,7 @@ const getPromptInputs = async ({
       message: 'Netlify functions folder:',
       default: defaultFunctionsDir,
     },
-  ]
+  ].filter(Boolean)
 
   if (recommendedPlugins.length === 0) {
     return inputs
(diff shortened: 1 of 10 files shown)


src/utils/init/utils.js
```diff
@@ -104,29 +126,35 @@ const getPromptInputs = async ({
   ]
 }
 
-const getBuildSettings = async ({ siteRoot, config, env, warn }) => {
-  const nodeVersion = await detectNodeVersion({ siteRoot, env, warn })
+const getBaseDirectory = ({ config }) => config.build.base || process.cwd()
```
erezrokah, line 129: This is an important change. We used to pass `site.root` everywhere, and now we pass `base` or `process.cwd`.
`site.root` is the value of `buildDir` as returned from `@netlify/config`:
https://github.com/netlify/cli/blob/bc47458f32bffe14e5ecf90d1a30b150931e907d/src/utils/command.js#L118

In a monorepo this can evaluate to the repository root, even when running the CLI from a subdirectory, in case of the following structure:
```
├── .git
├── index.html
├── netlify.toml
└── projects
    ├── project1
    │   ├── index.html
    └── project2
        └── index.html
```

This happens since we `find-up` `netlify.toml` when looking for the `buildDir`.

We can consider fixing it in `@netlify/config` too, but I'm not sure what the implications are yet.

ehmicky: The `buildDir` returned by `@netlify/config` should be the monorepo subdirectory providing the user `cd` into it.

The current logic when `@netlify/config` is called in the CLI is:
  - `repositoryRoot` should be the first parent `.git`. If none, `cwd` is used. ([link to code](https://github.com/netlify/build/blob/main/packages/config/src/options/repository_root.js))
  - If `cwd` is a subdirectory inside `repositoryRoot` and if it has a `.netlify` or `netlify.toml`, then it is used as default vaule for `build.base`. This is also performed for any directory above `cwd` but below `repositoryRoot` ([link to code](https://github.com/netlify/build/blob/main/packages/config/src/options/base.js))
  - `buildDir` is `build.base` if defined. Otherwise, it is `repositoryRoot` ([link to code](https://github.com/netlify/build/blob/9c8c1929566eb561bac2bfd69cbe388b33cdce0e/packages/config/src/files.js#L62))

If I'm reading your comment correctly, is the problem that the monorepo is not detected due to lacking of a `.netlify` or `netlify.toml` file?

erezrokah: - On a new repo (no `.netlify` or `netlify.toml`) in any of the directories (root or sub) `buildDir` will always evaluate to `repositoryRoot` whereas the CLI we needs it to be `cwd` when running `netlify init`.
- On a repo with `.netlify.toml` in the root (but not in any of the sub directories) `buildDir` will evaluate to `repositoryRoot` which is also undesired. In addition config values will be taken from the root `netlify.toml`.

We might want to always resolve configuration based on the `cwd` in the CLI instead of using `find-up`. I need to give it some thought.

ehmicky: I am wondering whether the problem might be that we are currently using the same variable `buildDir` with different meanings. At the moment, the `buildDir` returned by `@netlify/config` is meant as "repository root or monorepo package root" (not monorepo package subdirectory). On the other hand, `cwd` might locally be inside a subdirectory of a monorepo package (not the root).

Do you think the following would solve the original problem, or not?
  - Making `@netlify/config` lookup algorithm look for `package.json` (on top of `.netify` and `netlify.toml`)
  - In the CLI, use `buildDir` when the monorepo root is needed, and `process.cwd()` when it's not

verythorough:
Thanks for building this!

I tested locally from the branch and I noticed a couple things:

First, if the directory is nested more than one level deep (e.g. `/projects/project1`), my base directory was saved with lovely Windows backslashes in it, causing a build failure.
build-error
backslash

The second issue seems to be a separate from this PR, since it also happened with a UI-created site where I set the base after the first deploy. For publish and functions directories, the log says it's overriding with config file values, but there's no config file:
no-config-file

However, it _did_ choose the right path for both things, and the log line is correct about what the values look like in the UI. So everything works, it's just confusing because it mentions a non-existent config file.

Otherwise, it looks great. Looking forward to seeing this released!

erezrokah:
> First, if the directory is nested more than one level deep (e.g. `/projects/project1`), my base directory was saved with lovely Windows backslashes in it, causing a build failure.

Great catch - that explains why the Windows tests are failing 😄

> The second issue seems to be a separate from this PR, since it also happened with a UI-created site where I set the base after the first deploy. For publish and functions directories, the log says it's overriding with config file values, but there's no config file:

I'll track down this warning message and open an issue on the relevant repo

src/utils/init/utils.js
erezrokah: Thanks for the reply @ehmicky,
I'm not sure I understand what is the difference between `monorepo package root` and `monorepo package subdirectory`.
What would those be in the structure described [here](https://github.com/netlify/cli/pull/2401#discussion_r635478262).

The use case we're optimizing for at the moment (in monorepos) is that the user runs `netlify init` from the project/package directory (in the above structure it will be either `project1` or `project2` and the monorepo packages are configured as `workspaces: ["projects/**"]`).

> * Making `@netlify/config` lookup algorithm look for `package.json` (on top of `.netify` and `netlify.toml`)

Yes, that will help in monorepos that have a `package.json` file(s).

> * In the CLI, use `buildDir` when the monorepo root is needed, and `process.cwd()` when it's not

I would say always use `process.cwd()` in `netlify init` from the CLI, unless a user configures a `base` directory.
The latter can happen if a user already linked the site/have `netlify.toml` with a `base` directory configured.

ehmicky: > I'm not sure I understand what is the difference between monorepo package root and monorepo package subdirectory.

I mean it as (using the example above): monorepo root would be `/projects/project1` while monorepo subdirectory would be `/projects/project1/some/sub/directory`.

> The use case we're optimizing for at the moment (in monorepos) is that the user runs netlify init from the project/package directory (in the above structure it will be either project1 or project2 and the monorepo packages are configured as workspaces: ["projects/**"]).

I am wondering if the underlying problem would be our inability to detect monorepo roots in some situations? Our current logic in `@netlify/config` relies on either: 
  - `.netlify` or `netlify.toml` being present in the monorepo package (not the root)
  - user specifying the `base` directory 
  
The case you're describing is failing those requirements since it is lacking a `netlify.toml` in the monorepo packages and the user has not specified a `base` directory yet.

> I would say always use process.cwd() in netlify init from the CLI, unless a user configures a base directory.
The latter can happen if a user already linked the site/have netlify.toml with a base directory configured.

This makes sense. I would add: if a monorepo package has a `package.json` and is a parent to `process.cwd()`, we probably should use this instead.

It seems like this logic might be implemented by: 
  - Adding `package.json` detection to `@netlify/config` 
  - If its returned `buildDir` is different from its returned `repositoryRoot`, use `buildDir` in `netlify init` since it indicates a monorepo package has been detected
  - Otherwise, use `process.cwd()`

What do you think?

erezrokah: > This makes sense. I would add: if a monorepo package has a `package.json` and is a parent to `process.cwd()`, we probably should use this instead.

Yes that makes sense and solves the subdirectory use case you mentioned.

> What do you think?

💯  I'll submit a PR to `@netlify/config` to detect `package.json`.

For this PR it looks like we should change the condition here:
https://github.com/netlify/cli/pull/2401/files#diff-f89b1876ce20ea7649a59d7113b68804eae8cac4522ddd87d5ae9afc41ac1c66R130

to use `buildDir` if `repositoryRoot !== buildDir`, otherwise use `process.cwd()`

ehmicky: This sounds good!

erezrokah: Done in https://github.com/netlify/cli/pull/2401/commits/3ede87388c46f3fe70603c73b5ee2e24d932ae96

ehmicky (approved):

PR #2432, opened 2021-05-20
chore(telemetry): switch the telemetry backend via redirects

JGAntunes:
**- Summary**

This is a follow up of the work in https://github.com/netlify/cli-telemetry-service/issues/14 and developed under https://github.com/netlify/pod-workflow/issues/160

**- Test plan**

I've tested this work locally, however I'll probably give the redirects a try once the deploy preview is up and running 👍 

🐥

### DIFF
--- a/functions/telemetry.js
+++ b/functions/telemetry.js
@@ -4,7 +4,7 @@ const TELEMETRY_SERVICE_URL = 'https://cli-telemetry.netlify.engineering'
 
 // This function is a workaround for our inability to redirect traffic to a ntl function in another site
 // using redirects (see https://github.com/netlify/cli-telemetry-service/issues/14)
-const handler = function ({ path, httpMethod, headers, body }) {
+const handler = async function ({ path, httpMethod, headers, body }, context) {
   const upstreamPath = path.replace(/^\/telemetry\//, '/')
 
   // Filter out some headers that shouldn't be fwded
@@ -25,6 +25,7 @@ const handler = function ({ path, httpMethod, headers, body }) {
   // We don't wait for the telemetry service response because the CLI does not care about
   // it. We want to return as soon and as fast as possible while making these requests in
   // the background.
+  context.callbackWaitsForEmptyEventLoop = false
   response
     // eslint-disable-next-line promise/prefer-await-to-then
     .then((res) => console.log(`Telemetry service responded with ${res.status}`))
--- a/functions/telemetry.js
+++ b/functions/telemetry.js
@@ -1,0 +1,41 @@
+const fetch = require('node-fetch')
+
+const TELEMETRY_SERVICE_URL = 'https://cli-telemetry.netlify.engineering'
+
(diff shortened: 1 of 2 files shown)


ehmicky (approved):

ehmicky (approved):

netlify.toml
```diff
@@ -13,14 +13,14 @@
 [[redirects]]
   # Telemetry link
   from = "/telemetry/track"
-  to = "https://api.netlify-services.com/telemetry/track"
-  status = 200
+  to = "https://cli-telemetry.netlify.engineering/.netlify/functions/track"
+  status = 302
```
erezrokah, line 17: Redirect instead of rewrite seems to work, with the cost of the additional hop. Can you confirm this @JGAntunes?

JGAntunes: The problem with a `3xx` redirect is that the clients will be changing their request methods from `POST` to `GET`. E.g. [check the logs for the last hour](https://app.netlify.com/sites/cli-telemetry-proxy/functions/track)

erezrokah: 🤦

functions/telemetry.js
```diff
@@ -0,0 +1,41 @@
+const fetch = require('node-fetch')
+
+const TELEMETRY_SERVICE_URL = 'https://cli-telemetry.netlify.engineering'
+
+// This function is a workaround for our inability to redirect traffic to a ntl function in another site
+// using redirects (see https://github.com/netlify/cli-telemetry-service/issues/14)
+const handler = function ({ path, httpMethod, headers, body }) {
```
erezrokah, line 7: I think we need to explicitly use `async` in the signature otherwise we need to invoke the `callback` argument.

Actually we might need to use `callback` with `callbackWaitsForEmptyEventLoop = false` to get the outcome we want. See https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html.

With `callbackWaitsForEmptyEventLoop = false` there's the side effect of freezing the event loop

JGAntunes: > With callbackWaitsForEmptyEventLoop = false there's the side effect of freezing the event loop

yeah I was just reading through it and I don't think that will cut it also 😕 because we would ideally want the processing to continue but not be frozen until the next execution.

I think we might need to just wait for the response, at least for now.

JGAntunes: Yeah, tried it out under the latest commit, managing connection state becomes an issue:

image

Logs:
https://app.netlify.com/sites/cli/functions/telemetry?scope=deploypreview:2432
https://app.netlify.com/sites/cli-telemetry-proxy/functions/track

JGAntunes: Changed the logic to wait for the response 👍

JGAntunes: FYI, in terms of response time, getting between 600ms to 1s (for cold executions)

erezrokah (approved):

Issue #409, opened 2019-07-16
Any way to debug functions?


fineline:
Not a bug or feature request but an enquiry:

What context do javascript lambda functions run in when using netlify dev? Is there any way to pass the --inspect flag so a debugger can be connected like a standard node app? 

Thanks, and great work on netlify dev.

modermo:
Not sure if this is what @fineline needs, but for me, being able to attach a local instance of the chrome debugger so that I can step through a lambda function is in orders of magnitude more helpful than just `console.log()` statements.

sw-yx:
what do we need to do on our end to make that happen? never done this before. i guess node doesnt just do this by default? 😂

8eecf0d2:
I haven't looked at how `netlify-cli` works, but I'd assume the following will work:

```
node --inspect node_modules/.bin/netlify-cli dev
```

All that's really needed to make this work is to pass the `--inspect` flag to the node process which executes the actual function code (presumably the `netlify-cli` command itself).

If the above works, all that @sw-yx and the team would really need to do is provide their own flag (eg, `--node-flag-inspect`) which would pass the `--inspect` flag to the node process which executes the functions - this would only serve to clean up the syntax.

sudall:
@8eecf0d2 I tried `node --inspect node_modules/netlify-cli/bin/run dev`, but it doesn't seem to work. I think you might be right about the --inspect on the child lambda process that `netlify dev` is making.

sudall:
I have some more progress. I'm able to hit breakpoints in the built js files in Webstorm and Chrome Debugger. Note that the debugger has to point to port 9229 and not the port the lambdas are served from. https://nodejs.org/de/docs/guides/debugging-getting-started/.

I think the next step is to figure out how to generate source-maps with netlify-lambda so I can hit ts breakpoints.

sudall:
`npx --node-arg=--inspect netlify dev` also appears to work

sudall:
Okay I got it working with TypeScript. For anyone wondering:

```
    "build:lambda": "netlify-lambda build src/lambda --config ./webpack.config.dev.js",
    "dev": "npx --node-arg=--inspect netlify dev"
```

in the webpack.config.dev.js

```
module.exports = {
    devtool: 'inline-source-map',
    optimization: { minimize: false }
};
```

run `dev` script and attach a debugger

mcrowder65:
[...]

PR #2440, opened 2021-05-23
docs: VS code debugging config & MRE

MentalGear:
Vs Code debug config


MentalGear:
As proposed by @erezrokah
https://github.com/MentalGear/netlify-dev_debug_vs-code/pull/1#issuecomment-846510622

erezrokah:
Sorry for the force push @MentalGear, I though the missing status checks are because you were missing a rebase, but it's because of https://github.com/netlify/cli/blob/a2e10672b4695cd69d430e0a263d7c3e1d0175b5/.github/workflows/main.yml#L11

MentalGear:
> Sorry for the force push @MentalGear, I though the missing status checks are because you were missing a rebase, but it's because of
> 
> https://github.com/netlify/cli/blob/a2e10672b4695cd69d430e0a263d7c3e1d0175b5/.github/workflows/main.yml#L11

NP. So this pull request should be good to go, right ?

erezrokah:
Sorry for the delay @MentalGear, I'll review it today

docs/functions-dev.md
```diff
@@ -42,3 +42,66 @@ To do so set the `--inspect` Node.js option when starting the functions server:
...
+Put the below configuration in your repo's `.vscode/launch.json`.
+
+Then in `VS Code`:
+1. Open `Run and Debug` Sidebar. Make sure "Caught Exceptions" is deactivated.
+2. In the dropdown, select `Fullstack Debugging`
+3. Run the debugger. A chrome window should open. Wait until content is loaded (may take a few seconds for server to be ready).
+
+```jsonc
+{
+  // Use IntelliSense to learn about possible attributes.
```
erezrokah, line 58: Can we remove the code comments to brevity? 
It seems we don't need the `https://go.microsoft.com/fwlink/?linkid=830387` and the `FULLSTACK NETLIFY DEV DEBUGGING` steps are already described above

docs/functions-dev.md
```diff
@@ -42,3 +42,66 @@ To do so set the `--inspect` Node.js option when starting the functions server:
...
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  // ---- FULLSTACK NETLIFY DEV DEBUGGING ----
+  // 1. Make sure "Caught Exceptions" is deactivated in your VS Code debugger
+  // 2. Run 'Fullstack Debugging' from the debug interface
+  // 3. A chrome browser window will open, however it may take a few seconds
+  // before the server is ready and content is displayed.
+  "compounds": [
```
erezrokah, line 67: Do we need the Chrome client? This ties the configuration to a specific browser which might not be installed on some machines. Also I was prompted to install a VS Code extension to get it working.

I was able to debug the CLI by using the following minimal config:
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Launch Program",
      "skipFiles": ["<node_internals>/**"],
      "program": "${workspaceFolder}/node_modules/.bin/netlify",
      "args": ["dev"]
    }
  ]
}
```

Also we should instruct users to install the CLI locally via `npm install --save-dev netlify-cli` for this to work

docs/functions-dev.md
```diff
@@ -42,3 +42,66 @@ To do so set the `--inspect` Node.js option when starting the functions server:
...
+        "webpack:///*": "${workspaceRoot}/*"
+      }
+    },
+    {
+      "name": "netlify dev",
+      "type": "node",
+      "request": "launch",
+      "skipFiles": ["<node_internals>/**"],
+      "program": "${workspaceFolder}/node_modules/.bin/netlify",
+      "args": ["dev", "--inspect"],
```
erezrokah, line 90: I believe `"--inspect"` can be removed

docs/functions-dev.md
```diff
@@ -42,3 +42,66 @@ To do so set the `--inspect` Node.js option when starting the functions server:
...
+      "sourceMapPathOverrides": {
+        "webpack:///*": "${workspaceRoot}/*"
+      }
+    },
+    {
+      "name": "netlify dev",
+      "type": "node",
+      "request": "launch",
+      "skipFiles": ["<node_internals>/**"],
+      "program": "${workspaceFolder}/node_modules/.bin/netlify",
```
erezrokah, line 89: As mentioned before this only works after installing the CLI locally so we should instruct users to do so.

docs/functions-dev.md
```diff
@@ -42,3 +42,66 @@ To do so set the `--inspect` Node.js option when starting the functions server:
...
+      "args": ["dev", "--inspect"],
+      "resolveSourceMapLocations": [
```
erezrokah, line 91: I don't think this is required as a default configuration.

I would rather have another section that explains about source maps and the purpose of this setting, as this is a more advanced use case.

erezrokah (changes requested):
Hi @MentalGear, thank you for submitting the PR and sorry for the late reply.

I added some comments targeted at simplifying the default configuration as much as possible.

Please let me know what you think.

docs/functions-dev.md
MentalGear: As I understand, netlify dev is used with js/ts/node. I would think then that a) using node modules would be pretty default for this use case b) as is the need to exclude them from source maps.

For me at least, without this source-map config, I get lots of errors in the debug console on startup.  I remember also reading that without proper sourcemap config breakpoints might not work (I didn't test that).

However, also happy to spawn this out into it's own section if you prefer. :)

What concern me a more however is, that I noticed that errors are not forwarded to `debug console` or vs code.

docs/functions-dev.md
MentalGear: You're the expert on the internal working of it, so if it serves no (upcoming) purpose I'll remove it.

MentalGear:
Thanks for the review. I added to your comments above and updated the guide with the dependency note as well as a more straightforward numbered instruction.

docs/functions-dev.md
```diff
@@ -42,3 +42,80 @@ To do so set the `--inspect` Node.js option when starting the functions server:
 - On Mac/Linux run `NODE_OPTIONS=--inspect netlify functions:serve`
 
 Then attach any Node.js debugger to the CLI process to debug your functions.
+
+
+## VS Code Debugging
+
+### 1. Setup
+Install netlify dev as a local dev dependency:
```
erezrokah, line 50:
```suggestion
Install Netlify CLI as a local dev dependency:
```

docs/functions-dev.md
erezrokah: I think we can remove both in favor of a `to learn more about VSCode launch configurations, see (here)[https://code.visualstudio.com/docs/editor/debugging#_launch-configurations]`

Or something similar. That way we can be brief and provide users the additional information without duplicating it.

docs/functions-dev.md
erezrokah: The CLI ignores that flag so it serves no purpose (it's not a CLI flag, it's a Node.js flag).

docs/functions-dev.md
erezrokah: This section is about functions debugging and I see it serving 2 purposes:
1. Debugging only functions with `netlify functions:serve`
2. Debugging functions with redirects, headers, etc with `netlify dev`

As for frontend debugging, I don't think it's related to the original issue https://github.com/netlify/cli/issues/409, so can we descope it from this PR?

I agree this is a valid use case, so having another section for it would makes sense.
In that section I prefer to link to https://code.visualstudio.com/blogs/2016/02/23/introducing-chrome-debugger-for-vs-code or https://code.visualstudio.com/docs/editor/debugging instead of duplicating content that might go stale.

Please let me know if that makes sense, and sorry for not being clear enough on this from the start.

docs/functions-dev.md
erezrokah: > As I understand, netlify dev is used with js/ts/node

I'm not sure I understand this statement. `netlify dev` runs as a regular Node.js executable and doesn't require any source maps.

> For me at least, without this source-map config, I get lots of errors in the debug console on startup. I remember also reading that without proper sourcemap config breakpoints might not work (I didn't test that).
> 
> However, also happy to spawn this out into it's own section if you prefer. :)

Could this be related to frontend debugging? Have you seen this for functions too? If so, can you share an example repo? By default, breakpoints in JavaScript functions should work.

erezrokah (changes requested):
Thanks for the follow up @MentalGear.

I've added a few comments to clarify my intention with the initial review.

docs/functions-dev.md
MentalGear: All right, I will add link towards the official vs code guide and the netlify guide.

docs/functions-dev.md
MentalGear: Thanks for clarifying that as indeed I overlooked that the page is called "Serving functions locally" with only the subsection being about local function debugging.

Since you also agree that having fullstack debugging for netlify-cli is valid, I would then rather propose to simply spawn out the whole debugging topic into it's own page where we can have both functions (backend) and frontend (chrome client) debugging mentioned. What do you think ?

docs/functions-dev.md
MentalGear: I think VS code debugger may need access to the sourcemaps.

The error messages are showing up in the netlify dev debug console, not the client console, so definitely backend related.

You can reproduce it by using my MRE https://github.com/MentalGear/netlify-dev_debug_vs-code/issues and comment out in `launch.json`:
- `resolveSourceMapLocations `
- "console": "integratedTerminal" (this I added later as vs-code `debug console` doesn't forward all the std output as integratedTerminal does)

```jsonc
      // "console": "integratedTerminal" // avoids logs & errors from being suppressed
      // "resolveSourceMapLocations": [
      //   // NOTE: your linter may think this property shouldn't be here
      //   // yet it works for finding the source maps
      //   "${workspaceFolder}/**",
      //   "!**/node_modules/**"
      // ]
```

MentalGear (commented):
Thanks for replying.

I adjusted the content were necessary and replied to your questions. Overall, I'm in favor though to spawn out the whole debugging into its own page for best DX coverage. Let me know your thought.

docs/functions-dev.md
erezrokah: > Since you also agree that having fullstack debugging for netlify-cli is valid, I would then rather propose to simply spawn out the whole debugging topic into it's own page where we can have both functions (backend) and frontend (chrome client) debugging mentioned. What do you think

Yes that makes sense.

MentalGear:
I have spawn out the Debugging part into its own page and added the discussed changes. @erezrokah.

erezrokah:
Sorry for the very late follow up @MentalGear. I was finally able to dig more into this.
I think I was able to further reduce the launch configurations size and also avoid the usage of an extension.
Also, now VSCode will wait for the server to be up before launching the browser.

I didn't have any issues with source maps not loading with my Create React App project, so I'd assume their project specific and would want to avoid adding them in the default configuration.

If that's not the case can you please  guide how to reproduce that issue?

I've pushed my changes in https://github.com/netlify/cli/pull/2440/commits/adfe78f4da253acd77a7ba2aa6e72006fe32fbfa

See https://deploy-preview-2440--cli.netlify.app/vscode

docs/vscode.md
```diff
@@ -0,0 +1,67 @@
+---
+title: Usage with Visual Studio Code
```
rstavchansky, line 2: This may need a more specific title. Would this work?

Another option would be `Run and debug with Visual Studio Code`, per the comment below

```suggestion
title: Debugging with Visual Studio Code
```

docs/vscode.md
```diff
@@ -0,0 +1,67 @@
...
+---
+
+# Usage with Visual Studio Code
```
rstavchansky, line 5:
```suggestion
# Run and debug with Visual Studio Code
```

docs/vscode.md
```diff
@@ -0,0 +1,67 @@
...
+
+This guide will walk you through configuring VScode to run and debug `netlify dev`.
+That allows debugging your site and functions all in the same IDE.
```
rstavchansky, line 8: Small suggested wording tweak: 

```suggestion
This allows debugging your site and serverless functions all in the same IDE.
```

site/src/_app.js
```diff
@@ -67,6 +68,7 @@ const pageNames = {
   contributing: 'Contributing',
   'netlify-dev': 'Netlify Dev',
   'functions-dev': 'Functions Development',
+  vscode: 'Usage with VSCode',
```
rstavchansky, line 71: If changed above - 

```suggestion
  vscode: 'Run and debug with VSCode',
```

rstavchansky (commented):
Adding some minor suggestions from a docs perspective

erezrokah (approved):

PR #2450, opened 2021-05-24
test: improve process handling

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/issues/2209

Uses a more robust way to kill child processes in tests
Also fixes a bug in the functions tests (not passing the `port` parameter)

**- Test plan**

Existing tests

🙈

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,8 @@
     "strip-ansi": "^6.0.0",
     "temp-dir": "^2.0.0",
     "toml": "^3.0.0",
-    "tomlify-j0.4": "^3.0.0"
+    "tomlify-j0.4": "^3.0.0",
+    "tree-kill": "^1.2.2"
   },
   "ava": {
     "files": [
@@ -225,7 +226,7 @@
     "failFast": false,
     "failWithoutAssertions": false,
     "tap": false,
-    "timeout": "2m"
+    "timeout": "5m"
   },
   "oclif": {
     "bin": "netlify",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -225,7 +226,7 @@
     "failFast": false,
     "failWithoutAssertions": false,
     "tap": false,
-    "timeout": "2m"
+    "timeout": "5m"
```
erezrokah, line 229: A higher timeout allows us to see why the test is taking a long time to execute (instead of `ava` killing it)

tests/command.functions.test.js
```diff
@@ -155,7 +156,7 @@ test('should not create a new function directory when one is found', async (t) =
 })
 
 const DEFAULT_PORT = 9999
-const SERVE_TIMEOUT = 300000
+const SERVE_TIMEOUT = 60000
```
erezrokah, line 159: The previous timeout was 5m which doesn't make sense as the test timeout was 2m

tests/command.functions.test.js
```diff
@@ -242,7 +240,7 @@ test('should use settings from netlify.toml dev', async (t) => {
       })
       .buildAsync()
 
-    await withFunctionsServer({ builder }, async () => {
+    await withFunctionsServer({ builder, port }, async () => {
```
erezrokah, line 243: This was the bug in the test that was causing it to be unstable.

The CLI was using a custom port from `netlify.toml` but the test was waiting for the default port to be open.
It worked sometimes since we run the tests in parallel and we have a test that opens the default port.

ehmicky (approved):

Issue #2458, opened 2021-05-25
Analytics: track command duration and success/failure


erezrokah:
**Is your feature request related to a problem? Please describe.**

We only track command invocations at the moment - that allows us to know which commands are used/more popular.

We should also track command duration and success/failure rate so we can detect regressions.

**Describe the solution you'd like**

Add command duration and success/failure to analytics

**Describe alternatives you've considered**

N/A

**Can you submit a pull request?**

Yes

PR #2462, opened 2021-05-25
fix(telemetry): fix tracking of --telemetry-enable flag

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/issues/2458

Cleans up our telemetry code a bit and also fixes a bug with `netlify --telemetry-enable` not being tracked

**- Test plan**

Added 2 new tests for `--telemetry-disable` and `--telemetry-enable` using a mock API

image

### DIFF
--- a/src/hooks/init.js
+++ b/src/hooks/init.js
@@ -19,9 +19,7 @@ module.exports = async function initHooks(context) {
     globalConfig.set('telemetryDisabled', false)
     console.log('Netlify telemetry has been enabled')
     console.log('You can disable it anytime with the --telemetry-disable flag')
-    track('user_telemetryEnabled', {
-      force: true,
-    })
+    await track('user_telemetryEnabled')
     process.exit()
   }
 
--- a/src/utils/telemetry/index.js
+++ b/src/utils/telemetry/index.js
@@ -1,16 +1,16 @@
-const { spawn } = require('child_process')
 const path = require('path')
 const process = require('process')
 
-const ci = require('ci-info')
+const { isCI } = require('ci-info')
+const execa = require('execa')
 
 const getGlobalConfig = require('../get-global-config')
 
 const isValidEventName = require('./validation')
 
-const IS_INSIDE_CI = ci.isCI
-
-const DEBUG = false
+const isTelemetryDisabled = function (config) {
+  return config.get('telemetryDisabled')
+}
 
 const send = function (type, payload) {
   const requestFile = path.join(__dirname, 'request.js')
@@ -19,18 +19,18 @@ const send = function (type, payload) {
     type,
   })
 
-  if (DEBUG) {
-    console.log(`${type} call`, payload)
-    return Promise.resolve()
+  const args = [process.execPath, [requestFile, options]]
+  if (process.env.NETLIFY_TEST_TELEMETRY_WAIT === 'true') {
+    return execa(...args, {
+      stdio: 'inherit',
+    })
   }
 
   // spawn detached child process to handle send
-  spawn(process.execPath, [requestFile, options], {
+  execa(...args, {
     detached: true,
     stdio: 'ignore',
   }).unref()
-
-  return Promise.resolve()
 }
 
 const eventConfig = {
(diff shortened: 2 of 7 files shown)


src/hooks/init.js
```diff
@@ -19,9 +19,7 @@ module.exports = async function initHooks(context) {
     globalConfig.set('telemetryDisabled', false)
     console.log('Netlify telemetry has been enabled')
     console.log('You can disable it anytime with the --telemetry-disable flag')
-    track('user_telemetryEnabled', {
```
erezrokah, line 22: This was a bug since `track` is `async` (because it reads the global config). We need to `await` it otherwise `process.exit()` is called before starting the child process that sends the telemetry request.

src/hooks/init.js
```diff
@@ -19,9 +19,7 @@ module.exports = async function initHooks(context) {
...
-      force: true,
```
erezrokah, line 23: This is the only place `force` was passed. It's redundant as `track` re-reads `telemetryDisabled`

src/utils/telemetry/index.js
```diff
@@ -1,16 +1,16 @@
...
+const { isCI } = require('ci-info')
+const execa = require('execa')
 
 const getGlobalConfig = require('../get-global-config')
 
 const isValidEventName = require('./validation')
 
-const IS_INSIDE_CI = ci.isCI
-
-const DEBUG = false
```
erezrokah, line 13: `DEBUG` was always false

src/utils/telemetry/index.js
```diff
@@ -19,18 +19,18 @@ const send = function (type, payload) {
     type,
   })
 
-  if (DEBUG) {
-    console.log(`${type} call`, payload)
-    return Promise.resolve()
+  const args = [process.execPath, [requestFile, options]]
+  if (process.env.NETLIFY_TEST_TELEMETRY_WAIT === 'true') {
+    return execa(...args, {
```
erezrokah, line 24: We set `NETLIFY_TEST_TELEMETRY_WAIT` in our tests since we want to wait for the telemetry request

src/utils/telemetry/index.js
```diff
@@ -45,36 +45,17 @@ const eventConfig = {
...
-  let { cliId } = properties
-
-  if (!userId) {
-    userId = globalConfig.get('userId')
-  }
-
-  if (!cliId) {
-    cliId = globalConfig.get('cliId')
-  }
+  const [userId, cliId] = [globalConfig.get('userId'), globalConfig.get('cliId')]
```
erezrokah, line 58: I couldn't find a place we ass `track` `userId` and `cliId` as properties of `payload`, so we can always take them from `globalConfig

src/utils/telemetry/index.js
```diff
@@ -88,70 +69,39 @@ const track = async function (eventName, payload) {
     return false
   }
 
-  const defaultProperties = {
-    // cliId: cliId
-  }
-
-  delete properties.force
```
erezrokah, line 95: `force` flag was only passed once and it was redundant so I removed it

src/utils/telemetry/index.js
```diff
@@ -88,70 +69,39 @@ const track = async function (eventName, payload) {
...
-    name: userProfile.name,
-    email: userProfile.email,
+    name,
+    email,
     cliId,
-    telemetryDisabled: TELEMETRY_DISABLED,
   }
 
-  // remove force key
-  delete data.force
```
erezrokah, line 148: Same as other `force` related changes

src/utils/telemetry/index.js
```diff
@@ -88,70 +69,39 @@ const track = async function (eventName, payload) {
...
     anonymousId: cliId,
-    properties: { ...defaultProperties, ...properties },
+    properties: payload,
   }
 
   return send('track', defaultData)
 }
 
 const identify = async function (payload) {
-  const data = payload || {}
```
erezrokah, line 108: `identify` is only called once in https://github.com/netlify/cli/blob/b13d6d07afdfe4b932cb115daa161abf93fee12f/src/utils/command.js#L307 so `payload` is always defined.

src/utils/telemetry/index.js
```diff
@@ -88,70 +69,39 @@ const track = async function (eventName, payload) {
...
-    userId = globalConfig.get('userId')
-  }
-
-  if (!cliId) {
-    cliId = globalConfig.get('cliId')
-  }
-
-  const userProfile = globalConfig.get(`users.${userId}`)
+  const cliId = globalConfig.get('cliId')
+  const { userId, name, email } = payload
```
erezrokah, line 93: Those are all passed when calling `identify`

tests/telemetry.test.js
```diff
@@ -0,0 +1,38 @@
...
+
+const callCli = require('./utils/call-cli')
+const { withMockApi } = require('./utils/mock-api')
+
+const getCLIOptions = (apiUrl) => ({
+  env: {
+    NETLIFY_TEST_TRACK_URL: `${apiUrl}/track`,
+    NETLIFY_TEST_IDENTIFY_URL: `${apiUrl}/identify`,
+    NETLIFY_TEST_TELEMETRY_WAIT: true,
+    CI: undefined,
```
erezrokah, line 12: This is required so the tracking code will work in our GitHub actions

tests/utils/mock-api.js
```diff
@@ -15,7 +15,11 @@ const startMockApi = ({ routes }) => {
   const requests = []
   const app = express()
   app.use(bodyParser.urlencoded({ extended: true }))
-  app.use(bodyParser.json())
+  app.use(
+    bodyParser.json({
```
erezrokah, line 19: Since `request.js` doesn't send the content type header this is required.

tests/telemetry.test.js
```diff
@@ -0,0 +1,40 @@
...
+const { withMockApi } = require('./utils/mock-api')
+
+const getCLIOptions = (apiUrl) => ({
+  env: {
+    NETLIFY_TEST_TRACK_URL: `${apiUrl}/track`,
+    NETLIFY_TEST_IDENTIFY_URL: `${apiUrl}/identify`,
+    NETLIFY_TEST_TELEMETRY_WAIT: true,
+    PATH: process.env.PATH,
+  },
+  extendEnv: false,
```
erezrokah, line 16: This is required so the tracking code will work in our GitHub actions

JGAntunes (approved):
Awesome work! 🎉

PR #2472, opened 2021-05-26
fix(analytics): remove usage as hook

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/issues/2458

In some cases we were using `track` directly and in some cases as a `oclif` hook. Not sure why we need the hook - the only reason I could think of is if we wanted to use the built in `oclif` methods in the hooks (e.g `this.exit`), but even then I would prefer to pass the `oclif` instance as a context argument to `track`.

This PR removes the hook complexity in favor of using `track` directly.

I also moved the tracking to the beginning of the commands as some were existing early before calling `track`

See about hooks [here](https://oclif.io/docs/hooks#custom-events).

**- Test plan**

Existing tests and ~~I'll test this manually too~~ **I invoked each of the commands manually to verify it works**

image


src/commands/dev/index.js
```diff
@@ -175,14 +176,11 @@ const handleLiveTunnel = async ({ flags, site, api, settings, log }) => {
...
-    payload: {
-      command: 'dev',
-      projectType: settings.framework || 'custom',
-      live: flagsLib.live || false,
-    },
+const reportAnalytics = async ({ settings, flags }) => {
+  await track('command', {
+    command: 'dev',
+    projectType: settings.framework || 'custom',
+    live: flags.live || false,
```
erezrokah, line 183: I believe I introduces this bug a while back. We should pass in `flags.live` instead of using `flagsLib.live`

src/commands/sites/config.js
```diff
@@ -1,35 +0,0 @@
-const { Command, flags } = require('@oclif/command')
```
erezrokah, line 1: This command was hidden and not implemented

src/commands/status/index.js
```diff
@@ -4,12 +4,17 @@ const clean = require('clean-deep')
...
 
 const Command = require('../../utils/command')
+const { track } = require('../../utils/telemetry')
 
 class StatusCommand extends Command {
   async run() {
     const { globalConfig, api, site } = this.netlify
     const { flags } = this.parse(StatusCommand)
 
+    await track('command', {
```
erezrokah, line 14: The `status` command was missing a call to track it

JGAntunes (approved):
Awesome work @erezrokah 🚀

PR #2487, opened 2021-05-27
feat(telemetry): measure duration and status

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2458

Uses a new command base class to track duration and status to avoid duplication

**- Test plan**

I tested each command manually

image

### DIFF
--- a/src/commands/build/index.js
+++ b/src/commands/build/index.js
@@ -1,22 +1,24 @@
-const { flags } = require('@oclif/command')
+const { flags: flagsLib } = require('@oclif/command')
 
 const { getBuildOptions, runBuild } = require('../../lib/build')
 const Command = require('../../utils/command')
-const { track } = require('../../utils/telemetry')
 
 class BuildCommand extends Command {
   // Run Netlify Build
   async run() {
+    const { flags } = this.parse(BuildCommand)
+
+    this.setAnalyticsPayload({ dry: flags.dry })
+
     // Retrieve Netlify Build options
     const [token] = await this.getConfigToken()
+
     const options = await getBuildOptions({
       context: this,
       token,
-      flags: this.parse(BuildCommand).flags,
+      flags,
     })
     this.checkOptions(options)
-
-    await track('command', { command: 'build', dry: Boolean(options.dry) })
 
     const exitCode = await runBuild(options)
     this.exit(exitCode)
@@ -35,10 +37,11 @@ class BuildCommand extends Command {
 
 // Netlify Build programmatic options
 BuildCommand.flags = {
-  dry: flags.boolean({
+  dry: flagsLib.boolean({
     description: 'Dry run: show instructions without running them',
+    default: false,
   }),
-  context: flags.string({
+  context: flagsLib.string({
     description: 'Build context',
   }),
   ...BuildCommand.flags,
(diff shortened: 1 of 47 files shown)


src/commands/build/index.js
```diff
@@ -35,10 +37,11 @@ class BuildCommand extends Command {
 
 // Netlify Build programmatic options
 BuildCommand.flags = {
-  dry: flags.boolean({
+  dry: flagsLib.boolean({
     description: 'Dry run: show instructions without running them',
+    default: false,
```
erezrokah, line 42: I removed the casting of `Boolean(options.dry)` in favor of a default value (without a default value not passing `--dry` results in `flags.dry` being `undefined`)

src/commands/dev/index.js
```diff
@@ -316,6 +307,7 @@ DevCommand.flags = {
   live: flagsLib.boolean({
     char: 'l',
     description: 'start a public live session',
+    default: false,
```
erezrokah, line 310: Same as before I removed `flags.live || false` in favor of a default value

src/commands/functions/build.js
```diff
@@ -1,30 +1,24 @@
...
-    })
-
     const { config } = this.netlify
 
     const src = flags.src || config.build.functionsSource
     const dst = getFunctionsDir({ flags, config })
 
     if (src === dst) {
       this.log(`${NETLIFYDEVERR} Source and destination for function build can't be the same`)
-      process.exit(1)
```
erezrokah, line 27: This prevents from the analytics to be reported in the `finally` block

src/utils/telemetry/tracked-command.js
```diff
@@ -0,0 +1,51 @@
...
+  async init() {
+    this.analytics = { startTime: hrtime.bigint() }
+    await super.init()
+  }
+
+  setAnalyticsPayload(payload) {
+    this.analytics = { ...this.analytics, payload }
+  }
+
+  async finally(error) {
```
erezrokah, line 38: For this to work well we need commands to use `this.error` or `this.exit` which most commands do.
Some areas of the codebase use `process.exit` which we should change.
I think we can separate it into a dedicated PR as it will require some refactoring.

Also, we would need to verify that commands report errors properly, but we can do that gradually.

src/utils/telemetry/tracked-command.js
```diff
@@ -0,0 +1,51 @@
...
+  if (error === undefined || isSafeError(error)) {
+    return 'success'
+  }
+
+  return 'error'
+}
+
+class TrackedCommand extends Command {
+  async init() {
+    this.analytics = { startTime: hrtime.bigint() }
```
erezrokah, line 30: We could do this as a class properly to avoid someone extending and forgetting to call `super.init`.

However, based on [this code](https://github.com/oclif/command/blob/adc0b266777b515a26365d31c0e3a926b68d0c72/src/command.ts#L145) I think `init` should always be called, and it seems the correct place to start measuring command time.

We could also move it to `run` but then we would have to call `super.run` in every command which doesn't feel correct.

JGAntunes (approved):
🚀 Really looking forward to this! 😄

PR #2504, opened 2021-05-29
feat: add detector for SvelteKit

kvn-shn:
**- Summary**

Neither the [`svelte`](https://github.com/netlify/cli/blob/main/src/detectors/svelte.js) nor [`sapper`](https://github.com/netlify/cli/blob/main/src/detectors/sapper.js) detector fit to SvelteKit as discussed here https://github.com/sveltejs/kit/issues/1568 and its PR. That's why I added a new detector.

Unfortunately a small but -- in my opinion -- reasonable workaround is needed to make Netlify Dev useful together with SvelteKit: enforcing a prod build beforehand. But it looks like e.g. the [angular](https://github.com/netlify/cli/blob/main/src/detectors/angular.js) detector is doing something similar (for similar reasons?!)

**- Test plan**

I've run and tested it locally. Both the detection itself and the enforcement of a SvelteKit production build is working fine.

🐶 🐕

### DIFF
--- a/src/detectors/svelte-kit.js
+++ b/src/detectors/svelte-kit.js
@@ -1,0 +1,26 @@
+const { hasRequiredDeps, hasRequiredFiles, getYarnOrNPMCommand, scanScripts } = require('./utils/jsdetect')
+
+const FRAMEWORK_PORT = 3000
+
+module.exports = function detector() {
+  // REQUIRED FILES
+  if (!hasRequiredFiles(['package.json'])) return false
+  // REQUIRED DEPS
+  if (!hasRequiredDeps(['@sveltejs/kit'])) return false
+
+  // Everything below now assumes that we are within SvelteKit
+  // For more details refer to https://github.com/sveltejs/kit/tree/master/packages/adapter-netlify
+
+  const possibleArgsArrs = scanScripts({
+    preferredScriptsArr: ['dev'],
+    preferredCommand: 'svelte-kit dev',
+  })
+
+  return {
+    framework: 'svelte-kit',
+    command: getYarnOrNPMCommand(),
+    frameworkPort: FRAMEWORK_PORT,
+    possibleArgsArrs,
+    dist: 'static',
+  }
+}
--- a/src/detectors/svelte.js
+++ b/src/detectors/svelte.js
@@ -7,7 +7,8 @@ module.exports = function detector() {
   if (!hasRequiredFiles(['package.json'])) return false
   // REQUIRED DEPS
   if (!hasRequiredDeps(['svelte'])) return false
-  // HAS DETECTOR, IT WILL BE PICKED UP BY SAPPER DETECTOR, avoid duplication https://github.com/netlify/cli/issues/347
+  // AVOID FALSE-POSITIVES (having their own detectors and to avoid duplication https://github.com/netlify/cli/issues/347)
+  if (hasRequiredDeps(['@sveltejs/kit'])) return false
   if (hasRequiredDeps(['sapper'])) return false
 
   /** everything below now assumes that we are within svelte */


benmccann:
@kvn-shn it looks like one of the checks on this PR is failing because you need to prefix the PR title with something like `fix:` or `feat:`

kvn-shn:
Thanks for the hint. I just adjusted the commit but forgot to edit the PR too. (Wasn't clear to me what to do solely based on the cryptic "Missing required PR label of type:*" message so I thought someone else has to add a label... ;) )

src/detectors/svelte-kit.js
```diff
@@ -0,0 +1,32 @@
...
+  // Please note: Netlify Dev doesn't apply `_redirects` etc. if run via the preferred `svelte-kit dev` command. Since
+  // that is the only benefit of Netlify Dev compared to directly running SvelteKit, we'll simply enforce PROD mode:
+  if (!hasRequiredFiles(['build'])) {
+    throw new Error('SvelteKit detected. Please run `netlify build` first!')
+  }
+
+  const possibleArgsArrs = scanScripts({
+    preferredScriptsArr: ['preview'],
+    preferredCommand: 'svelte-kit preview',
+  })
```
erezrokah, line 23:
```suggestion
  // Everything below now assumes that we are within SvelteKit
  // For more details refer to https://github.com/sveltejs/kit

  const possibleArgsArrs = scanScripts({
    preferredScriptsArr: ['dev'],
    preferredCommand: 'svelte-kit dev',
  })
```

erezrokah (changes requested):
Thanks @kvn-shn, it would be great to support SvelteKit.

Can you please explain the requirement to run a production build first?

Based on https://svelte.dev/blog/sveltekit-beta it seems running the `dev` script should work (per my suggestion).

Also, if we do run the build script first won't that would be required each time the user changes something in their project?

src/detectors/svelte-kit.js
kvn-shn: Well, as I tried to explain, in my experience Netlify Dev simply won't apply e.g. custom `_redirects` rules if Svelte is started in dev mode. Though, for SvelteKit this would be one of the only benefits of using `netlify dev` instead of `svelte-kit dev`. That's why I made this workaround.

Yes, the user would have to run `svelte-kit build` before each time. But you're right. We should actually investigate first why Netlify Dev doesn't apply `_redirects` in this case...

erezrokah: Can you share an example repo? `netlify dev` looks for a redirects file both in the current directory and in the `dist` directory

benmccann: I would expect `preview` to also ignore `_redirects` because you're running the SvelteKit server which knows nothing about them. Just like `dev`, `preview` doesn't use the adapters. You'd have to run the Netlify CLI in order for `_redirects` to do anything (or put code in your `_handle` method that parses and handles the file)

kvn-shn: @erezrokah Okay, that's interesting. So normally `netlify dev` should work just fine (using `svelte-kit dev`), right? I'll try to investigate later on and give an update...

@benmccann It's not about what SvelteKit does (not) know. It's rather about Netlify Dev not working as it is supposed to.

erezrokah: > @erezrokah Okay, that's interesting. So normally `netlify dev` should work just fine (using `svelte-kit dev`), right?

Yes it should

kvn-shn: I'm dumbfounded as it's (now?) working as expected (with `svelte-kit dev` and all).

@erezrokah I adjusted it accordingly. Additionally setting the `dist` folder correctly.

erezrokah (approved):
Thanks for the follow up @kvn-shn, this is 🚀

PR #2102, opened 2021-04-08
feat: Adjust traffic-mesh call for new version

mraerino:
**- Summary**

**- Test plan**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #2529, opened 2021-06-01
feat: replace mesh-forward with new routing-local-proxy

NeoLegends:
**- Summary**

Hey all! 👋

Over the past couple of months we have been doing a couple of refactorings on our edge-routing (previously "traffic-mesh") layer:
- Edge Handlers are now no longer a part of a redirect rule. Instead, they are declared separately in the netlify.toml and mounted directly on a path.
- mesh-forward has been renamed to "routing-local-proxy" and has been rewritten to use the same HTTP server stack as routing-proxy, the server running edge handlers on the CDN.

This PR updates the CLI to use the new binary and fixes the tests to use the new format of specifying which edge handler to run for a given request.

**- Test plan**

Existing tests should pass. In addition, routing-local-proxy will now receive its own integration tests against the CLI in the edge-routing repository, such that we can catch regressions and generally avoid the mess that was mesh-forward.


NeoLegends:
Tests are failing because they are still using an outdated version of the binary. I've got the fixes in my local branch already, but they're not released yet.

----

e5fbc5c

So, my plan for the tests is to just make the CI on the edge-routing repo run the routing-local-proxy-specific tests of this repo, and then add a bunch of tests here. Therefore I "marked" all the relevant tests with the term "routing-local-proxy", so that in the edge-routing CI we can run only these ones (using `ava --match ...`) and not _all_ the tests the CLI has. WDYT of this approach?

NeoLegends:
Here we have a successful run of the CI in edge-routing against this branch, with the latest version of routing-local-proxy: https://app.circleci.com/pipelines/github/netlify/edge-routing/1369/workflows/94a2d592-2ae2-4e71-a7e3-6a7f203880a7

I'm releasing a new version of the rlp binary now so that the CI here is fixed and then this PR is good to be merged 👍🚀

erezrokah (changes requested):
Thanks @NeoLegends. this looks great from the code perspective.

Some questions before we merge and release this:
1. Can we close https://github.com/netlify/cli/pull/2102?
2. From a user perspective the only breaking change is using a different way to configure Edge Handers? Is there a place we can point users to learn about the new syntax? How will they know they need to change it?
3. How about doing a pre-release version of the CLI with this so we can verify the installation process of the published artifact?

NeoLegends:
1. Yes, I think #2102 can be closed.
1. Going forward users need to adjust their `netlify.toml` and also the JS code, so more changes are necessary (though they are not specific to the CLI, the final product will also require these changes to be made). The docs on netlify.com are currently in the process of being updated (https://github.com/netlify/docs/pull/1113) to reflect the new way of configuring / writing edge handlers. Given that the `--edgeHandlers` flag is still hidden (and AFAIK there has been no marketing announcement of the CLI being able to run them) I'm not sure we need to provide migration instructions and we'll be able to migrate the DevEx team and these few enterprise test customers that currently _can_ run edge handlers on the CDN (and know about the CLI being able to run them) "by hand".
1. SGTM! We now also have experimental support for M1 macs (notice the `@netlify/routing-local-proxy-darwin-arm64` package), so if we can get a CI run there that would be awesome. @rjbeers is also going to test routing-local-proxy on his private M1 laptop. I'd also very much like to add a lot more tests in another iteration / a follow-up PR (basically add a bunch of example edge handlers that should work). The tests around edge handlers we have currently are not enough to have a good reading on the quality of the product.

Does that all make sense to you? I'm happy to discuss if my reading on 2) is correct or not.

erezrokah:
That sounds good @NeoLegends, I've created another branch with the pre-release:
https://github.com/netlify/cli/tree/pre-release/eh
We don't have to merge it or anything - it's just for history purposes.
Also see the [tag](https://github.com/netlify/cli/releases/tag/v3.33.0-eh.0).

The pre-release should be available via `npm install -g netlify-cli@eh`, see https://www.npmjs.com/package/netlify-cli:
image

We can test it out, then merge and release.

Also when you merge please squash commit with a `feat:` type (I updated the title), otherwise our release automation won't create a new release from the change.

erezrokah (approved):

erezrokah:
> We can test it out, then merge and release.

This looks on my MacOS (x64) with Node.js 10 & 16

PR #2543, opened 2021-06-02
feat: Adds RedwoodJS Detector for netlify dev

dthyresson:
As part of adding framework detection for RedwoodJS with:

https://github.com/netlify/framework-info/blob/main/src/frameworks/redwoodjs.json

We also need to add a detector so that `netlify dev` will detect a redwoodjs-based site/app.

This PR endeavors to add this configuration such that netlify dev can be used with a `netlify.toml` file in a RedwoodJS app like:

```toml
[dev]
  # Note be sure to `netlify link` your site first
  command = "yarn rw dev"
  framework = "redwoodjs"
  targetPort = 8910 # The port for your application server, framework or site generator
  port = 8888 # The port that the netlify dev will be accessible on
```

Notes:

* we do use `yarn rw dev` as the command and `yarn rw` is set in this detector. We do know this might spawn an extra yarn, but so much of our docs use `yarn rw` and Redwood uses `yarn workspaces` so the familiarity is a benefit.
* we will still want to be able to overrive `targetPort = 8910` because people have the option to set that port in their `redwood.toml` config

### DIFF
--- a/src/detectors/redwoodjs.js
+++ b/src/detectors/redwoodjs.js
@@ -12,9 +12,11 @@ module.exports = function detector() {
 
   return {
     framework: 'redwoodjs',
-    command: 'yarn rw',
+    // yarn is a requirement for RedwoodJS:
+    // https://learn.redwoodjs.com/docs/tutorial/installation-starting-development
+    command: 'yarn',
     frameworkPort: FRAMEWORK_PORT,
-    possibleArgsArrs: ['dev'],
+    possibleArgsArrs: [['rw', 'dev']],
     dist: 'web/dist',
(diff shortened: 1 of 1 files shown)


erezrokah (approved):
Thanks @dthyresson, this looks good.

I moved the `rw` into the arguments to match the way we spawn processes here:
https://github.com/netlify/cli/blob/7e81d104435fa40e683eac0a78ef07371d5a201e/src/commands/dev/index.js#L60

I also added a comment to make it clear than `yarn` is a requirement.

As for configuring the port that is still possible via `netlify.toml`.

PR #2572, opened 2021-06-07
feat: enable stripping out headers early in the HTTP API handling stack

benlavalley:
When my Netlify function received a POST call with header ("Expect: 100-continue"), the request times out and throws this error before my function can even begin processing the event:

```
BadRequestError: request aborted
    at IncomingMessage.onAborted (/usr/local/lib/node_modules/netlify-cli/node_modules/body-parser/node_modules/raw-body/index.js:231:10)
    at IncomingMessage.emit (events.js:314:20)
    at abortIncoming (_http_server.js:533:9)
    at socketOnClose (_http_server.js:526:3)
    at Socket.emit (events.js:326:22)
    at TCP.<anonymous> (net.js:675:12)
```

This change supports removing unwanted headers entirely early in the processing pipeline by leveraging Netlify's existing support for adding headers within _headers or netlify.toml -- simply set any given header value to "removeHeader" to have it stripped.

Once this header is removed, processing the POST request with the offending expect header works just fine.

More background here:
https://linkylinky.netlify.app/vgm5xBjL

### DIFF
--- a/src/utils/proxy.js
+++ b/src/utils/proxy.js
@@ -283,6 +283,18 @@ const initializeProxy = function (port, distDir, projectDir) {
     headerRules = headersFiles.reduce((prev, curr) => Object.assign(prev, parseHeadersFile(curr)), {})
   })
 
+  proxy.before('web', 'stream', (req) => {
+    const requestURL = new URL(req.url, `http://${req.headers.host || 'localhost'}`)
+    const pathHeaderRules = headersForPath(headerRules, requestURL.pathname)
+    if (!isEmpty(pathHeaderRules)) {
+      Object.entries(pathHeaderRules).forEach(([key, val]) => {
+        if (val && val[0] === 'removeHeader' && req.headers[key]) {
+          delete req.headers[key]
+        }
+      })
+    }
+  })
+
   proxy.on('error', (err) => console.error('error while proxying request:', err.message))
   proxy.on('proxyReq', (proxyReq, req) => {
     if (req.originalBody) {


erezrokah (changes requested):
Hi @benlavalley, thank you for opening this PR to fix your use case!

Do you mind opening an issue linked to it with **a public repo reproducing the issue**?
I want to make sure I understand the problem correctly before we dive into a solution.

The current solution is adding a feature to our headers logic, and only does so in the CLI (not in our production logic), which we would like to avoid.

If the issue is with our local functions server not handling a specific header correctly it seems we should fix that instead of removing the header.
Also, what happens when the function is deployed? Are you experiencing the same behavior?

Finally do you mind sharing how the syntax for removing headers looks like in both `_headers` and `netlify.toml`?

benlavalley:
@erezrokah Issue here: https://github.com/netlify/cli/issues/2591

I am unable to reproduce this issue with my deployed Netlify production app, which is good news and at least means this issue was isolated to netlify-cli.

If you do need a public repository reproducing the issue I can still create one later this week, but ultimately all you need to do to reproduce is add any serverless function to your local Netlify app development environment (it's contents will not matter as it never receives the event or context values).

erezrokah:
See https://github.com/netlify/cli/pull/2599

benlavalley:
@erezrokah  Really nice to see - thanks for the rapid attention on this one!

erezrokah:
Thanks @benlavalley, you've made it very easy for us to fix with the detailed issue and PR :)

Issue #2557, opened 2021-06-06
Unhelpful error: TypeError: Cannot read property '0' of   undefined


sw-yx:
**Describe the bug**

typing in `ntl deploy --prod` gets this:

```
Deploy path:        /Users/swyx/Work/swyxdotio/public
Configuration path: /Users/swyx/Work/swyxdotio/netlify.toml
Deploying to main site URL...
 ›   Warning: 
 ›   {}
 ›
    TypeError: Cannot read property '0' of
     undefined
```

I've never seen this error before as a heavy netlify user. netlify console displayed this

![CleanShot 2021-06-07 at 01 30 07](https://user-images.githubusercontent.com/6764957/120934195-ebe55e00-c72f-11eb-9e73-dc60d99ea92c.png)

When I reran `ntl deploy`, it worked again. I think there is some bad error handling inside of the CLI but cant tell more because the stack trace is swallowed.

**Configuration**

- If possible, please copy/paste below your `netlify.toml`.

```
[build]
  command = "yarn build"
  publish = "public/"

[[headers]]
  for = "/dist/svelte/*"
  [headers.values]
    cache-control = '''
      public,
      max-age=31536000,
      immutable'''
```

- Please enter the following command in a terminal and copy/paste its output:

```bash
npx envinfo --system --binaries --npmPackages netlify-cli --npmGlobalPackages netlify-cli
```

```

  System:
    OS: macOS 11.3.1
    CPU: (8) x64 Apple M1
    Memory: 32.25 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 15.5.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 7.3.0 - /usr/local/bin/npm
  npmGlobalPackages:
    netlify-cli: 3.21.10
```

erezrokah:
Thanks for opening the issue @sw-yx ❤️ 
I believe this is somehow a duplicate of #2086.
In #2086 the error was coming from our `js-client`, but since then https://github.com/netlify/js-client/issues/157 was resolved, so now the offending code is part of the CLI.

Moving the code to the CLI **_should_** allow us to better debug this once we follow up on #2086 with some more refactoring.

Keeping the issue open for now.

eduardoboucas:
@ascorbic is this the error you were seeing?

ascorbic:
[...]

Issue #2086, opened 2021-04-06
Deployments failing with `TypeError: Cannot read property '0' of undefined`


DavidWells:
**Describe the bug**

Deployment randomly throws with `TypeError: Cannot read property '0' of undefined`.

This happens on and off and is hard to pin down. If I had to guess I'd say about 1 in 15 this error occurs.

```
> site@0.1.0 draft /Users/xyz/abc/site
> netlify deploy --dir out

Deploy path:        /Users/xyz/abc/site/out
Configuration path: /Users/xyz/abc/site/netlify.toml
Deploying to draft URL...
✔ Finished hashing 246 files
✔ CDN requesting 77 files
 ›   Warning:
 ›   {}
 ›
    TypeError: Cannot read property '0' of undefined
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! site@0.1.0 draft: `netlify deploy --dir out`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the site@0.1.0 draft script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

**To Reproduce**

This is a non-deterministic bug and is tricky to reproduce. 

Running `netlify deploy` again after encountering the issue usually works. 

**Configuration**

```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
    Memory: 456.26 MB / 16.00 GB
    Shell: 5.3 - /bin/zsh
  Binaries:
    Node: 12.7.0 - ~/.nvm/versions/node/v12.7.0/bin/node
    Yarn: 1.22.10 - ~/xyz/site/node_modules/.bin/yarn
    npm: 6.10.0 - ~/.nvm/versions/node/v12.7.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmPackages:
    netlify-cli: ^3.13.7 => 3.13.7
  npmGlobalPackages:
    netlify-cli: 3.13.4
```

**Expected behavior**

If this error occurs with the SDK automatically retry the deployment?

**Additional context**

If I recall correctly, when trying to track this down a while back, there is an issue with how the open API sdk handles specific types of errors. This results in this parsing issue and the error it throws. Hope this helps 😃

erezrokah:
[...]

PR #2576, opened 2021-06-07
fix(command-deploy): remove random spinner

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2086
Fixes https://github.com/netlify/cli/issues/2557

After adding some debug logs, specifically logging the error here:
https://github.com/netlify/cli/blob/fa7cd0057bc7b92fbc81ec7c6390a8e0a1e6fc21/src/utils/telemetry/tracked-command.js#L38

I was able to reproduce the issue and get the stacktrace for it. The offending code is here:
https://github.com/sindresorhus/ora/blob/c7d6dba4a723cfa640e6adb716cb6b5dad2c22d2/index.js#L249
`frames` appears to be `undefined`.

The underlying issue is that the following line:
https://github.com/netlify/cli/blob/fa7cd0057bc7b92fbc81ec7c6390a8e0a1e6fc21/src/commands/deploy.js#L6

adds `'default'` to the `cliSpinnerNames` array and that is not a valid spinner name.
Spinner names are validated here:
https://github.com/sindresorhus/ora/blob/c7d6dba4a723cfa640e6adb716cb6b5dad2c22d2/index.js#L171

but that code suffers from a similar bug of allowing `'default'`.

I removed using a random spinner since it feels confusing anyway, however if we want to keep it we can simply use `spinner: 'random'`.

**- Test plan**

Existing tests

🐒


eduardoboucas:
Wow! 😮

eduardoboucas (approved):

erezrokah:
Also https://github.com/sindresorhus/ora/pull/175

PR #2592, opened 2021-06-08
chore: refactor redirects file path logic

ehmicky:
Part of https://github.com/netlify/build/issues/2889

This is a small clarification of the logic used to find the files containing redirect rules.

### DIFF
--- a/src/utils/rules-proxy.js
+++ b/src/utils/rules-proxy.js
@@ -59,13 +59,8 @@ const getCountry = function () {
 
 const createRewriter = async function ({ distDir, projectDir, jwtSecret, jwtRoleClaim, configPath }) {
   let matcher = null
-  const configFiles = [
-    ...new Set(
-      [path.resolve(distDir, '_redirects'), path.resolve(projectDir, '_redirects'), configPath].filter(
-        (configFile) => configFile && configFile !== projectDir,
-      ),
-    ),
-  ]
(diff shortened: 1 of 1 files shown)


src/utils/rules-proxy.js
```diff
@@ -59,13 +59,8 @@ const getCountry = function () {
 
 const createRewriter = async function ({ distDir, projectDir, jwtSecret, jwtRoleClaim, configPath }) {
   let matcher = null
-  const configFiles = [
-    ...new Set(
-      [path.resolve(distDir, '_redirects'), path.resolve(projectDir, '_redirects'), configPath].filter(
-        (configFile) => configFile && configFile !== projectDir,
```
ehmicky, line 65: It seems to me `configFile` can never equal `projectDir`. Both are returned by `@netlify/config` as `configPath` (which can be `undefined`) and `buildDir` (which is never `undefined`) respectively. The first is always a regular file, the second is always a directory.

src/utils/rules-proxy.js
```diff
@@ -59,13 +59,8 @@ const getCountry = function () {
...
-      ),
-    ),
-  ]
+  const redirectsFiles = [...new Set([path.resolve(distDir, '_redirects'), path.resolve(projectDir, '_redirects')])]
```
ehmicky, line 62: In production, I believe only the publish directory (named `distDir` here) can be used for `_redirects`. That's at least what my manual test + the documentation website indicate.
`projectDir` is the build directory. do we have a reason why `netlify dev` allows `_redirects` in that directory, even though it is not valid in production?

Issue #2501, opened 2021-05-28
Handle Netlify configuration changes during build


ehmicky:
Plugin authors can now modify some configuration properties during builds (see https://github.com/netlify/build/issues/1193).

When using the `netlify deploy --build` command, any configuration used after the build should take this into account.

https://github.com/netlify/cli/blob/6a5fc64990d35c810a9d8ddbdf65110495185604/src/commands/deploy.js#L423

`@netlify/build` is now returning a `netlifyConfig` property which includes any potential changes. It is otherwise the same value as the one returned by `@netlify/config`.

Giving it a quick search, it seems like the following configuration properties might be used after build:
  - `build.publish`
  
https://github.com/netlify/cli/blob/6a5fc64990d35c810a9d8ddbdf65110495185604/src/commands/deploy.js#L51

  - `functionsDirectory`

https://github.com/netlify/cli/blob/6a5fc64990d35c810a9d8ddbdf65110495185604/src/commands/deploy.js#L100

  - `functions` configuration

https://github.com/netlify/cli/blob/6a5fc64990d35c810a9d8ddbdf65110495185604/src/function-builder-detectors/zisi.js#L236

I might be missing some of them though.

PR #2593, opened 2021-06-08
feat(command-deploy): refresh configuration when running with --build

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2501

**- Test plan**

TBD - add a new test

🐶

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -297,19 +297,20 @@ const runDeploy = async ({
 }
 
 const handleBuild = async ({ context, flags }) => {
-  if (flags.build) {
-    const [token] = await context.getConfigToken()
-    const options = await getBuildOptions({
-      context,
-      token,
-      flags,
-    })
-    const { exitCode, newConfig } = await runBuild(options)
-    if (exitCode !== 0) {
-      context.exit(exitCode)
-    }
-    return newConfig
+  if (!flags.build) {
+    return
   }
+  const [token] = await context.getConfigToken()
+  const options = await getBuildOptions({
+    context,
+    token,
+    flags,
+  })
+  const { exitCode, newConfig } = await runBuild(options)
+  if (exitCode !== 0) {
+    context.exit(exitCode)
+  }
+  return newConfig
 }
 
 const printResults = ({ flags, results, deployToProduction, log, logJson, exit }) => {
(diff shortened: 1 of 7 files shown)


tests/command.deploy.test.js
```diff
@@ -348,6 +348,56 @@ if (process.env.IS_FORK !== 'true') {
     })
   })
 
+  test.only('should refresh configuration when --build is passed', async (t) => {
```
ehmicky, line 351:
```suggestion
  test('should refresh configuration when --build is passed', async (t) => {
```

src/commands/deploy.js
```diff
@@ -296,6 +296,22 @@ const runDeploy = async ({
   }
 }
 
+const handleBuild = async ({ context, flags }) => {
+  if (flags.build) {
```
ehmicky, line 300: [sand] Should we use an early exit to avoid indentation?

tests/command.deploy.test.js
```diff
@@ -348,6 +348,56 @@ if (process.env.IS_FORK !== 'true') {
...
+            functions: { directory: 'functions' },
+            plugins: [{ package: './plugins/mutator' }],
+          },
+        })
+        .withBuildPlugin({
+          name: 'mutator',
+          plugin: {
+            onPreBuild: async ({ netlifyConfig }) => {
+              // eslint-disable-next-line node/global-require
+              const [fs, util] = [require('fs'), require('util')]
```
ehmicky, line 370: [sand] Would this work at the top-level?

ehmicky (approved):

tests/command.deploy.test.js
erezrokah: Nice catch!

tests/command.deploy.test.js
erezrokah: This translates to plugin code, so it having it top level in this file will result in missing requires in the plugin file.

src/commands/deploy.js
erezrokah: Done in https://github.com/netlify/cli/pull/2593/commits/21f6f2805f084201a37b5a93244022f11ae1d0d5

tests/command.deploy.test.js
ehmicky: :+1:

Issue #2641, opened 2021-06-11
`netlify dev` reload causes error: "Module did not self-register" with node-canvas


amerkay:
**Describe the bug**

Any time I **change any file** inside my netlify/functions/ folder, the [node-canvas](https://github.com/Automattic/node-canvas) module fails to load on the next request. I have to stop `netlify dev` and rerun it.

**To Reproduce**

Simple code to reproduce **uploaded here**:
**https://github.com/amerkay/netlify-cli-dev-bug**

1. Run locally using `netlify dev`.\
   Then check the output at http://localhost:8888/.netlify/functions/test-canvas
2. Change anything in the function's file `test-canvas.js`, like changing the text string. This triggers a function _reload_.\
   `"◈ .../netlify/functions/test-canvas.js modified, successfully reloaded!"`
3. Refresh the /test-canvas function in the browser, now you get the error:\
   **`"Error: Module did not self-register"`**

**Configuration**

Happens without any toml config.

```bash
npx envinfo --system --binaries --npmPackages netlify-cli --npmGlobalPackages netlify-cli

  System:
    OS: Linux 5.12 Manjaro Linux
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
    Memory: 3.95 GB / 15.56 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 12.22.0 - /usr/bin/node
    npm: 7.15.0 - /usr/bin/npm
  npmGlobalPackages:
    netlify-cli: 3.34.0

```

**Expected behavior**

I should not get the error "Module did not self-register"

**Additional context**

- Issue does not occur once deployed to Netlify.
- The issue happens not just on my Linux machine, but also on MacOS.
- Happens whether I use default bundler or `esbuild`.
- Happens with npm@6 or npm@7
- Example simple function **uploaded to** [**this repo**](https://github.com/amerkay/netlify-cli-dev-bug)

Much appreciated!

PR #2681, opened 2021-06-15
fix(command-dev): use decache instead of clearing the entire cache

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2641

Uses [decache](https://github.com/dwyl/decache) to clear only the relevant function dependencies instead of the entire Node.js cache (which includes the CLI modules as well).
`decache` solves #2641 by not clearing native modules:
https://github.com/dwyl/decache/blob/593197b33af666edf05b41299009fa57104547e0/decache.js#L66

**- Test plan**

Existing tests

🐱


package.json
```diff
@@ -109,6 +109,7 @@
     "cookie": "^0.4.0",
     "copy-template-dir": "^1.4.0",
     "debug": "^4.1.1",
+    "decache": "^4.6.0",
```
erezrokah, line 112: While the repo hasn't be updated in over a year, the package has many downloads:
https://www.npmjs.com/package/decache

src/utils/serve-functions.js
```diff
@@ -159,17 +160,11 @@ const logAfterAction = ({ path, action }) => {
   console.log(`${NETLIFYDEVLOG} ${path} ${action}, successfully reloaded!`)
 }
 
-const clearRequireCache = () => {
-  Object.keys(require.cache).forEach((key) => {
-    delete require.cache[key]
```
ehmicky, line 164: Really happy we're removing this, or at least abstracting it away.

ehmicky (approved):

eduardoboucas:
I'm late to the party, but I still wanted to say that I'm super happy about this! ❤️

PR #2692, opened 2021-06-15
chore: skip main CI workflow on release branches

eduardoboucas:


JGAntunes (approved):

erezrokah (approved):
I'm good with this solution until we can find a way to exit early from a workflow run

erezrokah (approved):
We can use `::set-output` like https://github.com/netlify/cli/blob/321e2b2e67c1a734cced78f97057e5776264f8d9/.github/workflows/main.yml#L72 but this works too

Issue #2754, opened 2021-06-24
Add offline mode for `build`


ascorbic:
**Is your feature request related to a problem? Please describe.**

Currently `netlify build` does not work unless the site has been already linked. If you want to build without linking to a site (e.g. if you're testing somebody else's repo, or building a site in CI), you need to include the `.netlify/state.json` file, either with a real siteId or a fake one, then use that alongside a fake `NETLIFY_TOKEN`.

**Describe the solution you'd like**

Add an `--offline` flag analogous to the flag for `dev`. When passed, it does not attempt to use the Netlify API, but just runs the build using local config.

**Describe alternatives you've considered**

Using a fake site id and fake token. Committing a real site id to git.

ehmicky:
Good idea. :+1: 
Note: `@netlify/build` and `@netlify/config` both [already have an `--offline` CLI flag](https://github.com/netlify/build/blob/27b0988ecce1b0122b9fb76a00e6a5058eb1c863/packages/build/src/core/flags.js#L162). We just need to pass it from Netlify CLI and ensure [the following checks](https://github.com/netlify/cli/blob/4d62dd4afb88c3a73105dc2016175370c4bfb866/src/commands/build/index.js#L28) are not performed when `offline` is `true`.

PR #2755, opened 2021-06-24
feat(command-build): add offline mode

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2754

The `offline` flag is already pass to `@netlify/config` here:
https://github.com/netlify/cli/blob/4d62dd4afb88c3a73105dc2016175370c4bfb866/src/utils/command.js#L158

**- Test plan**

TBD

🐕


ehmicky (approved):

ehmicky:
Note: `npm run docs` should probably be executed.

erezrokah:
> Note: `npm run docs` should probably be executed.

Ah! good point

src/lib/build.js
```diff
@@ -18,6 +18,7 @@ const getBuildOptions = ({
   telemetry: false,
   // buffer = true will not stream output
   buffer: json || silent,
+  offline,
```
erezrokah, line 21: I don't think we actually need to pass `offline` to `@netlify/build` since we use `cachedConfig`.

I'm still doing it for clarify

ascorbic:
Thank you so much!

PR #2758, opened 2021-06-24
fix(deps): revert update dependency @netlify/zip-it-and-ship-it to version 4.6.0

erezrokah:
Reverts `zisi` to `4.4.2`.


erezrokah:
Going to re-do this one

PR #2761, opened 2021-06-25
fix(command-lm): setting of PATH for Git LFS with fish shell

Be-ing:
This script gets installed to
~/.config/netlify/helper/path.fish.inc
and the executable is installed to
~/.config/netlify/helper/bin/git-credential-netlify

Without this fix, pushing after running `netlify lm:setup` fails with:
```
git: 'credential-netlify' is not a git command. See 'git --help'.
```


erezrokah (approved):
Thanks @Be-ing, great catch!

Issue #2806, opened 2021-06-30
Structure functions serving logic to accommodate different runtimes


eduardoboucas:
This is a cover issue for restructuring the functions serving logic in the CLI so that it better supports different runtimes.

PR #2811, opened 2021-06-30
refactor: move function builders to functions lib

eduardoboucas:
**- Summary**

Moves the function builders (netlify-lambda and zip-it-and-ship-it) to a new location, where all the JS runtime files will live.

Part of #2806.

**- Test plan**

N/A


ehmicky (approved):

erezrokah (approved):
I don't think we need to update this section https://cli.netlify.com/netlify-dev/#function-builders-function-builder-detection-and-relationship-with-netlify-lambda with the removal of the README.

Just sharing for context

erezrokah:
@eduardoboucas, that failing test is flaky. I'll look into it - if the PR passes on other OS I think it's ok

eduardoboucas:
> @eduardoboucas, that failing test is flaky. I'll look into it - if the PR passes on other OS I think it's ok

I'm not in a rush to merge this, so I don't mind waiting a bit more and restarting the CI if we want to be cautious. I'll leave it up to you. Thanks for looking into it!

PR #2816, opened 2021-07-01
refactor: move debounced file watcher to separate file

eduardoboucas:
**- Summary**

Moves function file watcher to a separate file.

Part of #2806.


src/lib/functions/runtimes/js/builders/zisi.js
```diff
@@ -34,6 +34,14 @@ const addFunctionToTree = (func, fileTree) => {
   fileTree.set(func.mainFile, { ...func, bundleFile, inputs })
 }
 
+const clearFunctionsCache = (functionsPath) => {
```
eduardoboucas, line 37: This is required because `decache` is removing from the cache the file that was changed, but when using the zip-it-and-ship-it builder we're serving a different set of files. We need to clear the cache for those too.

erezrokah (approved):

PR #2824, opened 2021-07-01
refactor(command-dev): extract starting a static server to a separate function

erezrokah:
Part of https://github.com/netlify/cli/pull/1704

This PR extracts some code to a separate function


src/commands/dev/index.js
```diff
@@ -28,24 +28,27 @@ const SERVER_POLL_INTERVAL = 1e3
...
+    },
+  })
 
-    await new Promise((resolve) => {
-      server.start(function onListening() {
-        log(`\n${NETLIFYDEVLOG} Server listening to`, settings.frameworkPort)
-        resolve()
-      })
+  await new Promise((resolve) => {
+    server.start(function onListening() {
```
ehmicky, line 42: Looking at https://github.com/nbluis/static-server/blob/38994a482a382019a99690cf7a5dfdbe5c332c61/server.js#L114, it seems like we might be able to do:

```js
await promisify(server.start.bind(server))()
log(`\n${NETLIFYDEVLOG} Server listening to`, settings.frameworkPort)
```

What do you think?

ehmicky (approved):

src/commands/dev/index.js
erezrokah: That work 🎉  https://github.com/netlify/cli/pull/2824/commits/70ad3a89062817fe9a961851870bcfb22049b7df

PR #2842, opened 2021-07-05
fix(command-dev): use waitPort built-in http request support

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/pull/1704

Uses `waitPort` built in support for testing `HTTP` requests instead of polling ourselves.

**- Test plan**

Existing tests, I'll verify with various frameworks manually too.

😼


src/commands/dev/index.js
```diff
@@ -93,35 +86,12 @@ const startFrameworkServer = async function ({ settings, log, exit }) {
       port: settings.frameworkPort,
       output: 'silent',
       timeout: FRAMEWORK_PORT_TIMEOUT,
+      ...(settings.disableLocalServerPolling ? {} : { protocol: 'http' }),
```
erezrokah, line 89: This option documented via the library types and somehow [here](https://github.com/dwmkerr/wait-port#usage) too.

erezrokah: @eduardoboucas can you verify this doesn't break the original fix (I remember you had issues with this when running `create-react-app`).

eduardoboucas: > @eduardoboucas can you verify this doesn't break the original fix (I remember you had issues with this when running `create-react-app`).

It seems to work fine! I remember the issue was with a Gatsby site and I'm not able to replicate it using this branch.

eduardoboucas (approved):

PR #2852, opened 2021-07-06
refactor: enable and fix fp/no-loops

tinfoil-knight:
**- Summary**

See #1891 

Enables the [fp/no-loops](https://github.com/jfmengels/eslint-plugin-fp/blob/master/docs/rules/no-loops.md) for `eslint-plugin-fp`.

- for loops were changed to use `forEach` or other array methods like `map`, `some` or `every` wherever appropriate.
- Retry logic was changed from using loop to being recursive in nature.
- 4 instances of for loop were ignored due to any of the following reasons:
  - `async/await` doesn't work with forEach
  -  recursive solutions would be functional but hard to read/understand
  - `Promise.allSettled` is not supported below `12.9.0` (Can replace making `await()` requests in for loops)

**- Test plan**

- `npx ava --serial --verbose` ran without any errors.


tinfoil-knight:
A missing `?.` to handle cases where value is undefined has caused this. Will fix it.

tinfoil-knight:
All failing tests have been fixed.

tinfoil-knight:
It seems that `?.` isn't supported yet in the netlify deploys we're using. Will need to change it.

tinfoil-knight:
All tests are passing except on (ubuntu-latest, bash, 10.x). The error seems to be in the `detect-server.js` file. I can't reproduce the issue but I'll try pushing a fix.

erezrokah:
> @erezrokah What version of `node` is the CI supporting for tests?

The CLI supports [Node.js >= 10.18.0](https://github.com/netlify/cli/blob/6ee72fa872fd16fd9641618ae869bf819bd68308/package.json#L14), The optional chaining operator `?.` is supported for [Node.js >= 14](https://node.green/#ES2020-features-optional-chaining-operator-----)

erezrokah:
> The error seems to be in the `detect-server.js` file. I can't reproduce the issue but I'll try pushing a fix.

You can use [`nvm`](https://github.com/nvm-sh/nvm) to install and switch to Node.js 10 locally.

src/detectors/utils/jsdetect.js
```diff
@@ -33,22 +33,15 @@ const getYarnOrNPMCommand = function () {
...
   const { dependencies, devDependencies } = getPkgJSON()
-  for (const depName of requiredDepArray) {
-    const hasItInDeps = dependencies && dependencies[depName]
-    const hasItInDevDeps = devDependencies && devDependencies[depName]
-    if (!hasItInDeps && !hasItInDevDeps) {
-      return false
-    }
-  }
-  return true
+  return requiredDepArray.every((depName) => {
```
erezrokah, line 36: Will the following work:

```js
const { dependencies, devDependencies } = getPkgJSON()
const allDependencies = { ...dependencies, ...devDependencies }
return requiredDepArray.every((depName) => allDependencies[depName])
```

src/utils/detect-server.js
```diff
@@ -67,10 +67,11 @@ const serverSettings = async (devConfig, flags, projectDir, log) => {
         return null
       }
     })
-    for (const detector of detectors) {
-      const detectorResult = detector(projectDir)
-      if (detectorResult) settingsArr.push(detectorResult)
-    }
+    detectors &&
+      detectors.forEach((detector) => {
```
erezrokah, line 71: How about using `map` and `filter`? `const settingsArr = detectors.map(...)`

src/utils/get-global-config.js
```diff
@@ -29,17 +29,18 @@ const getGlobalConfigOnce = async function () {
 }
 
 const getGlobalConfig = async function () {
-  const retries = 3
-  for (let retry = 1; retry <= retries; retry++) {
+  const maxAttempts = 3
+  const retry = async (retriesLeft) => {
```
erezrokah, line 33: Can we keep the old code? It seems more straightforward that using recursion

tests/utils/dev-server.js
```diff
@@ -71,21 +71,24 @@ const startServer = async ({ cwd, offline = true, env = {}, args = [] }) => {
 
 const startDevServer = async (options, expectFailure) => {
   const maxAttempts = 5
-  for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+
+  const retry = async (retriesLeft) => {
```
erezrokah, line 75: Same as the other `retry` function comment

tests/utils/site-builder.js
```diff
@@ -150,6 +150,8 @@ const createSiteBuilder = ({ siteName }) => {
       return builder
     },
     buildAsync: async () => {
+      // Replace by Promise.allSettled() [Supported for >=12.9.0]
```
erezrokah, line 153: `allSettled` is not the same as the this code.
We want to run the tasks in order, hence `await` inside the for loop

src/utils/headers.js
```diff
@@ -86,10 +86,10 @@ const parseHeadersFile = function (filePath) {
 
   let path
   let rules = {}
-  for (const { line, index } of lines) {
+  lines.forEach(({ line, index }) => {
```
erezrokah, line 89: I find the previous code easier to understand (`continue` vs `return`).

src/utils/gitignore.js
```diff
@@ -15,19 +15,19 @@ const parser = function (input, fn = (line) => line) {
   let section = { name: 'default', patterns: [] }
   const state = { patterns: [], sections: [section] }
 
-  for (const line of lines) {
+  lines.forEach((line) => {
```
erezrokah, line 18: I find the previous code easier to understand (`continue` vs `return`).

Also, looks like we only use `ensureNetlifyIgnore` from this file, so we can delete `parser`, `parse` and `format`.

src/utils/state-config.js
```diff
@@ -84,9 +84,9 @@ class StateConfig {
     const config = this.all
 
     if (args.length === 1) {
-      for (const keyPart of Object.keys(key)) {
+      Object.keys(key).forEach((keyPart) => {
```
erezrokah, line 87: How about using `Object.entries`?

src/commands/env/list.js
```diff
@@ -39,9 +39,9 @@ class EnvListCommand extends Command {
     const table = new AsciiTable(`Environment variables`)
 
     table.setHeading('Key', 'Value')
-    for (const [key, value] of Object.entries(environment)) {
+    Object.entries(environment).forEach(([key, value]) => {
```
erezrokah, line 42: Same as other comment about `addRowMatrix`

src/commands/env/import.js
```diff
@@ -64,9 +64,9 @@ class EnvImportCommand extends Command {
     const table = new AsciiTable(`Imported environment variables`)
 
     table.setHeading('Key', 'Value')
-    for (const [key, value] of Object.entries(importedEnv)) {
+    Object.entries(importedEnv).forEach(([key, value]) => {
```
erezrokah, line 67: Will `table.addRowMatrix(Object.entries(environment))` work? See [here](https://github.com/sorensen/ascii-table#instanceaddrowmatrixrows)

erezrokah (changes requested):
I've added a few more comments.

In general we should treat `fp/no-loops` as a guideline and not a strict rule, as sometimes it's better to have a `for` loop.

src/utils/detect-server.js
tinfoil-knight: We'd end up having the same issue: `Cannot read property 'map' of undefined` . An alternative to this could be `(detectors || []).someArrayMethod()`

src/detectors/utils/jsdetect.js
tinfoil-knight: Yeah.

tinfoil-knight: Yeah.

tests/utils/site-builder.js
tinfoil-knight: Oh. Okay. Removing the comment.

src/utils/get-global-config.js
tinfoil-knight: Alright.

src/utils/detect-server.js
tinfoil-knight: How does this look?
```
const detectors = detectorsFiles
      ? detectorsFiles.map((det) => {
          try {
            return loadDetector(det)
          } catch (error) {
            console.error(error)
            return null
          }
        })
      : []
 ```

Comment (tinfoil-knight) [reply src/utils/detect-server.js:71]: I've implemented this:
```
const detectors = detectorsFiles
      ? detectorsFiles.map((det) => {
          try {
            return loadDetector(det)
          } catch (error) {
            console.error(error)
            return null
          }
        })
      : []

    const settingsArr = detectors.map((detector) => detector(projectDir)).filter((el) => el)    
```

File: tests/utils/dev-server.js (hunk 4 above)
Comment (tinfoil-knight) [reply tests/utils/dev-server.js:75]: Okay.

File: src/commands/env/import.js (hunk 10 above)
Comment (tinfoil-knight) [reply src/commands/env/import.js:67]: Yeah.

File: src/commands/env/list.js (hunk 9 above)
Comment (tinfoil-knight) [reply src/commands/env/list.js:42]: Okay. `addRowMatrix` seems better for both instances of bulk row operations.

File: src/utils/headers.js (hunk 6 above)
Comment (tinfoil-knight) [reply src/utils/headers.js:89]: Yeah. `return` in `forEach` is a bit confusing.

File: src/utils/state-config.js (hunk 8 above)
Comment (tinfoil-knight) [reply src/utils/state-config.js:87]: Yeah. `Object.entries` seems more suitable now that the loop has been removed.

File: src/utils/gitignore.js (hunk 7 above)
Comment (tinfoil-knight) [reply src/utils/gitignore.js:18]: Done.

File: src/utils/detect-server.js (hunk 2 above)
Comment (tinfoil-knight) [reply src/utils/detect-server.js]: Removed the ternary since `detectorsFiles` comes from an internal system call (`fs.readDirSync()`) and will have an empty array if it doesn't find anything.

File: src/lib/functions/synchronous.js [hunk 11]
```diff
@@ -14,13 +24,9 @@ const handleSynchronousFunction = function (err, result, response) {
   }
 
   response.statusCode = result.statusCode
-  for (const key in result.headers) {
```
erezrokah, line 17: We can't use `headers.forEach` directly instead of `for in`. We need to get the entries first via `Object.entries(headers)`

tinfoil-knight: Okay. Thank you.

erezrokah (approved):
🚀

tinfoil-knight:
Thank you @erezrokah for merging my first PR in this repository. More PRs upcoming!

erezrokah:
Great work @tinfoil-knight, looking forward 👍

PR #2857, opened 2021-07-06
refactor(command-dev): use execa to launch framework server

erezrokah:
**- Summary**

Related to https://github.com/netlify/cli/pull/1704

Switches to using `execa` when launching the framework server to make it easier to know when the process exits. 

**- Test plan**

Existing tests + added a test for a specific use case when the command doesn't exists.
I'll also verify with various frameworks.

🐟


src/commands/dev/index.js
```diff
@@ -35,47 +34,53 @@ const startStaticServer = async ({ settings, log }) => {
...
-  ps.stderr.pipe(stripAnsiCc.stream()).pipe(process.stderr)
-
-  process.stdin.pipe(process.stdin)
-
-  const handleProcessExit = function (code) {
-    log(
-      code > 0 ? NETLIFYDEVERR : NETLIFYDEVWARN,
-      `"${[settings.command, ...settings.args].join(' ')}" exited with code ${code}. Shutting down Netlify Dev server`,
-    )
     process.exit(1)
```
erezrokah, line 79: We should avoid `process.exit` in general as it kills the CLI process abruptly.
For example it prevents [this code](https://github.com/netlify/cli/blob/c883b6e76aca955b3c4d04475ec96c61993827c9/src/utils/telemetry/tracked-command.js#L38) from running.

I'll handle it in another PR

tests/framework-detection.test.js
```diff
@@ -177,7 +177,32 @@ test('should start custom command if framework=#custom, command and targetPort a
 
     const error = await t.throwsAsync(() =>
       withDevServer(
-        { cwd: builder.directory, args: ['--command', 'cat non-existing', '--targetPort', '3000'] },
+        { cwd: builder.directory, args: ['--command', 'echo hello', '--targetPort', '3000'] },
```
erezrokah, line 180: Switched to `echo` as it exists on Windows too.

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
   log(`\n${NETLIFYDEVLOG} Server listening to`, settings.frameworkPort)
 }
 
+const NOT_FOUND_EXIT_CODE = 127
+
+const isNonExistingCommandError = (error) => {
+  // `ENOENT` is only returned for non Windows systems
+  // See https://github.com/sindresorhus/execa/pull/447
+  // exitCode === 127 is returned from npm/yarn
```
ehmicky, line 42: I might be wrong, but I think exit codes are program-specific, not OS-specific.
`127` is specific by Bash when calling a non-existing command (including npm or yarn).
The reason why Windows appears to behave differently is because CI tests use `cmd.exe` instead of Bash on Windows.

That being said, the best we can do to detect this is basically what this function is doing, so this looks good :+1:

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
     return await startStaticServer({ settings, log })
   }
 
   log(`${NETLIFYDEVLOG} Starting Netlify Dev with ${settings.framework || 'custom config'}`)
-  const commandBin = await which(settings.command).catch((error) => {
-    if (error.code === 'ENOENT') {
-      throw new Error(
-        `"${settings.command}" could not be found in your PATH. Please make sure that "${settings.command}" is installed and available in your PATH`,
+
+  const commandWithArgs = `${settings.command} ${settings.args.join(' ')}`
```
ehmicky, line 60: We might want to use `execa()` instead of `execa.command()`.
Note about `execa.command()`:
  - It does not run in a shell by default (like `execa()`)
  - If an argument has a significant space, it must be escaped with a backslash. In our case, I believe we would need to do this escaping on `settings.args`? Or we can just use `execa()` which does not have that problem.

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
+  // we use reject=false to avoid rejecting synchronously when the command doesn't exist
+  // we can't try->await->catch since we don't want to block on the framework server which
+  // is a long running process
+  const frameworkProcess = execa.command(commandWithArgs, { preferLocal: true, reject: false })
+  frameworkProcess.stdout.pipe(stripAnsiCc.stream()).pipe(process.stdout)
+  frameworkProcess.stderr.pipe(stripAnsiCc.stream()).pipe(process.stderr)
```
ehmicky, line 66: Would [the `all` option](https://github.com/sindresorhus/execa#all) help here?

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
-    if (error.code === 'ENOENT') {
-      throw new Error(
-        `"${settings.command}" could not be found in your PATH. Please make sure that "${settings.command}" is installed and available in your PATH`,
+
+  const commandWithArgs = `${settings.command} ${settings.args.join(' ')}`
+  // we use reject=false to avoid rejecting synchronously when the command doesn't exist
+  // we can't try->await->catch since we don't want to block on the framework server which
+  // is a long running process
+  const frameworkProcess = execa.command(commandWithArgs, { preferLocal: true, reject: false })
+  frameworkProcess.stdout.pipe(stripAnsiCc.stream()).pipe(process.stdout)
```
ehmicky, line 65: Just double-checking that we really need to strip ANSI sequences here?
Otherwise, we could just use `stdio: 'inherit'`, which would be simpler and more robust.

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
+  frameworkProcess.stderr.pipe(stripAnsiCc.stream()).pipe(process.stderr)
+  process.stdin.pipe(frameworkProcess.stdin)
+
+  // eslint-disable-next-line promise/catch-or-return,promise/prefer-await-to-then
+  frameworkProcess.then(async () => {
```
ehmicky, line 70: Should we use `await`?

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
+    // eslint-disable-next-line promise/always-return
+    if (result instanceof Error && isNonExistingCommandError(result)) {
+      log(
+        NETLIFYDEVERR,
+        `Failed launching framework server. Please verify ${chalk.magenta(`'${settings.command}' exists`)}`,
+      )
+    } else {
+      log(
+        exitCode > 0 ? NETLIFYDEVERR : NETLIFYDEVWARN,
+        `"${commandWithArgs}" exited with code ${exitCode}. Shutting down Netlify Dev server`,
```
ehmicky, line 82: The `error` rejected by `execa` contains some useful information beyond just the exit code, available in [`error.message`](https://github.com/sindresorhus/execa#all) or [`error.shortMessage`](https://github.com/sindresorhus/execa#shortmessage). Would we want to show it?

src/commands/dev/index.js
```diff
@@ -35,47 +34,59 @@ const startStaticServer = async ({ settings, log }) => {
...
+  frameworkProcess.stdout.pipe(stripAnsiCc.stream()).pipe(process.stdout)
+  frameworkProcess.stderr.pipe(stripAnsiCc.stream()).pipe(process.stderr)
+  process.stdin.pipe(frameworkProcess.stdin)
+
+  // eslint-disable-next-line promise/catch-or-return,promise/prefer-await-to-then
+  frameworkProcess.then(async () => {
+    const result = await frameworkProcess
+    const { exitCode = 0 } = result
+    // eslint-disable-next-line promise/always-return
+    if (result instanceof Error && isNonExistingCommandError(result)) {
```
ehmicky, line 74: We might want to check [the `failed` property](https://github.com/sindresorhus/execa#failed) instead of `instanceof Error`.

src/commands/dev/index.js
erezrokah: I used `execa.command` in preparation to use the output of `framework-info` which is a single string.
https://github.com/netlify/framework-info#example-nodejs

I'll switch to `execa` until we land https://github.com/netlify/cli/pull/1704

ehmicky: Oh, then `execa.command()` makes sense. :+1: 
We might just want to escape spaces with a backslash if they might contain any. Something along the lines of:

```js
settings.args.map(arg => arg.replace(/ /g, '\\ '))
```

src/commands/dev/index.js
erezrokah: `frameworkProcess` is for example the invocation of `gatsby develop`. We can't `await` it since it will block the CLI.
We still want to know when it exists since we should exit the CLI with an error when that happens.

ehmicky: Makes sense!
We might want to add a code comment for future reference.

src/commands/dev/index.js
erezrokah: Yes it can, should we pipe the the `all` to `stdout` in that case?

src/commands/dev/index.js
erezrokah: Seems like this was a [bug fix](https://github.com/netlify/cli/pull/828) for multiple bugs.
I'll check what happens without it.

src/commands/dev/index.js
ehmicky: Oh, that would not work, since we still want to redirect to both stdout and stderr. 
Please discard my comment :+1:

src/commands/dev/index.js
ehmicky: Would it be worth checking in a different PR?

erezrokah: Yes! See https://github.com/netlify/cli/issues/2879

src/commands/dev/index.js
erezrokah: It's commented here https://github.com/netlify/cli/pull/2857/files#diff-42c491dfbe209d7493ec82cd5d031be64f17cf6e622201768d29b3539194d9e1R61

I'll move part of that comment closer to the `then`

src/commands/dev/index.js
erezrokah: Done in https://github.com/netlify/cli/pull/2857/commits/c84f56c345fc68ffcb0859588b5ae76a54d5324e

src/commands/dev/index.js
erezrokah: I'm keeping `execa` for now in https://github.com/netlify/cli/pull/2857/commits/00eec5f8ef810d117e17fef2616e313ab7d785af

I think we might want to handle space escaping in `framework-info` so the command in the result is ready to be invoked

src/commands/dev/index.js
erezrokah: Done in https://github.com/netlify/cli/pull/2857/commits/0fd3527fad9bfd82cecba703188a832b82672e69.

`shortMessage` has enough information so we don't need to format the error ourselves. The other case is that the long running process exited successfully, so we need to format an error manually.

ehmicky (approved):

tests/snapshots/framework-detection.test.js.md
```diff
@@ -158,4 +168,4 @@ Generated by [AVA](https://avajs.dev).
     > start␊
     > react-scripts start␊
     ␊
-    ◈ "npm start" exited with code *. Shutting down Netlify Dev server`
+    ◈ Failed launching framework server. Please verify 'npm' exists`
```
erezrokah, line 171: This doesn't look good. Looking into it

erezrokah: Fixed in https://github.com/netlify/cli/pull/2857/commits/f41e5b1946e2bc6d317aefb85b81b0f3d7f166c5

Issue #2882, opened 2021-07-08
Netlify Dev server doesn't start with Vite


Hrishikesh-K:
**Describe the bug**

Netlify Dev server never starts when using Vite.

**To Reproduce**

1. Setup a basic Vite app as mentioned here: https://vitejs.dev/guide/#scaffolding-your-first-vite-project.
1. Run `netlify dev`.
1. See that the Vite dev server starts, but Netlify Dev server never starts.

**Configuration**

No extra configuration needed.

**Expected behavior**

Netlify Dev server at port 8888 should start like it starts with other SSGs like for example create react app.

**CLI Output**

```
◈ Netlify Dev ◈
◈ Ignored general context env var: LANG (defined in process)
◈ Starting Netlify Dev with vite

> dev
> vite


  vite v2.4.1 dev server running at:

  > Local: http://localhost:3000/
  > Network: use `--host` to expose

  ready in 217ms.
```

**Expected CLI Output**:

(This is from create react app)

```
◈ Netlify Dev ◈
◈ Ignored general context env var: LANG (defined in process)
◈ Starting Netlify Dev with create-react-app

> my-app@0.1.0 start
> react-scripts start

ℹ ｢wds｣: Project is running at http://192.168.0.109/
ℹ ｢wds｣: webpack output is served from 
ℹ ｢wds｣: Content not from webpack is served from <path>
ℹ ｢wds｣: 404s will fallback to /
Starting the development server...

Compiled successfully!

You can now view my-app in the browser.

  Local:            http://localhost:3000
  On Your Network:  http://192.168.0.109:3000

Note that the development build is not optimized.
To create a production build, use npm run build.


   ┌─────────────────────────────────────────────────┐
   │                                                 │
   │   ◈ Server now ready on http://localhost:8888   │
   │                                                 │
   └─────────────────────────────────────────────────┘
```

Reported by customer(s):

1. https://answers.netlify.com/t/functions-not-executing-with-vite/40474

Hrishikesh-K:
Hey team,

I believe this is happening again in Netlify CLI `6.14.7` with Vite `2.6.13` and Node `17.0.1`.

LuskeyNoah:
I also am having issues getting netlify dev to start on the latest Netlify CLI + node `17.0.1` on macOS Monterey. I am not using vite (using create react app) but the behavior is similar:

```
-> % netlify dev
◈ Netlify Dev ◈
◈ Ignored general context env var: LANG (defined in process)
[...]

PR #2883, opened 2021-07-08
fix(command-dev): don't poll for http requests for vite framework

erezrokah:
**- Summary**

Fixes https://github.com/netlify/cli/issues/2882 which is a regression because of https://github.com/netlify/cli/pull/2842
If we see more cases I'll revert that PR

**- Test plan**

Tested manually

🐟


ehmicky (approved):

unclejustin:
@erezrokah thanks for jumping on this so quickly! 🙇

PR #2900, opened 2021-07-11
chore: remove `safe-join` dependency

### LINKED ISSUE #2896: Remove license incompatible safe-join dependency
`netlify-cli` is currently using the [`safe-join`](https://www.npmjs.com/package/safe-join) package as a dependency. This package is unlicensed (i.e. has full propietary protection) and thus is incompatible with the MIT license `netlify-cli` uses. I think given that the code in that package is super small, it would make more sense just inlining a utility function that does the same. Despite the legal problem of using this package with incompatible licenses, having that dependency can cause issues in projects depending on `netlify-cli` who do automatically check their licenses to make sure only compatible licenses are used.

radiantly:
Closes #2896 

**- Summary**
 - Removes the unlicensed `safe-join` dependency
 - Instead, we remove the leading and trailing slashes from the path

**- Test plan**
:eyes: 

:eyes: :eyes:

### DIFF
--- a/src/utils/read-repo-url.js
+++ b/src/utils/read-repo-url.js
@@ -1,7 +1,6 @@
 const url = require('url')
 
 const fetch = require('node-fetch')
-const { safeJoin } = require('safe-join')
 
 // supported repo host types
 const GITHUB = Symbol('GITHUB')
@@ -27,7 +26,7 @@ const getRepoURLContents = async function (repoHost, ownerAndRepo, contentsPath)
   // naive joining strategy for now
   if (repoHost === GITHUB) {
     // https://developer.github.com/v3/repos/contents/#get-contents
-    const APIURL = safeJoin('https://api.github.com/repos', ownerAndRepo, 'contents', contentsPath)
+    const APIURL = `https://api.github.com/repos/${ownerAndRepo}/contents/${contentsPath}`
     try {
       const res = await fetch(APIURL)
       return await res.json()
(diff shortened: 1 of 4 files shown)


src/utils/read-repo-url.js
```diff
@@ -51,7 +50,10 @@ const parseRepoURL = function (repoHost, URL) {
   if (repoHost === GITHUB) {
     // https://developer.github.com/v3/repos/contents/#get-contents
     // what if it's not master? note that our contents retrieval may assume it is master
-    const [ownerAndRepo, contentsPath] = URL.path.split('/tree/master')
+    const [ownerAndRepo, contentsPath] = URL.path.split('/tree/master').map((path) =>
+      // remove the leading and trailing slashes
+      path.replace(/^\/?(.*?)\/?$/, '$1'),
+    )
```
erezrokah, line 56: Since URL is expected to be in the format `https://github.com/netlify-labs/all-the-functions/tree/master/functions/9-using-middleware`, how about doing the following?

```suggestion
    const [ownerAndRepo, contentsPath] = URL.path.slice(1).split('/tree/master/')
```

erezrokah (changes requested):
Thanks @radiantly for the quick follow up on the issue.

I made a suggestion, please let me know what you think.

src/utils/read-repo-url.js
radiantly: This is shorter and way more clearer - thanks!

erezrokah (approved):
Thanks @radiantly! I added a test case for the parse function

PR #2901, opened 2021-07-11
fix(command-dev): pass full lambda-local result functions server

erezrokah:
**- Summary**

Follow up on https://github.com/netlify/cli/pull/2826
Fixes an issue where functions headers from `lambda-local` are not passed down to the functions server.
Found it while reviewing https://github.com/netlify/cli/pull/2852

**- Test plan**

Added a test case for it

🐪

### DIFF
--- a/src/lib/functions/runtimes/js/index.js
+++ b/src/lib/functions/runtimes/js/index.js
@@ -54,7 +54,7 @@ const invokeFunction = async ({ context, event, func, timeout }) => {
   // If a function builder has defined a `buildPath` property, we use it.
   // Otherwise, we'll invoke the function's main file.
   const lambdaPath = (func.buildData && func.buildData.buildPath) || func.mainFile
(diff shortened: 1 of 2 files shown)


src/lib/functions/runtimes/js/index.js
```diff
@@ -54,15 +54,15 @@ const invokeFunction = async ({ context, event, func, timeout }) => {
...
+  const result = await lambdaLocal.execute({
     clientContext: context,
     event,
     lambdaPath,
     timeoutMs: timeout * SECONDS_TO_MILLISECONDS,
     verboseLevel: 3,
   })
 
-  return { body, statusCode }
+  return result
```
erezrokah, line 65: We can also pass just the stuff we're expecting.

eduardoboucas (approved):
Great catch, thanks!

PR #2920, opened 2021-07-12
fix(command-deploy): sync redirects after build

erezrokah:
**- Summary**

This is the CLI equivalent of https://github.com/netlify/build/pull/3259 and https://github.com/netlify/build/pull/3271

When a build plugin changes redirects we need to make sure those changes are deployed.
`netlify/build` returns a single `config.redirects` property with all the redirects from `netlify.toml` and `_redirects` in the correct priority order.

To make sure the new redirects are applied in the correct order this PR updates the deploy command's code to check if the redirects were modified and if they were:
1. Hides any existing `_redirects` file so it won't be deployed
2. Creates a new `netlify.toml` with the new redirects array (based on the existing one if there is one)
3. Runs the deploy
4. Unhides the `_redirects` if needed and deletes the created `netlify.toml` file

**- Test plan**

Added a new test

🐱


src/commands/deploy.js
```diff
@@ -366,6 +368,26 @@ const printResults = ({ flags, results, deployToProduction, exit }) => {
...
+const hideRedirects = async (deployFolder) => {
+  await renameAsyncNoError(`${deployFolder}/_redirects`, `${deployFolder}/.redirects`)
+}
+
+const unhideRedirects = async (deployFolder) => {
+  await renameAsyncNoError(`${deployFolder}/.redirects`, `${deployFolder}/_redirects`)
+}
+
+const syncRedirects = async ({ oldConfig, newConfig, deployFolder, configPath }) => {
+  if (oldConfig.redirects === newConfig.redirects) {
```
erezrokah, line 380: I've tested it and this evaluates to `true` is a plugin didn't change the redirects.
We could use `fast-deep-equal` too but since customers can have many redirects we might want to avoid it.

ehmicky:
Thanks a lot @erezrokah!
On Slack, @erezrokah and I discussed exploring whether we could re-use the logic in `@netlify/config` which is doing the same but in production. This would prevent code duplication and promote better separation of concerns, e.g. keeping TOML parsing/serializing to `@netlify/config`. This might require making then restoring a copy of the files modified by `@netlify/config` in production like `_redirects` and `netlify.toml`.

erezrokah:
This PR now depends on https://github.com/netlify/build/pull/3379

ehmicky (approved):
Yay to code re-use! :+1: 
Approving already, although this is pending on the PR in Netlify Build.

PR #2929, opened 2021-07-13
feat: add support for Go functions

eduardoboucas:
**- Summary**

Adds support for Go functions in Netlify Dev.

**- Test plan**

Added a new test.

840447-bebe-gopher-pico-fora-de-buraco-foto


tests/serving-functions-go.test.js
```diff
@@ -0,0 +1,104 @@
...
+const pWaitFor = require('p-wait-for')
+
+const { withDevServer } = require('./utils/dev-server')
+const got = require('./utils/got')
+const { withSiteBuilder } = require('./utils/site-builder')
+
+const WAIT_INTERVAL = 1800
+const WAIT_TIMEOUT = 30000
+
+test('Updates a Go function when a file is modified', async (t) => {
```
eduardoboucas, line 11: Feels a bit weird to be including the Go source files in the test file, but I'm not sure how else to do it.

erezrokah: We can have a subdirectory `go-functions` and read the code from there?
The benefit is that we'll get code formatting and highlighting.

This is good too 🚀

erezrokah (approved):

PR #2930, opened 2021-07-13
fix(serving-functions): wait for watcher to be ready after registering a new function

erezrokah:
**- Summary**

Enables the functions serving tests and fixes the issue that caused them to fail.
Since we didn't wait for `buildFunctionAndWatchFiles` to finish when running `registerFunction`, the following could happen:
1. We start the initial scan and discover a new function `hello.ts`
2. We register it and start building it
3. The user changes `hello.ts` (triggering a change event)
4. We finish building the function and setup the watcher for `hello.ts`

As a result the change event is skipped.

The current solution is to wait for `buildFunctionAndWatchFiles` to complete as a part of `registerFunction` which has a performance cost of having to wait for all functions to build before starting the functions server.
I couldn't think of another way to fix it, happy to get some more feedback on this.

**- Test plan**

Existing tests

🐒

### DIFF
--- a/src/lib/functions/netlify-function.js
+++ b/src/lib/functions/netlify-function.js
@@ -1,6 +1,11 @@
+const { Mutex, withTimeout } = require('async-mutex')
+
 const { difference } = require('../../utils/difference')
 
 const BACKGROUND_SUFFIX = '-background'
+
+// 3 minutes
+const MUTEX_TIMEOUT = 18e4
 
 class NetlifyFunction {
   constructor({
@@ -31,6 +36,7 @@ class NetlifyFunction {
     // List of the function's source files. This starts out as an empty set
     // and will get populated on every build.
     this.srcFiles = new Set()
+    this.mutex = withTimeout(new Mutex(), MUTEX_TIMEOUT)
   }
 
   // The `build` method transforms source files into invocable functions. Its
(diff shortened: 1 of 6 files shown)


src/lib/functions/netlify-function.js
```diff
@@ -39,25 +45,29 @@ class NetlifyFunction {
...
-      projectRoot: this.projectRoot,
-    })
-
-    this.buildQueue = buildFunction({ cache })
-
     try {
-      const { srcFiles, ...buildData } = await this.buildQueue
-      const srcFilesSet = new Set(srcFiles)
-      const srcFilesDiff = this.getSrcFilesDiff(srcFilesSet)
+      return await this.mutex.runExclusive(async () => {
```
erezrokah, line 49: This is just another way of doing the same thing. Happy to revert it

erezrokah: Reverted in https://github.com/netlify/cli/pull/2930/commits/14dfe1d7e029da6d48aa9fd6334031638686e487 for clarity

erezrokah:
Adding more context - we don't know which files to track until we finish building the functions for the first time.

As we don't want to wait for all functions to build before starting this function server closing in favor of https://github.com/netlify/cli/pull/2933 since it's not very likely that a user will manage to change a file before the first build.

We might find a way to improve this by tracking file changes on the directory before the first build and accounting for that, but that won't work for supporting files that are outside the functions directory.

PR #2931, opened 2021-07-13
refactor: enable and fix fp/no-proxy

tinfoil-knight:
**- Summary**

See #1891 

Enables the fp/no-proxy  for `eslint-plugin-fp`.

There was only a single instance of Proxy being used to disable colors for the [chalk](https://www.npmjs.com/package/chalk) package. The package now has a `level` option that can be used to disable the coloring as documented here: https://github.com/chalk/chalk#chalklevel.

The file: `src/utils/command.js` was using this custom chalk instance.

**- Test plan**

`npx ava -v -s` ran without errors. I tried out the `sites:list` command normally and with the `--json` flag. As expected `--json` produced discolored output.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,7 +22,6 @@ module.exports = {
     'fp/no-mutating-assign': 0,
     'fp/no-mutating-methods': 0,
     'fp/no-mutation': 0,
-    'fp/no-proxy': 0,
     'fp/no-this': 0,
     'import/max-dependencies': 0,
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
Thanks @tinfoil-knight nice catch. I got an error on a missing ctor with the previous code.
Looks like we should use https://github.com/chalk/chalk/tree/v4.1.1#chalklevel

The examples on the default branch are already updated for the switch to ESM

PR #2949, opened 2021-07-15
chore: skip main workflow on push event

eduardoboucas:
**- Summary**

I think we can skip the main workflow on the `push` event, since on this event we run `release-please` which will trigger its own run of the test suite as part of `npm publish`. Running the test suite again feels redundant?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,10 +1,6 @@
 name: Testing for CLI
 
 on:
-  push:
-    branches: [main]
-    paths-ignore:
-      - 'docs/**'
   pull_request:
     branches: [main]
     paths-ignore:


erezrokah:
> I think we can skip the main workflow on the `push` event, since on this event we run `release-please` which will trigger its own run of the test suite as part of `npm publish`. Running the test suite again feels redundant?

This will also skip the tests when merging any PR to the `main` branch no?

eduardoboucas:
> This will also skip the tests when merging any PR to the main branch no?

That should be covered by the `pull_request` event (https://github.com/netlify/cli/pull/2949/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R4). I think `push` only fires when you push directly to that branch?

erezrokah:
I don't thing `pull_request`[ triggers on merge](https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359).

When you merge a PR, you push a new commit to the repo.

I think we can keep `push` and add `paths-ignore` to the changelog? Assuming only release PRs update the changelog that should avoid running the tests when they are merged.

erezrokah:
Closing this as stale, I think we can re-visit per a more general discussion if we still want to run tests on branches that have the `Require branches to be up to date before merging` setting enabled.

PR #2971, opened 2021-07-20
fix: shorthand notation o for the offline flag fixes #2400

### LINKED ISSUE #2400: bug(command-dev): offline flag shorthand version `o` doesn't work
**Describe the bug**

See https://github.com/netlify/cli/issues/2399#issuecomment-841801570

`netlify dev --offline` works but not `netlify dev -o`

**To Reproduce**

1. Run `netlify dev -o` on a machine without network connection

**Configuration**

```bash
  System:
    OS: macOS 11.3
    Memory: 173.85 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.1.0 - ~/.nvm/versions/node/v16.1.0/bin/node
    npm: 7.11.2 - ~/.nvm/versions/node/v16.1.0/bin/npm
  npmGlobalPackages:
    netlify-cli: 3.29.15
```

**Expected behavior**

Shorthand version of `offline` should work

**Additional context**

Relevant code https://github.com/netlify/cli/blob/b6da1ac6e1d971c82e00986132da34d6c540a789/src/utils/command.js#L141
See workaround for `deploy`:
https://github.com/netlify/cli/blob/b6da1ac6e1d971c82e00986132da34d6c540a789/src/utils/command.js#L330

The proper solution is to not use `minimist` in the `Command` base class and only rely on `oclif` parsing.

tinfoil-knight:
**- Summary**

fixes https://github.com/netlify/cli/issues/2400

This PR makes a temporary fix for the shorthand notation (`-o`) of the offline flag not working due to parsing flags through `minimist` instead of the oclif parser. 

The fix is temporary since the ideal solution using the oclif parser might require substantial refactoring.

**- Test plan**

- `npx ava -v -s` ran without errors.
- Tested a log statement in the `getConfig` method to see if the value for the `offline` parameter was `true` for both `--offline` and `-o` flags.

### DIFF
--- a/src/utils/command.js
+++ b/src/utils/command.js
@@ -143,7 +143,7 @@ class BaseCommand extends TrackedCommand {
   }
 
   // Find and resolve the Netlify configuration
-  async getConfig({ cwd, host, offline = argv.offline, pathPrefix, scheme, state, token }) {
+  async getConfig({ cwd, host, offline = argv.offline || argv.o, pathPrefix, scheme, state, token }) {
     try {
       return await resolveConfig({
         config: argv.config,


erezrokah (approved):

erezrokah:
🚀

PR #3024, opened 2021-07-27
fix(command-status): fix typo Github -> GitHub

erezrokah:
**- Summary**

Follow up on https://github.com/netlify/cli/pull/3022.
Made another fix to the command and added a test

**- Test plan**

Added a new test

🐈


src/commands/status/index.js
```diff
@@ -44,7 +44,7 @@ class StatusCommand extends Command {
     const accountData = {
       Name: user.full_name,
       Email: user.email,
-      Github: ghuser,
+      GitHub: ghuser,
```
eduardoboucas, line 47: 🙏🏻

eduardoboucas (approved):

eduardoboucas:
The snapshot test is failing. Seems to be an issue with whitespace?

erezrokah:
> The snapshot test is failing. Seems to be an issue with whitespace?

Yes! Looking into it ❤️

erezrokah:
@eduardoboucas I [fixed](https://github.com/netlify/cli/pull/3024/commits/bffeca5166dfa3fad2febcaec326b8c5b42fb9d3) the test by using the `--json` flag. However, while doing so I realized this might be a breaking change if someone is accessing the `Github` property.

I'll see if we can get some data on how often its used.

erezrokah:
> @eduardoboucas I [fixed](https://github.com/netlify/cli/pull/3024/commits/bffeca5166dfa3fad2febcaec326b8c5b42fb9d3) the test by using the `--json` flag. However, while doing so I realized this might be a breaking change if someone is accessing the `Github` property.
> 
> I'll see if we can get some data on how often its used.

Update: the `--json` flag is undocumented https://cli.netlify.com/commands/status/ so I think we're safe

PR #3027, opened 2021-07-28
chore(ci): skip tests for release-please publish flow

erezrokah:
**- Summary**

This avoids re-running the tests before publishing on `release-please` workflows

**- Test plan**

Existing tests

🐪


.github/workflows/release-please.yml
```diff
@@ -30,6 +30,8 @@ jobs:
       - name: Install site dependencies
         run: npm run site:build:install
         if: ${{ steps.release.outputs.release_created }}
+      - name: Install core dependencies
```
erezrokah, line 33: Moved from the `package.json` script so we can use `run-s` and `run-local`.

For local publishing it's expected to for the developer to already those available.

eduardoboucas (approved):

PR #3028, opened 2021-07-28
refactor: warn, exit and error utils from oclif command class

tinfoil-knight:
**- Summary**

Ref: #2728 

Some logging and process control utilities (`log`, `warn`,`exit`, `error`) were being passed down to each method and helper functions which isn't ideal. Thus, `log` was refactored in #2827 & `warn`, `exit` & `error` have been refactored in this PR.

The utilities have been referenced from the `@oclif/command` package to not break any existing functionality. They are present in the `src/utils/command-helpers.js` file.

**- Test plan**

`npx ava --verbose --serial`


tinfoil-knight:
@erezrokah Please review the PR.

erezrokah:
> @erezrokah Please review the PR.

Hi @tinfoil-knight, sorry for the delay. The approach for creating these utilities is nice! The only concern is for them to change in the base class, so our implementation will drift. However I don't think it happens very often (without it being a breaking change).

Since this PR touches many files, it will take some time to review as I would like to invoke each code flow to make sure the change works (not all flows are covered by tests).

tinfoil-knight:
> Hi @tinfoil-knight, sorry for the delay. The approach for creating these utilities is nice! The only concern is for them to change in the base class, so our implementation will drift. However I don't think it happens very often (without it being a breaking change).
> 
> Since this PR touches many files, it will take some time to review as I would like to invoke each code flow to make sure the change works (not all flows are covered by tests).

That seems reasonable.

Thank you for the response.

erezrokah:
@tinfoil-knight, sorry for the delay to review this.
I was thinking of a way we can move forward and reduce the risk.
How about we break this PR into an initial PR that just adds the utility methods? Basically just [this change](https://github.com/netlify/cli/pull/3028/files#diff-01cdd11fb026ca768cbe8311004eae8a2aeca7937ddd50393cd9f5048ce06a26)

Then we can gradually introduce the usage of those methods into different areas of the codebase, and we won't need to keep this PR up to date with other changes.
That way we could also test each specific change separately. 

tinfoil-knight:
> @tinfoil-knight, sorry for the delay to review this.

Don't worry about it.

> How about we break this PR into an initial PR that just adds the utility methods? Basically just [this change](https://github.com/netlify/cli/pull/3028/files#diff-01cdd11fb026ca768cbe8311004eae8a2aeca7937ddd50393cd9f5048ce06a26)
> Then we can gradually introduce the usage of those methods into different areas of the codebase, and we won't need to keep this PR up to date with other changes.

Yeah. That seems like a good idea. I'll pick this and the other sub-PRs up then in a few days.

tinfoil-knight:
Started with https://github.com/netlify/cli/pull/3247

tinfoil-knight:
Continuing with https://github.com/netlify/cli/pull/3346.

Apologies for the late PR. Was bogged down with college work.

ehmicky:
The apology is on our side @tinfoil-knight for the delay reviewing your contribution. Thanks a lot for doing all of this! :heart: 

Also, college work always prevails! :)

Erez is on PTO so I'm going to help with reviewing your awesome contribution (which I find very helpful!). If you want to keep breaking this down into separate PRs, I'll make sure to review it as soon as possible. :+1:

tinfoil-knight:
Thank you.

tinfoil-knight:
Closing this since the work has been done in the referenced PRs. @ehmicky issue #2728 can also be closed now.

PR #3037, opened 2021-07-29
feat: add Rust functions runtime

eduardoboucas:
**- Summary**

Enables support for Rust functions when the `NETLIFY_EXPERIMENTAL_BUILD_RUST_SOURCE` environment variable is enabled.

**- Test plan**

I'll add tests separately.

🦀


src/lib/functions/runtimes/rust/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+const { readFileAsync } = require('../../../fs')
+const { getPathInProject } = require('../../../settings')
+const { runFunctionsProxy } = require('../../local-proxy')
+
+const build = async ({ func }) => {
+  const functionDirectory = dirname(func.mainFile)
+  const cacheDirectory = resolve(getPathInProject(['functions-serve']))
+  const targetDirectory = join(cacheDirectory, func.name)
+  const crateName = await getCrateName(functionDirectory)
```
eduardoboucas, line 18: We should probably move this to zip-it-and-ship-it in the future.

erezrokah (approved):

PR #3047, opened 2021-07-30
chore: add util for mocking `execa` calls

eduardoboucas:
**- Summary**

As we add support for different Netlify Functions runtimes, we need our test suite to capture their usage, namely in `netlify dev`.

However, using runtimes like Go and Rust involves having the CLI compiling binaries from source, which can mean quite expensive operations that will have a significant impact on the already heavy test suite. This is already noticeable with the single test we have for Go functions, and will be compounded once we add more tests and especially more runtimes, like Rust.

As discussed internally, the only viable approach seems to be shifting the responsibility of asserting the correctness of the function invocations on the various runtimes to the functions proxy repository, instead of the CLI. On the CLI side, we can simply assert the interaction between the various moving pieces (the filesystem, the proxy, and the functions server), mocking any expensive calls to compile binaries or invoke the proxy.

With that in mind, this PR adds a very thin layer on top of `execa` that allows it to be easily mocked in tests, even when CLI runs in a separate process. Here's how it works:

- A new runtime util was created at `src/lib/execa.js`. This util should replace the original npm module whenever we want the ability to mock an `execa` call. The util is literally just a `require` call that conditionally calls the original npm module or an alternative path, depending on the value of the `NETLIFY_CLI_EXECA_PATH` environment variable

- A new test util was created at `tests/utils/mock-execa.js`. It exports a `createMock` function that writes to disk a JavaScript file with the contents it receives and returns an object in the format `{ NETLIFY_CLI_EXECA_PATH: <path> }`. This object can then be supplied to the CLI's dev server via the `env` parameter.

To see an example of how it's used, see the modified `tests/serving-functions-go.test.js` file.

**- Test plan**

One test updated.

1200px-Mimus_polyglottos_adult_02_cropped


src/utils/execa.js
```diff
@@ -0,0 +1,9 @@
+const { env } = require('process')
+
+// This is a thin layer on top of `execa` that allows consumers to provide an
+// alternative path to the module location, making it easier to mock its logic
+// in tests (see `tests/utils/mock-execa.js`).
+// eslint-disable-next-line import/no-dynamic-require
+const execa = require(env.NETLIFY_CLI_EXECA_PATH || 'execa')
```
ehmicky, line 7: To keep static imports (which might be useful when switching to ES modules in the future), should we do something more like this?

```
const execaLib = require('execa')
const execa = env.NETLIFY_CLI_EXECA_PATH === undefined ? execaLib : require(env.NETLIFY_CLI_EXECA_PATH)
```

tests/serving-functions-go.test.js
```diff
@@ -42,58 +23,69 @@ func main() {
...
+          content: `<mock main.go>`,
         },
       ])
       .buildAsync()
 
+    const originalBody = 'Hello, world!'
+    const updatedBody = 'Hello, Netlify!'
+
+    const [execaMock, removeExecaMock] = await createExecaMock(`
+      const { writeFileSync } = require('fs')
```
ehmicky, line 43: Should we put this contents into a separate file, then read it?
This would allow keeping JavaScript syntax highlighting, linting, autocompletion, etc.

tests/utils/mock-execa.js
```diff
@@ -0,0 +1,30 @@
...
+    NETLIFY_CLI_EXECA_PATH: path,
+  }
+  const cleanup = () => removeMock(path)
+
+  return [env, cleanup]
+}
+
+const removeMock = async (path) => {
+  try {
+    await rmdirRecursiveAsync(path)
```
ehmicky, line 24: [sand] Should this pattern (noop `try`/`catch` + `rmdirRecursiveAsync()`) be moved to a new utility function in `./src/lib/fs`?

tests/serving-functions-go.test.js
```diff
@@ -102,12 +94,14 @@ require github.com/aws/aws-lambda-go v1.20.0`,
...
 
-                return response === 'Hello, Netlify!'
+                return response === updatedBody
               },
               { interval: WAIT_INTERVAL, timeout: WAIT_TIMEOUT },
             ),
           outputBuffer,
         )
+
+        await removeExecaMock()
```
ehmicky, line 104: Should this be in a `try`/`finally` block? To ensure cleanup is performed even if an exception is thrown before this.

tests/utils/mock-execa.js
eduardoboucas: Done in https://github.com/netlify/cli/pull/3047/commits/38a6c6a11e29462e2ac1d227339c295c7fcd5cb9.

src/utils/execa.js
eduardoboucas: But we'll still have a dynamic import (when we're requiring the env variable), so the change you're proposing just means having one instead of two. Do you still think it's worth doing?

tests/serving-functions-go.test.js
eduardoboucas: I did it this way because all test files in the CLI are self-contained and there's precedent for keeping the source of any associated function files in the test file itself vs. having a separate file (e.g. https://github.com/netlify/cli/blob/main/tests/serving-functions.test.js#L98-L120).

I agree with the benefits of what you're suggesting, but I personally don't think they justify breaking that convention. What do you think?

tests/serving-functions-go.test.js
eduardoboucas: Done in https://github.com/netlify/cli/pull/3047/commits/5426129adb724acfc244e626fea85c5eab2c4661.

src/utils/execa.js
ehmicky: Yes, exactly, that's what I meant: one instead of two.
It's not super important, but I am wondering for example if type inference on `execa` (once we start using TypeScript) might be better that way? Regardless, this is very nit-picky, so please feel free to discard :+1:

tests/serving-functions-go.test.js
ehmicky: Oh, ok, that's just me not being used to work on the CLI codebase. Since other test files are also doing this, I agree this makes more sense to keep it consistent.

That being said, at the codebase scale (outside of this PR), I am wondering whether not having to split into several files is enough of a benefit compared to having JavaScript static analysis (linting, Prettier, autocompletion, syntax highlighting, and more). It seems to me the benefits of the latter outweighs the former.

ehmicky (approved):

src/utils/execa.js
eduardoboucas: I'm happy either way. Done in https://github.com/netlify/cli/pull/3047/commits/bec7f55ace9a61823d9e7decb9584b5df0c22873.

tests/serving-functions-go.test.js
eduardoboucas: I think it's mostly about our approach of using a site builder that creates the fixtures dynamically vs. having a fixtures directory like we have in Netlify Build. I don't have a strong preference on one over the other, but I feel that having a mix of both, where we create part of the fixtures dynamically but still rely on certain files to exist, is not ideal.

We should probably discuss this in a separate forum, though! 😅

PR #3071, opened 2021-08-02
fix: start functions server if internal functions directory exists

### LINKED ISSUE #3064: Functions server does not start if functions directory is not set
**Describe the bug**

If the user has not set a functions directory, then the functions server is not started in `ntl dev`. This means that any functions in the internal directory are not served. This is because of [this check](https://github.com/netlify/cli/blob/60167575efbb0aa26d6ea5dbd14882509fe31275/src/lib/functions/server.js#L152)

eduardoboucas:
**- Summary**

Currently, CLI will only start the functions server if a functions directory is configured, even if the internal functions directory exists and contains valid functions.

This PR changes that, so that the server is start if at least one of the directories (public and internal) exists.

Closes #3064.


src/lib/functions/registry.js
```diff
@@ -186,8 +185,8 @@ class FunctionsRegistry {
 
       const func = new NetlifyFunction({
         config: this.config,
+        directory: directories.find((directory) => mainFile.startsWith(directory)),
```
eduardoboucas, line 188: This is the main change in this file. The registry constructor used to receive the path to the functions directory, but that's not necessary anymore. Instead, each function can figure out which of the paths in `directories` this function belongs to, and it passes that value down to the function instance.

ehmicky (approved):

ehmicky:
Note: CI tests are failing.

Issue #3025, opened 2021-07-27
Expand description of `netlify init`


KyleBlankRollins:
**Describe the bug**

The [current description for `netlify init`](https://cli.netlify.com/commands/init) makes it seem as if continuous integration and git are required. This can be confusing if people want to create a site that's not tied to those things.

Conversation started in the [Netlify forums](https://answers.netlify.com/t/how-to-creat-a-site-so-can-use-local-dev-but-no-git/41243).

Possible update to the description:
- “Configure continuous deployment for a existing site or create a new site without continuous deployment.”

erezrokah:
Hi @KyleBlankRollins, the `ntl init` command can also configure continuous deployment for a new site.

The `ntl init` command falls black to creating a new site in case the project is not a git repo (so we can't complete continuous deployment).

How about we point to the `sites:create` command instead? For example: 
`Configure continuous deployment for a new or existing site. To create a site without continuous deployment see the 'sites:create' command.`

KyleBlankRollins:
@erezrokah, that makes a lot of sense. :) I can open a PR if that's the preferred docs workflow for this repo.

erezrokah:
That would be great @KyleBlankRollins, thank you!

Please note the CLI docs are autogenerated from the code so you'll need to update the following line:
https://github.com/netlify/cli/blob/44122805e6a9d2faecf6c0efffa835c0f7d9571d/src/commands/init.js#L203

and run `npm run site:build` locally to sync the docs.

PR #3073, opened 2021-08-02
fix: update `InitCommand.description`

KyleBlankRollins:
**- Summary**
Closes [#3025](https://github.com/netlify/cli/issues/3025)

Make the `netlify init` docs description clearer.

**- Test plan**
I don't think I'm properly set up locally for tests. Getting some errors that mostly seem to be around authentication. I'm happy to dig in and figure these out if necessary, but wasn't sure if that's needed with a 1-line string change.


ehmicky (approved):
Thanks @KyleBlankRollins!

ehmicky:
Note: the CI tests were failing because [`npm run docs`](https://github.com/netlify/cli/blob/857c61bcf6347abdbab07cbb6c1daabb91917f73/CONTRIBUTING.md#pull-requests) must be run after changing any CLI command's description in order to update some automatically generated documentation files. I have fixed it :+1:

KyleBlankRollins:
Thank you, @ehmicky! I appreciate the additional context and help.

ehmicky:
Final fix: we have some tests which take a snapshot of the output of the `--help` flag. Those snapshots needed to be updated using `ava -u tests/command.help.test.js`.

src/commands/init.js
```diff
@@ -200,7 +200,7 @@ class InitCommand extends Command {
   }
 }
 
-InitCommand.description = `Configure continuous deployment for a new or existing site`
+InitCommand.description = `Configure continuous deployment for an existing site or create a new site without continuous deployment`
```
erezrokah, line 203: We might need to clarify this.
The `ntl init` command lets you:
1. Configure continuous deployment for an existing site
2. Create a new site with continuous deployment
3. Create a new site without continuous deployment

Creating a new site without continuous deployment is more of a fallback when the project is not configured as a `git` repo.
If users only want to create a site we can direct them to `ntl sites:create`

erezrokah (changes requested):
Thanks @KyleBlankRollins and @ehmicky.

I added a small comment, please let me know what you think

src/commands/init.js
KyleBlankRollins: Hmm. Do we want to not mention "Create a new site without continuous deployment" specifically then? Instead, pointing people to `ntl sites:create`?

For example: "Configure continuous deployment for a new or existing site. To create a new site without continuous deployment, use [`ntl sites:create`](https://cli.netlify.com/commands/sites#sitescreate)"

erezrokah: > For example: "Configure continuous deployment for a new or existing site. To create a new site without continuous deployment, use [`ntl sites:create`](https://cli.netlify.com/commands/sites#sitescreate)"

Yes! I think this is preferable if the user's intention is to solely create a new site.

Also, for the description we should use `netlify sites:create` instead of `ntl sites:create` for consistency with the other docs.

KyleBlankRollins:
@erezrokah, I ran into some issues with running the tests (again). `npm run docs` works fine, so I did that. However, I couldn't get `ava -u tests/command.help.test.js` to work (probably a PATH env var issue locally). And I couldn't get the tests to complete successfully.

I'd appreciate some help with figuring out the testing side of things.

erezrokah:
Thanks @KyleBlankRollins, I updated the snapshots by running `npx ava -u tests/command.help.test.js` (that should handle any path related issues).
Also I changed the text to point to `ntl sites:create`.

KyleBlankRollins:
Ack. I don't know why I forget about `npx`. It's so useful!

LGTM! Thanks, @erezrokah. Appreciate your help and guidance here. And @ehmicky!

Issue #3020, opened 2021-07-27
Improve DX for `ntl build` + `ntl deploy` command sequence


eduardoboucas:
## Problem

Some users will run the `build` command followed by `deploy` with the expectation of it being a two-step process, in which the first command creates a set of artefacts that the second command deploys. Unfortunately, this isn't always the case.

For example, imagine a site with a plugin that modifies certain configuration properties related to functions, via a `onPreBuild` hook. When running `build`, those changes will affect the functions bundling stage and the generated zip archives will have the expected contents. But if a `deploy` command follows, the functions bundling stage will run again, except this time the plugin will not run, so the changes to the configuration will not take place and a new set of zip archives (with the wrong result) will be deployed instead of the ones generated previously.

## Proposed solution

When functions are bundled as part of the `deploy` command, the zip archives are placed in a `.netlify/functions` directory. The `deploy` command could treat this directory as a bundling cache, so that if exists, it skips the bundling stage and directly uploads the archives that have been previously generated.

Caching is always difficult and we want to make sure we minimise the risk of deploying the wrong functions, so here are a few scenarios and how CLI would respond.

### Scenario 1: Using `ntl deploy` without `ntl build`

A user uses `ntl dev` or any other development server to develop their site locally. To deploy, they run their build command (e.g. `npm run build`) and then deploy to Netlify using `ntl deploy`. They don't use `ntl build` as part of their workflow.

In this scenario, CLI would continue to work in the same way, since the `deploy` command does not create a `.netlify/functions` directory, so the functions would be bundled each time.

### Scenario 2: Using `ntl deploy --build`

A user uses `ntl dev` or any other development server to develop their site locally. To deploy, they use the `netlify deploy --build` command, which will build their site and deploy it in one go.

In this case, we can safely skip the functions bundling (see https://github.com/netlify/cli/issues/2596), since we know for a fact that the build and the deployment are an atomic operation.

If they use `netlify build` followed by `netlify deploy`, then it's guaranteed that `.netlify/functions` will contain the correct functions by the time `deploy` starts.

### Scenario 3: Using `ntl deploy` followed by `ntl build`

A user uses `netlify dev` or any other development server to develop their site locally. When they're ready to deploy, they use the `ntl build` to build their site followed by `ntl deploy` to finalise the deployment. 

In this case, we know that `.netlify/functions` will contain the correct functions, bundled and ready for deployment, by the time `ntl deploy` starts, so we can skip a second bundling stage.

[...]

PR #3091, opened 2021-08-03
feat: add support for manifest file

### LINKED ISSUE #2596: Avoid re-bundling functions when running `netlify deploy --build`

When running `netlify deploy --build`, function bundling is happening twice.

The `deploy` should bypass its call to `zip-it-and-ship-it` when the `--build` flag is present, since we'll already have bundled the functions.

Yes

eduardoboucas:
**- Summary**

With this PR, CLI looks for a functions manifest file (introduced in Netlify Build v17.7.0), and uses it to avoid re-bundling functions in a `deploy` command.

The bundled functions from the manifest are used if:

1. Their timestamp is not older than 2 minutes, and
2. The `--skip-functions-cache` command was not used in the `deploy` command

Closes #2596 and #3020.

**- Test plan**

New tests added.

9669c59a7f3fe784bae4141bd93159e9


eduardoboucas:
I wonder if this should be a major version, since technically we're changing the behaviour of the `deploy` command.

@erezrokah @ehmicky what do you think?

src/utils/deploy/hash-fns.js
```diff
@@ -7,9 +7,63 @@ const pump = promisify(require('pump'))
...
+
+      statusCb({
+        type: 'functions-manifest',
+        msg: 'Using bundled functions from manifest file (use --bundle to override)',
+        phase: 'stop',
+      })
+
+      return functions
+    } catch (error) {
+      statusCb({
```
ehmicky, line 38: Should this be split into two separate errors handlers: one if `require()` fails, and one if manifest expired?

ehmicky (approved):

src/utils/deploy/hash-fns.js
eduardoboucas: Since it's not a fatal error and not something actionable by the user (other than using `--bundle`), I personally don't see the need to add too much detail, especially as it would complicate the logic a little bit.

I'm happy to do it if you feel strongly about it, though.

ehmicky: This makes sense :+1:

docs/commands/deploy.md
```diff
@@ -98,6 +98,7 @@ netlify deploy
 - `timeout` (*string*) - Timeout to wait for deployment to finish
 - `trigger` (*boolean*) - Trigger a new build of your site on Netlify without uploading local files
 - `build` (*boolean*) - Run build command before deploying
+- `bundle` (*boolean*) - Ignore functions manifest file, forcing functions to always be bundled as part of the deployment
```
eduardoboucas, line 101: I'm not too sure about the name of this flag. I think it should be clear that it overrides the default behaviour, so perhaps it should be something like `--force-bundle` or `--ignore-cache`? 🤔

ehmicky: We might want to see it more from the perspective of the users. For example, the users would not know what the manifest file is. From their perspective, they might only care about whether their Functions are uploaded or not.

I might be missing some details, but what do you think about something like: `--skip-functions-cache` with a description along the lines of "Deploy the Functions even if they have been already deployed". I think this is most likely inaccurate and I might be not understanding the full intent behind this flag, so please feel free to correct me.

eduardoboucas: Agreed 100%! I've changed it to `--skip-functions-cache` and removed any references to "manifest file" from the user-facing messages.

tests/command.deploy.test.js
```diff
@@ -565,6 +566,124 @@ if (process.env.NETLIFY_TEST_DISABLE_LIVE !== 'true') {
...
+
+    await withSiteBuilder('site-with-functions-manifest', async (builder) => {
+      await builder
+        .withNetlifyToml({
+          config: {
+            build: { publish: 'out' },
+            functions: { directory: 'functions' },
+          },
+        })
+        .withCopiedFile({
```
erezrokah, line 585: Nice!

erezrokah (approved):

eduardoboucas:
Added another test in https://github.com/netlify/cli/pull/3091/commits/567cfba673f44a9df9707292c44a6bc11748984d to assert that we don't deploy pre-bundled functions if the manifest file is older than the configured TTL.

PR #3105, opened 2021-08-05
chore: pin release-please version

erezrokah:
This is an attempted workaround for https://github.com/google-github-actions/release-please-action/issues/337

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -12,7 +12,7 @@ jobs:
         with:
           private-key: ${{ secrets.TOKENS_PRIVATE_KEY }}
           app-id: ${{ secrets.TOKENS_APP_ID }}
-      - uses: GoogleCloudPlatform/release-please-action@v2
+      - uses: GoogleCloudPlatform/release-please-action@d3d5b411d22e25052bb20e73a0ec1b156d67d944
         id: release
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         with:
           private-key: ${{ secrets.TOKENS_PRIVATE_KEY }}
           app-id: ${{ secrets.TOKENS_APP_ID }}
-      - uses: GoogleCloudPlatform/release-please-action@v2
+      - uses: GoogleCloudPlatform/release-please-action@d3d5b411d22e25052bb20e73a0ec1b156d67d944
```
BenWildeman, line 15: you could do this instead:
```suggestion
      - uses: GoogleCloudPlatform/release-please-action@v2.28
```
this is what we did for our repos that relies on this. you can find the previous tags here: https://github.com/google-github-actions/release-please-action/tags

ehmicky (approved):

.github/workflows/release-please.yml
erezrokah: Thanks @BenWildeman, that works too. Using a commit sha has the advantage of being immutable, e.g. someone can remove/update the tag.

Not saying it happens often, but for a workaround I rather use a sha,

PR #3125, opened 2021-08-09
feat: install function template deps in site-level `package.json`

eduardoboucas:
**- Summary**

Currently, dependencies of a function template are installed in the sub-directory created for the function. This is a behaviour that we discourage, since we advise users to install their function dependencies in the site-level `package.json`.

With this PR, CLI looks for a site-level `package.json` and installs the function dependencies there.

**- Test plan**

Added 2 new tests.

images


tests/command.functions.test.js
```diff
@@ -97,6 +97,152 @@ test('should create a new function directory when none is found', async (t) => {
...
+      })
+
+      handleQuestions(childProcess, createFunctionQuestions)
+
+      await childProcess
+
+      // eslint-disable-next-line import/no-dynamic-require, node/global-require
+      const { dependencies } = require(`${builder.directory}/package.json`)
+
+      // NOTE: Ideally we should be running this test with a specific template,
```
eduardoboucas, line 166: This is annoying. 😡

src/commands/functions/create.js
```diff
@@ -301,12 +303,50 @@ const downloadFromURL = async function (context, flags, args, functionsDir) {
...
+  }
+
+  // eslint-disable-next-line import/no-dynamic-require, node/global-require
+  const { dependencies: siteDependencies, devDependencies: siteDevDependencies } = require(sitePackageJson)
+  const dependencies = getNpmInstallPackages(siteDependencies, functionDependencies)
+  const devDependencies = getNpmInstallPackages(siteDevDependencies, functionDevDependencies)
+  const npmInstallPath = path.dirname(sitePackageJson)
+
+  if (dependencies.length !== 0) {
+    await execa('npm', ['i', dependencies.join(' '), '--save'], { cwd: npmInstallPath })
```
ehmicky, line 340:
```suggestion
    await execa('npm', ['i', ...dependencies, '--save'], { cwd: npmInstallPath })
```

ehmicky, line 340: Should errors be handled?

ehmicky, line 340: Should the output be shown to the users, i.e. using `stdio: 'inherit'`? This might help with making them realize the command is not stuck. This might also help with debugging.

ehmicky, line 340: Should `--no-audit` and/or `--no-fund` be used?

src/commands/functions/create.js
```diff
@@ -301,12 +303,50 @@ const downloadFromURL = async function (context, flags, args, functionsDir) {
...
-const installDeps = function (functionPath) {
-  return new Promise((resolve) => {
-    cp.exec('npm i', { cwd: path.join(functionPath) }, () => {
-      resolve()
-    })
-  })
+// Takes an existing list of packages and a list of packages needed by a
+// function, and computes the list of npm modules with version range that
+// must be installed.
+const getNpmInstallPackages = (existingPackages = {}, neededPackages = {}) =>
```
ehmicky, line 309: Would this work for Yarn users?

src/commands/functions/create.js
eduardoboucas: They're currently bubbling up, so the user will see them if something goes wrong. Do you have something different in mind?

src/commands/functions/create.js
```diff
@@ -301,12 +303,50 @@ const downloadFromURL = async function (context, flags, args, functionsDir) {
...
+    await execa('npm', ['i', dependencies.join(' '), '--save'], { cwd: npmInstallPath })
+  }
+
+  if (devDependencies.length !== 0) {
+    await execa('npm', ['i', devDependencies.join(' '), '--save-dev'], { cwd: npmInstallPath })
+  }
+
+  // We installed the function's dependencies in the site-level `package.json`,
+  // so there's no reason to keep the function's `package.json`. We delete it.
+  fs.unlinkSync(functionPackageJson)
```
ehmicky, line 349: Is that `package.json` created by users? If so, they might not expect it being deleted.

src/commands/functions/create.js
eduardoboucas: The current implementation works with npm only. I agree that it should be revisited, but I don't think that belongs in this PR.

src/commands/functions/create.js
```diff
@@ -301,12 +303,50 @@ const downloadFromURL = async function (context, flags, args, functionsDir) {
...
+
+// When installing a function's dependencies, we first try to find a site-level
+// `package.json` file. If we do, we look for any dependencies of the function
+// that aren't already listed as dependencies of the site and install them. If
+// we don't do this check, we may be upgrading the version of a module used in
+// another part of the project, which we don't want to do.
+const installDeps = async ({ functionPackageJson, functionPath, functionsDir }) => {
+  // eslint-disable-next-line import/no-dynamic-require, node/global-require
+  const { dependencies: functionDependencies, devDependencies: functionDevDependencies } = require(functionPackageJson)
+  const sitePackageJson = await findUp('package.json', { cwd: functionsDir, type: 'file' })
```
ehmicky, line 322: [sand] `type: 'file'` is the default value, so might be omitted.

src/commands/functions/create.js
ehmicky: No, this is probably good enough :+1:

src/commands/functions/create.js
ehmicky: Makes sense :+1:

src/commands/functions/create.js
eduardoboucas: I personally like to be explicit, but I can remove if you feel strongly about it.

src/commands/functions/create.js
eduardoboucas: Done in https://github.com/netlify/cli/pull/3125/commits/fd8aafd709fc6751bdb5e62df454c534afdba014.

src/commands/functions/create.js
ehmicky: Either works :+1:

src/commands/functions/create.js
eduardoboucas: I see where you're coming from, but won't that add too much clutter? People would be typing a CLI command for creating a function and all of a sudden they'd see a ton of output from npm without realising what's happening, which I think would make for a confusing experience. I think that catching errors and showing the appropriate message if something goes wrong should handle most problematic cases?

ehmicky: Note: I forgot to comment, but this was not a stylistic issue, but an escaping one. I believe the previous code would not have worked with multiple dependencies, because `npm install` would have received a single argument with spaces in it, as opposed to multiple arguments.

ehmicky: Yes, you're right, this is better DX :+1:

ehmicky (approved):

src/commands/functions/create.js
eduardoboucas: Actually, done in https://github.com/netlify/cli/pull/3125/commits/3aec5a7f41f3ea488e90d5634e8644cf1bd40133.

src/commands/functions/create.js
eduardoboucas: Done in https://github.com/netlify/cli/pull/3125/commits/3aec5a7f41f3ea488e90d5634e8644cf1bd40133.

eduardoboucas:
@ehmicky I had to add https://github.com/netlify/cli/pull/3125/commits/d2cb4e1ab469aabe5a66524f9d1adee8f54045b3 because the tests were failing on Node 10. Can you give a final 👍🏻 ?

ehmicky (approved):

PR #3127, opened 2021-08-10
feat: add TypeScript function template

### LINKED ISSUE #3132: `functions:create` creates functions with incorrect name
**Describe the bug**

The `functions:create` command allows the user to choose the name of the function to be created, which dictates the name of the sub-directory where the templates will be copied to.

However, the original name of the files will be kept, so if the template is called `foo` and I choose `bar` as the name of the function, the function file will be placed as `[functions directory]/foo/bar.ts`. This isn't a valid function path.

**Expected behavior**

We need to rename the function file to match the name of the sub-directory, so that we end up with `[functions directory]/bar/bar.ts`.

eduardoboucas:
**- Summary**

Adds a new prompt to the `functions:create` command that allows users to select the language of their function. A TypeScript is also included.

https://user-images.githubusercontent.com/4162329/129007739-b832fc93-686d-44f1-ab16-ee4e314b8640.mov

Closes #3132.

**- Test plan**

Updated existing tests.

360_F_265707361_N79UdgpbERwsjHZxIWRixuodCKqYsBF5


src/commands/functions/create.js
```diff
@@ -145,57 +153,47 @@ const formatRegistryArrayForInquirer = function (lang) {
...
-  const [
-    jsreg,
-    // tsreg, goreg
-  ] = [
-    'js',
-    // 'ts', 'go'
-  ].map(formatRegistryArrayForInquirer)
   const specialCommands = [
-    new inquirer.Separator(`----[Special Commands]----`),
+    new inquirer.Separator(),
```
eduardoboucas, line 157: Purely cosmetic change. I personally don't see a reason to draw special attention to these options by using a "Special Commands" header and surrounding the options with `***`.

src/commands/functions/create.js
```diff
@@ -145,57 +153,47 @@ const formatRegistryArrayForInquirer = function (lang) {
...
-      name: `*** Report issue with, or suggest a new template ***`,
+      name: `Report issue with, or suggest a new template`,
       value: 'report',
       short: 'gh-report',
     },
+    new inquirer.Separator(),
   ]
-  const { chosentemplate } = await inquirer.prompt({
-    name: 'chosentemplate',
+  const { language } = await inquirer.prompt({
```
eduardoboucas, line 170: This is the new prompt that will ask users to select the language before presenting the available templates.

src/commands/functions/create.js
```diff
@@ -419,16 +426,19 @@ const scaffoldFromTemplate = async function (context, flags, args, functionsDir)
     fs.unlinkSync(path.join(functionPath, '.netlify-function-template.js'))
     // rename the root function file if it has a different name from default
     if (name !== templateName) {
-      fs.renameSync(path.join(functionPath, `${templateName}.js`), path.join(functionPath, `${name}.js`))
+      const oldPath = path.join(functionPath, `${templateName}.${lang}`)
```
eduardoboucas, line 429: This operation must happen for all languages, not just JS.

ehmicky (approved):
Looking good! :+1:

PR #3134, opened 2021-08-11
fix: use variables in function template filenames

eduardoboucas:
**- Summary**

Leverages [`copy-template-dir`](https://www.npmjs.com/package/copy-template-dir) to ensure that the function directory follows the function name, as opposed to renaming the files after they're created.

The entry file for each template is named `{{name}}.js`, which will get replaced with the name of the function when it's created.


ehmicky (approved):
:+1:

PR #3152, opened 2021-08-12
feat: prints functions proxy output

eduardoboucas:
**- Summary**

Takes the stderr output from the local functions proxy and pipes it to CLI's stdout. This allows functions to print data to the CLI console.


ehmicky (approved):

ehmicky:
Note: CI tests are failing

src/lib/functions/local-proxy.js
```diff
@@ -34,8 +36,11 @@ const runFunctionsProxy = ({ binaryPath, context, directory, event, name, timeou
     '--timeout',
     `${timeout}s`,
   ]
+  const proxyProcess = execa(functionsProxyPath, parameters)
+
+  proxyProcess.stderr.pipe(stdout)
```
ehmicky, line 41: What about checking for `proxyProcess.stderr && proxyProcess.stderr.pipe` existence? This would remove the need to update the mock.

eduardoboucas: Hmm. 🤔 

My instinct is to say that I would rather keep the runtime code simple and shift any complexity to the test suite, and in reality there's no reason to perform that check other than the fact that we sometimes mock `execa` in tests.

Having said that, I don't have a strong preference here, so I can change it if you like.

ehmicky: > I would rather keep the runtime code simple and shift any complexity to the test suite,

Yes, that's a good point. :+1:

PR #3180, opened 2021-08-16
feat: add Go function template

eduardoboucas:
**- Summary**

Adds a Go function template to the `functions:create` command. _Go_ will only be shown as an option in the initial language prompt if the `NETLIFY_EXPERIMENTAL_BUILD_GO_SOURCE` environment variable is set to `true`. This is the same variable we're using to enable the local development flow for Go functions, which we will remove once support in our build system becomes official.

https://user-images.githubusercontent.com/4162329/129536693-2e60de42-0094-41e6-a645-785e6657c234.mov


erezrokah (approved):
🚀

PR #3189, opened 2021-08-16
feat: use `netlify-headers-parser`

ehmicky:
Part of https://github.com/netlify/build/issues/2890

This uses [`netlify-headers-parser`](https://github.com/netlify/netlify-headers-parser) inside Netlify CLI in order to:
  - Prevent code duplication between `@netlify/config` and `netlify dev`
  - Be consistent with how redirects are parsed in `netlify dev` (with `netlify-redirects-parser`)
  - Isolate the headers parsing logic so it's easier to make it use the back-end parsing logic in the future
  - Benefit from the several bug fixes and minor improvements added by `netlify-headers-parser`

I have updated the automated tests and also tested it manually.

### DIFF
--- a/src/utils/headers.test.js
+++ b/src/utils/headers.test.js
@@ -4,7 +4,7 @@ const test = require('ava')
 
 const { createSiteBuilder } = require('../../tests/utils/site-builder')
 
-const { parseHeadersFile, headersForPath, matchesPath } = require('./headers')
+const { parseHeaders, headersForPath, matchesPath } = require('./headers')
 
 const headers = [
   { path: '/', headers: ['X-Frame-Options: SAMEORIGIN'] },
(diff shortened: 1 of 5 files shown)


src/utils/headers.test.js
```diff
@@ -70,77 +70,98 @@ test.after(async (t) => {
...
   })
 })
 
-test('_headers: validate rules', (t) => {
-  const rules = parseHeadersFile(path.resolve(t.context.builder.directory, '_headers'))
-  t.deepEqual(rules, {
-    '/': {
-      'X-Frame-Options': ['SAMEORIGIN'],
+test('_headers: validate rules', async (t) => {
+  const rules = await parseHeaders({ headersFiles: [path.resolve(t.context.builder.directory, '_headers')] })
```
ehmicky, line 84: `parseHeaders()` now returns an array instead of an object.
The source code and tests were updated to take it into account.

src/utils/headers.test.js
```diff
@@ -70,77 +70,98 @@ test.after(async (t) => {
...
 
-test('_headers: headersForPath testing', (t) => {
-  const rules = parseHeadersFile(path.resolve(t.context.builder.directory, '_headers'))
+test('_headers: headersForPath testing', async (t) => {
+  const rules = await parseHeaders({ headersFiles: [path.resolve(t.context.builder.directory, '_headers')] })
   t.deepEqual(headersForPath(rules, '/'), {
-    'X-Frame-Options': ['SAMEORIGIN'],
-    'X-Frame-Thing': ['SAMEORIGIN'],
+    'X-Frame-Options': 'SAMEORIGIN',
+    'X-Frame-Thing': 'SAMEORIGIN',
```
ehmicky, line 137: Header array values are now joined with `,` which is the production behavior. 
The current `netlify dev` behavior is to set several HTTP headers instead.
So this fixes a difference between production and local experience.

JGAntunes (approved):

PR #3229, opened 2021-08-21
fix: add newline when editing the .gitignore

jlengstorf:
without a newline, appending to the `.gitignore` can result in problems

for example:

```
ntl init # => appends .netlify to .gitignore with no newline

echo ".cache" >> .gitignore
```

now the .gitignore looks like this:
```
.netlify.cache
```

adding a newline fixes this problem

**- Summary**

**- Test plan**


erezrokah (approved):
Thanks @jlengstorf, sorry for the late review

PR #3250, opened 2021-08-25
chore(docs-vs-debugging): add information on debugging with esbuild

### LINKED ISSUE #3248: feat(docs-vs-debugging): add information on debugging with `esbuild`

Related to https://github.com/netlify/cli/blob/f5743f02bf3bb091b243862a3ff49e2481c817b9/docs/vscode.md#add-vscode-launch-configurations

Per @eduardoboucas 

>when we use esbuild, the files that we serve are actually different from the files people write (we transpile them and serve them from a different directory). we need to tell VS Code to transform the locations of breakpoints accordingly. we need this line in the VS Code launch file configuration: `"outFiles": ["${workspaceFolder}/.netlify/functions-serve/**/*.js"]`

Update the docs with a note about `esbuild` debugging

Keeping the docs as they are and point to https://answers.netlify.com/t/debugging-typescript-functions/42816

Yes

Patil2099:
Fixes #3248


eduardoboucas:
Thanks for picking this up so quickly, @Patil2099!

I've just tested debugging with the `outFiles` property _without_ using esbuild and it still works as expected, so I assume VS Code gracefully handles the case of `outFiles` pointing to a directory that doesn't exist or doesn't contain the expected sourcemaps.

So I wonder if we could simply add it to the recommended `launch.json`, as opposed to saying it should be added when using esbuild. Some people might not know that they're using esbuild, so it'd be great to simplify the flow as much as possible.

(cc @erezrokah)

erezrokah (approved):
Updated the PR based on https://github.com/netlify/cli/pull/3250#issuecomment-905645556

PR #3278, opened 2021-08-31
fix(command-dev): ensure no orphaned child process on Windows

### LINKED ISSUE #3251: netlify dev leaves behind zombie processes on Windows
**Describe the bug**

`netlify dev` leaves behind zombie processes **on windows** after ctrl-c. This leaves copies of dev tools like tsc, esbuild, nextjs, etc in watch mode running after the dev session is terminated.

**To Reproduce**

https://github.com/kamsar/nfcli-sigint-zombies#readme

**Configuration**

See repro repository above for steps and reproducible scenario. Tests ran on Windows 10 19042, node 14.15.4, npm and yarn both. It seems to occur on any process spawned by netlify dev - tsc, npm-run-all, esbuild, node scripts, etc.

**Expected behavior**

`netlify-dev` should not leave behind zombie processes - and it did not used to, this started happening a month or two ago.

Xenonym:
**- Summary**

Fixes #3251.

On Windows, the `dev` command does not properly terminate the framework process on exit, and leaves them running even though `netlify-cli` has exited.

This is because `execa` defaults to `windowsHide: true`, which seems to interfere with terminating child processes: sindresorhus/execa#433.

Other projects have also encountered this issue with execa's defaults and Windows: snowpackjs/snowpack#1022, lerna/lerna#2946.

Let's set `windowsHide: false` so that the framework process will terminate properly on exit. 

**- Test plan**

Run `netlify dev` with this PR and attempt to exit on Windows (e.g. via Ctrl-C). No framework processes should be left running after exit.

🦓

### DIFF
--- a/src/commands/dev/index.js
+++ b/src/commands/dev/index.js
@@ -61,7 +61,13 @@ const startFrameworkServer = async function ({ settings, exit }) {
   log(`${NETLIFYDEVLOG} Starting Netlify Dev with ${settings.framework || 'custom config'}`)
 
   // we use reject=false to avoid rejecting synchronously when the command doesn't exist
-  const frameworkProcess = execa.command(settings.command, { preferLocal: true, reject: false, env: settings.env })
+  const frameworkProcess = execa.command(settings.command, {
+    preferLocal: true,
+    reject: false,
+    env: settings.env,
+    // windowsHide needs to be false for child process to terminate properly on Windows
+    windowsHide: false,
+  })
   frameworkProcess.stdout.pipe(stripAnsiCc.stream()).pipe(process.stdout)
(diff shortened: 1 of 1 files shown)


ehmicky:
Thanks a lot for looking into this @Xenonym.

This is a rather tricky one. Both `true` and `false` values for `windowsHide` have downsides. The two following comments are pointing to many issues discussing this in more details ([here](https://github.com/sindresorhus/execa/issues/433#issuecomment-674411152) and [there](https://github.com/sindresorhus/execa/issues/433#issuecomment-675556994)).

The `CTRL-C` issue has actually been mentioned in the following core Node.js issue (https://github.com/nodejs/node/issues/29837).

Overall, I think using `windowsHide: false` might be the lesser evil, so let's give this a go :+1:

ehmicky (approved):

PR #3316, opened 2021-09-09
fix: use original `event.path` when redirecting to Functions

ehmicky:
When `netlify dev` redirects to a Function, we keep the original request URL as a `X-Netlify-Original-Pathname` HTTP request header.

https://github.com/netlify/cli/blob/cc84d2ab4da62c8fc523b30362d243c044d71c57/src/utils/proxy.js#L250

This is used by the Functions server to set the `event.path` so it matches the URL before the redirection happened.

https://github.com/netlify/cli/blob/cc84d2ab4da62c8fc523b30362d243c044d71c57/src/lib/functions/server.js#L65

However, this does not work when the redirect is a fallback after a JWT authentication failed. In that case, we set the request URL as `/.netlify/non-existing-path` in order to serve it as a fallback. However, `/.netlify/non-existing-path` ends up being interpreted as the original URL, which is then set to `event.path`. This is incorrect.

This PR fixes this by keeping track of the original URL before `/.netlify/non-existing-path` is set.

### DIFF
--- a/src/utils/proxy.js
+++ b/src/utils/proxy.js
@@ -130,6 +130,7 @@ const serveRedirect = async function ({ req, res, proxy, match, options }) {
     return handleAddonUrl({ req, res, addonUrl: urlForAddons })
   }
 
+  const originalURL = req.url
   if (match.exceptions && match.exceptions.JWT) {
     // Some values of JWT can start with :, so, make sure to normalize them
     const expectedRoles = new Set(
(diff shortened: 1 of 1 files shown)


src/utils/proxy.js
```diff
@@ -247,7 +242,7 @@ const serveRedirect = async function ({ req, res, proxy, match, options }) {
     }
 
     if (isFunction(options.functionsPort, req.url)) {
-      req.headers['x-netlify-original-pathname'] = reqUrl.pathname
+      req.headers['x-netlify-original-pathname'] = reqToURL(req, originalURL).pathname
```
ehmicky, line 245: We need the path, but without the query parameters and hash.
One way to do this using `req.url` (which includes all of those but not the origin) is to parse it with `new URL()`. This re-uses the `new URL()` logic from above to do this.

erezrokah (approved):
🚀

PR #3346, opened 2021-09-16
refactor: replace inherited error and exit utils in command.addons

tinfoil-knight:
**- Summary**

This PR focuses specifically on the `src/commands/addons` folder.

**- Test plan**

`npx ava --verbose --serial` ran without any failing tests.


ehmicky (approved):
This is looking good @tinfoil-knight. Thanks a lot for this contribution!

PR #3349, opened 2021-09-17
refactor: replace inherited log utils in src/command/(dev|sites|open)

tinfoil-knight:
**- Summary**

This PR focuses specifically on the `dev`,`open`,`sites` folder in `src/commands`.
Utilities outside this folder have been changed as less as possible to avoid growing the diff size in a single PR.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Thanks @tinfoil-knight!

PR #3347, opened 2021-09-16
refactor: replace inherited logging, exit utils in commands.functions…

tinfoil-knight:
… and adjacent utils

**- Summary**

#3028  was too large for review and is being broken down.

This PR is a follow-up on #3247 and uses the separate logging and process control utilities created in it to remove the inherited ones.

This PR focuses specifically on the src/commands/functions folder. Utilities outside this folder have been changed as less as possible to avoid growing the diff size in a single PR.

**- Test plan**

`npx ava --verbose --serial` ran without any failing tests.

PR #3350, opened 2021-09-17
refactor: replace catch(_) with catch in try-catch blocks

tinfoil-knight:
**- Summary**

While reviewing #3347, @ehmicky suggested to replace `catch(_)` with `catch` since the `_` isn't really needed. 
This PR implements that same suggestion for the entire repository.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Nice!

PR #3366, opened 2021-09-20
refactor: replace inherited log, exit utils in src/commands.(status|api|deploy)

tinfoil-knight:
**- Summary**

This PR focuses specifically on the `status`, `api` , `deploy` commands in `src`.
Utilities outside this folder have been changed as less as possible to avoid growing the diff size in a single PR.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Nice! :tada:

PR #3368, opened 2021-09-21
feat: error 400 on invalid function names

### LINKED ISSUE #2494: Dev doesn't break on invalid function name
**Describe the bug**

Functions names are required to consist of alphanumerics characters, hyphen and underscores only. If deploying (`netlify deploy`) a function that contains a character other than these, deploy will fail with a `422 JSONHTTPError`. Running locally with `netlify dev` permits invalid filenames.

**To Reproduce**

1. Create a new function with a period in it (e.g. `hello.world.js`)
2. Run `netlify dev`
3. Check everything working
4. Run `netlify deploy` 
5. Will fail with `Warning: JSONHTTPError: Invalid function names.`

```bash
System:
  OS: macOS 11.4
  CPU: (12) x64 Intel(R) Core(TM) i5-10600 CPU @ 3.30GHz
  Memory: 579.63 MB / 16.00 GB
  Shell: 5.8 - /bin/zsh
Binaries:
  Node: 16.1.0 - /usr/local/bin/node
  npm: 7.11.2 - /usr/local/bin/npm
npmGlobalPackages:
  netlify-cli: 3.31.3
```

**Expected behavior**

That as closely as possible `netlify dev` would replicate what is expected when running `netlify deploy` (i.e. would fail and return an error stating invalid function name.)

**CLI Output**

```
Deploying to main site URL...
✔ Finished hashing 8 files and 2 functions
⠋ CDN diffing files... ›   Warning: JSONHTTPError: Invalid function names. Name should consist of only alphanumeric characters, hyphen & underscores 422
 ›   Warning: 
 ›   {
 ›     "name": "JSONHTTPError",
 ›     "status": 422,
 ›     "json": {
 ›       "code": 422,
 ›       "message": "Invalid function names. Name should consist of only alphanumeric characters, hyphen & underscores"
 ›     }
 ›   }
 ›
⠙ CDN diffing files...    JSONHTTPError: Unprocessable Entity
```

**Additional context**

I have not seen any mention of function naming conventions in the [Netlify Doc / Functions](https://docs.netlify.com/functions/overview/ )

netlify-team-account-1:
closes #2494

**- Summary**

This PR makes `ntl dev` behave like Bitballoon when it comes to function names. It will throw a 400 response whenever an invalid function name (valid being alphanumeric + hyphen + underscores) is used.

**- Test plan**

Unit test was added.


src/lib/functions/server.js
```diff
@@ -38,13 +38,22 @@ const buildClientContext = function (headers) {
...
+const isValidNetlifyFunctionName = (name) => /^[\w-]+$/.test(name)
+
 const createHandler = function ({ functionsRegistry }) {
   return async function handler(request, response) {
     // handle proxies without path re-writes (http-servr)
     const cleanPath = request.path.replace(/^\/.netlify\/functions/, '')
     const functionName = cleanPath.split('/').find(Boolean)
     const func = functionsRegistry.get(functionName)
 
+    if (!isValidNetlifyFunctionName(functionName)) {
```
netlify-team-account-1, line 51: There's two checks we make here:

1. Does this function exist? (is there a corresponding file?)
2. Is the function name valid? (regex test)

In which order should we perform them? Should we error for invalid names, even if the function doesn't exist? Or should we only check the name if there's a corresponding file? Is one preferable DX-wise?

netlify-team-account-1:
tests are green 🥳 @eduardoboucas plz review

src/lib/functions/server.js
```diff
@@ -37,13 +37,23 @@ const buildClientContext = function (headers) {
   }
 }
 
+// same as https://github.com/netlify/bitballoon/blob/fbd7881e6c8e8c48e7a0145da4ee26090c794108/app/models/deploy.rb#L482
+// eslint-disable-next-line unicorn/better-regex
```
eduardoboucas, line 41: Why do we need to disable this rule?

src/lib/functions/server.js
```diff
@@ -37,13 +37,23 @@ const buildClientContext = function (headers) {
...
+const isValidNetlifyFunctionName = (name) => /^[A-Za-z0-9_-]+$/.test(name)
```
eduardoboucas, line 42: We have a class that represents a function (see https://github.com/netlify/cli/blob/main/src/lib/functions/netlify-function.js). Rather than having the validation logic in the server, what do you think about moving it to the `NetlifyFunction` class and have a `isValidName` instance method?

src/lib/functions/server.js
eduardoboucas: I think we should check for the existence of the file first, and only then check whether it's a valid function name. If you expect a function to exist and it doesn't, I feel that's the problem you need to address first, regardless of whether you're choosing a valid name for it.

src/lib/functions/server.js
netlify-team-account-1: b/c I'd like to keep the syntax exactly as in BitBalloon, to make it obvious that it's the same.

src/lib/functions/server.js
netlify-team-account-1: addressed in 80749db

src/lib/functions/server.js
netlify-team-account-1: addressed in 80749db

eduardoboucas (approved):
LGTM!

PR #3369, opened 2021-09-21
refactor: replace gotCatch404 with "throwHttpErrors: false"

netlify-team-account-1:
**- Summary**
While working on #3368, I stumbled over the `gotCatch404` helper function, which works around Got's behaviour to throw on non-successful responses. The more idiomatic way to disable this behaviour is the built-in `throwHttpErrors: false` option. This has the benefit that tests now won't fail because of a thrown error, but instead because of failed assertions (wrong status code!), resulting in test failure messages that are easier to trace.

**- Test plan**


eduardoboucas (approved):
Nice! 🎉

PR #3373, opened 2021-09-21
refactor: replace inherited log,exit utils in src/commands

tinfoil-knight:
**- Summary**

This PR focuses specifically on the `init`, `login`, `logout`, `switch`, `watch` commands in `src`.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
:heart:

PR #3381, opened 2021-09-22
refactor: replace log,exit in utils of dev,functions,link command

tinfoil-knight:
**- Summary**

This PR focuses on utilities related to the dev, functions and link commands. Changes are a bit sparse and not contained in a specific command folder like the previous PRs.

This is probably the 2nd last PR in this series and then we should be done.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Thanks @tinfoil-knight!

PR #3388, opened 2021-09-23
chore: remove deprecated bodyParser in favor of express methods

tinfoil-knight:
**- Summary**

The [body-parser](https://www.npmjs.com/package/body-parser) package is not needed post `express v4.17.0` as `express` already has the inbuilt methods as specified in the Methods section here: https://expressjs.com/en/api.html#express.

It has been replaced in the main `cli` package and in the function templates.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Good catch :+1:

PR #3392, opened 2021-09-23
refactor: split core function of getAgent for better testing

tinfoil-knight:
**- Summary**

While working on #3382, @ehmicky suggested that we could break down the `getAgent` function in `src/lib/http-agent.js` into 2 different functions (one that only returns successful or error responses and another which handles logging and process control) to allow better testing.

**- Test plan**

`npx ava --verbose --serial`

Note: Tests for the http-proxy now use the core function `tryGetAgent` instead of the earlier `getAgent` function which has now been turned into a wrapper.

### DIFF
--- a/src/lib/http-agent.js
+++ b/src/lib/http-agent.js
@@ -30,7 +30,7 @@ const AGENT_PORT_TIMEOUT = 50
 
 const tryGetAgent = async ({ httpProxy, certificateFile }) => {
   if (!httpProxy) {
-    return
+    return {}
   }
 
   let proxyUrl
@@ -88,20 +88,15 @@ const tryGetAgent = async ({ httpProxy, certificateFile }) => {
 }
 
 const getAgent = async ({ httpProxy, certificateFile }) => {
-  const result = await tryGetAgent({ httpProxy, certificateFile })
-  if (!result) {
-    return
-  }
-  if (result.error) {
-    log(NETLIFYDEVERR, result.error, result.message || '')
+  const { error, warning, agent, message } = await tryGetAgent({ httpProxy, certificateFile })
+  if (error) {
+    log(NETLIFYDEVERR, error, message || '')
     exit(1)
   }
-  if (result.warning) {
-    log(NETLIFYDEVWARN, result.warning, result.message || '')
+  if (warning) {
+    log(NETLIFYDEVWARN, warning, message || '')
   }
-  if (result.agent) {
-    return result.agent
-  }
+  return agent
 }
 
 module.exports = { getAgent, tryGetAgent }
--- a/src/lib/http-agent.js
+++ b/src/lib/http-agent.js
@@ -3,7 +3,7 @@ const { URL } = require('url')
 const { HttpsProxyAgent } = require('https-proxy-agent')
 const waitPort = require('wait-port')
 
-const { log } = require('../utils/command-helpers')
+const { log, exit } = require('../utils/command-helpers')
 const { NETLIFYDEVERR, NETLIFYDEVWARN } = require('../utils/logo')
 
 const fs = require('./fs')
@@ -28,7 +28,7 @@ const DEFAULT_HTTPS_PORT = 443
 // 50 seconds
 const AGENT_PORT_TIMEOUT = 50
 
-const getAgent = async ({ httpProxy, certificateFile, exit }) => {
+const tryGetAgent = async ({ httpProxy, certificateFile }) => {
   if (!httpProxy) {
     return
   }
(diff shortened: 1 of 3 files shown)


src/lib/http-agent.js
```diff
@@ -85,7 +83,25 @@ const getAgent = async ({ httpProxy, certificateFile, exit }) => {
   }
 
   const agent = new HttpsProxyAgentWithCA(opts)
-  return agent
+  response = { ...response, agent }
+  return response
+}
+
+const getAgent = async ({ httpProxy, certificateFile }) => {
+  const result = await tryGetAgent({ httpProxy, certificateFile })
```
ehmicky, line 91: [pebble] By returning an empty object instead of `undefined`, we could use destructuring:

```js
  const { error, warning, message, agent } = await tryGetAgent({ httpProxy, certificateFile })
```

src/lib/http-agent.js
```diff
@@ -28,23 +28,21 @@ const DEFAULT_HTTPS_PORT = 443
...
   }
 
   let proxyUrl
   try {
     proxyUrl = new URL(httpProxy)
-  } catch (error) {
-    log(NETLIFYDEVERR, `${httpProxy} is not a valid URL`)
-    exit(1)
+  } catch {
+    return { error: `${httpProxy} is not a valid URL` }
```
ehmicky, line 40: The `error` instance might be useful here for debugging, i.e. we might want to return it?

```js
  } catch (error) {
    return { error: `${httpProxy} is not a valid URL`, message: error.message }
```

src/lib/http-agent.js
```diff
@@ -85,7 +83,25 @@ const getAgent = async ({ httpProxy, certificateFile, exit }) => {
...
+    return
+  }
+  if (result.error) {
+    log(NETLIFYDEVERR, result.error, result.message || '')
+    exit(1)
+  }
+  if (result.warning) {
+    log(NETLIFYDEVWARN, result.warning, result.message || '')
+  }
+  if (result.agent) {
```
ehmicky, line 102: [pebble] Since `result.agent` is either `undefined` or an `Agent` instance, this line might be removed, i.e. we might be able to always `result.agent`, regardless of whether it is defined or not.

src/lib/http-agent.test.js
```diff
@@ -3,36 +3,31 @@ const http = require('http')
 const test = require('ava')
 const { createProxyServer } = require('http-proxy')
 const { HttpsProxyAgent } = require('https-proxy-agent')
-const sinon = require('sinon')
```
ehmicky, line 6: :tada:

src/lib/http-agent.test.js
```diff
@@ -3,36 +3,31 @@ const http = require('http')
...
 
-test(`should exit with error on invalid url`, async (t) => {
+test(`should return error on invalid url`, async (t) => {
   const httpProxy = 'invalid_url'
-  const exit = sinon.stub()
-  exit.withArgs(1).throws('error')
-
-  await t.throwsAsync(getAgent({ httpProxy, exit }))
+  const result = await tryGetAgent({ httpProxy })
+  t.truthy(result.error)
```
ehmicky, line 16: Those tests are so much clearer without mocking! :tada:

ehmicky:
This is looking great @tinfoil-knight!
Just a few minor comments.

src/lib/http-agent.js
tinfoil-knight: I'm not sure about returning the entire error object in response considering that most errors occurring in this function are known before-hand.

src/lib/http-agent.js
tinfoil-knight: Yeah.

src/lib/http-agent.js
tinfoil-knight: Yep. Changed this post destructuring.

tinfoil-knight:
A few commands always time out on the windows CI. (This has been happening across multiple unrelated PRs.)

Tests:
- `tests » serving-functions » Adds a new TypeScript function when a function file is created - esbuild`
- `tests » command.lm » netlify lm:info`

src/lib/http-agent.js
ehmicky: This would be returning only the error message string, not the entire error object.
The purpose would be to show why the `httpProxy` is not a valid URL: is the scheme missing? is there some other syntax error? 
However, this is not a blocker, so please feel free to ignore :+1:

ehmicky:
About the CI: I have noticed too, those tests seem to be flaky, I am not quite sure why :(

ehmicky (approved):
:tada:

src/lib/http-agent.js
tinfoil-knight: Oh! I misunderstood. Returning the error message definitely makes sense.

I think I got confused since a few places in this file already had the response with the error message and thought that you wanted to add the stack too.

tinfoil-knight: Will add it in another PR soon since this one is merged.

PR #3396, opened 2021-09-24
refactor: split core function of loadDotEnvFiles for better testing

tinfoil-knight:
**- Summary**

While working on #3382, @ehmicky suggested that we could break down the `loadDotEnvFiles` function in `src/utils/dot-env.js` into 2 different functions (one that only returns successful or error responses and another which handles logging) to allow better testing.

**- Test plan**

`npx ava --verbose --serial`


ehmicky (approved):
Thanks!

PR #3445, opened 2021-10-04
feat: support functions within type=module packages (Closes #3394)

netlify-team-account-1:
**- Summary**

`lambda-local`, the library we use to emulate Lambda, currently errors for functions that are inside of a directory with `type=module` in their package.json.

**- Test plan**

- added a reproduction test

**- A picture of a cute animal (not mandatory but encouraged)**

PR #3483, opened 2021-10-12
chore: add eslint-plugin-sort-destructure-keys

netlify-team-account-1:
**- Summary**

Idea from here: https://github.com/netlify/cli/pull/3445#discussion_r727003403

This PR introduces the `eslint-plugin-sort-destructure-keys` plugin, which forces all destructurings (including function parameter destructuring) to be sorted alphabetically.

**- Test plan**


eduardoboucas:
I feel that I should add some context here, since I wrote the initial comment.

Sorting object keys alphabetically is not a convention we've adopted. Personally, I think it makes code easier to read, especially when you have a long list of properties (which is often the case in codebases like Netlify Build and zip-it-and-ship-it). We've had [informal discussions](https://netlify.slack.com/archives/CN8S7NB29/p1628844711123500) on the subject and the majority of people seem to feel the same way.

When I write new code, I adopt that convention, and I tend to ask people to keep that order when they're adding a new property to an object that's already sorted. That was the example [here](https://github.com/netlify/cli/pull/3445/files#diff-41e7ff2f2dc80310c60a5be107c9b017bd1d3bbf9b986ed524de02d71d76a4f1R23).

If we're introducing this as a linting rule, it feels like we should do it in https://github.com/netlify/eslint-config-node, but we should definitely involve everyone from the guild to make sure they're comfortable with it.

(cc @netlify/netlify-dev)

eduardoboucas:
> I personally prefer grouping properties by their meaning or relatedness instead.

My issue with that is that "relatedness" is a subjective criteria, which means not only that it's unenforceable using automated tools, but also it's possible that the sorting will help whoever wrote the code and nobody else.

netlify-team-account-1:
@eduardoboucas awaiting your review, then we can merge this :)

netlify-team-account-1:
> would we add it to CLI only at first to see how it goes and then make it part of https://github.com/netlify/eslint-config-node? 

That sounds like a good action plan.

PR #3490, opened 2021-10-13
chore: remove read pkg up

netlify-team-account-1:
**- Summary**

as discussed in https://github.com/netlify/cli/pull/3445#discussion_r727002278, recreating our usage of `read-pkg-up` isn't very hard, especially considering we don't need any of the normalisation steps that it does. This PR removes the package. While this doesn't have any immediately-tangible benefits (`read-pkg-up` still is in our transitive deps, as can be seen in the diff of `npm-shrinkwrap`), it's consistent with our goal of keeping dependency number lean, to reduce install time.

(depends on #3445 to be merged first)

**- Test plan**

test suite still passes.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -106,7 +106,6 @@
         "prettyjson": "^1.2.1",
         "pump": "^3.0.0",
         "raw-body": "^2.4.1",
-        "read-pkg-up": "^7.0.1",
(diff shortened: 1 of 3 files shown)


netlify-team-account-1:
here's another argument to the debate of "to depend or not to depend, that is the question": https://github.com/netlify/cli/pull/3491

PR #3493, opened 2021-10-13
chore: remove pause() calls

netlify-team-account-1:
None of the `pause()` calls in serving-functions.test.js do something in the context of their individual test. This commit removes them. Locally, all tests ran through fine without them - let's see if that still holds true on CI :)

**- Summary**

**- Test plan**


erezrokah:
Going to close this as stale. Please re-open if I got this wrong

PR #2895, opened 2021-07-09
fix(docs): clarify live site creation from tests closes #2541

tinfoil-knight:
**- Summary**

resolves https://github.com/netlify/cli/issues/2541

This PR clarifies that the deploy test suite use a live site to run integration tests which require an active Netlify account. 

To make things simpler to configure, the `IS_FORK` environment variable (which disables these tests) has been renamed to `NETLIFY_TEST_DISABLE_LIVE`. 

**- Test plan**

`npx ava -v -s` for https://github.com/netlify/cli/commit/a134b72409550aec4676a992d47dcca668322eb9 on Node.js 10

PR #3563, opened 2021-11-01
fix: typo regarding tests in contributing doc

tinfoil-knight:

#### Summary

A weird typo was introduced in #2895. Spotted it while reading the Contributing Docs today. Just fixing that.


erezrokah (approved):
Thanks for the follow up @tinfoil-knight, nice catch!

PR #3574, opened 2021-11-03
fixes the netlify lm:install tests for different shells (like zsh)

### LINKED ISSUE #3573: The CLI tests are not running when the source path is longer in `netlfiy lm:install`
**Describe the bug**

**To Reproduce**

Steps to reproduce the behaviour:

have a longer source path on the `netlify lm:install` command:

CleanShot 2021-11-03 at 10 18 40

1. Run `https://github.com/netlify/cli.git`
2. Navigate to the cloned repository
3. Run `node  ./node_modules/.bin/ava tests/command.lm.test.js`
4. Test is failing with the following error:

CleanShot 2021-11-03 at 10 19 36

The issue seems located inside `tests/command.lm.test.js:L74` where the regex is getting the source path out of the `lm:install` command.

```javascript
    // tests/command.lm.test.js:L74
    const [source] = cliResponse.match(/source.+inc/)
```

Due to the fact that the print width or theme (with zsh) is displaying the border as `||` the box that can be seen in the first screenshot is deciding the output and the `cliResponse` for the `const cliResponse = await callCli(['lm:install'], t.context.execOptions)` has the following output:

CleanShot 2021-11-03 at 10 55 04

therefore the regex is matching the pipe as well and passing a wrong source to the next command inside the test

lukasholzer:
#### Summary

Fixes #3573 

When executing the test with a `zsh` shell the formatting of the box was off, so that the regex matched a wrong string. Therefore the further execution of the test was not working. 

Updated the regular expression to match exactly the path regardless of what is between the word source and the absolute path.

      passes our tests.

image

### DIFF
--- a/tests/command.lm.test.js
+++ b/tests/command.lm.test.js
@@ -71,8 +71,16 @@ test.serial('netlify lm:install', async (t) => {
     t.true(stdout.startsWith('git-credential-netlify'))
   } else {
     t.true(cliResponse.includes('Run this command to use Netlify Large Media in your current shell'))
-    const [source] = cliResponse.match(/source.+inc/)
-    const { stdout } = await execa.command(`${source} && git-credential-netlify version`, {
+    // The source path is always an absolute path so we can match for starting with `/`.
+    // The reasoning behind this regular expression is, that on different shells the border of the box inside the command output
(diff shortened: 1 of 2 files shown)


ehmicky (approved):
:tada:

PR #3584, opened 2021-11-04
fix(dev): fixes an issue where errors in the `dev` command where not shown

### LINKED ISSUE #3572: Let netlify dev output an error if you are not logged in with the right account

I switch between Netlify accounts in my CLI and today it took me a long time to realize that I'm on the wrong account. `netlify dev` would just stop after injecting environment vars and it didn't tell me why.

the CLI should show an error/warning if you're trying to run `netlify dev` on a site that is linked to another account that you are currently logged into.

You have to know/guess what is the problem.

**Additional context**

This is how it looks like when it stops without any relevant info.

Screenshot 2021-11-03 at 09 38 00

No.

lukasholzer:

#### Summary

Fixes #3572

This caused the CLI to exit silently without any further information for the user about the problem.

I guess this was a typo on importing the wrong function as the signature of the `exit` function only can get an `exitCode: number` and the error method accepts a `string`

CleanShot 2021-11-04 at 10 52 26

The function was used with a string so I guess it was the wrong import.
*TypeScript would have prevented this issue ;)*

```javascript
const getAddons = async ({ api, site }) => {
  try {
    const addons = await api.listServiceInstancesForSite({ siteId: site.id })
    return addons
  } catch (error) {
    failAndExit(`Failed retrieving addons for site ${chalk.yellow(site.id)}: ${error.message}. ${ERROR_CALL_TO_ACTION}`)
  }
}
```

image


lukasholzer:
When I switch now to an account which is not connected with my site I get the following error:

CleanShot 2021-11-04 at 10 56 43@2x

erezrokah (approved):
Great catch @lukasholzer!

We did some [refactoring](https://github.com/netlify/cli/pull/3381/files#diff-42c491dfbe209d7493ec82cd5d031be64f17cf6e622201768d29b3539194d9e1L184) around this code recently and we missed this.

PR #3592, opened 2021-11-04
fix: fix a regression where the creation of a new github token failed inside the init command

### LINKED ISSUE #3588: Github authentication for `netlify init` is not working due to a regression.
**Describe the bug**

**To Reproduce**

1. remove the global `~/Library/Preferences/netlify/config.json`
2. run `netlify init` in a new project
3. Select Authentication via netlify app
4. Use Github as the authentication method.
5. It will fail with `401 Bad Credentials`

Regression was introduced with this change: https://github.com/netlify/cli/pull/3569/files#diff-f64dd88efec835a92fdd4fd9b786ef5faf32c0c732656935ec33a443efb2c912R39

The problem is that the return type of the `getGitHubToken` changed from `Promise<Record<string, string>>` to `Promsie<URLSearchParams>` which cannot be set inside the globalConfig.

Just to note (TypeScript would have prevented this)

**Expected behavior**

Authentication via Github should work

**CLI Output**

```
Adding deploy key to repository...
 ›   Error: Failed retrieving GitHub repository information with error: Bad credentials.
```

lukasholzer:

#### Summary

Fixes #3588

      passes our tests.


src/utils/gh-auth.js
```diff
@@ -31,14 +31,26 @@ const promptForAuthMethod = async () => {
...
+ */
 const authWithNetlify = async () => {
   const port = await getPort({ port: SERVER_PORT })
   const { promise: deferredPromise, reject: deferredReject, resolve: deferredResolve } = createDeferred()
 
   const server = http.createServer(function onRequest(req, res) {
     const parameters = new URLSearchParams(req.url.slice(req.url.indexOf('?') + 1))
     if (parameters.get('token')) {
-      deferredResolve(parameters)
+      const recordOfParams = [...parameters].reduce((prev, [key, value]) => ({ ...prev, [key]: value }), {})
```
tinfoil-knight, line 52: @lukasholzer While going through the fix for this bug yesterday, I found out that the `cli` repo already has a polyfill for `fromEntries` (which is what we're trying to do here).

You can use it like this:
`const fromEntries = require('@ungap/from-entries')`

`const recordOfParams = fromEntries([...parameters])`

lukasholzer: thx for pointing out :)

src/lib/functions/form-submissions-handler.js
```diff
@@ -38,7 +39,7 @@ const createFormSubmissionHandler = function ({ functionsRegistry, siteUrl }) {
         encoding: ct.parameters.charset,
       })
 
-      fields = getObjectFromParams(new URLSearchParams(bodyData.toString()))
+      fields = fromEntries([...new URLSearchParams(bodyData.toString())])
```
ehmicky, line 42:
```suggestion
      fields = fromEntries(new URLSearchParams(bodyData.toString()))
```

[sand] Tiny simplification since `URLSearchParams()` is iterable and `Object.fromEntries()` can use iterables as argument.

src/utils/gh-auth.js
```diff
@@ -31,14 +32,25 @@ const promptForAuthMethod = async () => {
...
+ */
 const authWithNetlify = async () => {
   const port = await getPort({ port: SERVER_PORT })
   const { promise: deferredPromise, reject: deferredReject, resolve: deferredResolve } = createDeferred()
 
   const server = http.createServer(function onRequest(req, res) {
     const parameters = new URLSearchParams(req.url.slice(req.url.indexOf('?') + 1))
     if (parameters.get('token')) {
-      deferredResolve(parameters)
+      deferredResolve(fromEntries([...parameters]))
```
ehmicky, line 53:
```suggestion
      deferredResolve(fromEntries(parameters))
```

[sand] See above

src/utils/open-browser.js
```diff
@@ -33,4 +33,4 @@ const openBrowser = async function ({ silentBrowserNoneError, url }) {
   }
 }
 
-module.exports = openBrowser
+module.exports = { openBrowser }
```
ehmicky, line 36: The following consumers should be updated with the new signature:

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/commands/addons/auth.js#L4

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/commands/deploy.js#L24

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/commands/dev/index.js#L20

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/commands/open/admin.js#L3

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/commands/open/site.js#L3

https://github.com/netlify/cli/blob/9165fe3e69b69dab015fdd930feed8ebec89181a/src/utils/command.js#L14

https://github.com/netlify/cli/pull/3592/files#diff-9f1c90587f2f1996609d5d02f01ef29eba18d163b06171d7b11a651376e40461R7

ehmicky (commented):
Thanks @lukasholzer!

The automated tests seem to be failing, which might be related to.

src/utils/open-browser.js
lukasholzer: Thx have overseen that 👍

ehmicky (approved):

PR #3595, opened 2021-11-05
feat: add a prompt to modify users .gitignore (closes netlify#2038)

### LINKED ISSUE #2038: feat: add a log message when we modify users `.gitignore`

See https://github.com/netlify/cli/issues/2029#issuecomment-803581401

Modifying users files without letting them know is not very nice. We should at least log that we're changing `.gitignore`

When the CLI modifies `.gitignore` we should print a log message reflecting the change to the user

We could also prompt the user to approve/deny the change

Yes

ngonsoto:

#### Summary

Fixes #2038

Edits to .gitignore file are not shown to the user. This PR adds a prompt if the .netlify folder is not found in the .gitignore file and asks the user if it wants to add it.

Captura de pantalla 2021-11-05 105606

---

      passes our tests.

8741538e4321434fe9f765f0769c78892


lukasholzer (changes requested):
@ngonsoto Thank you very much for your PR and taking the time to enhance our CLI! 
I want to apologize upfront for being so late with my review 😬 

From a consumer perspective there is no reason why the `.netlify` folder should not be added to the `.gitignore`, or at least we don't want to promote it. As all our build outputs are there and they should not be committed.

So the user input here would be in 99% the cases a "yes" – I think we can drop it for a better user experience and instead providing a meaningful log message that we adapted the users `.gitignore` which is stated as solution inside the issue:

> Describe the solution you'd like
> 
> When the CLI modifies .gitignore we should print a log message reflecting the change to the user

Furthermore a prompt would cause issues on the link command in the CI so we would need to adapt it with a flag to pass without asking. The simpler and better approach in my humble opinion is to informing the user.

Therefore can you please update the PR with a log message?

ngonsoto:
Hi @lukasholzer, no problem. I'll be updating the PR with the requested changes.

src/utils/gitignore.js
```diff
@@ -30,6 +32,8 @@ const ensureNetlifyIgnore = async function (dir) {
   }
   /* Not ignoring .netlify folder. Add to .gitignore */
   if (!ignorePatterns || !ignorePatterns.patterns.some((pattern) => /(^|\/|\\)\.netlify($|\/|\\)/.test(pattern))) {
+    log()
+    log('Adding local .netlify folder to .gitignore file...')
```
lukasholzer, line 36:
```suggestion
    log('Adding local .netlify folder to .gitignore file.')
```

lukasholzer (approved):
Awesome! Thank you very much for going over and adapting the PR 🌟

PR #3615, opened 2021-11-09
fix: show Background Function plan warning only if site is connected

eduardoboucas:
#### Summary

The `netlify dev` command is showing a warning message when a Background Function is used and the `capabilities` object shows that the current plan doesn't support that feature. However, it's possible that the `capabilities` object is not populated due to the user being signed out or running the command using the `--offline` flag.

This PR addresses that by passing an `isConnected` flag to the functions registry, which infers whether the site is connected by looking at the `siteUrl` property. If that property is falsy, CLI won't show the warning because it can't know whether the user has a plan with access to the feature or not.

Closes https://github.com/netlify/pod-serverless/issues/31.

### DIFF
--- a/src/lib/functions/registry.js
+++ b/src/lib/functions/registry.js
@@ -12,9 +12,10 @@ const runtimes = require('./runtimes')
 const { watchDebounced } = require('./watcher')
 
 class FunctionsRegistry {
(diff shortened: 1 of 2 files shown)


netlify-team-account-1 (approved):
:rocket:

PR #3620, opened 2021-11-11
fix(dev): fix wrong rawUrl in the dev command on redirect

### LINKED ISSUE #3609: When invoking function, `event.rawUrl` value is different from production deploy
**Describe the bug**

When using a function via a 200 rewrite, the value of `event.rawUrl` is the function URL. In production, the value is the original request URL.

**To Reproduce**

- Create a function that logs `event.rawUrl`
- Add a rewrite that points to that function URL
- Run with `ntl dev`, observing URL value
- Deploy to production and observe URL value

**Expected behavior**

The CLI and production values should match. I would prefer the CLI's value (i.e. the function target), but changing prod would be a breaking change so it's the CLI that needs to change (and ideally both should provide another header or field that includes the target URL)

lukasholzer:

#### Summary

Fixes #3609

Fixes an issue where the `rawUrl` was the original URL once a redirect was created to the function

---

      passes our tests.

image


mraerino (approved):
LGTM

mraerino (dismissed):

src/lib/functions/server.test.js
```diff
@@ -0,0 +1,80 @@
...
+})
+
+test.after((t) => {
+  t.context.zisiStub.restore()
+})
+
+test('should get the url as the `rawUrl` inside the function', async (t) => {
+  await request(app)
+    .get('/hello')
+    // .set('x-netlify-original-pathname', 'orig')
```
eduardoboucas, line 53: Did you mean to leave this here?

eduardoboucas (approved):
🚀

src/lib/functions/server.test.js
lukasholzer: 😬 oh no this one slipped through :D

lukasholzer: already updated it ;)

eduardoboucas: Awesome, thanks!

eduardoboucas (approved):

PR #3631, opened 2021-11-12
fix: resemble Netlify Production logic for base64 encoding more closely

### LINKED ISSUE #3315: Functions event.isbase64Encoded discrepancy in CLI vs Production
**Describe the bug:**

The production server encodes the function's `event.body` to `base64` except for a few content-types as mentioned here: https://answers.netlify.com/t/changed-behavior-in-function-body-encoding/19000.

When submitting a file using a (non-Netlify) form to a Netlify Function, we have to use `multipart/form-data` content-type which is encoded according to the above post. Thus, we need to decode it as:

```js
Buffer.from(event.body, 'base64').toString('utf8')
```

However, when doing the same in CLI, the body is not encoded and available easily as `event.body`.

It can easily be worked around like:

```js
if (event.isbase64Encoded) {
  Buffer.from(event.body, 'base64').toString('utf8')
} else {
  event.body
}
```

But, it leads to confusion.

**To reproduce:**

1. Clone this commit: https://github.com/Hrishikesh-K/formFile/tree/1f7d002c18eff4830614d6d405d84cee4dd106e8.
2. Navigate to the cloned repository.
3. Run `netlify dev`.
4. Try to submit the form with any random file and log `event.body`. Then, log the same in production environment.

**Configuration:**

`netlify.toml`:

```toml
[build]
  command = "echo Hi!"
  publish = "dist"

[functions]
  directory = "functions"
  node_bundler = "esbuild"
```

System info:

```bash
  System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 746.86 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.9.0 - /usr/local/bin/node
    npm: 7.21.1 - /usr/local/bin/npm
```

**Expected behaviour:**

The CLI should encode `event.body` to `base64`.

**Additional context:**

Slack discussion (INTERNAL ONLY): https://netlify.slack.com/archives/C023PC3D08J/p1631194079008400

netlify-team-account-1:
#### Summary

Fixes #3315

Updates the CLI's decision logic for encoding incoming requests as base64 or not, to better match what Netlify production infrastructure is doing.

---

      passes our tests.


src/lib/functions/utils.js
```diff
@@ -21,8 +19,40 @@ const detectAwsSdkError = ({ error }) => {
 
 const formatLambdaError = (err) => chalk.red(`${err.errorType}: ${err.errorMessage}`)
 
+const exceptionsList = new Set([
```
eduardoboucas, line 22: I'm assuming this list comes from production code? If so, do you mind adding a comment to that effect?, just so people know where to look if they're working on this?

src/lib/functions/utils.js
```diff
@@ -21,8 +19,40 @@ const detectAwsSdkError = ({ error }) => {
...
+ */
 const shouldBase64Encode = function (contentType) {
-  return Boolean(contentType) && BASE_64_MIME_REGEXP.test(contentType)
+  if (!contentType) {
+    return true
+  }
+
+  contentType = contentType.toLowerCase()
+
+  if (contentType.startsWith('text/')) {
```
eduardoboucas, line 43: Any reason why we can't move these to a single condition?

src/lib/functions/utils.js
netlify-team-account-1: Sure thing! It stems from https://github.com/netlify/proxy/blob/main/pkg/functions/request.go#L105, which I put into one of the commit's bodies - but a comment is a good idea, too.

```suggestion
// should be equivalent to https://github.com/netlify/proxy/blob/main/pkg/functions/request.go#L105
const exceptionsList = new Set([
```

src/lib/functions/utils.js
netlify-team-account-1: I like how each of the conditions is its own case, à la "if it's a text, don't encode; if it's another content-type encoded as json or xml, don't encode; here's a set of exceptions we don't encode". Moving them into a single condition would make that grouping harder to see. Can you elaborate on why we should move them into a single condition?

eduardoboucas: Personally, I think having 3 termination points when you can have 1 makes it harder to trace and debug. I also don't see how having a single condition makes the grouping harder to see. It's not a big deal though, and I respect the personal preference if you think it adds more clarity.

eduardoboucas (approved):

Xenonym:
Turns out I inadvertently tried to fix this in #3564 😅, this PR should supersede that I think.

PR #3632, opened 2021-11-12
fix(init): fixes an issue where the init command exited with bad credentials

### LINKED ISSUE #3488: Github credentials have expired. CLI throwing error but not prompting for updated credentials.
**Describe the bug**

When I try and init a new project using the netlify-cli, everything works as planned until the cli attempts to link the netlify project to the associated git repository. At that point it throws the following error:

**Error: Failed retrieving GitHub repository information with error: Bad credentials**

I've looked through the CLI docs but I'm not sure how to reset my Github credentials.

**To Reproduce**

For me, this occurs during the `netlify init` process with a new netlify project and an existing (linked) github repository. Again, the netlify project is created but the `init` process ends in error and I must visit the netlify.com interface to link the Github project.

**Expected behavior**

I suppose I expected the CLI to prompt me to update my Github credentials or for the documentation to explain how I can do this.

**CLI Output**

```
Error: Failed retrieving GitHub repository information with error: Bad credentials
```

lukasholzer:

#### Summary

Fixes #3488

Fixes an issue where the init command failed with bad credentials error if the github token was not valid:

CleanShot 2021-11-12 at 14 56 11

With the newly introduced changes, the validity check of the token does not only check if there is any value it checks it by requesting the GitHub API.

If GitHub returns an error we generate now new credentials like it was never connected:

CleanShot 2021-11-12 at 14 57 32

---

      passes our tests.


src/utils/gh-auth.js
```diff
@@ -108,15 +113,19 @@ const authWithToken = async () => {
...
+ * Get a github token
+ * @returns {Promise<Token>} Returns a Promise with a token object
+ */
 const getGitHubToken = async () => {
   log('')
 
   const withNetlify = await promptForAuthMethod()
   if (withNetlify) {
-    return await authWithNetlify()
+    return authWithNetlify()
```
erezrokah, line 125: See [here](https://github.com/goldbergyoni/nodebestpractices#-212-always-await-promises-before-returning-to-avoid-a-partial-stacktrace) on the reason we use `return await`

src/utils/init/config-github.js
```diff
@@ -6,34 +6,55 @@ const { getGitHubToken: ghauth } = require('../gh-auth')
...
-  if (isValidToken(githubToken)) {
-    return githubToken.token
+  if (githubToken && githubToken.user && githubToken.token) {
+    try {
+      const octokit = getGitHubClient(githubToken.token)
+      const { status } = await octokit.rest.users.getAuthenticated()
+      if (status < 400) {
+        return githubToken.token
+      }
+    } catch {}
```
erezrokah, line 41: Should we add a log message telling the user their token is expired?

erezrokah (changes requested):
This looks great @lukasholzer and I love the type definitions ❤️ 

Added two comments for your consideration

src/utils/gh-auth.js
lukasholzer: Thx have not known about the code guidelines 👍  will read through :)

src/utils/init/config-github.js
lukasholzer: Will do! Good point :)

src/utils/gh-auth.js
erezrokah: Yes, this should be a lint rule actually (e.g. the opposite of https://eslint.org/docs/rules/no-return-await)

erezrokah: Update: I wonder why https://eslint.org/docs/rules/require-await didn't catch this. We have it [configured](https://github.com/netlify/eslint-config-node/blob/5dd59a59882b0a9cee04e376d3126a72da036daa/.eslintrc.js#L126)

src/utils/init/config-github.js
```diff
@@ -193,7 +217,7 @@ module.exports = async function configGithub({ context, repoName, repoOwner, sit
   })
   await saveNetlifyToml({ repositoryRoot, config, configPath, baseDir, buildCmd, buildDir, functionsDir })
 
-  const octokit = getGitHubClient({ token })
+  const octokit = await getGitHubClient(token)
```
erezrokah, line 220: Is this `await` needed (same for the tests)?

erezrokah (approved):
Added one more comment to clarify.

Looks 🚀

src/utils/init/config-github.js
lukasholzer: oh this is a old left over will change

lukasholzer: What do you mean with the same for the tests? In the tests a different function is tested `getGitHubToken` this is actually async

erezrokah: Ah, I missed that part (the fact it's async in the tests). You can ignore that part of my comment

erezrokah (approved):

PR #3697, opened 2021-11-24
chore!: migrate to esmodules

### LINKED ISSUE #3514: Use pure ES modules
See background at https://github.com/netlify/team-dev/issues/36

Once https://github.com/netlify/cli/issues/3512 is done and released, we should use pure ES modules and make a major release.

This is more than just switching from CommonJS to `import`/`export`. See [this list](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) for other changes which might be involved. This should be broken in many PRs, as much as possible, to lower the risk. Also, non-breaking changes (such as adding file extensions in imports, or loading JSON files differently) should be done before the breaking changes (such as using `import`/`export` statements).

Since we can make a major release, upgrading would require a user action and no sites would be broken.

lukasholzer:

#### Summary

Fixes #3514 

THIS IS WIP TO TRACK THE CURRENT STATE

---

      passes our tests.

image


src/lib/functions/registry.js
```diff
@@ -1,15 +1,19 @@
 // @ts-check
-const { mkdir } = require('fs').promises
-const { env } = require('process')
+import { promises } from 'fs'
+import { env } from 'process'
 
-const { NETLIFYDEVERR, NETLIFYDEVLOG, chalk, log, warn } = require('../../utils')
-const { getLogMessage } = require('../log')
+import { listFunctions } from '@netlify/zip-it-and-ship-it'
```
lukasholzer, line 5: @ehmicky or @erezrokah maybe you have a better idea but the require was in the constructor for performance reasons.

I don't know any better way that does not require huge refactorings as a constructor cannot be async or an unawaited promise can cause race conditions.

Do you know how large the performance impact will be with that?

src/utils/execa.js
```diff
@@ -1,12 +1,13 @@
...
-const execaLib = require('execa')
+import execaLib from 'execa'
 
 // This is a thin layer on top of `execa` that allows consumers to provide an
 // alternative path to the module location, making it easier to mock its logic
 // in tests (see `tests/utils/mock-execa.js`).
 
 // eslint-disable-next-line import/no-dynamic-require
+// TODO: does not work either!
 const execa = env.NETLIFY_CLI_EXECA_PATH ? require(env.NETLIFY_CLI_EXECA_PATH) : execaLib
```
lukasholzer, line 11: @ehmicky do you have a better idea as turning that into an async function? Do we need that still?

src/lib/functions/registry.js
ehmicky: Could this be moved out of the constructor?
For example:

```js
const zipItAndShipItPromise = import('@netlify/zip-it-and-ship-it')

class FunctionsRegistry {
  // Does not set `listFunctions` anymore
  constructor(...) {...}
  ...

  async scan(directories) {
    const { listFunctions } = await zipItAndShipItPromise
    const functions = listFunctions()
  },
```

On my machine, ZISI takes 500ms to load. However, it is using some modules shared by other CLI dependencies, so some of those 500ms are spent loading modules which would need to be loaded anyway.

Still, 500ms is somewhat significant, so we might want to keep it this performance optimization, since it is fairly simple to do so. What do you think?

Also cc @eduardoboucas who probably introduced this in https://github.com/netlify/cli/pull/2826

src/utils/execa.js
ehmicky: `NETLIFY_CLI_EXECA_PATH` feels a little hacky to me, and seems to be only needed for testing purpose, so if there is a way to remove it, we should try that.

src/lib/functions/registry.js
erezrokah: I believe it was me do did the optimization in https://github.com/netlify/cli/pull/2585, so it can have a big difference.
I recommend running some benchmarks to verify.

erezrokah:
Closing in favor of https://github.com/netlify/cli/pull/4141.
Will keep the branch as a reference

PR #3717, opened 2021-11-26
fix: Remove oclif framework and fixes an issue with sites:create

### LINKED ISSUE #3716: Remove oclif and replace with commander
The full problem and discussion about which tool can be found on our internal issue: https://github.com/netlify/pod-workflow/issues/343

lukasholzer:

#### Summary

Fixes https://github.com/netlify/pod-workflow/issues/343

Fixes #3716

## Good to know when reviewing

- [commanders syntax for specifying a value on an option](https://github.com/tj/commander.js#options) or argument tells if it is required or not:
   `.option('--data [data]')` brackets make data optional can be undefined as well
   `.option('--data <data>')` angle brackets make passing a data required.
- checkout the updated docs, as they changed a little bit (best place to identify changes)

## Refactorings:

1. Refactored each command into its own folder
   - Each command (parent or sub) has its own file and exports a `create...Command` function that has one parameter the `program` and returns one.
   - Each command folder has a barrel file (`index.js`) that exports the `create....Command` function of the main command (not the sub ones) and if needed the action (like in the `init` case where the init command is reused).
2. The BaseCommand and the Telemetry are now merged in one class `src/commands/base-command.js` and every command extends from it.
3. replaced most `src/lib/fs.js` methods like `readFileAsync` through `fs.promises.readFile` as there is no need to promisify them.
4. The completion is completely written up from scratch as the old one was a plugin for oclif

#### Commands to refactor:

- [x] main entry with update notifier
- [x] api
- [x] addons (Not sure on how to test them manually)
     - [x] auth
     - [x] config
     - [x] create
     - [x] delete
     - [x] list
- [x] build
- [x] completion
- [x] deploy
- [x] dev
     - [x] exec
     - [x] trace
- [x] env
     - [x] get
     - [x] import
     - [x] list
     - [x] set
     - [x] unset
- [x] functions
     - [x] build
     - [x] create
     - [x] invoke
     - [x] list
     - [x] serve
- [x] help
- [x] init
- [x] link
- [x] lm
     - [x] info
     - [x] install
     - [x] setup
     - [x] uninstall
- [x] login
- [x] logout
- [x] open
- [x] sites
     - [x] create
     - [x] delete
     - [x] list
- [x] status
     - [x] hooks
- [x] switch
- [x] unlink
- [x] watch
- [x] base command
- [x] tracking base

### Other things to do

- [x] create other styling for default help page to be visually the same like the oclif one
- [x] Fix the oclif `error`, `warning` and `exit` things
- [x] override the exit and do the send tracking there
- [x] add debug log scopes
- [x] check for open todos in code `// TODO: ....`

image


lukasholzer:
Sorry for the size 🙈

lukasholzer:
@erezrokah removed `cli-ux` as well from the dependencies as it had a dependency to `oclif` and therefore causing incident: https://github.com/netlify/cli/issues/3788

Tried this version and works with `yarn add`

erezrokah (changes requested):
Hi @lukasholzer, this is great work.

I left some comments and keep doing so, but I'll let you go over the existing ones while I continue the review to unblock you.

src/commands/open/index.js
```diff
@@ -1,41 +1,5 @@
...
-
-    // Show help on empty sub command
-    if (isEmptyCommand(flags, args)) {
-      showHelp(this.id)
-    }
-
-    if (flags.site) {
-      await OpenSiteCommand.run()
-    }
-    // Default open netlify admin
```
erezrokah, line 22: There is behavior change in this command. `ntl open` should show the help and then open the admin page by default.

With this PR, `ntl open` only shows the help. Can we keep the old behavior?

src/commands/open/open.js
```diff
@@ -0,0 +1,38 @@
...
+const open = async (options, command) => {
+  if (!options.site || !options.admin) {
+    command.help()
+  }
+
+  if (options.site) {
+    await openSite(options, command)
+  }
+  // Default open netlify admin
+  await openAdmin(options, command)
```
erezrokah, line 18: Not sure why this is not working, it seems `commander` exits after the `help` method

src/commands/sites/sites-create.js
```diff
@@ -0,0 +1,187 @@
...
+
+  track('sites_created', {
+    siteId: site.id,
+    adminUrl: site.admin_url,
+    siteUrl,
+  })
+
+  if (options.withCi) {
+    log('Configuring CI')
+    const repoData = await getRepoData()
```
erezrokah, line 128:
```suggestion
    const repoData = await getRepoData({})
```

Unrelated bug introduced in https://github.com/netlify/cli/pull/2827

src/commands/sites/sites-create.js
```diff
@@ -0,0 +1,187 @@
...
+    await configureRepo({ context: this, siteId: site.id, repoData, manual: options.manual })
```
erezrokah, line 129:
```suggestion
    await configureRepo({ command, siteId: site.id, repoData, manual: options.manual })
```

site/scripts/generate-command-data.js
```diff
@@ -1,113 +1,62 @@
...
-const { globby } = require('markdown-magic')
-
-const getFlagType = (flagData) => {
-  if (flagData.type === 'option') {
-    return flagData.options ? flagData.options.join(' | ') : 'string'
+// @ts-check
+const { createMainCommand } = require('../../src/commands')
+const { sortOptions } = require('../../src/utils')
+
+const program = createMainCommand()
```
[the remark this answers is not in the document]
lukasholzer: thx!

src/commands/sites/sites-create.js
lukasholzer: I would solve it like this in `src/utils/get-repo-data.js`:

```
/**
 *
 * @param {object} config
 * @param {string} [config.remoteName]
 * @returns
 */
const getRepoData = async function ({ remoteName } = {}) {
```

src/commands/sites/sites-create.js
lukasholzer: good catch thx!

src/commands/open/open.js
lukasholzer: Yea `command.help()` shows the help and exits I refactored to `log(command.helpInformation())` this returns the help information as string.

src/commands/open/index.js
lukasholzer: I did not notice but already fixed :)

src/commands/base-command.js
```diff
@@ -0,0 +1,501 @@
...
+   * IMPORTANT this function will be called for each command!
+   * Don't do anything expensive in there.
+   * @param {string} name The command name
+   * @returns
+   */
+  createCommand(name) {
+    return (
+      new BaseCommand(name)
+        // If  --silent or --json flag passed disable logger
+        .addOption(new Option('--json', 'Output return values as JSON').hideHelp(true))
```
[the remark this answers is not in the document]
lukasholzer: oh have overseen it! already added

erezrokah (approved):
I think this is good to go @lukasholzer 🚀

PR #3766, opened 2021-12-03
feat: remove unused `--name` flag

netlify-team-account-1:
While testing #3717, I wondered what the `--name` flag of `ntl functions:list` does. Looking through the code, I can't find a single usage of the flag. Maybe it's a remnant from some copy & paste.

Technically we remove a flag, which would be a breaking change - but since it doesn't do anything, I think it's fine to assume nobody uses it. Is it still a breaking change?


netlify-team-account-1:
@lukasholzer I'll hold this PR open until #3717 is merged, so you don't get any weird merge conflicts because of it.

lukasholzer:
@netlify-team-account-1 I will take over this and migrate it to the new commander framework

erezrokah (approved):

PR #3863, opened 2021-12-20
fix(functions): fixes an issue where the functions could not be served

### LINKED ISSUE #3854: Presence of a [dev] block in Netlify TOML causes local dev to fail for functions
**Describe the bug**

The presence of a `[dev]` block in `netlify.toml` causes functions to fail with what looks like a zisi error:
```
Request from ::1: GET /.netlify/functions/hello
TypeError: Cannot destructure property 'inputs' of '(intermediate value)' as it is undefined.
    at buildFunction (/Users/jlengstorf/.nvm/versions/node/v16.13.0/lib/node_modules/netlify-cli/src/lib/functions/runtimes/js/builders/zisi.js:39:11)
    at NetlifyFunction.build (/Users/jlengstorf/.nvm/versions/node/v16.13.0/lib/node_modules/netlify-cli/src/lib/functions/netlify-function.js:53:42)
    at FunctionsRegistry.buildFunctionAndWatchFiles (/Users/jlengstorf/.nvm/versions/node/v16.13.0/lib/node_modules/netlify-cli/src/lib/functions/registry.js:67:38)
```

**To Reproduce**

1. Create an empty folder
2. Add a "hello world" TypeScript function
    ```ts
    import { Handler } from "@netlify/functions";
    
    const handler: Handler = async (event, context) => {
      return {
        statusCode: 200,
        body: JSON.stringify({ message: "Hello World" }),
      };
    };
    
    export { handler };
    ```
3. Add a netlify.toml with a `[dev]` block
    ```toml
    [build]
      functions = "netlify/functions"
      publish = "public"
    
    [dev]
      functions = "netlify/functions"
      publish = "public"
    ```
4. Run `ntl dev`
5. Try to execute the function

Removing the `[dev]` block solves the issue

**Configuration**

```toml
[build]
  functions = "netlify/functions"
  publish = "public"

[dev]
  functions = "netlify/functions"
  publish = "public"
```

```bash
  System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 92.36 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    npm: 8.3.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  npmGlobalPackages:
    netlify-cli: 8.1.8
```

**Expected behavior**

**CLI Output**

If applicable, add the CLI output to help explain your problem.

**Additional context**

lukasholzer:

#### Summary

Fixes #3854

      passes our tests.

image


eduardoboucas (commented):
If I understand correctly, this is doing the same check twice: once in the registry when creating the `NetlifyFunction` instance, and again in the constructor of `NetlifyFunction`. Is that necessary?

lukasholzer:
@eduardoboucas yes you are absolutely right this wasn't the most elegant approach 😂  – I think I had to less coffee this morning (already updated)

eduardoboucas (dismissed):
LGTM!

lukasholzer:
@eduardoboucas seems that some tests are failing need to investigate further maybe I broke something with the change

src/lib/functions/server.test.js
```diff
@@ -17,6 +18,9 @@ const functionsPath = `functions`
 /** @type { express.Express} */
 let app
 
+// mock the file-system to not write on the real one with `mkdir` inside the registry.js
+mock({ node_modules: mock.load(join(__dirname, '../../../node_modules')) })
```
lukasholzer, line 22: tests were failing before as we were doing an mkdir un-mocked on the real fs which is not possible to write inside `/my-functions/functions` on CI therefore I mocked the entire filesystem as the test does not need one.

src/lib/functions/registry.js
```diff
@@ -143,7 +144,10 @@ class FunctionsRegistry {
     log(`${NETLIFYDEVLOG} ${chalk.green('Loaded')} function ${chalk.yellow(name)}.`)
   }
 
-  async scan(directories) {
+  async scan(relativeDirs) {
+    const directories = relativeDirs.filter(Boolean).map((dir) => (isAbsolute(dir) ? dir : join(this.projectRoot, dir)))
```
lukasholzer, line 148: @erezrokah @ehmicky I think this is the fastet way to fix it -> Wondering if we can fix it on the long term inside the netlify config.

erezrokah: Yes, this should be address by https://github.com/netlify/build/issues/3620.

eduardoboucas (approved):

erezrokah (approved):

PR #3876, opened 2021-12-20
feat(command-env): add migrate subcommand

### LINKED ISSUE #3575: feature: add env:transfer CLI command

I would like to transfer UI environment variables from one site to another

```
# from current linked site to a new site
netlify env:transfer <to-site-id-or-name>

# from one site to another 
netlify env:transfer <from-site-id-or-name> <to-site-id-or-name>
```

Manually transferring variables via copy/paste

**Additional context**

https://github.com/netlify/product/issues/836 (internal link)

ashalfarhan:

#### Summary

Fixes #3575

This PR will add the ability to transfer environment variables from one site to another.
If you want to send env vars from one site to another you can run this subcommand.
Given 2 arguments, the basic usage is:
- If both arguments are provided, then the first argument will be the sender, and the second one is the destination
- If only one argument is provided, then that will be the destination, and the sender should be the current directory site, otherwise, it will exit with a helpful error message
- If there's no env from `siteIdA`, then it will exit because nothing to transfer

So users can use this command without opening their browser to copy and paste the env vars from one to another.

### Notes:
- Currently this just only works with site id.

---

      passes our tests.

koala5


erezrokah (commented):
@ashalfarhan, sorry for the very later review.

I changed the command a bit to use options instead of arguments.
It makes the purpose of each site id more explicit.
I also added error handling to handle when the CLI can't get the site from the id.
Finally, I think `transfer` can be confusing for this command, as it doesn't remove the variables from the `from` site (I should have figured this out sooner).

We could try calling it `env:migrate` or `env:clone`, what do you think?

ashalfarhan:
> @ashalfarhan, sorry for the very later review.
> 
> I changed the command a bit to use options instead of arguments. It makes the purpose of each site id more explicit. I also added error handling to handle when the CLI can't get the site from the id. Finally, I think `transfer` can be confusing for this command, as it doesn't remove the variables from the `from` site (I should have figured this out sooner).
> 
> We could try calling it `env:migrate` or `env:clone`, what do you think?

Yeah, I think it's better to use options instead of arguments.
And sorry to make you work for the error handling 🙏, because I am still not sure should I just log the error or I should do something, so I just let the error be thrown.

Btw `env:migrate` is more relevant for this feature I think.

erezrokah (approved):
🎸

PR #3885, opened 2021-12-21
feat: add support for deploying scheduled functions

eduardoboucas:
#### Summary

Reads the `schedule` property from the functions manifest and compiles the resulting schedules into a `function_schedules` property that is passed to the `createSiteDeploy`/`updateSiteDeploy` API methods.

dog-clock


src/utils/deploy/hash-fns.js
```diff
@@ -107,7 +107,13 @@ const hashFns = async (
     normalizedPath: path.basename(functionPath, path.extname(functionPath)),
     runtime,
   }))
+  const functionSchedules = functionZips.reduce((acc, { name, schedule }) => {
```
netlify-team-account-1, line 110: [dust] could this be map + filter instead? this usage of `reduce` makes it hard to skim for me.

netlify-team-account-1 (dismissed):

netlify-team-account-1 (approved):

src/utils/deploy/hash-fns.js
eduardoboucas: Done in https://github.com/netlify/cli/pull/3885/commits/d2cfbb8467edbce2c1cdc1fe28b08007bf81ce74.

PR #3938, opened 2022-01-03
chore: add CodeQL Action

XhmikosR:

#### Summary

Should help with the security issues in the future.

---

      passes our tests.

---

Feel free to adapt the `on` event or schedule to match your needs.


XhmikosR:
To clarify in case you are not aware, you can ignore specific errors/warnings that are intentional (like used in tests) in the security tab.

CodeQL is not a 1-to-1 replacement to [LGTM](https://lgtm.com/projects/g/netlify/cli/?mode=list), but at least it covers the security-related issues.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,33 @@
+name: 'CodeQL'
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  schedule:
```
erezrokah, line 10: Do we need the `schedule` and `workflow_dispatch`?

What is the use cases for them?

erezrokah (changes requested):
This is @XhmikosR, wondering if we need the schedule and dispatch triggers

.github/workflows/codeql.yml
XhmikosR: Well, the schedule is what it literally means: run the workflow on a schedule. You can adapt this or remove it.

`workflow_dispatch` allows to run the workflow on demand, it was recently added on Actions as a feature.

erezrokah: Sorry, let me clarify - do they provide added value assuming we run this for each PR and push to `main`?

erezrokah (approved):
🚀

PR #3971, opened 2022-01-07
fix: identity default for function:invoke w/o arg

### LINKED ISSUE #3954: Infer default value for Identity headers in `functions:invoke` command

When using the `functions:invoke` command, I'm always asked for input about whether or not to have emulated Identity headers:

```
% ntl functions:invoke test
◈ "port" flag was not specified. Attempting to connect to localhost:8888 by default
? Invoke with emulated Netlify Identity authentication headers? (pass --identity/--no-identity to override) (Y/n)
```

It would be great if we could use a default option, requiring no additional input from the user, and offer overrides via the existing flags.

The default should be no Identity unless the function is one of the Identity triggers (see https://docs.netlify.com/functions/trigger-on-events/#available-triggers).

karagulamos:
Fixes #3954 

The fix eliminates the prompt that users of the Netlify CLI get when calling the `functions:invoke` command without identity header flags (i.e. `--identity` and `--no-identity`). In addition, when the `functions:invoke` is called without an identity flag, a `--no-identity` default is assumed.

Although now redundant, I started with a simple test to validate that we receive a response from the function when no identity flag is passed. However, I left it in the code to catch regressions should requirements change in the future.

Pup Vader


eduardoboucas (dismissed):
Great work, thanks!

eduardoboucas (approved):

erezrokah:
@eduardoboucas can we merge and release this?

Issue #3981, opened 2022-01-09
Error on post install script


mrloop:
**Describe the bug**

`yarn global add netlify-cli` fails with 

```
2022-01-09T06:12:14.3245718Z Command: node ./scripts/postinstall.js
2022-01-09T06:12:14.3246467Z Arguments: 
2022-01-09T06:12:14.3247497Z info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
2022-01-09T06:12:14.3253330Z Directory: /home/runner/.config/yarn/global/node_modules/netlify-cli
2022-01-09T06:12:14.3254522Z Output:
2022-01-09T06:12:14.3255630Z LIBERTY LIBERTY LIBERTY
2022-01-09T06:12:14.3256543Z LIBERTY LIBERTY LIBERTY
2022-01-09T06:12:14.3257372Z LIBERTY LIBERTY LIBERTY
2022-01-09T06:12:14.3262258Z                                    !             H|H|H|H|H           H__________________________________             H|§|§|§|H           H|* * * * * *|---------------------|             H|§|∞|§|H           H| * * * * * |---------------------|             H|§|§|§|H           H|* * * * * *|---------------------|             H|H|H|H|H           H| * * * * * |---------------------|             H|H|H|H|H           H|---------------------------------|          ===============        H|---------------------------------|            /| _   _ |          H|---------------------------------|            (| O   O |)          H|---------------------------------|            /|   U   |          H-----------------------------------             |  =/  |           H              _..._/            H              _|I/|_            H      _______/| H |/_______    H     /           / /          H    |          | | /         |  H    |          ||o||          |  H    |    |     ||o||     |    |  H    |    |     ||o||     |    |  H   Carl Pilcher  
[...]

PR #3995, opened 2022-01-10
fix: pin colors to 1.4.0

erezrokah:

#### Summary

Fixes https://github.com/netlify/cli/issues/3981

We don't use colors as a direct dependency, however by installing it as one, we can force package managers to use the unbroken version.

```
npm ls colors

├── colors@1.4.0
├─┬ node-version-alias@1.0.1
│ └─┬ all-node-versions@8.0.0
│   └─┬ fetch-node-website@5.0.3
│     └─┬ cli-progress@3.9.1
│       └── colors@1.4.0 deduped
├─┬ prettyjson@1.2.1
│ └── colors@1.4.0 deduped
└─┬ winston@3.3.3
  └─┬ logform@2.3.0
    └── colors@1.4.0 deduped
```

---

      passes our tests.

🐒


lukasholzer (approved):
LGTM!

XhmikosR:
I don't think this PR fixes the issue... Since this project is using a lock file, this add a redundant dependency which does not enforce the package managers to respect it in any way.

One solution would be to use `overrides` and `resolutions` with whatever downsides they have.

The best solution would be to bug the direct parents of `colors` to pin the dependency on their side, until npm finally removes the offending packages.

erezrokah:
This is a not ideal solution for `yarn`. With `npm` we didn't encounter the issue as we have a shrinkwrap file (which `yarn` doesn't support).
The idea is to rely on package managers de-deduping mechanism to install the correct version.

>One solution would be to use overrides and resolutions with whatever downsides they have.

Those work at the consumer level, and not at the publisher level as far as I know (overrides is quite new as well and only supported in latest `npm`).

>The best solution would be to bug the direct parents of colors to pin the dependency on their side, until npm finally removes the offending packages.

💯 And we'll follow up with that.

XhmikosR:
> The idea is to rely on package managers de-deduping mechanism to install the correct version.

But this change here does not do this. You can try it and you'll see that whatever you have in the root project as a dependnecy, is ignored downlevel.

XhmikosR:
@erezrokah I think you can update to the latest prettyjson which has colors now pinned too.

PR #3996, opened 2022-01-10
chore(tests): remove `util.promisify`

XhmikosR:

#### Summary

---

      passes our tests.


XhmikosR:
Unsure how to make tests run to verify the change is OK... it seems this fails in `test:affected`

lukasholzer (approved):
Thank you very much for finding those old legacy parts in our codebase and fixing them! 💪 

🚀

XhmikosR:
NP, glad I could help :)

BTW can you confirm this is fine? CI cannot run due to an issue with the affected test files.

lukasholzer:
Have you tried running the test locally?

`npx ava tests/command.deploy.test.js`

```
✖ No tests found in tests/command.deploy.test.js
```

XhmikosR:
Yeah, it has to do with the affected files script, I believe. So, running the whole suite should be enough, but I can't run it on my machine because ava cripples my dev VM :/

lukasholzer:
mhmm locally it works like a charm:

CleanShot 2022-01-12 at 13 04 16@2x

erezrokah (approved):

erezrokah:
Failures are related to https://github.com/netlify/cli/blob/8c73e8442d94ce51be24cce5ff6cef23f56b3e03/.github/workflows/main.yml#L69.

The deploy tests hit a live API, so we use a Netlify API token in CI.
However, that token is not available to forked pull requests hence we don't run those for forked PRs, causing this failure.

See https://github.com/netlify/cli/blob/8c73e8442d94ce51be24cce5ff6cef23f56b3e03/tests/command.deploy.test.js#L45.

I think I'll add a simple always passing test to workaround this issue.

erezrokah (approved):

PR #3998, opened 2022-01-10
fix: pin colors to 1.4.0

erezrokah:
Follow up to https://github.com/netlify/cli/pull/3995, which actually pins `colors`

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -30,7 +30,7 @@
         "chokidar": "^3.0.2",
         "ci-info": "^3.0.0",
         "clean-deep": "^3.0.2",
-        "colors": "^1.4.0",
+        "colors": "1.4.0",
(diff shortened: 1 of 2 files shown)


lukasholzer (approved):

erezrokah:
Closing as it doesn't work (tested with RC release https://github.com/netlify/cli/commits/releases/rc.0/8.6.17)

erezrokah:
> @erezrokah because yarn does not respect shrinkwrap json?

I was trying to rely on the de-duping mechanism to install the correct version, i.e. when `yarn` resolves which dependencies to install, it should install the version that matches all the relevant ranges. By pinning it, I thought I could "trick" yarn to install a the specific version.
For some reason it doesn't work.

Looks like it's better to wait, see https://github.com/netlify/cli/issues/3981#issuecomment-1008797313

PR #4003, opened 2022-01-10
fix: ignore function files with .ZIP extension

### LINKED ISSUE #4001: When encountering a zip file, ignore it and show a log message that we haven't loaded the function
**Describe the bug**

The issue is already described here https://github.com/netlify/zip-it-and-ship-it/issues/637. I'm reopening it here because it's related to the CLI registry module.

**Expected behavior**

When loading functions invoked via `netlify functions:serve`, we should ignore function files ending with a ZIP extension and log a detailed message for the user as shown below:

image

@eduardoboucas, I have created a PR for this issue here https://github.com/netlify/cli/pull/4003.

karagulamos:
Fixes #4001 

When serving functions via `netlify functions:serve`, the functions registry incorrectly loads function bundles (i.e. zip archive) created by https://github.com/netlify/zip-it-and-ship-it/issues/637. This results in a cryptic and unrecoverable error when we try to access the functions.

This fix ensures the registry ignores function bundles and logs a helpful message to the console.

@eduardoboucas 

image

image


src/lib/functions/registry.js
```diff
@@ -141,6 +143,18 @@ class FunctionsRegistry {
       )
     }
 
+    // This fixes the bug described here https://github.com/netlify/zip-it-and-ship-it/issues/637
+    // If the current function's file has a '.zip' extension, we ignore it to prevent unrecoverable
+    // errors when it's called and log a descriptive message of the issue to the console.
+    if (func.mainFile.toLowerCase().endsWith(ZIP_EXTENSION)) {
+      log(
+        `${NETLIFYDEVERR} Unable to load the function ${terminalLink(chalk.yellow(name), func.url)}. Invalid file '${
```
eduardoboucas, line 151: Rather than framing this as an error and saying that it's an invalid file, I think I would say something like "Skipped bundled function <X>. Unzip the archive to load it from source", and maybe use `NETLIFYDEVWARN` instead of `NETLIFYDEVERR`? My thinking is that this is not a fatal error, and we should try to give people instructions on how to self-serve a solution.

What do you think?

src/lib/functions/registry.js
```diff
@@ -141,6 +143,18 @@ class FunctionsRegistry {
       )
     }
 
+    // This fixes the bug described here https://github.com/netlify/zip-it-and-ship-it/issues/637
+    // If the current function's file has a '.zip' extension, we ignore it to prevent unrecoverable
+    // errors when it's called and log a descriptive message of the issue to the console.
+    if (func.mainFile.toLowerCase().endsWith(ZIP_EXTENSION)) {
```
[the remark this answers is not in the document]
karagulamos: Thanks for the tip. Done.

src/lib/functions/registry.js
karagulamos: Yes, I think yours reads better than what I had, and I agree we should treat this as a warning since it's no longer a crash scenario. Thanks 🥇

src/lib/functions/registry.js
```diff
@@ -141,6 +143,13 @@ class FunctionsRegistry {
       )
     }
 
+    // This fixes the bug described here https://github.com/netlify/zip-it-and-ship-it/issues/637
+    // If the current function's file is a zip bundle, we ignore it and log a helpful message.
+    if (extname(func.mainFile) === ZIP_EXTENSION) {
+      log(`${NETLIFYDEVWARN} Skipped bundled function. Unzip the archive to load it from source.`)
```
eduardoboucas, line 149: I think it would still be useful to show the name of the function.

```suggestion
      log(`${NETLIFYDEVWARN} Skipped bundled function ${chalk.yellow(name)}. Unzip the archive to load it from source.`)
```

eduardoboucas (approved):

src/lib/functions/registry.js
karagulamos: > I think it would still be useful to show the name of the function.

💯

eduardoboucas (approved):
🚀

Issue #3941, opened 2022-01-04
Dependencies to remove/replace


XhmikosR:
**Is your feature request related to a problem? Please describe.**

In an attempt to reduce the number of dependencies, I've made a couple of PRs (#3933, #3939, #3940).

**Describe the solution you'd like**

I think the following could be implemented to reduce the number of dependencies a little more:

1. replace `minimist` with `commander` in builders/netlify-lambda.js; commander is [bigger](https://packagephobia.com/result?p=commander%2Cminimist) compared to minimist, but commander is already a dependency
2. replace `omit.js` with Object spread?
3. replace/remove `read-pkg-up` with `find-up` + own read functions
4. maybe replace `find-up` with `escalade` https://packagephobia.com/result?p=find-up%2Cescalade%2Cread-pkg-up
5. `del` can be replaced with native when Node.js 12.x is dropped (it can be replaced already but because `fs.rm`/`fs.rmDir` has changed it would require a workaround)

There are probably more packages that could be replaced, but the above should already help a bit. Feel free to add any other candidates to the list.

**Describe alternatives you've considered**

-

**Additional context**

-

**Can you submit a pull request?**

Maybe.

> Pull requests are welcome! If you would like to help us add this feature, please check our
> [contributions guidelines](../../CONTRIBUTING.md).

erezrokah:
I marked this as a `good first issue`. If someone is planning to pick this up, please specify which bullet point you'd like to address first

cordery:
Almost 500MB now.   895 packages in in /opt/homebrew/Cellar/netlify-cli/16.9.1/libexec/lib/node_modules/netlify-cli/node_modules/.

Seems a bit excessive.

```
du -hs /opt/homebrew/Cellar/netlify-cli/16.9.1/
496M	/opt/homebrew/Cellar/netlify-cli/16.9.1/
```

Jayllyz:
Some dependencies that could be replaced:

`chalk`-> `picocolors`
`execa`-> `tinyexec`
`sora`-> `nanospinner` or `tiny-spinner`

See https://github.com/es-tooling/module-replacements

sindresorhus:
> chalk-> picocolors

I recommend reading this: https://github.com/chalk/chalk#why-not-switch-to-a-smaller-coloring-package

> ora-> nanospinner or tiny-spinner

The dangers of moving to a smaller package is that there is a reason a package is larger, it handles more edge-cases, which the small package eventually will discover. And so the cycle continues.

[...]

PR #4017, opened 2022-01-12
fix(deps): remove del dependency

XhmikosR:
So, apparently there is already code for `rm`. Testing to see if there are any deprecation warnings/errors.

The `src/lib.fs.js` file could be simplified later when Node.js 12.x support is dropped, removing semver and the conditional `rm`.

Also, if you have similar code in your other packages which cli depends one, it'd be nice to apply the same patch. If not, when Node.js 12.x support is dropped, you will be able to remove the del package from the other packages.

Refs #3941.

---

#### Summary

---

      passes our tests.


XhmikosR:
I don't see any deprecation warnings on CI, but I could use more eyes to be safe. :)

src/lib/fs.js
```diff
@@ -40,7 +39,7 @@ const rmdirRecursiveAsync = async (path) => {
   if (gte(NODE_VERSION, '14.14.0')) {
     return await rm(path, { force: true, recursive: true })
   }
-  await del(path, { force: true })
+  await rmdir(path, { force: true, recursive: true })
```
ehmicky, line 42: Unlike `fs.promises.rm()` and `del`, `fs.promises.rmdir()` [does not have a `force` option](https://nodejs.org/api/fs.html#fspromisesrmdirpath-options). Running it against a non-existing file throws `ENOENT`, but only starting from Node 16.0.0

The `recursive` option with `fs.promises.rmdir()` [is deprecated](https://nodejs.org/api/fs.html#fspromisesrmdirpath-options). Starting from Node 16.0.0, using it prints an error message:

```
(node:24125) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```

`fs.promises.rm()` requires Node 14.14.0.

We could possibly fix this by using a utility function along the lines of:

```js
const { promises: fs } = require('fs')

export const removeDir = function (path) {
  return fs.rm === undefined ? fs.rmdir(path, { recursive: true }) : fs.rm(path, { force: true, recursive: true })
}
```

However, I am wondering if we should just wait to drop Node 12, and use `fs.promises.rm()` then? What do you think?
Also cc @lukasholzer

XhmikosR: Correct me if I'm wrong, since there's code that's using `fs.rm` if Node.js is >= 14.14.0, then this path will be followed only on Node.js >= 12.20 and < 14.14.0.

I can't talk about `force`, I'll take your word for it :)

ehmicky: Oh, I did not look at the context, only the line of code, and failed to notice this was already inside a test helper with a Node.js version condition :facepalm: Thanks for pointing this out.

This looks good then, except the `force` option can be removed since it does not exist. :+1:

XhmikosR: I removed `force` but I'm unsure about the change since the 2 code paths don't seem to be equivalent without `force`...

XhmikosR: I think the behavior should be the same, but I can't guarantee it. I'm fine with dropping this PR until Node.js 12.x support is dropped and you can clean up this file even more + do this in all of your project.

src/lib/fs.js
```diff
@@ -40,7 +39,7 @@ const rmdirRecursiveAsync = async (path) => {
   if (gte(NODE_VERSION, '14.14.0')) {
     return await rm(path, { force: true, recursive: true })
   }
-  await del(path, { force: true })
+  await rmdir(path, { recursive: true })
```
lukasholzer, line 42: This is breaking the behavior. It does not support the force argument as this is only supported in `rm` which is only avialable from `14.14.0`

Which means if it would be used to delete a file it would fail:

https://nodejs.org/api/fs.html#fspromisesrmdirpath-options

>  Using fsPromises.rmdir() on a file (not a directory) results in the promise being rejected with an ENOENT error on Windows and an ENOTDIR error on POSIX.
>
> To get a behavior similar to the rm -rf Unix command, use fsPromises.rm() with options { recursive: true, force: true }.

As our test coverage is sadly to low I would not do this change as we cannot predict the side effects. I would try to update the minimum node version as fast as possible to `14.14.0` to use `rm`.

cc @erezrokah @ehmicky

lukasholzer (changes requested):

XhmikosR:
Closing for the aforementioned reasons. Just something more to do when Node.js 12.x is dropped :)

src/lib/fs.js
ehmicky: It appears that `fs.promises.rmdir()` [does not throw `ENOENT`](https://github.com/netlify/cli/pull/4017#discussion_r783122286) on missing directories in Node <14.14.0, but I agree with @lukasholzer this might be safer to wait until we drop Node 12. :+1:

Issue #3655, opened 2021-11-18
Auto-detected frameworks cannot use the static server


ascorbic:
**Describe the bug**

If a framework has an entry in framework-info it disables the static server, meaning that any framework that needs this cannot use auto-detection, so misses out on the other benefits of auto-detection, such as build command.

**To Reproduce**
Cannot be reproduced because it would require adding a framework to framework-info, but the code is here: 
https://github.com/netlify/cli/blob/main/src/utils/detect-server-settings.js#L260

This is related to #3654, and could be fixed with a similar solution to that, or more simply by starting the static server if the framework specifies no dev command and/or target port.

erezrokah:
My comment [here](https://github.com/netlify/cli/issues/3654#issuecomment-972756052) is relevant for this issue to

erezrokah:
@ascorbic I opened https://github.com/netlify/framework-info/issues/657.

Once we have that I think this issue will be about adding `useStaticServer: true` here https://github.com/netlify/cli/blob/f5323c45fec92c3a1e4604f472ea6ee9e468b67b/src/utils/detect-server-settings.js#L295 when a command/port is missing.

PR #4020, opened 2022-01-12
feat(command-dev): support frameworks without a port

erezrokah:

#### Summary

Fixes https://github.com/netlify/cli/issues/3655

Also has https://github.com/netlify/framework-info/pull/663

If the detected framework doesn't have a port, the CLI will still run the command, and fallback to using a static server.

---

      passes our tests.


lukasholzer (approved):
Awesome this should now support the remix use case! :) 

Thx for adding a test ;)

lukasholzer (dismissed):

erezrokah:
@lukasholzer I had to fix one of our tests utilities in https://github.com/netlify/cli/pull/4020/commits/6de6b32b8b95a5d714bbc5ae20e74b9c32106583 to get Windows tests to pass.

The problem is that on Windows the CLI will print that the server is ready before exiting with an error due to a non existing command. When the framework doesn't have a port, we don't really have a good way to know when the framework command is ready.

lukasholzer (approved):

PR #4024, opened 2022-01-12
test: add e2e tests for cli

### LINKED ISSUE #2430: chore: add tests on the published npm artifact
Since we automate `npm` publishing, it should be fairly easy to add tests that ensure the published artifact installs/works correctly.

We can achieve this by `npm` or `yarn` installing the published version in CI after we publish it.

> We might need to retry the installation as sometimes there's a delay between the time `npm publish` finishes to the time the package is available

lukasholzer:

#### Summary

Fixes https://github.com/netlify/pod-workflow/issues/339
Fixes #2430

Runs a big matrix of tests with installing the CLI on the specified package manager (`npm`, `yarn`, `pnpm`) and runs the help command.

This is run on the 3 major operating systems and in two node versions (12 and latest)

CleanShot 2022-01-12 at 17 03 50

---

      passes our tests.

image

cc @ehmicky


package.json
```diff
@@ -194,6 +191,7 @@
     "form-data": "^4.0.0",
     "got": "^11.8.1",
     "graphviz": "^0.0.9",
+    "husky": "^4.3.8",
```
lukasholzer, line 194: Was the only way to get commit lint passing (without it it prompted me the following error: `npm ERR! could not determine executable to run`)

cc @ehmicky

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,59 @@
...
+  e2e:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
+    env:
+      DEBUG: true
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+        node-version: ['12.x', '*']
+        package-manager: ['npm', 'yarn', 'pnpm']
```
erezrokah, line 23: Could we move the `package-manager` inside the test suite?
I'm concerned this will take up quite a few CI machines.

Since we install the CLI locally we can create a directory for each package manager and install the CLI locally to that directory.

erezrokah (changes requested):
This is very cool @lukasholzer and long over due for us.

Added a few comments

npm-shrinkwrap.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "netlify-cli",
   "version": "8.9.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
```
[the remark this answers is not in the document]
lukasholzer: I tried to fix the breaking npm ci due to some weired dependency issue https://netlify.slack.com/archives/CBC2U9MMG/p1642427032006000

I will revert it :)

tools/e2e/run.js
```diff
@@ -0,0 +1,105 @@
...
+    timeout: '5m',
+    updateSnapshots: undefined,
+    workerArgv: [packageManager],
+  })
+
+  api.on('run', (plan) => {
+    reporter.startRun(plan)
+  })
+
+  const runStatus = await api.run({ filter: tests.map((test) => ({ pattern: test })) })
```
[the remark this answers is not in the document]
lukasholzer: I've tried it without but had issues with conflicts but I think maybe adding a seperate config for e2e test files would work.

We don't want to start the e2e tests by running plain `npx ava` as they will fail and need the registry started up through running `npm run e2e`

.github/workflows/e2e-test.yml
lukasholzer: Yea no problem just wanted to achieve the fastest result :D

e2e.config.cjs
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
lukasholzer, line 1: This configuration needs to be on the same level as the package.json

I decided to use a different format for e2e tests (`e2e.js`) to not be picked up by normal ava that can be run by `npx ava` as these tests require the npm registry that is only spawned by running `npm run e2e`

e2e/install.e2e.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+const prepare = async (folderName) => {
+  const folder = join(process.env.E2E_TEST_WORKSPACE, folderName)
+  await mkdir(folder, { recursive: true })
+  return folder
+}
+
+Object.entries(packageManagerConfig).forEach(([packageManager, { install: installCmd, lockFile }]) => {
+  /** @type {import('ava').TestInterface} */
+  const testSuite = packageManagerExists(packageManager) ? test : test.skip
```
lukasholzer, line 26: test will be skipped if package manager is not installed (useful if someone has not installed pnpm for example locally)

erezrokah (approved):
This is great 🚀

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,57 @@
...
+          cache-dependency-path: 'npm-shrinkwrap.json'
+          check-latest: true
+      - name: Install dependencies
+        run: npm ci --no-audit
+      - name: Install pnpm
+        run: npm install -g pnpm
+      - run: npm run e2e
+      - name: Notify Slack
+        uses: 8398a7/action-slack@v3
+        if: failure()
```
erezrokah, line 44: We probably want to send this notification only if this is a cron job

erezrokah: Maybe `if: github.event_name == 'schedule' will work

lukasholzer: I thought we want to have all failings? 🤔 maybe only if branch is main?

lukasholzer: because schedule is running on main as well

erezrokah: For PRs and `main` we should see those as failed status checks and can also make them required checks.

For Slack notifications, we'd want to capture the ones that are caused by a breaking change to one of our dependencies like we had with `colors.js`.

My concern is that if every failure is reported to the Slack channel it might get too busy.
 
 We can go with send a notification if it fails on `main` and adjust accordingly, WDYT?

lukasholzer: Going with your proposed schedule check :) if we oversee something we can always get more restrictive but I agree it should not get noisy

erezrokah (approved):

XhmikosR:
FYI this results in errors https://github.com/netlify/cli/runs/4855874596?check_suite_focus=true

But they don't make the action fail. I can also reproduce the errors locally without a lockfile.

XhmikosR:
This fixes the errors for me plus flattens the deps tree: https://github.com/netlify/cli/compare/main...XhmikosR:deps

PR #4027, opened 2022-01-13
fix: add `schedule` property to normalised config

eduardoboucas:
#### Summary

Include `schedule` in the list of properties that is passed to zip-it-and-ship-it. This fixes an issue where schedules defined in `netlify.toml` are not picked up.

should-dogs-watch-tv-1636567174


lukasholzer (approved):
LGTM

PR #4029, opened 2022-01-13
chore: automatic update contributors list

lukasholzer:

#### Summary

Fixes https://github.com/netlify/pod-workflow/issues/355

To keep our package.json contributors field up to date to value the effort of each individual person who is contributing to the CLI I'm adding a github actions that runs on every PR to update the contributors field.

---

      passes our tests.

image


lukasholzer:
cc @ehmicky @eduardoboucas

tools/contributors.mjs
```diff
@@ -0,0 +1,113 @@
...
+  contributorList
+    .filter(({ type }) => type === 'User')
+    .map((user) => octokit.request('GET /users/{username}', { username: user.login }).then(({ data }) => data)),
+)
+
+// generate a list of strings with name email and website
+const packageJsonContributors = contributors.map((user) => {
+  const web = (user.twitter_username && `https://twitter.com/${user.twitter_username}`) || user.blog
+  let fullName = user.name || user.login
+  let email = mailList.get(user.login) || mailList.get(user.name) || user.email
```
lukasholzer, line 85: Maybe someone has a better idea of getting the email address sadly there is nothing matching from the github api so I have to manual match them with the users

ehmicky: Could we only show the person's URL, without the email address? This is what [`all-contributors`](https://github.com/all-contributors/cli) does for example. This might be better from a privacy standpoint, and might help you solve that implementation problem.

Note: the `all-contributors` codebase might be useful to browse too to compare how they are getting the user information.

lukasholzer: Ahh thx :) I was just trying to get the information that was already provided in the `package.json`

.github/workflows/contributors.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '*'
+          cache: 'npm'
+          cache-dependency-path: 'npm-shrinkwrap.json'
+          check-latest: true
+      - name: Install dependencies
+        run: npm ci --no-audit
+      - name: Generate Contributors
```
erezrokah, line 23: We might be able to use https://github.com/actions/github-script#run-a-separate-file-with-an-async-function for this.

It will save us from installing the `octokit` dependencies directly.

erezrokah: We'll need to wait for https://github.com/actions/github-script/issues/168 or use `commonJS` syntax

tools/contributors.mjs
```diff
@@ -0,0 +1,113 @@
...
+const [GITHUB_TOKEN] = process.argv.slice(2)
+
+if (!GITHUB_TOKEN) {
+  throw new Error('Please provide the GITHUB_TOKEN as argument to the command: node ./tools/contributors.mjs <token>')
+}
+
+const PagedOctokit = Octokit.plugin(paginateRest)
+const octokit = new PagedOctokit({ auth: GITHUB_TOKEN })
+
+const contributorList = await octokit.paginate('GET /repos/{owner}/{repo}/contributors', {
```
erezrokah, line 68: This will give us the existing contributors list, so I'm not sure if it includes first time contributors.
That is, if a first time user opens a PR, their name won't be added in that PR, but in the next on (I think), which can be confusing.

We probably want to look at the `git log` for the current commit, and use that as a reference

erezrokah: Maybe a better way will be to take the PR author from the GitHub action event and use that one

erezrokah:
> @erezrokah I though moving this maybe in a seperate own github action that we can share across the organization? any thoughts on this? Will post it in our (internal) slack

Yes please! This is a great idea

erezrokah (dismissed):
@lukasholzer I stole the code from this PR and moved it to https://github.com/netlify/update-contributors-action.

Also handled the case of adding the current actor https://github.com/netlify/update-contributors-action/blob/0409f900ca26df7f0b876b98144ebd6f833c36ea/src/update.ts#L83

erezrokah (approved):

Issue #4039, opened 2022-01-14
Bug(command-build): unable to set build context via the CONTEXT env flag


erezrokah:
**Describe the bug**

Previously users could set the build context via an environment variable by running `CONTEXT=deploy-preview ntl build`.
This doesn't work anymore, see https://github.com/netlify/cli/pull/3899#issuecomment-1008658116.
Also, running `ntl build` sets the context to `build`.

**To Reproduce**

1. Run `ntl build` and see output, context should be `production`.

or

1. Run `CONTEXT=deploy-preview ntl build` and see output, context should be `deploy-preview`.

**Configuration**

```bash
npx envinfo --system --binaries --npmPackages netlify-cli --npmGlobalPackages netlify-cli

  System:
    OS: macOS 12.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 3.57 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/.yarn/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  npmGlobalPackages:
    netlify-cli: 8.8.2
```

**Expected behavior**

Context should be set by env variable, and default to `production`

**CLI Output**

```
➜  netlify-build-reproductions git:(main) ntl build                                                                                
​
────────────────────────────────────────────────────────────────
  Netlify Build                                                 
────────────────────────────────────────────────────────────────
​
❯ Version
  @netlify/build 26.1.3
​
❯ Flags
  dry: false
  offline: false
​
❯ Current directory
  /Users/erezrokah/Code/github/netlify/netlify-build-reproductions
​
❯ Config file
  /Users/erezrokah/Code/github/netlify/netlify-build-reproductions/netlify.toml
​
❯ Context
  build
```

**Additional context**

Discovered while working on https://github.com/netlify/cli/pull/3899

PR #4040, opened 2022-01-14
fix(command-build): allow setting context via CONTEXT env, default to production

erezrokah:

#### Summary

Fixes https://github.com/netlify/cli/issues/4039

`this.name()` is always defined, so it overrides the environment variable (which is evaluated in `@netlify/config`). Also `@netlify/config` defaults to `production` context, so we should pass `undefined` to allow the default behavior.

---

      passes our tests.


src/commands/base-command.js
```diff
@@ -470,7 +470,10 @@ class BaseCommand extends Command {
       return await resolveConfig({
         config: options.config,
         cwd,
-        context: options.context || this.name(),
+        context:
+          options.context || process.env.CONTEXT || ['dev', 'dev:exec', 'dev:trace'].includes(this.name())
```
erezrokah, line 474: This was done in the child dev command classes when we had `oclif`. Not sure if there's a better way to do it now.

src/commands/base-command.js
```diff
@@ -470,7 +470,10 @@ class BaseCommand extends Command {
       return await resolveConfig({
         config: options.config,
         cwd,
-        context: options.context || this.name(),
+        context:
+          options.context ||
+          process.env.CONTEXT ||
+          (['dev', 'dev:exec', 'dev:trace'].includes(this.name()) ? 'dev' : undefined),
```
erezrokah, line 476: This was previously done in `oclif` child class, not sure if there's a better way to do it now.

lukasholzer: I don't think either. I would add a small comment here in the code to outline why we are checking for (`(['dev', 'dev:exec', 'dev:trace']` to help our future selfs when touching this code again 😂

lukasholzer (dismissed):
Thanks for adding the tests and having further safety nets :)

src/commands/base-command.js
erezrokah: Good point, done in https://github.com/netlify/cli/pull/4040/commits/7ba63676d452b0a5fe8e4a1a6eaa5a2329ebdaf2

lukasholzer (approved):
Thx for adding the comment! 🎉

PR #4041, opened 2022-01-14
chore: fix Husky

ehmicky:
See https://github.com/netlify/js-client/pull/659

This fixes how we use Husky with its latest version.

---

      passes our tests.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,7 +20,7 @@ jobs:
           cache-dependency-path: 'npm-shrinkwrap.json'
           check-latest: true
       - name: Install dependencies
-        run: npm install --production --no-audit
+        run: npm ci --no-audit
       - name: Get size
         run: du -sk node_modules | cut -f1 > .delta.packageSize && echo "kb (Package size)" >> .delta.packageSize
       - name: Run Delta
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -16979,9 +16979,9 @@
       }
     },
     "node_modules/rollup": {
-      "version": "2.63.0",
-      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.63.0.tgz",
-      "integrity": "sha512-nps0idjmD+NXl6OREfyYXMn/dar3WGcyKn+KBzPdaLecub3x/LrId0wUcthcr8oZUAcZAR8NKcfGGFlNgGL1kQ==",
+      "version": "2.64.0",
+      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.64.0.tgz",
+      "integrity": "sha512-+c+lbw1lexBKSMb1yxGDVfJ+vchJH3qLbmavR+awDinTDA2C5Ug9u7lkOzj62SCu0PKUExsW36tpgW7Fmpn3yQ==",
       "bin": {
         "rollup": "dist/bin/rollup"
       },
@@ -32592,9 +32592,9 @@
       }
     },
     "rollup": {
-      "version": "2.63.0",
-      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.63.0.tgz",
-      "integrity": "sha512-nps0idjmD+NXl6OREfyYXMn/dar3WGcyKn+KBzPdaLecub3x/LrId0wUcthcr8oZUAcZAR8NKcfGGFlNgGL1kQ==",
+      "version": "2.64.0",
+      "resolved": "https://registry.npmjs.org/rollup/-/rollup-2.64.0.tgz",
+      "integrity": "sha512-+c+lbw1lexBKSMb1yxGDVfJ+vchJH3qLbmavR+awDinTDA2C5Ug9u7lkOzj62SCu0PKUExsW36tpgW7Fmpn3yQ==",
       "requires": {
         "fsevents": "~2.3.2"
       }
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "url": "https://github.com/netlify/cli/issues"
   },
   "scripts": {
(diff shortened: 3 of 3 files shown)


erezrokah (dismissed):

.github/workflows/benchmark.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           cache-dependency-path: 'npm-shrinkwrap.json'
           check-latest: true
       - name: Install dependencies
-        run: npm install --production --no-audit
+        run: npm ci --no-audit
```
ehmicky, line 23: `husky` is now an indirect dependency (through `@netlify/eslint-config-node`) instead of being direct.

Binaries of indirect dependencies are symlinked by `npm` to the top-level `node_modules/.bin/` thanks to deduping. However, this seems to only work without the `--production` flag.

Also, this is more consistent with the other GitHub actions which all use `npm ci --no-audit`. Please note this is only for the GitHub action which computes the package size.

erezrokah (approved):

.github/workflows/benchmark.yml
XhmikosR: Doesn't this give the wrong result now, though? It now includes devDependencies :/

ehmicky: Oh right!

XhmikosR:
@ehmicky this breaks production installs which probably is the same reason 8.9.0 release failed :/

I tried to install the latest main version from GitHub and it fails with the same error as the release action:

```
C:\Users\xmr\Desktop>npm i -g github:netlify/cli#main
npm WARN deprecated source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
npm WARN deprecated flatten@1.0.3: flatten is deprecated in favor of utility frameworks such as lodash.
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
npm WARN deprecated source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated statsd-client@0.4.7: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated node-pre-gyp@0.13.0: Please upgrade to @mapbox/node-pre-gyp: the non-scoped node-pre-gyp package is deprecated and only the @mapbox scoped package will recieve updates in the future
npm ERR! code 1
npm ERR! git dep preparation failed
npm ERR! command C:\Program Files\nodejs\node.exe C:\Program Files\nodejs\node_modules\npm\bin\npm-cli.js install --force --cache=C:\Users\xmr\AppData\Local\npm-cache --prefer-offline=false --prefer-online=false --offline=false --no-progress --no-save --no-audit --include=dev --include=peer --include=optional --no-package-lock-only --no-dry-run
npm ERR! npm WARN using --force Recommended protections disabled.
npm ERR! npm ERR! code 1
npm ERR! npm ERR! path C:\Users\xmr\AppData\Local\npm-cache\_cacache\tmp\git-cloneCg974T
npm ERR! npm ERR! command failed
npm ERR! npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c husky install node_modules/@netlify/eslint-config-node/.husky/
npm ERR! npm ERR! 'husky' is not recognized as an internal or external command,
npm ERR! npm ERR! operable program or batch file.
npm ERR!
npm ERR! npm ERR! A complete log of this run can be found in:
npm ERR! npm ERR!     C:\Users\xmr\AppData\Local\npm-cache\_logs\2022-01-15T07_29_00_041Z-debug.log

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\xmr\AppData\Local\npm-cache\_logs\2022-01-15T07_29_00_267Z-debug.log
```

Maybe it's time to reconsider reverting the husky update after all? It's been just a pain in the butt :/

ehmicky:
Thanks for finding this out @XhmikosR!
I have created https://github.com/netlify/cli/pull/4046 to fix this. Let's try to first just add `husky` as a `devDependency`, before reverting, as I believe this should fix this problem.

ehmicky:
The PR is merged now, which should fix the problem. :+1:

XhmikosR:
I don't think that's the reason, unfortunately :/

In production you do `npm prune --production` which probably conflicts.

XhmikosR:
You could probably try `npm pack` to see this failing.

ehmicky:
Great catch again!
Adding `husky` solved the problem in the other repositories, but we run some `prepack` and `postpack` npm scripts with Netlify CLI. `prepack` runs before `prepare`, so `husky` is missing since it is a `devDependencies`.

So, at the moment, `npm publish` will fail with an `husky: not found` error. Note: this bug should not impact any users yet, since we haven't released anything. However, it prevents us from publishing new releases.
  
I am wondering whether we could just remove those `prepack` and `postpack` scripts. :thinking: 
`npm publish` already exclude `devDependencies`, so there is no need to prune them. 
I tried running `npm pack --dry` before and after removing `prepack` and `postpack`, and the same bundle seems to be produced, with the same files.

.github/workflows/benchmark.yml
XhmikosR: @ehmicky How about doing `npm prune --prod` after `npm ci`? It's not ideal but it should be close to what it was before :)

XhmikosR:
You could probably remove them, but there must be some reason they exist.

Could be they were needed in the past or something. Or they were added as a precaution measure so that the lock file is as small as possible.

.github/workflows/benchmark.yml
ehmicky: Done in https://github.com/netlify/cli/pull/4066

PR #4043, opened 2022-01-15
chore(deps): remove read-pkg-up

XhmikosR:
Refs #3941.

Note:

Opening this for feedback. I have tried to make this foolproof, but it might not be 100% the same, for example you might not want to throw an error if `JSON.parse` fails here, so, I'll let you decide. :)

PS. I haven't checked if there's similar code in other parts of the codebase. If so, it's probably best abstracting this, using it everywhere and testing it properly.

---

If this lands, the same should be made in the remaining 2 production packages that depend on `read-pkg-up`:

```
C:\Users\xmr\Desktop\cli>npm ls read-pkg-up
netlify-cli@8.9.0 C:\Users\xmr\Desktop\cli
+-- @netlify/build@26.1.3
| `-- read-pkg-up@7.0.1
`-- @netlify/framework-info@8.0.1
  `-- read-pkg-up@7.0.1
```

---

#### Summary

---

      passes our tests.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -95,7 +95,6 @@
         "prettyjson": "^1.2.1",
         "pump": "^3.0.0",
         "raw-body": "^2.4.1",
-        "read-pkg-up": "^7.0.1",
         "semver": "^7.3.5",
         "source-map-support": "^0.5.19",
         "static-server": "^2.2.1",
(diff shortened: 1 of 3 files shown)


erezrokah:
Thanks @XhmikosR, I'm not sure we'd like to inline this dependency.
`read-pkg-up` is small, well maintained, comes with tests and types.

Also it uses `read-pkg` that has other benefits like data normalization, see https://github.com/sindresorhus/read-pkg#why

Closing this for now. Please comment if you think I got this wrong.

PR #4064, opened 2022-01-17
fix: live tunnel arch issue

### LINKED ISSUE #3561: "netlify dev --live" error
This is what happens when I run "netlify dev --live".

      flexone@penguin:~/code/flatiron/projects/project-1-weather-app$ netlify dev --live
      ◈ Netlify Dev ◈
      ◈ Ignored general context env var: LANG (defined in process)
      ◈ No app server detected. Using simple static server
      ◈ Running static server from "project-1-weather-app"
      ◈ Functions server is listening on 46415
      
      ◈ Static server listening to 3999
          Error: Command failed with exit code 2: /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client 
          version
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: 1: 
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: H1�H1��8P�8P��@P�@P��iF: not found
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: 1: 
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: ��: not found
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: 1: 
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: ELF: not found
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: 1: 
          /home/flexone/.config/netlify/tunnel/bin/live-tunnel-client: Syntax error: Unterminated quoted string

I've removed netlify-cli and reinstalled it, through npm, and I created a brand new site at netlify and ran "netlify init" to relink my project.

"netlify dev" works...

lukasholzer:

#### Summary

Fixes #3561

This will now stop working for debian arm64 so we need to build the image first before merging.

The matching PR for cross compiling the different CPU architectures https://github.com/netlify/live-tunnel/pull/112
---

      passes our tests.


src/lib/exec-fetcher.js
```diff
@@ -1,13 +1,15 @@
 // @ts-check
 const path = require('path')
 const process = require('process')
+const { stringify } = require('querystring')
```
ehmicky, line 4: [sand] What are your thoughts on using `new URL()` and `url.searchParams.set()` instead?
This is mostly equivalent, except it uses the DOM API instead of Node.js core API. What are your thoughts on this?

ehmicky (dismissed):

src/lib/exec-fetcher.js
lukasholzer: Yea why not I don't have a preference to be honest -> is there any benefit?

ehmicky: If we had more code to share with the front-end, it would be better. And Node.js has been moving in that direction lately of using more and more the DOM API, as opposed to to its own core Node.js modules (`URI`, `AbortSignal`, `EventTarget`, `globalThis`, ES modules, `Blob`, `ReadableStream`/`WritableStream`/`TransformStream`, `Worker`, Web Crypto, `performance`, `structuredClone`), which seems a good thing at a higher level, and is also closer to the direction Deno is heading.

However, this is more of a conceptual benefit in this case, since we do not intend to run this file anywhere but on a terminal in this case, so either way could work.

bin/run
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 /* eslint-disable promise/prefer-await-to-then,promise/prefer-await-to-callbacks,eslint-comments/disable-enable-pair */
 const process = require('process')
+const { format } = require('util')
```
erezrokah, line 4: Can we remove this line?

erezrokah (dismissed):
This looks good to me.
I'll @ehmicky and @lukasholzer resolve https://github.com/netlify/cli/pull/4064/files#r786137881

ehmicky:
https://github.com/netlify/cli/pull/4064/files#r786137881 was more of a side note, please feel free to dismiss! :)

lukasholzer:
@erezrokah who is the best one to review the matching PR in the live tunnel repo? netlify/live-tunnel#112

src/lib/exec-fetcher.js
ehmicky: I just realized this too: `querystring` is marked as "legacy" [in the Node.js documentation](https://nodejs.org/api/querystring.html#query-string).

lukasholzer:
just tried the new binary for arm64 on my m1 mac and it works 🥳

lukasholzer:
@charliegerard I hope it works now on the raspberry pi ;) 

I added now following CPU architectures: https://github.com/netlify/live-tunnel-client/releases/tag/v0.2.10

charliegerard:
omg thank you so much for working on it!! 😃 🙌  Right on time for a talk I have to give next week 🥳 I'll test it a bit later today!

src/lib/exec-fetcher.js
lukasholzer: I'll change it for you ;)

erezrokah (approved):

PR #4084, opened 2022-01-19
chore: fixes an issue with the e2e tests

### LINKED ISSUE #4082: e2e tests are passing even though they are failing
**Describe the bug**

> 
> FYI this results in errors https://github.com/netlify/cli/runs/4855874596?check_suite_focus=true
> 
> But they don't make the action fail. I can also reproduce the errors locally without a lockfile.
> 
> _Originally posted by @XhmikosR in https://github.com/netlify/cli/issues/4024#issuecomment-1015577829_

CleanShot 2022-01-19 at 11 10 02

lukasholzer:

#### Summary

Fixes #4082

---

      passes our tests.

### DIFF
--- a/tools/e2e/run.js
+++ b/tools/e2e/run.js
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-/* eslint-disable eslint-comments/disable-enable-pair, promise/prefer-await-to-callbacks */
 const { join } = require('path')
 const process = require('process')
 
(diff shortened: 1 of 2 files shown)


lukasholzer:
now exiting with the appropriate exit code

CleanShot 2022-01-19 at 12 18 27

erezrokah (approved):

XhmikosR:
Shouldn't the test fail when there are unmet peer dependencies too? https://github.com/netlify/cli/runs/4866793722?check_suite_focus=true#step:6:376

lukasholzer:
It does not block the user (as it does not crash the install) Therefore the test should not fail.

Only if something is blocking the user it should fail the test. It's basically a safeguard so that we know if we would break something for a user :)

PR #4101, opened 2022-01-21
test: disable flaky tests on windows

lukasholzer:

#### Summary

Disable flaky tests on windows:

These two where always failing and blocking 
https://github.com/netlify/cli/runs/4896037137?check_suite_focus=true

I curated a list of flaky tests and those two are failing more often
https://www.notion.so/netlify/Flaky-CLI-Test-1d125bd140da4d1dbdb05ca02113d46d (internal link)

      passes our tests.

### DIFF
--- a/tests/command.init.test.js
+++ b/tests/command.init.test.js
@@ -1,4 +1,5 @@
 const { readFile } = require('fs').promises
+const process = require('process')
 
 const test = require('ava')
 const cleanDeep = require('clean-deep')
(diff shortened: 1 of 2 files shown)


lukasholzer:
FYI @XhmikosR

erezrokah (approved):

XhmikosR:
Doesn't seem to fix all failures :/ https://github.com/netlify/cli/runs/4898013668?check_suite_focus=true

There must be a way to circumvent the affected tests script BTW. In this PR only a few tests ran :)

lukasholzer:
I forgot this one updated already a pr: https://github.com/netlify/cli/pull/4102

lukasholzer:
@XhmikosR why should we find a way around the affected test script? this was good that only this test was running as no others where affected by the changset.

The problem is when the lockfile or the package.json is touched all tests are run as it can potentially influence all (which causes often troubles as running only few of them is not causing issues only the majority of running in paralell is causing fs issues especially on windows)

XhmikosR:
Because in this case you couldn't test everything due to the nature of that script :)

PR #4106, opened 2022-01-21
feat(dev): document how to disable browser auto launch

### LINKED ISSUE #4100: Add an option to turn off autoLaunch in dev command

I really would like to have an option to be able to turn off the `autoLaunch` behaviour of the dev command.
I understand [why by default it's turned on](https://github.com/netlify/cli/pull/729#issuecomment-591701356), but I think that it would be a nice addition to add a way to turn it off if needed.

I would like to add (I can already open a PR if accepted) an option called `--autoLaunch` on the dev command. By default it would be `true` so that there would be no breaking change, but if provided a false value, the browser won't be auto opened.

I know that this can already be done via `toml` file, but I think it would be a nice addition to the CLI itself

Yes.

albertodeago:

#### Summary

Fixes #4100

Document how to avoid the browser being auto opened when using `netlify dev`.
- Add docs in the command
- Add docs in the Netlify Dev usage page

Feel free to suggest better text or anything

Screenshots of the modified docs

image

image

---

      passes our tests.

he ⬇  wrote the docs so if there are errors it's not my fault
image


src/commands/dev/dev.js
```diff
@@ -335,7 +335,13 @@ const createDevCommand = (program) => {
         'specify the path to a local GeoIP location database in MMDB format',
       ).hideHelp(),
     )
-    .addExamples(['netlify dev', 'netlify dev -d public', 'netlify dev -c "hugo server -w" --targetPort 1313'])
+    .addExamples([
+      'netlify dev',
+      'netlify dev -d public',
+      'netlify dev -c "hugo server -w" --targetPort 1313',
+      'BROWSER=firefox netlify dev # auto open firefox instead of preferred browser',
```
erezrokah, line 342: This bit should work (setting to firefox) should work, but currently doesn't due to an upstream bug:
https://github.com/michaellzc/better-opn/pull/38

erezrokah (approved):

src/commands/dev/dev.js
erezrokah: You're right, better to remove it for now

erezrokah (changes requested):
Requested changes until the removal of `BROWSER=firefox`

erezrokah (approved):

PR #4130, opened 2022-01-24
chore: update netlify-onegraph-internal

sgrove:
#### Summary

Updates the `netlify-onegraph-internal` package with better codegen for some frameworks and subscriptions.


erezrokah:
Thanks @sgrove, going to close this in favor of https://github.com/netlify/cli/pull/4131, and you can approve and merge that one your self.

Dependencies updates should be handled by renovate automatically, and you can force trigger an update in https://github.com/netlify/cli/issues/986

PR #4132, opened 2022-01-25
fix(command-dev): optionally pass settings to getNetlifyGraphConfig

biilmann:
#### Summary

When starting `netlify dev --graph` in a new remix project, I would run into:

```
 ›   Error: Could not acquire required 'port': '3000'
```

Turns out this happens because `getNetlifyGraphConfig` runs after the `dev`
command has already started a server on port 3000 and then use `detectServerSettings`
which includes a call to `acquirePort` and cause the startup to fail.

This makes sure that `detectServerSettings` only runs one during `netlify dev`


erezrokah (approved):
Thanks @biilmann 🚀

PR #4145, opened 2022-01-29
feat(functions): add scheduled functions templates

karagulamos:
This PR is an update to the CLI function templates as follows

1. Added a **scheduled-function** template for Javascript
2. Added a **scheduled-function** template for TypeSecript

image


src/functions-templates/javascript/scheduled-function/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "{{name}}",
+  "version": "1.0.0",
+  "description": "netlify functions:create - scheduled function in javascript",
```
eduardoboucas, line 4:
```suggestion
  "description": "netlify functions:create - scheduled function in JavaScript",
```

src/functions-templates/typescript/scheduled-function/{{name}}.ts
```diff
@@ -0,0 +1,15 @@
+import { Handler, schedule } from '@netlify/functions';
+
+const formatAsDateTime = (date: Date) => `${date.toLocaleDateString()} ${date.toLocaleTimeString()}`;
+
+const handler: Handler = () => {
+    const currentExecutionTime = formatAsDateTime(new Date(Date.now()));
```
eduardoboucas, line 6: The `new Date()` constructor uses the current date by default, there's no need to supply `Date.now()`.

```suggestion
    const currentExecutionTime = formatAsDateTime(new Date());
```

src/functions-templates/typescript/scheduled-function/{{name}}.ts
```diff
@@ -0,0 +1,15 @@
...
+    console.log(`Function executed at ${currentExecutionTime}.`);
+
+    return {
+        statusCode: 200
+    };
+};
+
+module.exports.handler = schedule("* * * * *", handler);
```
eduardoboucas, line 14: [boulder] We should use [ESM exports](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export) in TypeScript. See the [other TypeScript templates](https://github.com/netlify/cli/blob/main/src/functions-templates/typescript/hello-world/%7B%7Bname%7D%7D.ts).

eduardoboucas (changes requested):
Nice one! Noted a couple of things that we should address, but otherwise looks good.

src/functions-templates/javascript/scheduled-function/{{name}}.js
```diff
@@ -0,0 +1,12 @@
+const { schedule } = require('@netlify/functions')
+
+const formatAsDateTime = (date) => `${date.toLocaleDateString()} ${date.toLocaleTimeString()}`
+
+module.exports.handler = schedule('* * * * *', async () => {
+  const currentExecutionTime = formatAsDateTime(new Date())
+  console.log(`Function executed at ${currentExecutionTime}.`)
```
minivan, line 7: [dust] Looks good! I would use `Function executed at ${new Date()}`, without the conversion. I'd expect the users to delete the code pretty much immediately and adjust to the functions _they_ want to write, so it would be great for them to have to delete less code

minivan (dismissed):

src/functions-templates/javascript/scheduled-function/{{name}}.js
eduardoboucas: Agreed!

src/functions-templates/typescript/scheduled-function/{{name}}.ts
karagulamos: Lazy copying at work here 🙊

src/functions-templates/javascript/scheduled-function/{{name}}.js
karagulamos: This makes sense. I've applied the changes to  theJS & TS templates.

src/functions-templates/typescript/scheduled-function/{{name}}.ts
```diff
@@ -0,0 +1,9 @@
+import { Handler, schedule } from '@netlify/functions';
+
+export const handler: Handler = schedule("* * * * *", () => {
```
eduardoboucas, line 3: I don't think you need to specify the `Handler` type here because it should be inferred from `schedule`.

eduardoboucas: Also, I think we're trying to promote the human-friendly syntax (e.g. `@hourly`) over a cron string. Perhaps we could use that here? (@minivan is this right?)

karagulamos: I agree we should make this as human-friendly as possible. I'll update the template with these suggestions.

karagulamos: > Also, I think we're trying to promote the human-friendly syntax (e.g. `@hourly`) over a cron string. Perhaps we could use that here? (@minivan is this right?)

Perhaps we could include a comment of example human-friendly cron syntaxes. This way, users can try these out themselves and look up the rest in our documentation (which doesn't exist yet). What do you think?

src/functions-templates/javascript/scheduled-function/{{name}}.js
```diff
@@ -0,0 +1,18 @@
...
+//
+// Other supported extensions:
+// - @yearly: once a year, on January 1st 00:00 (0 0 1 1 *)
+// - @monthly: every month, on the first day of the month, at 00:00 (0 0 1 * *)
+// - @weekly: every Monday, 00:00 (0 0 * * 0)
+// - @daily: once a day, at 00:00 (0 0 * * *)
+// - @hourly: every hour, at 0 seconds (0 * * * *)
+//
+module.exports.handler = schedule('* * * * *', async () => {
+  console.log(`Function executed at ${new Date()}.`)
```
eduardoboucas, line 13: I think everything we add to a template should serve a specific purpose (e.g. to show that the function receives a certain parameter), and I'm not sure this `console.log` in its current form is aligned with that.

We could show the date of the next invocation, using the `next_run` field in `event.body` (after JSON-parsing).

What do you think?

src/functions-templates/javascript/scheduled-function/{{name}}.js
```diff
@@ -0,0 +1,18 @@
+const { schedule } = require('@netlify/functions')
+
+// Sample function scheduled to run every minute using conventional cron syntax.
```
eduardoboucas, line 3: This comment is great info, but I worry that people will feel intimidated by the amount of information in it. People shouldn't need to learn cron expressions to make use of this feature, which is why I suggested in a previous review that we defaulted to something like `@hourly`.

To accommodate more advanced use cases, I would add a comment saying that we support any cron expression and include a link to something like https://crontab.guru/ to allow people to create one, or familiarise themselves with the concept.

Skn0tt: I remember some discussion around what cron explainer service we should point to here. maybe it's time to fork crontab.guru? :D

karagulamos: I agree with defaulting the cron expression to @hourly, which eliminates the noisy comments in the code. Additionally, I think it makes more sense to refer users to the Labs documentation about [Scheduled Functions](https://ntl.fyi/sched-func) as it feels better adapted to explain what they're about and provides info about supported cron extensions.

eduardoboucas: @karagulamos Agree! I think a link to https://ntl.fyi/sched-func is all we need.

src/functions-templates/javascript/scheduled-function/{{name}}.js
karagulamos: 💯 agree.

eduardoboucas (approved):

PR #4158, opened 2022-01-31
chore: migrate affected tests to esm

lukasholzer:

#### Summary

part of the partial esm migration

---

      passes our tests.


package.json
```diff
@@ -73,7 +73,7 @@
     "postinstall": "node ./scripts/postinstall.js"
   },
   "config": {
-    "eslint": "--ignore-path .gitignore --cache --format=codeframe --max-warnings=0 \"{src,tools,scripts,site,tests,.github}/**/*.{mjs,cjs,js,md,html}\" \"*.{mjs,cjs,js,md,html}\" \".*.{mjs,cjs,js,md,html}\"",
+    "eslint": "--ignore-path .gitignore --cache --format=codeframe --max-warnings=0 \"{src,scripts,site,tests,.github}/**/*.{mjs,cjs,js,md,html}\" \"*.{mjs,cjs,js,md,html}\" \".*.{mjs,cjs,js,md,html}\"",
```
lukasholzer, line 76: needed to exclude `tools` from eslint as it hangs up on one file:

`npx eslint ./tools/project-graph/file-visitor.mjs` does not work cc @ehmicky maybe you have a clou for me it seems that some lint rules collide here

ehmicky: This seems to be a problem with `eslint-plugin-import` running very slowly when the file being linted has an `import ts from 'typescript'` statement. This bug report might be the same problem: https://github.com/import-js/eslint-plugin-import/issues/2340

It seems like the following fixes the problem: changing:

```js
import ts from 'typescript'
```

To:

```js
const { default: ts } = await import(`${'typescript'}`)
```

lukasholzer: @ehmicky yea you are right it passed locally after 13min 🤯 nevertheless I think we should not write our code different because a lint rule is not playing well. I would rather opt then for disabling this lint rule. IMO

erezrokah (approved):

erezrokah (approved):

PR #4163, opened 2022-01-31
fix: supply next_run for all scheduled functions calls

netlify-team-account-1:

#### Summary

Before this PR, the `next_run` property was only passed in for calls from `ntl functions:invoke`, but not for calls from the Browser or CURL requests. Since we want to allow those requests as first-class citizens, this PR moves over the emulation logic to the Netlify Dev side (as opposed to the `functions:invoke` side).


src/lib/functions/netlify-function.js
```diff
@@ -1,11 +1,20 @@
...
+const CronParser = require('cron-parser')
+
 const { error: errorExit } = require('../../utils/command-helpers')
 
 const BACKGROUND_SUFFIX = '-background'
 
 // Returns a new set with all elements of `setA` that don't exist in `setB`.
 const difference = (setA, setB) => new Set([...setA].filter((item) => !setB.has(item)))
 
+const getNextRun = function (schedule) {
```
eduardoboucas, line 11: [dust] Any reason for to have a second `getNextRun` method outside of the class, as opposed to having this logic in `NetlifyFunction.prototype.getNextRun`?

eduardoboucas (approved):
LGTM! Left a non-blocking comment, but it's good to 🚀

src/lib/functions/netlify-function.js
netlify-team-account-1: no specific reason 🤷 copied over from the other file, where there wasn't a prototype to place this on. also easier testable, potentially.

PR #4165, opened 2022-02-01
test: tryout jest as test framework

lukasholzer:

#### Summary

As discussed in our node guild sync I've tried out jest as test runner to see if it is less flaky on the CI.

one thing I already see is that tests are getting more readable through the expectation syntax of jest. IMHO

Furthermore the snapshot capability is better with jest as well and having the possibility to nest tests in `describes` with dedicated `beforeEach` has some advantages as well.

The true power though is the easiness of the mocking capabilities that jest provides. Which exceeds the capabilities with sinon by far. 

As the mocking is built into the test framework you don't need to mock something first before a file is imported which is a killer feature IMO.

---

      passes our tests.


lukasholzer:
It did not solve our issue with the flaky test doing this migration.

Basically I was using `jest-codemods` that can automatic migrate ava to jest: https://github.com/skovhus/jest-codemods

If needed in the future this can be picked up cc @erezrokah

PR #4167, opened 2022-02-02
chore: inject the Authlify Token in `netlify dev`

anmonteiro:

#### Summary

- Reported by @sgrove
- mirrors what was done in https://github.com/netlify/proxy/pull/719 for local dev:
  - in builds, inject the `ONEGRAPH_AUTHLIFY_TOKEN` variable
  - in function invocations, adds `authlifyToken` to the event

---

      passes our tests.

squirrel yelling at husky hooks


src/commands/dev/dev.js
```diff
@@ -253,7 +254,29 @@ const dev = async (options, command) => {
...
+    authlifyJWT = generateAuthlifyJWT(netlifyToken, siteInfo.authlify_token_id, site.id)
+  }
+
+  await injectEnvVariables({
+    env: Object.assign(
+      command.netlify.cachedConfig.env,
+      authlifyJWT == null
+        ? {}
+        : {
+            ONEGRAPH_AUTHLIFY_TOKEN: {
```
anmonteiro, line 271: follows the format expected by `injectEnvVariables`. Not sure if there's a helper to create this?

src/lib/functions/server.js
```diff
@@ -105,6 +106,11 @@ const createHandler = function ({ functionsRegistry }) {
       rawQuery,
     }
 
+    if (config.authlify && config.authlify.authlifyTokenId != null) {
```
anmonteiro, line 109: follows https://github.com/netlify/proxy/pull/719 by injecting the token whenever there's an authlify token ID, and not only when `--graph` is passed.

sgrove (approved):
The only concern I have is that we're exposing `authlifyToken` to end users when it should probably be `netlifyGraphToken`. Let's say that's fine for the beta period, and then migrate users off of it before GA.

anmonteiro:
Tests seem flakey and unrelated to my changes -- they fail in different OSes in different runs, for different tests 😢.

I'm planning on merging this later today if there are no objections. I'll follow up with a fix if indeed there are broken things.

src/commands/dev/dev.js
```diff
@@ -253,7 +254,29 @@ const dev = async (options, command) => {
     )
   }
 
-  await injectEnvVariables({ env: command.netlify.cachedConfig.env, site })
+  const startNetlifyGraphWatcher = Boolean(options.graph)
+  const netlifyToken = await command.authenticate()
```
erezrokah, line 258: This will open up a browser and authenticate if users are not logged in.
This is not the expected behavior for the CLI as `ntl dev` should not require it.

src/commands/dev/dev.js
erezrokah, line 271: Where do we want `ONEGRAPH_AUTHLIFY_TOKEN` injected to?

The PR description says builds and functions.
This injects it into functions and dev command, but not builds

erezrokah (changes requested):
Hi @sgrove and @anmonteiro.

I added 2 comments

src/commands/dev/dev.js
anmonteiro: It's injected in builds via Bitballoon. 

Are there local builds where I should inject the variable too?

src/commands/dev/dev.js
anmonteiro: we'll need the netlify bearer token for the user. What's an alternative way of getting it?

erezrokah: That's the way to get it. We should only do it when running with `graph` I believe

anmonteiro: There’s also the API Authentication use case. I can do it only in the presence of an authlify token ID, then. Appreciate the review!

src/commands/dev/dev.js
erezrokah: The CLI has a build command https://github.com/netlify/cli/blob/e76355677415300eb3c69be17aeba76c177ec53d/src/commands/build/build.js#L24

The CLI gets all the configuration, including environment variables from `netlify/config`, so I wonder if it will be easier to implement this PR in that package.
See the code to retrieve site information in https://github.com/netlify/build/blob/dc76f4b27a9fe92395c7d20346ce1655f8340d8c/packages/config/src/api/site_info.js

anmonteiro: I feel like this is more specific to the CLI than the generic config package.

@erezrokah are you on board if we merge this with the current functionality to unblock Mathias's upcoming talk and make the build improvements later?

erezrokah (approved):
Hi @anmonteiro, I'm good with unblocking.

I think it will great to follow up with a test case for this, do you mind opening an issue for it?

Do you mind opening another issue for supporting this in `ntl build`? If we inject the auth token in our build system, will should inject it to `ntl build` too.

erezrokah:
This is released in https://github.com/netlify/cli/releases/tag/v8.15.5

anmonteiro:
Thanks, @erezrokah! I opened https://github.com/netlify/cli/issues/4182 and https://github.com/netlify/cli/issues/4181 per your suggestions.

PR #4169, opened 2022-02-03
Graph improvements

sgrove:
#### Summary

Incorporates feedback from users on the `graph` part of the cli to make improvements on framework detection, path detection, codegen, and adding some (temporary but useful during dev) netlify.toml options to override the defaults when necessary until we have that down solid.


anmonteiro (approved):

erezrokah:
For future reference, for a commit to trigger a release (and to be included in the changelog) it should be prefixed by a `fix:` or a `feat:`.
This is evident by the failing  `conventionalcommits.org` status check.

See more in https://github.com/googleapis/release-please/tree/bd74a5b7b622a6cec0d64cbf9b9e01102dc835b2#how-should-i-write-my-commits.

TLDR: squash it with `fix: graph improvements`

sgrove:
@erezrokah To be clear, when you say "it", are you referring to the commits, or to the title of the PR? I saw the evident failure of the `conventionalcommits.org` status check, but I _thought_ my commits had followed the correct pattern, and it gave me no indication as to _what the failure was_.

If it's just the PR title, I'll be sure to follow that as well in the future.

erezrokah:
> If it's just the PR title, I'll be sure to follow that as well in the future.

It's both actually. If you have a single commit the PR title is taken from it by default, otherwise GitHub uses the branch name. Hence we validate both commit messages and PR titles.
The latter is important as if you have multiple commits the squash commit message it based on the PR title.

TLDR: what counts is the squash commit message, as it's used for changelog and release generation.

PR #4183, opened 2022-02-04
feat: Show edit graph link on dev startup, send operations library edits to UI, and run prettier (if available) on code output

### LINKED ISSUE #4180: feat: Run prettier automatically on generated code

We generate various bits of code from the `graph` piece of the CLI, but we won't know the project's specific settings, so the user always has to run prettier on the result to get it to conform to their expectations.

If prettier is available in the path, it would be a better DX to run it over the resulting files so that the output seamlessly conforms with the user's expectations.

We could also bundle prettier with the CLI and run it ourselves which would reduce the overhead of invoking another process, but the the bundle size increase + locking into a given version of prettier doesn't seem worth the benefits.

Yes

sgrove:

#### Summary

Fixes #4180 

Makes editing the Graph operations library file locally work while also sending updates to the ui to make the DX much smoother via file monitoring.

It also runs prettier (if available) on the output so the code conforms to the user's local expectations.

Shows a link when starting `netlify dev --graph` that takes the user to the graph editing UI (requested by @biilmann)

---

### DIFF
--- a/src/lib/one-graph/cli-netlify-graph.js
+++ b/src/lib/one-graph/cli-netlify-graph.js
@@ -4,7 +4,7 @@ const process = require('process')
 
 const { GraphQL, InternalConsole, NetlifyGraph } = require('netlify-onegraph-internal')
 
-const { detectServerSettings, error, getFunctionsDir, log, warn } = require('../../utils')
+const { detectServerSettings, error, execa, getFunctionsDir, log, warn } = require('../../utils')
 
 const { printSchema } = GraphQL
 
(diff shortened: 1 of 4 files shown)


anmonteiro (approved):

src/lib/one-graph/cli-netlify-graph.js
```diff
@@ -241,6 +241,31 @@ const ensureFunctionsPath = (netlifyGraphConfig) => {
...
+let disablePrettierDueToPreviousError = false
+
+const runPrettier = async (filePath) => {
+  if (disablePrettierDueToPreviousError) {
+    return
+  }
+
+  const command = `prettier --write ${filePath}`
+  try {
+    const commandProcess = execa.command(command, {
```
ehmicky, line 253: Small note: if `filePath` has a space in its path (which is common on Windows, e.g. `C:\Program Files`), it will be interpreted as two different arguments by `execa.command()`.
Instead, using `execa('prettier', ['--write', filePath])` will prevent this problem.

sgrove:
Oh, great note, I'll fix that up tomorrow. Thank you!

Issue #4196, opened 2022-02-07
CLI tracking of successful commands does not work currently


lukasholzer:
**Describe the bug**

Currently we only report crashed commands with the telemetry and there a culprit is in there that it does not differentiate between commands

**To Reproduce**

1. Run any command with debug flag `DEBUG* ntl <command>`
2. Check the logs for 

![CleanShot 2022-02-07 at 14 18 10](https://user-images.githubusercontent.com/11156362/152795389-2236483c-2aef-4e2d-94d6-c64363450357.png)

They won't appear for the command:

The issue consists out of two problems:

1. successful invocations are not reported
2. the wrong command name will be reported:

Issue 1. can be solved inside the `bin/run` file:

```javascript

  /* eslint-disable promise/prefer-await-to-then, promise/always-return */
  program
    .parseAsync(process.argv)
    .then(() => { program.onEnd() })
    .catch((error_) => program.onEnd(error_))
  /* eslint-enable promise/prefer-await-to-then, promise/always-return */
```

Just report on the `.then` as well not only on `.catch`

Issue 2. is here: https://github.com/netlify/cli/blob/main/src/commands/base-command.js#L309 that `this.name` is always `netlify` for each command instead of the real command.

PR #4227, opened 2022-02-08
fix: send analytics on successful commands

erezrokah:

#### Summary

Fixes https://github.com/netlify/cli/issues/4196

~~Will think on how to test this and add a test~~ Done

---

      passes our tests.


src/commands/base-command.js
```diff
@@ -300,13 +300,14 @@ class BaseCommand extends Command {
     const duration = getDuration(startTime)
     const status = error_ === undefined ? 'success' : 'error'
 
-    debug(`${this.name()}:onEnd`)(`Status: ${status}`)
-    debug(`${this.name()}:onEnd`)(`Duration: ${duration}ms`)
+    const command = Array.isArray(this.args) ? this.args[0] : this.name()
```
erezrokah, line 303: This seems rather defensive, but I rather not have the CLI error if something changes with `commander`

tests/integration/240.telemetry.test.js
```diff
@@ -49,9 +49,40 @@ test.serial('should track --telemetry-enable', async (t) => {
 test('should send netlify-cli/<version> user-agent', async (t) => {
   await withMockApi(routes, async ({ apiUrl, requests }) => {
     await callCli(['api', 'listSites'], getCLIOptions(apiUrl))
-    t.true(requests.length !== 0)
+    const request = requests.find(({ path }) => path === '/api/v1/track')
```
erezrokah, line 52: This test was failing on my machine as the mock api can receive requests in a different order.
Updated the test to better search the analytics request

bin/run
```diff
@@ -25,5 +25,10 @@ if (require.main === module) {
   const program = createMainCommand()
 
   // eslint-disable-next-line promise/prefer-await-to-then
-  program.parseAsync(process.argv).catch((error_) => program.onEnd(error_))
+  program
+    .parseAsync(process.argv)
+    .then(() => {
+      program.onEnd()
```
ehmicky, line 31: Note: if `onEnd()` throws, it will be called twice. Is that ok?

ehmicky (approved):

bin/run
erezrokah: Great point! I don't see how it can throw https://github.com/netlify/cli/blob/c2040bb9cd846d52e38eed44232872da7bbcdc91/src/commands/base-command.js#L298

We can put everything in `onEnd` to make it clear that it shouldn't throw, WDYT?

erezrokah:
This PR shows a package size of 508MB for some reason. Going to open an issue to investigate

erezrokah:
Size change unrelated to this change. See PR to fix in https://github.com/netlify/cli/pull/4247.

Looks like https://github.com/netlify/cli/pull/4233 did it since netlify build had a pinned version of `zisi` (didn't dive into it too much)

PR #4240, opened 2022-02-09
fix: handle content-type headers with charset

### LINKED ISSUE #4237: Netlify CLI server does not properly handle content-type to determine base64 body
Currently when the server checks `shouldBase64Encode()` it's passing in the `content-type` header. This function makes a series of checks and compares to an exclusion list. However, if the content-type header is something like `application/json; charset=utf-8` segment, this causes the exclusion list to not match and always returns true.

Oddly, even though `isBase64Encoded` is true, it doesn't seem to encode it *to* base64 data. So when processing the event and checking `isBase64Encoded` and converting *from* base64, it results in invalid binary data.

This was found while debugging a Remix/Stripe integration that only failed on the Netlify adapter. See this Discord thread for more details.

https://discord.com/channels/770287896669978684/940734316298797076/940739970874867793

anmonteiro:

#### Summary

Fixes #4237

---

      passes our tests.

### DIFF
--- a/src/lib/functions/utils.js
+++ b/src/lib/functions/utils.js
@@ -38,6 +38,8 @@ const shouldBase64Encode = function (contentType) {
     return true
   }
 
+  const [contentTypeSegment] = contentType.split(';')
+  contentType = contentTypeSegment
   contentType = contentType.toLowerCase()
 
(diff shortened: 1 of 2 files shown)


ascorbic (approved):
looks good!

PR #4268, opened 2022-02-12
feat: `netlify dev --graph`: watch config.toml and reload automatically

anmonteiro:

#### Summary

Fixes https://github.com/netlify/netlify-graph-project-planning/issues/16

---

      passes our tests.


src/commands/base-command.js
```diff
@@ -10,6 +11,7 @@ const merge = require('lodash/merge')
 const jsClient = import('netlify')
 const netlifyConfigPromise = import('@netlify/config')
 
+const { watchDebounced } = require('../lib/functions/watcher')
```
erezrokah, line 14: Should we move this `watcher` utility to a more generic location?
It can be confusing to have our base command use a functions utility.
We have a `command-helpers.js` file that can be more suitable.

src/commands/base-command.js
```diff
@@ -429,6 +431,18 @@ class BaseCommand extends Command {
     const apiOpts = { ...apiUrlOpts, agent }
     const globalConfig = await getGlobalConfig()
     const { NetlifyAPI } = await jsClient
+    const configWatcher = new events.EventEmitter()
+
+    watchDebounced(configPath, {
```
erezrokah, line 436: I would prefer if we do this only when `--graph` is supplied for now, and not for all commands as it can have unexpected behavior.
For example this assumes the new config is propagated to where it's used, however in some cases we only pass a subset of the config. For example:
https://github.com/netlify/cli/blob/7b6b94a2fb6bf1258b5f43aec770f768399512ad/src/commands/dev/dev.js#L288
https://github.com/netlify/cli/blob/7b6b94a2fb6bf1258b5f43aec770f768399512ad/src/commands/dev/dev.js#L302

Also we would need to check if we need to close the watcher (the one returned from `watchDebounced`), as unclosed file watchers have created issues in our tests in the past (mostly flakiness)

erezrokah (changes requested):
Thanks @anmonteiro, added a comment.

I think it would be better to introduce this functionality only when the `graph` flag is passed to reduce the risk of this breaking other commands.

src/commands/base-command.js
eduardoboucas: One option would be to not write the updated config to the `netlify.config` object (https://github.com/netlify/cli/pull/4268/files#diff-fc2c4aaa55fa6705ff0cf8ecdb6637ee37c932821b806fb3bef0a3b72afa309aR443), and instead supply it as a parameter to the `change` event.

So `netlify.config` would be immutable, which means it shouldn't affect the functionality of existing commands, and whoever wants to subscribe to the `change` event would take the updated config from the event and update it wherever it's needed.

erezrokah: Thanks @eduardoboucas, that's a better approach as it will allow us to gradually adopt this change

anmonteiro: thanks, first step done in 9304cb6e

src/commands/base-command.js
anmonteiro: done in bb026a5b9d17555aa1f1d9170abf4f5d12c68398

src/utils/command-helpers.js
```diff
@@ -198,6 +202,34 @@ const normalizeConfig = (config) =>
     ? { ...config, build: omit(config.build, ['publish', 'publishOrigin']) }
     : config
 
+const DEBOUNCE_WAIT = 100
+
+const watchDebounced = async (target, { depth, onAdd = () => {}, onChange = () => {}, onUnlink = () => {} }) => {
+  const watcher = chokidar.watch(target, { depth, ignored: /node_modules/, ignoreInitial: true })
+
+  await pEvent(watcher, 'ready')
```
ehmicky, line 210: `chokidar` can [emit `error` events](https://github.com/paulmillr/chokidar#methods--events). 
Not a blocker for this PR, but we might consider handling those.

erezrokah (approved):

ehmicky (approved):
This looks good to me, with some notes for follow-up PRs.
Also, some tests are failing, but I'm unsure whether this is related to this PR or not.

src/commands/base-command.js
```diff
@@ -429,6 +431,17 @@ class BaseCommand extends Command {
...
+      depth: 1,
+      onChange: async () => {
+        const { config: newConfig } = await actionCommand.getConfig({ cwd, state, token, ...apiUrlOpts })
```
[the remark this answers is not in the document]
anmonteiro: This is very well spotted, and it also looks a little harder than a quickfix. I'm going to open an issue to track it.

ehmicky (dismissed):

sgrove (approved):

PR #4285, opened 2022-02-14
feat: mark graph sessions as inactive so they don't show up in any UI

sgrove:
#### Summary

Fixes https://github.com/netlify/netlify-graph-project-planning/issues/44 (the cli portion)

We need to make sure we create sessions with the appropriate metadata, that we execute a heartbeat at server-specified intervals, and that on process exit we try our hardest to mark the session as inactive so that it can be removed from any UI list immediately, and also so that the server can GC sessions and associated resources.

Because the sessions may end up being GC'd, there's also logic to check if a session is still available at startup, and to create a new one without user intervention if not.

---


src/commands/dev/dev.js
```diff
@@ -100,27 +108,47 @@ const runCommand = (command, env = {}) => {
...
+      // eslint-disable-next-line promise/always-return
+      if (!cleanupStarted) {
+        cleanupStarted = true
+        const cleanupFinished = Promise.all(cleanupWork.map((cleanup) => cleanup()))
+        return cleanupFinished
+      }
+    })
+    // eslint-disable-next-line promise/prefer-await-to-then,promise/always-return
+    .then(() => {
+      process.exit(1)
```
sgrove, line 142: I'm a little bit confused by the presence of this line - it exits too rapidly, and prevents the signal handlers below from actually completing currently. I've left it in looking for some guidance, since I'm not sure why it's here.

src/commands/dev/dev.js
```diff
@@ -100,27 +108,47 @@ const runCommand = (command, env = {}) => {
...
+    })
   ;['SIGINT', 'SIGTERM', 'SIGQUIT', 'SIGHUP', 'exit'].forEach((signal) => {
-    process.on(signal, () => {
+    process.on(signal, async () => {
+      if (!cleanupStarted) {
+        // eslint-disable-next-line no-unused-vars
+        const cleanupFinished = await Promise.all(cleanupWork.map((cleanup) => cleanup()))
```
lukasholzer, line 148: Can we synchronise the cleanup work? Not sure if the promise in an `process.on` will be awaited propperly

sgrove: We cannot synchronize it, it's an IO call. The promise is properly awaited though.

src/commands/dev/dev.js
```diff
@@ -100,27 +108,47 @@ const runCommand = (command, env = {}) => {
...
+      } else {
+        const errorMessage = result.failed
+          ? `${NETLIFYDEVERR} ${result.shortMessage}`
+          : `${NETLIFYDEVWARN} "${command}" exited with code ${result.exitCode}`
+
+        log(`${errorMessage}. Shutting down Netlify Dev server`)
+      }
+    })
+    // eslint-disable-next-line promise/prefer-await-to-then
+    .then(() => {
```
erezrokah, line 132: Can we avoid this promise chaining and move all this code under the first `then`?

sgrove: How do we await the `Promise.all` inside of the first `then`?

sgrove: Hah, I see, it's an async function... on it!

sgrove:
@erezrokah cleaned up and centralized the `process.exit` logic.

This also fixes a bug that's probably been in the CLI for awhile where hitting `Ctrl-C` wouldn't show:

> ◈ Command was killed with SIGINT (User interruption with CTRL-C): yarn start. Shutting down Netlify Dev server

Because of the two separate `process.exit` invocations.

src/commands/dev/dev.js
```diff
@@ -78,6 +79,28 @@ const isNonExistingCommandError = ({ command, error: commandError }) => {
...
+const cleanupWork = []
+
+let cleanupStarted = false
+
+/**
+ * @param {object} input
+ * @param {number=} input.exitCode The exit code to return when exiting the process after cleanup
+ */
+const cleanupBeforeExit = async ({ exitCode }) => {
+  if (cleanupStarted) {
```
ehmicky, line 94: [sand] Can we use `if (!cleanupStarted)` to avoid the first empty branch?

src/commands/dev/dev.js
```diff
@@ -78,6 +79,28 @@ const isNonExistingCommandError = ({ command, error: commandError }) => {
...
+    // If cleanup has started, then wherever started it will be responsible for exiting
+  } else {
+    cleanupStarted = true
+    // eslint-disable-next-line no-unused-vars
+    const cleanupFinished = await Promise.all(cleanupWork.map((cleanup) => cleanup()))
+    process.exit(exitCode)
```
ehmicky, line 100: Should this be in a `try`/`finally` code to ensure it is called even if one of the `cleanupWork` is rejected?

sgrove: Will do

src/lib/one-graph/cli-client.js
```diff
@@ -51,13 +57,31 @@ InternalConsole.registerConsole(internalConsole)
...
+  // 30 minutes
+  const defaultHeartbeatFrequency = 1_800_000
   let shouldClose = false
+  let nextMarkActiveHeartbeat = defaultHeartbeatFrequency
+
+  const markActiveHelper = async () => {
+    const fullSession = await OneGraphClient.fetchCliSession({ authToken: netlifyToken, appId, sessionId })
+    // @ts-ignore
+    const heartbeatIntervalms = fullSession.session.cliHeartbeatIntervalMs || defaultHeartbeatFrequency
+    nextMarkActiveHeartbeat = heartbeatIntervalms
```
ehmicky, line 76: [sand] The above two lines can be merged by removing the intermediary `heartbeatIntervalms` variable.

sgrove: I specifically pulled it out so that it's obvious what the intention of that variable is via its name.

src/commands/dev/dev.js
```diff
@@ -78,6 +79,30 @@ const isNonExistingCommandError = ({ command, error: commandError }) => {
...
+ * @param {object} input
+ * @param {number=} input.exitCode The exit code to return when exiting the process after cleanup
+ */
+const cleanupBeforeExit = async ({ exitCode }) => {
+  // If cleanup has started, then wherever started it will be responsible for exiting
+  if (!cleanupStarted) {
+    cleanupStarted = true
+    try {
+      // eslint-disable-next-line no-unused-vars
+      const cleanupFinished = await Promise.all(cleanupWork.map((cleanup) => cleanup()))
```
ehmicky, line 99: [sand] Since `cleanupFinished` is not used, should it be removed?

sgrove:
@ehmicky Ok, addressed PR comments

ehmicky (approved):

src/lib/one-graph/cli-client.js
ehmicky: Alright :+1:

PR #4308, opened 2022-02-15
fix: graph language

sgrove:
#### Summary

Fixes `language` in `netlify.toml/graph` to be a scalar and not an array, which was causing the wrong language to be emitted for graph handlers

---

### DIFF
--- a/src/lib/one-graph/cli-netlify-graph.js
+++ b/src/lib/one-graph/cli-netlify-graph.js
@@ -217,8 +217,7 @@ const getNetlifyGraphConfig = async ({ command, options, settings }) => {
   const moduleType =
     (userSpecifiedConfig.moduleType && userSpecifiedConfig.moduleType.split(path.sep)) ||
     defaultFrameworkConfig.moduleType
-  const language =
-    (userSpecifiedConfig.language && userSpecifiedConfig.language) || autodetectedLanguage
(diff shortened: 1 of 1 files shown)


src/lib/one-graph/cli-netlify-graph.js
```diff
@@ -217,8 +217,7 @@ const getNetlifyGraphConfig = async ({ command, options, settings }) => {
   const moduleType =
     (userSpecifiedConfig.moduleType && userSpecifiedConfig.moduleType.split(path.sep)) ||
     defaultFrameworkConfig.moduleType
-  const language =
-    (userSpecifiedConfig.language && userSpecifiedConfig.language.split(path.sep)) || autodetectedLanguage
+  const language = (userSpecifiedConfig.language && userSpecifiedConfig.language) || autodetectedLanguage
```
erezrokah, line 220:
```suggestion
  const language = userSpecifiedConfig.language || autodetectedLanguage
```

It seems `userSpecifiedConfig.language` is duplicated?

erezrokah (changes requested):
Hi @sgrove, left one comment

erezrokah (approved):

erezrokah:
@sgrove when you squash commit this PR, make sure to use `fix:` as the prefix to trigger a new release

Issue #4178, opened 2022-02-04
chore: find a better more automatic way to distribute our tests


erezrokah:
Follow up to https://github.com/netlify/cli/pull/4164.

In https://github.com/netlify/cli/pull/4164 we split our tests across multiple machines using an `ava` feature that allows it.
The feature works by sorting specs by file names, dividing those into chunks and sending each chunk to a different machine.
As a result of `ava`'s distribution strategy and since some of our specs are slower than others, I manually distributed the tests using a naming scheme (prefixing with numbers).

See a suggestion in https://github.com/netlify/cli/pull/4164#discussion_r798675380.
Another possible option is to have a lint rule to enforce a maximum number of tests in a spec, to keep each spec execution time low

erezrokah:
Opened https://github.com/avajs/ava/issues/2965 for reference

sarahetter:
Moving away from ava

PR #4321, opened 2022-02-17
feat: env_files in netlify.toml

### LINKED ISSUE #4272: CLI flag or config to prevent .env variables being injected

I'm running a Vite + Vue 3 application with an API client that has to point to different URLs depending on the environment. Vite supports setting these in `.env`, `.env.local`, etc. These files load in a priority order so that `.env.local` values override `.env` values. However, if the environment variable is already set in the system, the `.env` and `.env.local` values will be ignored; they won't override.

Unfortunately, since `netlify dev` injects the `.env` values. These effectively become system environment variables and the `.env.local` is unable to override them.

There are several solutions:

1. A CLI flag or a config file option to disable injecting `.env`
2. A way to specify which dotenv files to exclude during the injection
3. A more fine-grained way to specify which system environment variables to exclude during injection

My current solution involves putting my local environment variable in to `.env` and setting the production override in `.env.production`. This works because the build is done with `NODE_ENV` as `production`. This works, but isn't an ideal way to organize environment variables.

**Additional context**

No

iib0011:
#### Summary

Fixes #4272

We can now set a list in netlify.toml in the dev section like this

`env_files = [ ".env", ".env.local" ]`

We can also set

`env_files = [ ]`

to avoid any env file being injected


src/utils/init/utils.js
```diff
@@ -198,6 +198,7 @@ const getNetlifyToml = ({
   #  command = "yarn start" # Command to start your dev server
   #  port = 3000 # Port that the dev server will be listening on
   #  publish = "dist" # Folder with the static content for _redirect file
+  #  env_priority = [ ".env", ".env.local", ".env.development", ".env.development.local" ]
```
erezrokah, line 201: I think we can remove this line as it's not expected to be changed that often.

src/utils/dot-env.js
```diff
@@ -19,8 +19,9 @@ const loadDotEnvFiles = async function ({ projectDir }) {
   return response.filter((el) => el.file && el.env)
 }
 
-const tryLoadDotEnvFiles = async ({ projectDir }) => {
-  const dotenvFiles = ['.env', '.env.development']
+const defaultEnvPriority = ['.env', '.env.development', '.env.local', '.env.development.local']
```
erezrokah, line 22: Can we keep the current defaults to avoid a breaking change in behavior?

src/commands/dev/dev.js
```diff
@@ -463,6 +465,7 @@ const createDevCommand = (program) => {
     .option('-o ,--offline', 'disables any features that require network access')
     .option('-l, --live', 'start a public live session', false)
     .option('--functionsPort <port>', 'port of functions server', (value) => Number.parseInt(value))
+    .option('-e, --env <env>', 'set the env files priority', false)
```
erezrokah, line 468: I'm not sure a flag is needed at the moment, so we can keep it out of scope for this PR.
The issue at hand is limited to having it configured in the `netlify.toml` `[dev]` block.

src/commands/dev/dev.js
```diff
@@ -325,8 +325,10 @@ const dev = async (options, command) => {
       '--trafficMesh and -t are deprecated and will be removed in the near future. Please use --edgeHandlers or -e instead.',
     )
   }
-
-  await injectEnvVariables({ env: command.netlify.cachedConfig.env, site })
+  if (options.env) {
+    devConfig.envPriority = options.env.split(' ')
```
erezrokah, line 329: If we do keep the flag (see my other comment on scope), I suggest using a comma separated list, you can see and example in https://github.com/tj/commander.js/tree/02a124c7d58dbae2ef11f9284b2c68ad94f6dc8b#custom-option-processing.

WDYT?

erezrokah (changes requested):
Thanks @iib0011, you're on 🔥 

I suggest keeping this feature limited to the `[dev]` block for now, and also name the property `envFiles` (to match our naming convention) and communicate this is not only about priority, but also which files to use (if any).

site/scripts/docs.mjs
```diff
@@ -9,7 +9,11 @@ import stripAnsi from 'strip-ansi'
 import { generateCommandData } from './generate-command-data.mjs'
 
 const rootDir = fileURLToPath(new URL('../..', import.meta.url))
-const markdownFiles = [join(rootDir, 'README.md'), join(rootDir, 'docs/**/**.md')]
+const isWindows = platform === 'win32'
```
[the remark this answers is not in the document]
iib0011: yes I will

src/utils/init/utils.js
iib0011: ok

src/utils/dot-env.js
iib0011: ok

src/commands/dev/dev.js
iib0011: I can try with the comma separated list

iib0011: Done

src/commands/dev/dev.js
erezrokah: @iib0011 what do you think about removing the `env` flag for now.
The issue is about configuring it in the `[dev]` block. Also if we add the flag we should add it in all relevant commands like `dev,dev:exec,functions:create,functions-serve`.

How about considering doing it in a separate PR?

erezrokah:
Thanks for following up with the changes @iib0011, add another comment.
Also do you mind documenting this under https://github.com/netlify/cli/blob/808a54c9900faf144349df1acd21f3884b7e8098/docs/netlify-dev.md#netlifytoml-dev-block and adding a test similar to https://github.com/netlify/cli/blob/808a54c9900faf144349df1acd21f3884b7e8098/tests/integration/300.command.dev.test.js#L162?

iib0011:
okay

erezrokah:
> @erezrokah should I create a test in each .dev.test file? And can you pls explain to me why there are 8 different test files

That's a very good question. You can create a test in the file I linked. See more in https://github.com/netlify/cli/issues/4178 for the reason we have many test files

iib0011:
thanks. I just commited the test

docs/netlify-dev.md
```diff
@@ -141,6 +141,7 @@ Netlify Dev will attempt to detect the site generator or build command that you
   targetPort = 3000 # # The port for your application server, framework or site generator
   port = 8888 # The port that the netlify dev will be accessible on
   publish = "dist" # If you use a _redirect file, provide the path to your static content folder
+  env_files = [ ".env", ".env.development" ] # The env files to use in priority order
```
erezrokah, line 144:
```suggestion
  envFiles = [ ".env", ".env.development" ] # The env files to use in priority order
```

tests/integration/300.command.dev.test.js
```diff
@@ -258,5 +258,54 @@ testMatrix.forEach(({ args }) => {
       })
     })
   })
+  test(testName('should pass .env vars to function', args), async (t) => {
+    await withSiteBuilder('site-with-env-files', async (builder) => {
+      builder
+        .withNetlifyToml({ config: { env: { env_files: ['.env.development', '.env'] } } })
```
erezrokah, line 264:
```suggestion
        .withNetlifyToml({ config: { dev: { envFiles: ['.env.development', '.env'] } } })
```

tests/integration/300.command.dev.test.js
```diff
@@ -258,5 +258,54 @@ testMatrix.forEach(({ args }) => {
...
+        const builderResponse = await got(`${server.url}/.netlify/builders/env`).text()
+        t.is(builderResponse, 'FROM_ENV_FILE')
+      })
+    })
+  })
+  test(testName('should use .env vars instead of .env.development vars', args), async (t) => {
+    await withSiteBuilder('site-with-build-environment', async (builder) => {
+      builder
+        .withNetlifyToml({
+          config: { env: { env_files: ['.env.development', '.env'] }, functions: { directory: 'functions' } },
```
erezrokah, line 290:
```suggestion
          config: { dev: { envFiles: ['.env.development', '.env'] }, functions: { directory: 'functions' } },
```

erezrokah (changes requested):
Please see my latest comments. I think that's the reason the tests are failing

iib0011:
@erezrokah It is fixed now

tests/integration/300.command.dev.test.js
```diff
@@ -258,5 +258,29 @@ testMatrix.forEach(({ args }) => {
       })
     })
   })
+  test(testName('should use .env vars instead of .env.development vars', args), async (t) => {
+    await withSiteBuilder('site-with-env-files', async (builder) => {
+      builder
+        .withNetlifyToml({
+          config: { dev: { envFiles: ['.env.development', '.env'] }, functions: { directory: 'functions' } },
```
erezrokah, line 265: Based on the code, the the order from left to right, is low priority to high priority.
That's the reason the test is failing (it was only passing because of a bug in the test, see https://github.com/netlify/cli/pull/4321/commits/a239de2e0c3f6b3a86d08ad04d7b37cec2b939e9

iib0011: I see, you're right. In the test I purposedly inverted the order.

erezrokah: This https://github.com/netlify/cli/pull/4321/commits/97380fa949a914ca7382e9821c8e723d48354b49 should fix it

iib0011: Finally. Yes it does

erezrokah (approved):

Soviut:
@erezrokah @iib0011 Tested and it works well.

https://docs.netlify.com/configure-builds/file-based-configuration/ This will need to be documented here.

Soviut:
@awacode21 it's usable since the 9.7.0 release, however, the documentation is lagging behind. As far as I can tell, the docs are not available as a public repo to make PRs to.

Note that the priority is **highest to lowest**; highest on the left of the array, lowest on the right. So a variable defined in an env file will override any conflicting variables in the files to the right of it in the array. 

```toml
  envFiles = [".env.netlify.local", ".env.netlify"]
```

awacode21:
@Soviut thank you for the detailed explanation and your help :-)

PR #4327, opened 2022-02-18
feat(command-deploy): allow support for deploying by site name

### LINKED ISSUE #157: Regression: cannot provide human-readable site name
With the previous version of the CLI, I was deploying to a specific site using

```bash
netlify deploy -p dist/ -s planet-stac
```

With the new version of the CLI, I now need to use

```bash
netlify deploy --prod --dir dist/ --site 95a54956-1c9b-4548-87bc-<redacted>
```

Is it possible to use the human-readable name when deploying sites using the new CLI?

### LINKED ISSUE #3033: Support deploying by site name

I often need to deploy one project (the term “project” here refers to a git repository that contains the source code for a webapp) to more than one Netlify site.

The current CLI is great for deploying to one Netlify site, but the experience could be improved when needing to deploy to more than one site. I believe that adding support for deploying by site name could improve the ergonomics of this use case.

Site names are created by developers and often more memorable than UUIDs. For my use case, the different site names for each project would be generated using a predictable pattern. For example: given a project `my-app`, there could be two Netlify sites: `my-app-A` and `my-app-B`. This would allow a developer to quickly run `netlify deploy` passing the appropriate site name. For example:

```
netlify deploy --site my-app-A
```

My understanding is that site names are unique, so there should always be a 1:1 mapping between a site name and its ID.

This mapping could be cached locally at `./netlify/state.json`. Perhaps:

```json
{
  "sites": {
    "my-app-A": "uuid1",
    "my-app-B": "uuid2"
  }
}
```

If site names do not map 1:1 to IDs, then perhaps a new concept could be introduced. Let's call this concept "targets" (although the name could be anything). The CLI command for this might be:

```
netlify deploy --target my-app-A
```

And the associated `state.json`:

```json
{
  "targets": {
    "my-app-A": "uuid1",
    "my-app-B": "uuid2"
  }
}
```

Targets would only exist on the developer's machine: this would not require a change to the Netlify backend system. The way that it would work is that the CLI would look for a matching ID for the given target name within `state.json`. If a match is not found, then it begins the process of creating the new site. Once the site is created, its site ID is stored within `state.json` under the associated target. On subsequent deploys, that site ID would be passed to the `netlify deploy` command to redeploy the site.

If there's any interest in either of these approaches, then additional changes could be made for DX. For example, if `"sites"` (or `"targets"`) exists within `state.json`, but `siteId` does not, and a user runs `netlify deploy` without any flags, then it could prompt the user if they wish to deploy one of those specific sites or if they wish to make a new one.

It's possible to implement either the `sites` or `targets` behavior described above in a wrapper CLI. That's my current plan. This isn't ideal because I'll be creating a new directory, perhaps `.deploy`, to keep the cached info in, and managing a separate `state.json` object. I'd love it if the Netlify CLI supported this directly!

**Additional context**

Related: #2199 

I'd be happy to if there's interest from the team.

---

Thank you so much for taking the time to read this!

tinfoil-knight:

#### Summary

Fixes #3033
Fixes #157

Currently deploying a specific directory to a site on netlify requires looking up the site id and providing it as `netlify deploy --site <very-long-id>`. Site Ids aren't very memorizable or readable. 

This PR adds support for deploying sites by their names like `netlify deploy --site my-first-site`.

---

      passes our tests.


src/commands/deploy/deploy.js
```diff
@@ -462,6 +464,20 @@ const deploy = async (options, command) => {
   await command.authenticate(options.auth)
 
   let siteId = options.site || site.id
+
+  // replace site name by id if a match is found
+  if (options.site) {
+    const isSiteId = uuidValidate(options.site)
```
tinfoil-knight, line 470: AFAIK netlify apis creates uuidv4. Should we validate the version here too if the API is strict about it?

erezrokah: Thanks for looking into this @tinfoil-knight.

I checked with our backend folks and it's better not to assume anything on the structure of the id.

How about we change [this line](https://github.com/netlify/cli/blob/9c8925222abc20aef50a600e83ae3ad6b7b0779d/src/commands/deploy/deploy.js#L468) to something like:
```
const [{ data: site, error: siteError }, sites] = await Promise.all([
        api
          .getSite({ siteId })
          .then((data) => ({ data }))
          .catch((error) => ({ error })),
        api.listSites({ name: siteId, filter: 'all', per_page: 1 }),
      ])
```

Then if `sites[0]` exists use that one, if not use the result of `getSite` (or error out)

tinfoil-knight: Yeah. This seems better. I'll make the change.

src/commands/deploy/deploy.js
```diff
@@ -462,10 +463,26 @@ const deploy = async (options, command) => {
...
+        api
+          .getSite({ siteId })
+          .then((data) => ({ siteFoundById: data }))
+          .catch((error_) => ({ siteError: error_ })),
+        api.listSites({ name: options.site, filter: 'all', per_page: 1 }),
+      ])
+
+      if (siteFoundById) {
+        siteData = siteFoundById
+      } else if (siteFoundByName && siteFoundByName.name === options.site) {
```
tinfoil-knight, line 480: **Note for Reviewer:**

The additional `siteFoundByName.name === options.site` check is here since the API can return partially matching results and we want exact matches only.

erezrokah: This is a great catch!
What happens if the we have multiple matches, the first one being a partial match?
I don't think the order is guaranteed.

So we retrieve all matches using https://github.com/netlify/cli/blob/bfadc7658518577c069d5e92baddf346f92dca63/src/lib/api.js#L85 and check for an exact match?

tinfoil-knight: If the order isn't guaranteed, then we should check multiple sites.

Plus, I don't think it should cause a performance issue since we're filtering by site name which should limit the number of results returned significantly.

erezrokah (approved):
🚀 Another great PR, @tinfoil-knight ❤️

Issue #4359, opened 2022-02-23
v9.6.0 causes `out of memory` error when deploying


gavinhenderson:
**Describe the bug**

Since upgrading to `v9.6.0` (from `9.5.0`) I consistently get a Javascript out of memory error when running `netlify deploy`.

I have tried upping the node memory limit incase I was just I was just slightly over the memory limit but I still get the error even after doubling the memory limit.

Nothing else has changed in my setup, if i drop the version it builds fine

**To Reproduce**

The best way to reproduce the error is run it within GitHub actions which runs the deploy. 

You can see the runs that consistently fail with the new version heere here: https://github.com/AceCentre/acecentre.org.uk/actions/workflows/deploy.yml?query=branch%3Aupdate-netlify

**Configuration**

My build command is:

```
yarn netlify deploy --alias=${{ needs.CalculateDeployKey.outputs.deploy-key }} --build --context deploy-preview --debug
```

<details>
 <summary>netlify.toml</summary>

```toml

 [[redirects]]
  from = "/trustees"
  to = "/about/trustees"
  status = 301

[[redirects]]
  from = "/account"
  to = "/my-acecentre"
  status = 301

[[redirects]]
  from = "/ace-centre-learning/my-courses"
  to = "/my-acecentre/courses"
  status = 301

[[redirects]]
  from = "/about/staff-member/*"
  to = "/about/staff/:splat"
  status = 301

[[redirects]]
  from = "/project/*"
  to = "/projects/:splat"
  status = 301

[[redirects]]
  from = "/news/*"
  to = "/blog/:splat"
  status = 301

[[redirects]]
  from = "/project/*"
  to = "/projects/:splat"
  status = 301

[[redirects]]
  from = "/news"
  to = "/blog"
  status = 301

[[redirects]]
  from = "/getting-started/faqs-frequently-asked-questions"
  to = "/getting-started"
  status = 301

[[redirects]]
  from = "/services/training"
  to = "/learning"
  status = 301

[[redirects]]
  from = "/services/information"
  to = "/services/advice-information"
  status = 301

[[redirects]]
  from = "/services/information/feedback/"
  to = "/form/information-appointment-feedback"
  status = 301

[[redirects]]
  from = "/privacy"
  to = "/page/privacy"
  status = 301

[[redirects]]
  from = "/ace-centre-privacy-policy-3"
  to = "/page/ace-centre-privacy-policy-3"
  status = 301

[[redirects]]
[...]

PR #4362, opened 2022-02-23
fix: move config watcher to graph command

anmonteiro:

#### Summary

Tentative fix for #4359

---

      passes our tests.


src/commands/dev/dev.js
```diff
@@ -486,8 +511,10 @@ const createDevCommand = (program) => {
       'netlify dev',
       'netlify dev -d public',
       'netlify dev -c "hugo server -w" --targetPort 1313',
+      'netlify dev --graph',
```
anmonteiro, line 514: drive-by adding a new example

src/commands/base-command.js
```diff
@@ -437,27 +430,10 @@ class BaseCommand extends Command {
     const globalConfig = await getGlobalConfig()
     const { NetlifyAPI } = await jsClient
 
-    const configWatcher = new events.EventEmitter()
-
-    // Only set up a watcher if we know the config path.
-    if (configPath) {
```
erezrokah, line 443: What do you think about keeping this in the base command, and only applying for `dev`?
I'm thinking we might want this for `functions:serve` and possibly other long running commands

src/commands/dev/dev.js
erezrokah: You'll need to run `npm run doc` to sync the docs with this change (and fix the CI)

src/commands/base-command.js
anmonteiro: I'd prefer to do that work when we need that for `functions:serve`. There are (as evidenced by this PR) wrinkles to figure out wrt memory usage and I believe exposing this in `BaseCommand` could give false confidence that this is ready for wide application.

sgrove (approved):
lgtm - thanks for helping out our users!

PR #4395, opened 2022-02-28
fix(functions): stop providing clientContext to background functions

### LINKED ISSUE #4383: Background functions provide a clientContext and user locally, but not in production
**Describe the bug**

Background functions don't check for an authorization header token in production, so there's never a `clientContext` or any way to validate that an Identity authenticated user is sending the request. This isn't documented, but it's mentioned in a community answer: https://answers.netlify.com/t/clientcontext-user-missing-in-background-functions/46166/2

This can be hacked around pretty simply (I'm using a regular function to auth and then passing a secret to the background function), but it's a surprise when Netlify Dev _does_ authenticate requests to background functions. So you put your authenticated background function live, it fails in production, and then you can't see why because _background functions don't provide logs_ 🙈 (except sometimes they do??)

I can open a PR to simply remove `clientContext` for background functions [here](https://github.com/netlify/cli/blob/88985c452484c64a1b5763d28e459c46913a330a/src/lib/functions/server.js#L123), which I’ve validated locally, but I didn't see an obvious way to test the inner workings of background functions in the current testing setup. Happy to open a PR with just the fix, or with tests too if anyone has thoughts on how to go about it.

**To Reproduce**

1. Run a background function locally
2. Send a request with a valid identity token in the authorization header
3. Check for `context.clientContext.user` in the handler

**Expected behavior**

`context.clientContext` should be nullish, matching production

Thank you for making the cli amazing! While no pictures of animals were requested, I've provided complimentary hippos.

image

erquhart:
#### Summary

Fixes #4383 

Removes `context.clientContext` for background functions to match production behavior.

---

      passes our tests.

Ah, animal pictures go on the PR, that's right. Here, have some more hippos:
image


erezrokah (approved):
Thanks @erquhart, looks great.

I added a test to capture this use case.

PR #4486, opened 2022-03-28
feat: improve functions error messages

wangoimacharia:
#### Summary

Extends [Improve Function Error Messages](https://github.com/netlify/pod-serverless/issues/78) to CLI. 

This PR adds a template for use when parsing errors from functions:
cli-error-page

---

      passes our tests.

cute animal


minivan:
I'm curious how this fares with functions that are invoked by clients that don't accept HTML. If it works, can we add a test case for that?

src/lib/functions/synchronous.js
```diff
@@ -1,5 +1,7 @@
 // @ts-check
 const { Buffer } = require('buffer')
+const fs = require('fs')
+const pathModule = require('path')
```
eduardoboucas, line 4: [sand] We typically use `path` for the variable that holds the `require` call, or we use object destructuring to get individual methods (e.g. `const { join } = require('path')`). Could we try to do one of those here for consistency?

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,26 @@ const handleSynchronousFunction = function (err, result, response) {
...
-const formatLambdaLocalError = (err) => `${err.errorType}: ${err.errorMessage}\n  ${err.stackTrace.join('\n  ')}`
+const formatLambdaLocalError = (err) =>
+  JSON.stringify({
+    errorType: err.errorType,
+    errorMessage: err.errorMessage,
+    trace: err.stackTrace,
+  })
 
 const handleErr = function (err, response) {
+  const errorTemplateFile = fs.readFileSync(pathModule.join(__dirname, './templates/function-error.html'), 'utf-8')
```
eduardoboucas, line 49: [boulder] I think we should try to read the template once and keep it in memory, as opposed to reading it from disk each time we process an error.

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,26 @@ const handleSynchronousFunction = function (err, result, response) {
...
+  let updatedErrorTemplate = errorTemplateFile
+
   detectAwsSdkError({ err })
 
   response.statusCode = 500
-  const errorString = typeof err === 'string' ? err : formatLambdaLocalError(err)
-  response.end(errorString)
+
+  const regexPattern = /<!--@ERROR-DETAILS-->/g
+  updatedErrorTemplate = updatedErrorTemplate.replace(regexPattern, errorString)
```
eduardoboucas, line 58: Do you think we could extract the logic for rendering the template into a separate function?

eduardoboucas (commented):
Looks good in general — great job!

I left some comments. Btw, if you're wondering about the sand/boulder prefixes, it refers to our [feedback ladder](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/#-sand-non-blocking-but-requires-future-consideration).

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,26 @@ const handleSynchronousFunction = function (err, result, response) {
   response.end()
 }
 
-const formatLambdaLocalError = (err) => `${err.errorType}: ${err.errorMessage}\n  ${err.stackTrace.join('\n  ')}`
+const formatLambdaLocalError = (err) =>
+  JSON.stringify({
+    errorType: err.errorType,
+    errorMessage: err.errorMessage,
+    trace: err.stackTrace,
+  })
```
sean-roberts, line 46: [sand] For readability, multiline arrow functions should be wrapped in braces with a return unless they start on the same like as the arrow.

eduardoboucas: @sean-roberts our global linting rules enforce the opposite.

Screenshot 2022-03-28 at 15 16 46

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,26 @@ const handleSynchronousFunction = function (err, result, response) {
...
   detectAwsSdkError({ err })
 
   response.statusCode = 500
-  const errorString = typeof err === 'string' ? err : formatLambdaLocalError(err)
-  response.end(errorString)
+
+  const regexPattern = /<!--@ERROR-DETAILS-->/g
+  updatedErrorTemplate = updatedErrorTemplate.replace(regexPattern, errorString)
+
+  response.end(updatedErrorTemplate)
```
sean-roberts, line 60: [boulder] we should check to see if the request's `Accept` header includes `text/html` and only if it does should we return this HTML body. Otherwise we return what we had been returning previously

src/lib/functions/synchronous.js
sean-roberts: grrrroosss :) Can you start it on the above line? 
```suggestion
const formatLambdaLocalError = (err) => JSON.stringify({
    errorType: err.errorType,
    errorMessage: err.errorMessage,
    trace: err.stackTrace,
  })
```

eduardoboucas: Formatting is handled automatically by Prettier. 😄

sean-roberts: yeah I figured, we should look at allowing that but this is up to the team. It's certainly a non-traditional code approach that makes it harder to read.

src/lib/functions/synchronous.js
wangoimacharia: Does [this block](https://github.com/netlify/cli/pull/4486/files#diff-a533fc859ae76f3974782aff19d74d697ff83bf2cec5c6ce8f04f742d5946f93R48-R53) suffice? Moved it out so it's only called when we need to render the template.

wangoimacharia:
> I'm curious how this fares with functions that are invoked by clients that don't accept HTML. If it works, can we add a test case for that?

@minivan I've changed it to revert to the original error format if client does not accept HTML. Behaviour before this change, using curl and API client, pasted the entire template out.

Current behaviour:
handles-non-html-clients

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,34 @@ const handleSynchronousFunction = function (err, result, response) {
...
+        errorType: err.errorType,
+        errorMessage: err.errorMessage,
+        trace: err.stackTrace,
+      })
+    : `${err.errorType}: ${err.errorMessage}\n ${err.stackTrace.join('\n ')}`
 
-const handleErr = function (err, response) {
+const renderErrorTemplate = (errString) => {
+  const regexPattern = /<!--@ERROR-DETAILS-->/g
+  const errorTemplateFile = readFileSync(join(__dirname, './templates/function-error.html'), 'utf-8')
```
eduardoboucas, line 52: We're still reading the file each time we need to render a new error page, right? I think we could have `errorTemplateFile` in the global scope of this file, and then whenever you need to render a template you do something like:

```js
errorTemplateFile = errorTemplateFile || readFile(...)
```

So if the file has already been read, use the in-memory value. If not, read it and assign it for future invocations. Does that make sense?

wangoimacharia: Yes and done.

src/lib/functions/synchronous.js
```diff
@@ -36,14 +38,36 @@ const handleSynchronousFunction = function (err, result, response) {
...
+        errorType: err.errorType,
+        errorMessage: err.errorMessage,
+        trace: err.stackTrace,
+      })
+    : `${err.errorType}: ${err.errorMessage}\n ${err.stackTrace.join('\n ')}`
 
-const handleErr = function (err, response) {
+const renderErrorTemplate = async (errString) => {
+  const regexPattern = /<!--@ERROR-DETAILS-->/g
+  let errorTemplateFile
```
eduardoboucas, line 52: Since you're declaring `errorTemplateFile` in the scope of `renderErrorTemplate`, it will always be falsy when you evaluate it in L54.

I think you want to move this declaration out of `renderErrorTemplate`?

eduardoboucas (changes requested):

eduardoboucas (approved):
🚀

PR #4491, opened 2022-03-28
feat: add local dev support for Go scheduled functions

karagulamos:
#### Summary

The PR is part of an ongoing effort to [improve functions support for Go](https://github.com/netlify/pillar-runtime/issues/253). In this specific case, we sprinkle the CLI with magic dust to enable the detection of scheduled functions in Go and improve the dev experience for anyone wishing to try this out locally.

With this update, users should now see the default message for locally run scheduled functions in their browser 

image

You can find relevant code [here](https://github.com/karagulamos/go-scheduled-functions/blob/master/src/hello-cronny/main.go). 

**NOTE**: `next_run` is only present in `request.Body` if the Go function is scheduled.

### DIFF
--- a/src/lib/functions/netlify-function.js
+++ b/src/lib/functions/netlify-function.js
@@ -25,7 +25,7 @@ class NetlifyFunction {
     runtime,
     settings,
     timeoutBackground,
-    timeoutSynchronous
+    timeoutSynchronous,
   }) {
     this.config = config
     this.directory = directory
--- a/src/lib/functions/netlify-function.js
+++ b/src/lib/functions/netlify-function.js
@@ -42,7 +42,7 @@ class NetlifyFunction {
     // name.
     this.isBackground = name.endsWith(BACKGROUND_SUFFIX)
 
-    const scheduledFunc = config.functions[name]
+    const scheduledFunc = config && config.functions && config.functions[name]
     this.schedule = scheduledFunc && scheduledFunc.schedule
 
     // List of the function's source files. This starts out as an empty set
--- a/src/lib/functions/netlify-function.js
+++ b/src/lib/functions/netlify-function.js
@@ -42,8 +42,8 @@ class NetlifyFunction {
     // name.
     this.isBackground = name.endsWith(BACKGROUND_SUFFIX)
 
-    const scheduledFunc = config && config.functions && config.functions[name]
-    this.schedule = scheduledFunc && scheduledFunc.schedule
+    const functionConfig = config.functions && config.functions[name]
+    this.schedule = functionConfig && functionConfig.schedule
 
     // List of the function's source files. This starts out as an empty set
     // and will get populated on every build.
--- a/src/lib/functions/netlify-function.js
+++ b/src/lib/functions/netlify-function.js
@@ -25,7 +25,7 @@ class NetlifyFunction {
     runtime,
     settings,
     timeoutBackground,
-    timeoutSynchronous,
+    timeoutSynchronous
   }) {
     this.config = config
     this.directory = directory
@@ -41,7 +41,7 @@ class NetlifyFunction {
     // Determines whether this is a background function based on the function
     // name.
     this.isBackground = name.endsWith(BACKGROUND_SUFFIX)
-    this.schedule = null
+    this.schedule = config.functions[name]?.schedule
 
     // List of the function's source files. This starts out as an empty set
     // and will get populated on every build.
@@ -91,7 +91,7 @@ class NetlifyFunction {
 
       this.buildData = buildData
       this.srcFiles = srcFilesSet
-      this.schedule = schedule
+      this.schedule = schedule || this.schedule
 
       return { srcFilesDiff }
     } catch (error) {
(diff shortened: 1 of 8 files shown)


src/lib/functions/runtimes/go/index.js
```diff
@@ -37,11 +40,32 @@ const checkGoInstallation = async ({ cwd }) => {
...
 const getBuildFunction = ({ func }) => {
   const functionDirectory = dirname(func.mainFile)
   const binaryPath = tempy.file(isWindows ? { extension: 'exe' } : undefined)
 
-  return () => build({ binaryPath, functionDirectory })
+  return async () => {
+    // From the current function directory, we look up parent directories for the netlify.toml file.
+    // If we find one, we parse it to see if a schedule is defined for the current function. It's okay
+    // if we don't find any, we just assume the function is not scheduled (i.e. undefined).
+    const schedule = await parseForSchedule({ cwd: functionDirectory, functionName: func.name })
```
mraerino, line 60: i don't like that this loads the file everytime even though the file was already loaded before.

i can see that `FunctionRegistry` calls `NetlifyFunction.build` and has the config already present.
consider passing the config down into this method instead of parsing the file anew.

mraerino:
doesn't zisi parse the in-code schedule configuration?

eduardoboucas:
Let's think about how the schedule is declared and how it makes its way to us:

- **In the TOML**: We have access to this information upfront (i.e. before building a function), because we have an in-memory representation of the config. This is what we need to pass to zip-it-and-ship-it, as mentioned in https://github.com/netlify/cli/pull/4491#pullrequestreview-924291951.

- **In the code**: We don't have access to this information until we build a function, so we need to get this data from zip-it-and-ship-it. If you remove this, only the TOML declarations will be considered. The data coming from zip-it-and-ship-it for Go functions will be redundant, since it's not possible for the function's build command to generate additional schedule data (because we don't support in-source declarations in Go). However, that's not true for JavaScript/TypeScript, so we need to make sure we consider data coming from both sources (i.e. config and ZISI), so the best way is to always pass the config to ZISI and use the value it returns as the single source of truth, because it will do the job of considering both sources.

Does that make sense?

karagulamos:
> Does that make sense?

Figured it out, but this makes a lot of sense, thanks.

karagulamos:
This is resolved.

eduardoboucas (approved):
🚀

src/lib/functions/netlify-function.js
```diff
@@ -41,7 +41,7 @@ class NetlifyFunction {
     // Determines whether this is a background function based on the function
     // name.
     this.isBackground = name.endsWith(BACKGROUND_SUFFIX)
-    this.schedule = null
+    this.schedule = config.functions[name]?.schedule
```
eduardoboucas, line 44: FYI we don't support the `?.` operator with the current Node target.

tests/integration/630.serving-functions-go.test.js
```diff
@@ -112,3 +112,72 @@ test('Updates a Go function when a file is modified', async (t) => {
     }
   })
 })
+
+// Reproduction test to verify the abscence/presence of a Go scheduled function
+test('Detects a Go scheduled function using netlify-toml config', async (t) => {
+  const [execaMock, removeExecaMock] = await createExecaMock(`
+    const { writeFileSync } = require('fs')
+
+    const handler = (...args) => {
```
eduardoboucas, line 121: @karagulamos I wonder if we could assert that the function receives `nextRun` in the body?

src/lib/functions/netlify-function.js
karagulamos: Yeah, I noticed sadly.

tests/integration/630.serving-functions-go.test.js
eduardoboucas: Not a blocker though, so please feel free to do that on a separate PR if you prefer.

karagulamos: Oh, it's resolved.

src/lib/functions/netlify-function.js
```diff
@@ -41,7 +41,9 @@ class NetlifyFunction {
     // Determines whether this is a background function based on the function
     // name.
     this.isBackground = name.endsWith(BACKGROUND_SUFFIX)
-    this.schedule = null
+
+    const scheduledFunc = config && config.functions && config.functions[name]
```
eduardoboucas, line 45: I believe `config` is always defined.

```suggestion
    const scheduledFunc = config.functions && config.functions[name]
```

eduardoboucas: sand: You don't know that this is a scheduled function, so I think `scheduledFunc` is not a great name for this variable. Perhaps something like `functionConfig` would be more accurate?

karagulamos: > I believe `config` is always defined.

I added the additional check because a few tests were failing.

eduardoboucas (approved):
🚢 🚢 🚢

PR #4495, opened 2022-03-30
feat: added scheduled functions template for Go

karagulamos:
#### Summary

The PR is part of efforts to [improve functions support for Go](https://github.com/netlify/pillar-runtime/issues/253) and includes a template for Go scheduled functions.


src/functions-templates/go/scheduled-function/Makefile
```diff
@@ -0,0 +1,7 @@
+build:
+	mkdir -p functions
+	GOOS=linux
+	GOARCH=amd64
+	GO111MODULE=on
+	GOBIN=${PWD}/functions go get ./...
+	GOBIN=${PWD}/functions go install ./...
```
mraerino, line 7: is this needed? i thought the CLI was able to build go functions on its own now?

src/functions-templates/go/scheduled-function/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/someone/{{name}}
```
mraerino, line 1: do you want to indicate that they'll likely have to modify this?

src/functions-templates/go/scheduled-function/go.sum
```diff
@@ -0,0 +1,21 @@
...
+github.com/shurcooL/sanitized_anchor_name v1.0.0/go.mod h1:1NzhyTcUVG4SuEtjjoZeVRXNmyL/1OwPU0+IJeTBvfc=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.6.1 h1:hDPOHmpOpP40lSULcqw7IrRb/u7w6RpDC9399XyoNd0=
+github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/urfave/cli/v2 v2.2.0/go.mod h1:SE9GqnLQmjVa0iPEY0f1w3ygNIYcIJ0OKPMoW2caLfQ=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 h1:tQIYjPdBoyREyB9XMu+nnTclpTYkz2zFM+lzLJFO4gQ=
+gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
```
mraerino, line 21: we might not want to commit the `go.sum` and instead create it once the user instantiates the template

src/functions-templates/go/scheduled-function/netlify.toml
```diff
@@ -0,0 +1,12 @@
+[build]
+command = "make build"
+functions = "functions"
+
+[build.environment]
+GO_VERSION = "1.17.6"
```
mraerino, line 6: how will we update this? wouldn't it make more sense to use `1.17` so we always get the latest patch version? (not sure if our build system supports that)

mraerino (commented):
cool stuff

src/functions-templates/go/scheduled-function/go.mod
karagulamos: This `{{name}}` placeholder is automatically replaced by the name configured for the template. I can add a comment if that would make it obvious.

mraerino: yeah i was mostly talking about the `someone` here

src/functions-templates/go/scheduled-function/Makefile
karagulamos: Correct, it should be fine to remove.

src/functions-templates/go/scheduled-function/go.sum
karagulamos: Not sure it works like this at the moment, but we could review this in another PR.

mraerino (approved):
cool! did you make sure to test this end-to-end by trying it locally and then deploying it?

karagulamos:
> cool! did you make sure to test this end-to-end by trying it locally and then deploying it?

Yeah, I did. Works locally and online.

PR #4547, opened 2022-04-18
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:

#### Summary

Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

---

      passes our tests.

Image from iOS (1)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,6 +6,10 @@ coverage:
     javascript:
       enable_partials: true
   status:
-    project: false
-    patch: false
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
 comment: false


danez:
I think this is a good idea. Any other opinions @eduardoboucas @lukasholzer @jackiewmacharia?

lukasholzer (approved):
Oh this is a great catch! Awesome Would love to see actual reports 🥳 

As long this is not blocking and is only informal I think we should give it a try 👍

PR #4553, opened 2022-04-20
fix: report an error whe running site:create nameflag is invalid

### LINKED ISSUE #4503: Too long names are not reported
### Describe the bug

When creating a new site and providing a name via parameter, too long names are ignored, but should be reported instead.

### Steps to reproduce

```
netlify sites:create --account-slug myslug --name aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```

Starts the interactive site creation dialog (max seems to be 63 chars)

### Configuration

_No response_

### Environment

```
  System:
    OS: macOS 12.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 4.34 GB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 17.5.0 - ~/.asdf/installs/nodejs/17.5.0/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.4.1 - ~/.asdf/plugins/nodejs/shims/npm
  npmGlobalPackages:
    netlify-cli: 9.16.0
```

Lxxyx:

#### Summary

Fixes #4503

---

      passes our tests.


src/commands/sites/sites-create.js
```diff
@@ -85,6 +85,19 @@ const sitesCreate = async (options, command) => {
   }
 
   const { name: nameFlag } = options
+
+  // netlify sites:create --name
+  if (nameFlag === true) {
+    error('Please specify site name, example: --name <site_name>')
+    return
+  }
```
danez, line 93: I think instead of this custom code it would be better to change the `commander` configuration at the end of this file to make the argument required:
```diff
-    .option('-n, --name [name]', 'name of site')
-    .option('-a, --account-slug [slug]', 'account slug to create the site under')
+    .option('-n, --name <name>', 'name of site')
+    .option('-a, --account-slug <slug>', 'account slug to create the site under')
```

This won't allow `--name` and  `--account-slug` to be used without argument.

tests/integration/140.command.sites.test.js
```diff
@@ -179,3 +180,29 @@ test.serial('should return an array of templates with name, source code url and
...
+    })
+
+    const program = new BaseCommand('netlify')
+
+    createSitesCreateCommand(program)
+
+    const missingNameError = await t.throwsAsync(async () => {
+      await program.parseAsync(['', '', 'sites:create', '--name'])
+    })
+    t.truthy(missingNameError.message.includes('Please specify site name'))
```
danez, line 200: I do not think we need this test anymore, once we make the arguments mandatory.

As a sidenote: If you write tests it might be better to split the two tests you created into two test functions. Because right now when the first one fails the second one won't be executed at all.

danez (changes requested):
Thanks for your PR.

This looks good to me, only one thing I like to be changed (see inline comment).
Once that is done I can merge.

danez (approved):

PR #4554, opened 2022-04-20
feat: warn about orphaned Edge Function declarations

eduardoboucas:
#### Summary

When a request matches an Edge Function for which there is a config declaration but no matching function file, we should show a warning message to assist developers with the process of creating a function.

Example:

Screenshot 2022-04-20 at 12 43 07


wangoimacharia (approved):
Works great locally!

PR #4566, opened 2022-04-26
feat: add geolocation data to Netlify Dev

eduardoboucas:
#### Summary

Adds geolocation data to Netlify Dev when using Edge Functions. It retrieves the geolocation data from https://netlifind.netlify.app/ and caches it locally for 24h, to avoid making a request every time the app starts.

When the request fails, or when CLI is used with the `--offline` flag, we use a mock location (San Francisco).

A new `--geo` flag is added to the `dev` command with the following values:

- `cache` (default): Uses the geolocation cached locally if it's not older than 24h. Otherwise, it retrieves it from the API.
- `update`: Retrieves geolocation data from the API, even if there is a cached version.
- `mock`: Does not make any requests to the API and uses the mock location.


ascorbic:
It might be good to add a check for user-agent in the API, to stop somebody deciding to use it as a free, unlimited geoip service

eduardoboucas:
> It might be good to add a check for user-agent in the API, to stop somebody deciding to use it as a free, unlimited geoip service

The service will return geolocation data only for the requesting IP, as opposed to accepting an IP address as input, so it will be less useful as a service? Also, I'm not sure that checking the UA will help, since it can be easily spoofed.

ascorbic:
The scenario I'm thinking about is somebody using it in their site to get the visitor's IP. It wouldn;t be spoofable then as it's the user's browser.

ascorbic:
I do forget that CORS exists for purposes other than wasting mornings on broken header configurations.

mraerino (dismissed):
LGTM

still slightly terrified of reviewing non-typescript code 🙈

ascorbic:
It's not quite the same, but `// @ts-check` with JSDoc types are better than nothing.

mraerino (approved):

PR #4567, opened 2022-04-28
feat: add HTTPS support for Edge Functions in Netlify Dev

eduardoboucas:
#### Summary

Fixes an issue where origin requests in Edge Functions will fail when running Netlify Dev with HTTPS enabled.

Builds on top of https://github.com/netlify/edge-functions-bootstrap/pull/30 and https://github.com/netlify/edge-bundler/pull/7 (the tests will fail until both of these are merged and released).

Closes https://github.com/netlify/pillar-runtime/issues/323.


minivan (dismissed):

Skn0tt (commented):
AFAICT, this will require every developer cloning this repository to run `npm run certs` as part of setting up their development environment. Could we use something like https://www.npmjs.com/package/openssl-nodejs (or execa) to generate the certificates as part of the test run? Or is that too resource-intensive?

eduardoboucas:
> AFAICT, this will require every developer cloning this repository to run `npm run certs` as part of setting up their development environment. Could we use something like https://www.npmjs.com/package/openssl-nodejs (or execa) to generate the certificates as part of the test run? Or is that too resource-intensive?

@Skn0tt I've added `certs` to the `test:dev` script in https://github.com/netlify/cli/pull/4567/commits/c43991f49fbd972900099549a43393d9114582c7. I think that should cover it?

Skn0tt:
Yes, that's good 👍

Skn0tt (approved):

PR #4604, opened 2022-05-12
fix: pass correct protocol to functions server

eduardoboucas:
#### Summary

When using Netlify Dev with HTTPS, the `event.rawUrl` property in serverless functions incorrectly references the HTTP URL. This PR fixes that.

Closes https://github.com/netlify/pod-compute/issues/86

### DIFF
--- a/src/lib/functions/server.js
+++ b/src/lib/functions/server.js
@@ -1,4 +1,5 @@
 // @ts-check
+const { get } = require('dot-prop')
 const jwtDecode = require('jwt-decode')
 
 const {
@@ -89,7 +90,7 @@ const createHandler = function (options) {
       {},
     )
     const rawQuery = new URLSearchParams(request.query).toString()
-    const protocol = options.config.dev.https ? 'https' : 'http'
+    const protocol = get(options, 'config.dev.https') ? 'https' : 'http'
     const url = new URL(requestPath, `${protocol}://${request.get('host') || 'localhost'}`)
     url.search = rawQuery
     const rawUrl = url.toString()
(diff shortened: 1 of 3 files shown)


src/lib/functions/server.js
```diff
@@ -89,7 +89,8 @@ const createHandler = function (options) {
       {},
     )
     const rawQuery = new URLSearchParams(request.query).toString()
-    const url = new URL(requestPath, `${request.protocol}://${request.get('host') || 'localhost'}`)
+    const protocol = options.config.dev.https ? 'https' : 'http'
+    const url = new URL(requestPath, `${protocol}://${request.get('host') || 'localhost'}`)
```
mraerino, line 93: so the server doesn't listen on 80 anymore when you enable https?

eduardoboucas: We have multiple servers running: a main server that then forwards requests to a functions server. The main server is listening on 443, the functions server is listening on 80.

This will continue to be the case, all we're changing is the event data that is sent to functions. Rather than using the protocol from the functions server (which will always be 80), we're using the protocol of the main server.

Skn0tt (approved):
lgtm

src/lib/functions/server.js
mraerino: but if you have two servers, you can't really use the config setting right?
since you need to set the protocol based on which server the request came in.
i'd recommend using a HTTP header for this.

eduardoboucas: Only the main/outer server is exposed to customers. The functions server is an internal implementation detail. So the `rawUrl` value of functions should always match the URL of the outer server.

PR #4615, opened 2022-05-18
feat: support for Edge Functions debugging

jackiewmacharia:

#### Summary

Fixes https://github.com/netlify/pillar-runtime/issues/357
Dependent on: https://github.com/netlify/edge-bundler/pull/31

This PR adds flags to run with `netlify dev` to extend `deno --inspect` and `deno --inspect-brk` for Edge Functions debugging. The `--edgeInspect` and `--edgeInspectBrk` have the `-e` and `-E` flags respectively as aliases. Each can take a custom host and port that defaults to the default <HOST:PORT> if none is specified.

To test this out, clone `feat/debug-edge-functions-flags` branch of [edge-bundler](https://github.com/netlify/edge-bundler/tree/feat/debug-edge-functions-flags); run `npm install` and `npm run build` in edge-bundler to set it up; edit CLI package.json to use the cloned version of edge-bundler eg `"@netlify/edge-bundler": "file:../edge-bundler"` and run some of the below examples in an Edge Functions project.

Examples:
```
  $ netlify dev --edgeInspect
  $ netlify dev --edgeInspect=127.0.0.1:9229
  $ netlify dev --edgeInspectBrk
  $ netlify dev --edgeInspectBrk=127.0.0.1:9229
```

---

      passes our tests.

cute animal


src/commands/dev/dev.js
```diff
@@ -522,11 +561,26 @@ const createDevCommand = (program) => {
...
+    .addOption(
+      new Option('-e, --edgeInspect [inspectHostPort]', 'enable Deno inspect with optional inspectHostPort').conflicts(
+        'edgeInspectBrk',
+      ),
+    )
+    .addOption(
+      new Option(
+        '-E, --edgeInspectBrk [inspectHostPort]',
+        'enable Deno inspect-brk with optional inspectHostPort',
+      ).conflicts('edgeInspect'),
```
eduardoboucas, line 573: Nice use of `.conflicts()`! ✨

src/commands/dev/dev.js
```diff
@@ -312,6 +322,32 @@ const startPollingForAPIAuthentication = async function (options) {
...
+ * @param {boolean|string} edgeInspectBrk
+ * @returns {InspectSettings}
+ */
+const generateInspectSettings = (edgeInspect, edgeInspectBrk) => {
+  const enabled = Boolean(edgeInspect) || Boolean(edgeInspectBrk)
+  const pause = Boolean(edgeInspectBrk)
+  const getAddress = () => {
+    if (edgeInspect) {
+      // .slice(1) to remove the = in `=127.0.0.1:9229`
+      return typeof edgeInspect === 'string' ? edgeInspect.slice(1) : undefined
```
eduardoboucas, line 336: I'm confused as to why you need to slice this. Commander is supposed to handle this automatically for you.

I've just pulled your branch, tried using `ntl dev --edgeInspect=127.0.0.1:1234` and `ntl dev --edgeInspect 127.0.0.1:1234` and in both cases `options.edgeInspect` has the right contents, with no need to slice.

Am I missing something?

src/commands/dev/dev.js
```diff
@@ -522,11 +561,26 @@ const createDevCommand = (program) => {
         .hideHelp(),
     )
     .addOption(new Option('--graph', 'enable Netlify Graph support').hideHelp())
+    .addOption(
+      new Option('-e, --edgeInspect [inspectHostPort]', 'enable Deno inspect with optional inspectHostPort').conflicts(
```
eduardoboucas, line 565: I made a couple of suggestions here. First, we're using `<parameter>` instead of `[parameter]`. I also think we can simplify `inspectHostPort` to just `address` and then use the description to explain the format.

Finally, I've reworded things a bit, let me know what you think.

```suggestion
      new Option('-e, --edgeInspect <address>', 'enable the V8 Inspector Protocol for Edge Functions, with an optional address in the host:port format').conflicts(
```

src/commands/dev/dev.js
```diff
@@ -522,11 +561,26 @@ const createDevCommand = (program) => {
...
+        'edgeInspectBrk',
+      ),
+    )
+    .addOption(
+      new Option(
+        '-E, --edgeInspectBrk [inspectHostPort]',
```
eduardoboucas, line 571: Same as above.

```suggestion
        '-E, --edgeInspectBrk <address>,
```

src/commands/dev/dev.js
```diff
@@ -522,11 +561,26 @@ const createDevCommand = (program) => {
...
+        'enable Deno inspect-brk with optional inspectHostPort',
```
eduardoboucas, line 572: Same as above.

```suggestion
        'enable the V8 Inspector Protocol for Edge Functions and pause execution on the first line of code, with an optional address in the host:port format,
```

eduardoboucas (commented):
Left a few comments, but looking great generally! ✨

src/commands/dev/dev.js
danez: I think square brackets indicate an optional argument, whereas `<>` indicates required. At least looking I see this in the docs: https://tj.github.io/commander.js/#Command.prototype.option

eduardoboucas: Ah, you're right! My bad.

src/commands/dev/dev.js
jackiewmacharia: Aha, this was because I was (wrongfully) using the short form that [doesn't support the equal sign](https://github.com/tj/commander.js/issues/1691) ending up with `edgeInspect: '=127.0.0.1:3030'` if `ntl dev -e=127.0.0.1:1234` is used.

I've updated the help flag with more examples on how to use the short forms:
```
'netlify dev -e 127.0.0.1:9229',
'netlify dev -e127.0.0.1:9229',
'netlify dev -E 127.0.0.1:9229',
'netlify dev -E127.0.0.1:9229',
```
I don't know if this is sufficient though because the error that pops up when `-e=127.0.0.1:1234` or `-E=127.0.0.1:1234` is ran is what's below from deno. The `--help` flag suggestion will bring up the above examples though when ran with `netlify dev`.  🤔

```
error: Invalid value for '--inspect=<HOST:PORT>...': invalid IP address syntax

For more information try --help

Caused by:
    invalid IP address syntax
```

src/commands/dev/dev.js
```diff
@@ -522,11 +559,33 @@ const createDevCommand = (program) => {
...
+      'netlify dev -e',
+      'netlify dev -e 127.0.0.1:9229',
+      'netlify dev -e127.0.0.1:9229',
+      'netlify dev --edgeInspect',
+      'netlify dev --edgeInspect=127.0.0.1:9229',
+      'netlify dev -E',
+      'netlify dev -E 127.0.0.1:9229',
+      'netlify dev -E127.0.0.1:9229',
+      'netlify dev --edgeInspectBrk',
+      'netlify dev --edgeInspectBrk=127.0.0.1:9229',
```
eduardoboucas, line 588: [dust] I don't think we need to add an example for every single permutation, otherwise we'll end up with a huge list in the help command. How do you feel about keeping just the ones below?

```suggestion
      'netlify dev --edgeInspect',
      'netlify dev --edgeInspect=127.0.0.1:9229',
      'netlify dev --edgeInspectBrk',
      'netlify dev --edgeInspectBrk=127.0.0.1:9229',
```

eduardoboucas (approved):
Great work! Left a minor, non-blocking comment.

src/commands/dev/dev.js
jackiewmacharia: I added the short form ones just in case someone uses the `=` sign with `-e` or `-E` and it fails, they see an example of how go use them with the `--help` flag. This is because something like `-e=127.0.0.1:1234` gives us `edgeInspect: '=127.0.0.1:3030'` because short forms don't support the equal sign causing it to fail.

eduardoboucas: You can do this by specifying a custom parser. See https://github.com/tj/commander.js/#custom-option-processing.

jackiewmacharia: @eduardoboucas this is now done, the review also got reset.

src/commands/dev/dev.js
```diff
@@ -312,6 +323,46 @@ const startPollingForAPIAuthentication = async function (options) {
...
+
+  return {
+    enabled,
+    pause,
+    address: getAddress(),
+  }
+}
+
+const validateShortFlagArgs = (args) => {
+  if (args.includes('=')) {
```
eduardoboucas, line 351: [dust] Not a big deal, but the particular case of people using  short form flag with a `=` would only happen if the arg _starts_ with a `=`, not when it exists elsewhere in the arg?

```suggestion
  if (args.startsWith('=')) {
```

eduardoboucas (approved):
Great work! 🚀

eduardoboucas (approved):

PR #4619, opened 2022-05-20
fix: correctly watch function files specified in the config option `included_files`

### LINKED ISSUE #4454: Changes to Included Files should restart/rebuild function
### Describe the bug

When developing a function that uses the `included_files` feature, those files are bundled into the function for access. However, when developing the function, I would go back and forth between editing the included file and the function itself. Only when I updated the included function did it rebuild.

### Steps to reproduce

Create a function that uses the functions.included_files toml config like below:

```
[functions]
  included_files = ["templates/*.*"]
```

When in netlify dev, updates to the matched files will not be presented unless the function is touched to trigger a rebuild/restart.

### Configuration

```
[functions]
  included_files = ["templates/*.*"]
```

### Environment

```
System:
    OS: macOS 12.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 60.16 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.1 - /opt/homebrew/opt/nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - /opt/homebrew/opt/nvm/versions/node/v16.13.1/bin/yarn
    npm: 8.1.2 - /opt/homebrew/opt/nvm/versions/node/v16.13.1/bin/npm
  npmGlobalPackages:
    netlify-cli: 9.13.0
```

danez:
#### Summary

Fixes #4454

Adds the files/globs from the `included_files` config option to the watcher.

This requires a new version of zisi with netlify/zip-it-and-ship-it#1098 , which is not released yet.

### DIFF
--- a/src/lib/functions/registry.js
+++ b/src/lib/functions/registry.js
@@ -76,7 +76,7 @@ class FunctionsRegistry {
       log(`${NETLIFYDEVLOG} ${chalk.magenta('Reloading')} function ${chalk.yellow(func.name)}...`)
     }
 
-    const { error_, srcFilesDiff } = await func.build({ cache: this.buildCache })
+    const { error_, includedFiles, srcFilesDiff } = await func.build({ cache: this.buildCache })
 
     if (error_) {
       log(
@@ -113,7 +113,9 @@ class FunctionsRegistry {
     // If there is no watcher for this function but the build produced files,
     // we create a new watcher and watch them.
     if (srcFilesDiff.added.size !== 0) {
-      const newWatcher = await watchDebounced([...srcFilesDiff.added], {
+      const filesToWatch = [...srcFilesDiff.added, ...includedFiles]
+
+      const newWatcher = await watchDebounced(filesToWatch, {
         onChange: () => {
           this.buildFunctionAndWatchFiles(func, { verbose: true })
         },
--- a/tests/integration/330.serving-functions.test.js
+++ b/tests/integration/330.serving-functions.test.js
@@ -951,7 +951,8 @@ test('Ensures watcher watches included files', async (t) => {
         .buildAsync()
 
       // wait for the watcher to rebuild the function
-      await pause(500)
+      const delay = 1000
+      await pause(delay)
 
       t.true(outputBuffer.some((buffer) => /.*Reloaded function hello.*/.test(buffer.toString())))
       await tryAndLogOutput(async () => {
--- a/tests/integration/330.serving-functions.test.js
+++ b/tests/integration/330.serving-functions.test.js
@@ -891,4 +891,73 @@ test('Populates the `event` argument', async (t) => {
     })
   })
 })
+
+test('Ensures watcher watches included files', async (t) => {
+  await withSiteBuilder('function-with-included-files-watch', async (builder) => {
+    await builder
+      .withContentFiles([
+        {
+          path: 'files/one.json',
+          content: `{"data": "one"}`,
+        },
+        {
+          path: 'files/two.json',
+          content: `{"data": "two"}`,
+        },
+      ])
+      .withFunction({
+        path: 'hello.js',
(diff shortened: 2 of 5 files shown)


src/lib/functions/registry.js
```diff
@@ -113,7 +113,18 @@ class FunctionsRegistry {
...
+      const config = func.config.functions
+      const filesToWatch = [
+        ...srcFilesDiff.added,
+        // Include the manually specified files from config. These files might not be
+        // source files, but might still be used/loaded in the function and therefore
+        // we need to watch them to rebuild once they change
+        ...((config && config[func.name] && config[func.name].included_files) || []).filter(
+          (glob) => !glob.startsWith('!'),
+        ),
+        ...((config && config['*'] && config['*'].included_files) || []).filter((glob) => !glob.startsWith('!')),
```
danez, line 125: These two lines really shouldn't be here, but all the alternatives were worse imho.
Moving the logic to zip-it-ship-it,and adding the globs to the result of the build command, seemed slightly wrong, because the result names the files src files, which these files really aren't necessary.
Adding a new node api to do this in zisi also seems wrong because its api shouldn't do config management.

Ideally there should be a package `@netlify/config` that handles all this. Then cli and zisi (and potentially other packages) could use it.

src/lib/functions/registry.js
```diff
@@ -113,7 +113,18 @@ class FunctionsRegistry {
...
     // we create a new watcher and watch them.
     if (srcFilesDiff.added.size !== 0) {
-      const newWatcher = await watchDebounced([...srcFilesDiff.added], {
+      const config = func.config.functions
+      const filesToWatch = [
+        ...srcFilesDiff.added,
+        // Include the manually specified files from config. These files might not be
+        // source files, but might still be used/loaded in the function and therefore
+        // we need to watch them to rebuild once they change
+        ...((config && config[func.name] && config[func.name].included_files) || []).filter(
```
eduardoboucas, line 122: Hmm, like you said I don't love the fact that we're duplicating the logic we have in zip-it-and-ship-it. Maybe the data coming back from [`listFunctions`](https://github.com/netlify/cli/blob/eb3f87d3dc842388258d580cff04b7cc9dacb87a/src/lib/functions/registry.js#L186-L192) already contains the resolved list of files? If not, perhaps we could add a new property with that information?

tests/integration/330.serving-functions.test.js
```diff
@@ -891,4 +891,73 @@ test('Populates the `event` argument', async (t) => {
...
+            content: `{"data": "four"}`,
+          },
+        ])
+        .buildAsync()
+
+      // wait for the watcher to rebuild the function
+      const timeout = 500
+      await new Promise((resolve) => {
+        setTimeout(resolve, timeout)
+      })
```
eduardoboucas, line 957: [dust] You can use [the `pause()` helper](https://github.com/netlify/cli/blob/c8281ec992c958a053d297b21894a0b8f137cd2b/tests/integration/utils/pause.js).

```suggestion
      await pause(timeout)
```

tests/integration/330.serving-functions.test.js
```diff
@@ -891,4 +891,73 @@ test('Populates the `event` argument', async (t) => {
...
+
+      t.true(outputBuffer.some((buffer) => /.*Reloaded function hello.*/.test(buffer.toString())))
```
eduardoboucas, line 959: Is it worth also asserting that the output of the function reflects the updated files?

eduardoboucas (approved):
Looks great! ✨

tests/integration/330.serving-functions.test.js
danez: done

eduardoboucas (dismissed):

Issue #4625, opened 2022-05-25
Flaky tests umbrella ☂️


danez:
Our tests are failing from time to time with different errors. These errors are unrelated to the actual changes in PRs.

As a first step, I thought it would be a good idea to collect these errors.

- Add errors as separate comments.
- Add logs where this happened to the comment. (I guess maximum 5 or so)

danez:
> Error: listen EADDRINUSE: address already in use :::\<PORT\>

https://github.com/netlify/cli/runs/6597781685?check_suite_focus=true

danez:
> yarn → should install the cli and run the help command
  ✖ Timed out while running tests

https://github.com/netlify/cli/runs/6598249957?check_suite_focus=true

danez:
> EBUSY: resource busy or locked, rmdir ...

Windows

https://github.com/netlify/cli/runs/6612669436?check_suite_focus=true

danez:
> Error: EPERM: operation not permitted, rename 'C:\\Users\\runneradmin\\AppData\\Roaming\\netlify\\Config\\config.json.3515056301' -> 'C:\\Users\\runneradmin\\AppData\\Roaming\\netlify\\Config\\config.json'`,

Windows

https://github.com/netlify/cli/runs/6612669436?check_suite_focus=true

JWhist:
Eleventy test frequently fails with connection refused error

<img width="1004" alt="Screen Shot 2022-06-29 at 12 24 54 PM" src="https://user-images.githubusercontent.com/63815686/176487661-81f0a090-07df-46ea-a0ff-14ad79e599e7.png">

danez:
Closing, it is a lot better now, and we should handle the ones that come up.

PR #4637, opened 2022-05-27
chore: use verdaccio memory plugin to speed up e2e tests

danez:
#### Summary

Refs #4625 

Use memory plugin in verdaccio, to speed up e2e tests and make them less likely timeout.
The memory plugin stores the packages in memory, instead of disk.


tools/e2e/setup.mjs
```diff
@@ -88,26 +95,21 @@ export const startRegistry = async () => {
  * }
  */
 export const setup = async () => {
-  const { storage, url } = await startRegistry()
+  const { url } = await startRegistry()
   const workspace = await mkdtemp(`${tmpdir()}${sep}e2e-test-`)
 
-  const npmrc = join(homedir(), '.npmrc')
+  const npmrc = fileURLToPath(new URL('../../.npmrc', import.meta.url))
```
lukasholzer, line 101: the `.npmrc` file has to be always in the homedir and depending on your checkout location this will not work

tools/e2e/setup.mjs
```diff
@@ -88,26 +95,21 @@ export const startRegistry = async () => {
...
   const registryWithAuth = `//${url.hostname}:${url.port}/:_authToken=dummy`
   let backupNpmrc
 
   /** Cleans up everything */
   const cleanup = async () => {
-    // restore ~/.npmrc
```
lukasholzer, line 100: As we modify the npmrc file we need to clean it up again after everything was done or an error occured

tools/e2e/setup.mjs
danez: This is the location of the `.npmrc` file that is committed to the repository. During publishing this adds the auth token to it.

tools/e2e/setup.mjs
danez: see other comment

danez:
Okay I reworked it, it is now caching the verdaccio storage between builds.

tools/e2e/setup.mjs
```diff
@@ -91,23 +98,18 @@ export const setup = async () => {
...
-      writeFileSync(npmrc, backupNpmrc)
-    } else {
-      await rmdirRecursiveAsync(npmrc)
-    }
     // remote temp folders
-    await rmdirRecursiveAsync(storage)
     await rmdirRecursiveAsync(workspace)
   }
 
+  env.npm_config_registry = url
```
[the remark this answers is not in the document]
lukasholzer: Does this work on linux as well? Had some issues on windows and linux when I recall correctly?

lukasholzer (approved):
Thanks for reworking looks good to me! 🐳

tools/e2e/setup.mjs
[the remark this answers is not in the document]
danez: `execa` which we use to call the package managers, does automatically inherit all environment variables.

the weird part is: as we call the e2e tests through an npm script this environment variable is already set, but with the proper npmjs registry. This is why I guess in the past we had to use `--registry` when calling the package managers, as the environment variable always takes precedence over any other config format besides cli options.

.github/workflows/e2e-test.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    types: [synchronize, opened, reopened, edited]
```
danez, line 7: I changed this, as the e2e tests would run whenever the main body or title was edited, which I don't think makes sense. I saw that `edited` is also triggered when the base branch changes, but I guess that is a less common scenario and in that case kodiak would merge anyway.

lukasholzer (approved):

PR #3904, opened 2021-12-23
feat: add Netlify Graph support

sgrove:
#### Summary

Basic, preliminary support for Netlify Graph

PR #4652, opened 2022-05-30
fix: remove unneeded graphql dependency

danez:
The dependency is never directly used and is also not a peer dependency of anything

This was added in #3904, but never used from the beginning afaics. It seems all Graphql stuff goes through `netlify-onegraph-internal`, which has a dependency on graphql.


sgrove:
Handled a later PR

PR #4658, opened 2022-06-01
fix: fix scheduled function template to use async

danez:
#### Summary

Without the async keyword, typescript instantly errors as the handler (according to the types
of `schedule()` ) needs to return a Promise.

I also tried deploying a function that is not async and lambda throws an error.


eduardoboucas:
oof, nice catch! 👏🏻

PR #4660, opened 2022-06-02
chore: remove Go scheduled functions template

karagulamos:
#### Summary

I'm removing the Go scheduled functions template from the CLI because it requires additional effort to use and/or deploy. For now, users can create/scaffold regular Go functions and turn them into scheduled functions via `netlify.toml`.


danez:
Just out of interest, what is the additional effort that is needed?

eduardoboucas (approved):

PR #4685, opened 2022-06-09
fix: gracefully fail Edge Functions in case of deno binary issues

### LINKED ISSUE #4621: [Android] Netlify dev failed with redirects/edge functions
### Describe the bug

```
Could not read downloaded binary
    at DenoBridge.downloadBinary (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/bridge.js:33:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async DenoBridge.getBinaryPath (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/bridge.js:116:32)
    at async Module.serve (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/server/server.js:52:5)
    at async prepareServer (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/src/lib/edge-functions/proxy.js:142:22)
Error: Could not read downloaded binary
    at DenoBridge.downloadBinary (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/bridge.js:33:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async DenoBridge.getBinaryPath (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/bridge.js:116:32)
    at async Module.serve (file:///data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/@netlify/edge-bundler/dist/server/server.js:52:5)
    at async prepareServer (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/src/lib/edge-functions/proxy.js:142:22)

/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/netlify-redirector/lib/redirects.js:116
      throw ex;
      ^
abort({}) at Error
    at jsStackTrace (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/netlify-redirector/lib/redirects.js:1070:13)
    at stackTrace (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/netlify-redirector/lib/redirects.js:1087:12)
    at process.abort (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/netlify-redirector/lib/redirects.js:8502:44)
    at process.emit (node:events:527:28)
    at processEmit [as emit] (/data/data/com.termux/files/usr/lib/node_modules/netlify-cli/node_modules/signal-exit/index.js:199:34)
    at emit (node:internal/process/promises:140:20)
    at processPromiseRejections (node:internal/process/promises:274:27)
    at processTicksAndRejections (node:internal/process/task_queues:97:32)
(Use `node --trace-uncaught ...` to show where the exception was thrown)
```

Here's the logs. I was trying to just get a dev environment running on my phone and it's not working.

### Steps to reproduce

```bash
netlify dev
```

### Configuration

```toml
[build]
publish = "public"
functions = "functions"

[[redirects]] 
from = "/api/*"
to= "/.netlify/functions/:splat"
status = 200

[dev]
port = 2222
```

### Environment

```
System:
    OS: android
    CPU: Unknown
    Memory: 1003.25 MB / 3.66 GB
  Binaries:
    Node: 17.9.0 - /data/data/com.termux/files/usr/bin/node
    npm: 8.10.0 - /data/data/com.termux/files/usr/bin/npm
  npmGlobalPackages:
    netlify-cli: 10.3.0
```

jackiewmacharia:

#### Summary

Fixes #4621 

Requires netlify/edge-bundler#42

`netlify dev` breaks in case one runs it on a platform (e.g linux aarch64) not yet officially supported by deno, to avoid this we need to gracefully fail Edge Functions making it possible to continue using `netlify dev`.

---

To test this:
- Clone `fix/add-error-state-to-after-download` branch of [edge-bundler](https://github.com/netlify/edge-bundler/tree/fix/add-error-state-to-after-download); run `npm install` and `npm run build` in edge-bundler to set it up; edit CLI's package.json to use the cloned version of edge-bundler eg `"@netlify/edge-bundler": "file:../edge-bundler" `
- Break the deno installation process by making the following changes in edge-bundler:
    - comment out this [block of code](https://github.com/netlify/edge-bundler/blob/main/src/bridge.ts#L164-L178) to force the download step to get called each time
    -  change [this line](https://github.com/netlify/edge-bundler/blob/main/src/bridge.ts#L67) to `if (downloadedVersion !== undefined) {` to cause the binary download step to throw an error.
    - change [this line](https://github.com/netlify/edge-bundler/blob/main/src/bridge.ts#L71) to `await this.writeVersionFile(downloadedVersion || '')` to get past TypeScript type check errors
    - run `npm run build` to make changes available CLI
- In a project use the local CLI to run the dev command e.g `~/netlify/cli/bin/run dev`
- It should surface the error without breaking rest of dev command process, test this by accessing a serverless function

Error string for review: `› There was a problem setting up the Edge Functions environment and it's unfortunately not possible to run Edge Functions from the CLI on this platform. More on supported platforms here: https://deno.land/manual/getting_started/installation. Error: <error.message from error object>`

Screenshot:

Screenshot 2022-06-09 at 18 51 34

      passes our tests.

cute animal


danez (dismissed):
works 🌻
Screenshot 2022-06-10 at 14 29 16

src/lib/edge-functions/proxy.js
```diff
@@ -141,34 +148,49 @@ const prepareServer = async ({
...
+      getUpdatedConfig,
+      internalFunctions,
+      projectDir,
+      runIsolate,
+    })
 
-  return { registry, runIsolate }
+    return { registry, runIsolate }
+  } catch (error) {
+    log(
```
eduardoboucas, line 182: I think we're assuming that this error will be due to an incompatible platform being used with Deno, but we don't know if that's the case. We're wrapping a fair amount of logic with this `try/catch`, so an error could come from different places, and we risk showing an error message about an unsupported platform to someone running edge functions on a supported platform, which will add to the confusion as opposed to help.

It feels to me that we want to address two separate things:

1. Allow the edge functions initialisation to fail without crashing CLI
2. Showing a useful error message in case people try to use edge functions with an unsupported platform

For point 1, I think a `try/catch` is indeed the right option — although personally I would do it on the `initializeProxy` side and not on `prepareServer`, because as it stands `prepareServer` is a Promise that _resolves_ an error (as opposed to rejecting).

For point 2, I would actually make CLI check the platform being used and output a specific message in that case. This way we can guarantee that we only show that message to people that indeed have an unsupported platform.

src/lib/edge-functions/proxy.js
```diff
@@ -141,34 +148,49 @@ const prepareServer = async ({
...
+      `${chalk.red(
+        BANG,
+      )} There was a problem setting up the Edge Functions environment and it's unfortunately not possible to run Edge Functions from the CLI on this platform. More on supported platforms here: https://deno.land/manual/getting_started/installation. Error: ${
```
eduardoboucas, line 185: We try to generate a `ntl.fyi` shortlink for user-facing links.

minivan: is `ntl.fyi/install-deno` good enough?

eduardoboucas: Yeah! 🚀

src/lib/edge-functions/proxy.js
jackiewmacharia: Point 1: I've moved try/catch to `initializeProxy` but it results in the the error being logged with every request made by cli (screenshot below)

Point 2: I've moved error message to [edge-bundler](https://github.com/netlify/edge-bundler/pull/42) so we only log `error.message`.

Screenshot 2022-06-14 at 16 30 46

src/lib/edge-functions/proxy.js
minivan: asked the marketing team [here](https://netlify.slack.com/archives/C1EENQ18U/p1655214472502759)

src/lib/edge-functions/proxy.js
eduardoboucas: > Point 1: I've moved try/catch to `initializeProxy` but it results in the the error being logged with every request made by cli (screenshot below)

I don't know how you've implemented that, but we should be able to keep track of whether we've already printed the error message or not?

```js
const initializeProxy = async (...) => {
  const server = prepareServer(...)

  let hasServerError = false

  return async (req) => {
    if (req.headers[headers.Passthrough] !== undefined || !hasEdgeFunctions || hasServerError) {
      return
    }

    try {
      // Get geolocation and server
    } catch (error) {
      // print error

      hasServerError = true
    }
}
```

This would also avoid calling the geolocation logic on subsequent requests if the server has errored, but it's mostly about respecting the error handling patterns, since `prepareServer` could throw an error when something fails, as opposed to resolving with an error object.

Not a blocker though, so I'll defer to you whether to do this or not.

eduardoboucas (dismissed):

src/lib/edge-functions/proxy.js
```diff
@@ -74,56 +78,62 @@ const initializeProxy = async ({
...
-      return
-    }
-
-    req[headersSymbol] = {
-      [headers.Functions]: functionNames.join(','),
-      [headers.ForwardedHost]: `localhost:${mainPort}`,
-      [headers.Passthrough]: 'passthrough',
-      [headers.RequestID]: generateUUID(),
-      [headers.IP]: LOCAL_HOST,
+    try {
```
eduardoboucas, line 88: I'm sorry if this wasn't clear in my previous comment, but I was trying to suggest that you only wrap the `Promise.all` call in the `try/catch`, not the entire block. Otherwise we're back to the problem of wrapping too much code and potentially swallowing errors.

Something like:

```ts
let ops

try {
  ops = await Promise.all([
    getGeoLocation({ mode: geolocationMode, offline, state }),
    server,
  ])
} catch (err) {
  error(err, { exit: false })

  return
}

const [geoLocation, { registry }] = ops
```

jackiewmacharia: Aaah I see, done.

src/lib/edge-functions/proxy.js
```diff
@@ -74,16 +78,27 @@ const initializeProxy = async ({
...
+    let promiseResult
+
+    try {
+      promiseResult = await Promise.all([getGeoLocation({ mode: geolocationMode, offline, state }), server])
+    } catch (error_) {
+      error(error_ instanceof Error ? error_ : format(error_), { exit: false })
+      hasServerError = true
+    }
+
+    if (promiseResult === undefined) {
```
eduardoboucas, line 97: This only happens if we fall into the `catch` above, right? Can we move it there? I think it would be a bit easier to read.

src/lib/edge-functions/proxy.js
```diff
@@ -74,16 +78,27 @@ const initializeProxy = async ({
...
-    const [geoLocation, { registry }] = await Promise.all([
-      getGeoLocation({ mode: geolocationMode, offline, state }),
-      server,
-    ])
+    let promiseResult
+
+    try {
+      promiseResult = await Promise.all([getGeoLocation({ mode: geolocationMode, offline, state }), server])
+    } catch (error_) {
+      error(error_ instanceof Error ? error_ : format(error_), { exit: false })
```
eduardoboucas, line 93: Can you add some context to why we need this check and the `format` call here? The `error` help is [prepared to handle an `Error` object](https://github.com/netlify/cli/blob/main/src/utils/command-helpers.js#L180).

eduardoboucas (commented):
It's looking great! Left just a couple more questions.

jackiewmacharia:
@eduardoboucas  this one is now updated with edge-bundler changes and is now ready.

eduardoboucas (commented):
Great work!

PR #4689, opened 2022-06-10
Jj/add edge func template

jenae-janzen:

#### Summary

Fixes #<[99](https://github.com/netlify/pod-compute/issues/99)>

Adds edge function templates to `functions:create` command.

Opening as a draft PR because there are still a few things to change, but I wanted to get some feedback at this point and make sure I'm on track.

At this point:
[X] When a user runs `functions:create` they're prompted to choose between edge and serverless functions
[X] When the user selects edge functions, we check for and create the netlify/edge-functions directory
[X] The user can select JS or TS and see a list of available edge function template
[X] The template a user chooses is cloned into their edge functions directory
[X] The edge function is appended to the end of their `netlify.toml` file 

To do:
[] Right now, the edge function templates need a bit of tweaking -- the edge function examples[ here](https://github.com/netlify/edge-functions-examples/tree/main/netlify/edge-functions) aren't quite compatible with how templates are scaffolded, so I only added one simple example for each 
[] Currently, there aren't any URL's edge functions can be cloned from -- handled the error in the `downloadFromUrl` function, but they're still given the option when choosing the templates
[] Need to verify the edge functions directory and netlify.toml get created in correct place
[] Need to write tests to make sure everything works as expected and nothing else broke, make sure all of the flags still work, etc
[] Update docs?

---

      passes our tests.


jenae-janzen:
Sorry this PR's a bit of a mess 😅  There are a lot of changes I made + changes from main. If it's too difficult to review, I can start over with a fresh PR or walk through the code changes involved!

PR #4698, opened 2022-06-13
feat: check for deno vs code ext and download when prompted

JWhist:

#### Summary

Fixes # https://github.com/netlify/pillar-runtime/issues/356

Adds a check on first run to see if user has Deno VS Code extension installed, and if not, prompts them to ask if they would like to install it.  If they select yes, it installs the Deno VS Code Extension.  This helps with some syntax highlighting.

---

Looking for a bit of help in testing this.  I uninstalled my local version a few times and tried that way, but looking for advice on what would be a standard way to do this?  Maybe a container or something?  Do we have a docker image we use for such cases?

      passes our tests.


eduardoboucas (changes requested):

src/recipes/vscode/index.js
```diff
@@ -44,15 +75,15 @@ const run = async ({ config, repositoryRoot }) => {
     return
   }
 
+  if (!(await hasDenoVSCodeExt())) {
```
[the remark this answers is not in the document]
JWhist: Sounds good to me.  What is going on is that the tests are failing because VSCode is not installed on the github actions runners, and so the execa call errors out.  I suppose there is a scenario where a user could have a .vscode folder but not actually have it installed and it triggers the recipe to run?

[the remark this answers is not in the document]
eduardoboucas: > I guess my question is, if they do not have VSCode installed, why should I still display a helpful message guiding them to the Deno Extension?

If this code is running, it means that either we detected that the terminal is VSCode's or the user typed the `vscode` recipe.

In both cases, there are high chances that they do have VSCode installed and the error is due to something else: the `code` binary is not in the shell path, we failed to parse the output, etc.

But even if they don't have it installed, pointing them to the extension feels more helpful than just saying "there was an error", because the URL they'll land on will eventually tell them that they need VSCode installed and offer them the option to do it.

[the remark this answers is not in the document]
eduardoboucas: As for the tests, you can mock execa if needed. We do it in some tests (you can search for `execa` in `*.test.js` files).

src/recipes/vscode/index.js
```diff
@@ -45,14 +67,22 @@ const run = async ({ config, repositoryRoot }) => {
...
-
-    log(`${NETLIFYDEVLOG} VS Code settings file ${fileExists ? 'updated' : 'created'}.`)
+    if (!(await hasDenoVSCodeExt())) {
+      const { confirm: denoExtConfirm } = await getDenoExtPrompt()
+      if (denoExtConfirm) getDenoVSCodeExt()
+    }
+  } catch {
     log(
-      `${NETLIFYDEVWARN} If you don't have the Deno VS Code extension, install it by visiting ${chalk.blue(
+      `${NETLIFYDEVWARN} Unable to install Deno VS Code extension, install it by visiting ${chalk.blue(
```
eduardoboucas, line 76: Small change to make this a bit more friendly.

```suggestion
      `${NETLIFYDEVWARN} Unable to install Deno VS Code extension. To install it manually, visit ${chalk.blue(
```

src/recipes/vscode/index.js
```diff
@@ -27,6 +29,26 @@ const getEdgeFunctionsPath = ({ config, repositoryRoot }) =>
 
 const getSettingsPath = (repositoryRoot) => join(repositoryRoot, '.vscode', 'settings.json')
 
+const hasDenoVSCodeExt = async () => {
+  const { stdout: extensions } = await execa('code', ['--list-extensions'])
+  return extensions.split('\n').includes('denoland.vscode-deno')
+}
+
+const getDenoVSCodeExt = async () => {
+  await execa('code', ['--install-extension', 'denoland.vscode-deno']).stdout.pipe(process.stdout)
```
eduardoboucas, line 38: Should we also pipe `stderr`?

eduardoboucas (commented):
Looks great! Would you like to add some tests?

src/recipes/vscode/index.js
danez: I think we do not even need to manually pipe stdout (or stderr for this matter) as this is the default behavior of `execa`. At least by looking at the options: https://github.com/sindresorhus/execa#stdout-1

So if we remove the `stdout.pipe(...)` part does it still work?

JWhist: In my local testing, if I remove the pipe commands it does not log anything, I would have to save the stdout from the response like const { stdout } = await execa... then console.log(stdout) etc.  Which is a roundabout way of doing the same thing as piping it imo.

eduardoboucas: I think the default behaviour is to capture the output into the return value of the `execa` call. If you also want to pipe it into stdout/stderr, you need to make that explicit.

src/recipes/vscode/index.js
```diff
@@ -27,6 +28,26 @@ const getEdgeFunctionsPath = ({ config, repositoryRoot }) =>
...
+  const { stdout: extensions } = await execa('code', ['--list-extensions'], { stderr: 'inherit' })
+  return extensions.split('\n').includes('denoland.vscode-deno')
+}
+
+const getDenoVSCodeExt = async () => {
+  await execa('code', ['--install-extension', 'denoland.vscode-deno'], { stdio: 'inherit' })
+}
+
+const getDenoExtPrompt = () => {
+  const message = 'The Deno VSCode extension is recommended. Would you like to install it now?'
```
eduardoboucas, line 41: I think this is the correct spelling, and what we've been using elsewhere in this file.

```suggestion
  const message = 'The Deno VS Code extension is recommended. Would you like to install it now?'
```

tests/integration/snapshots/600.framework-detection.test.js.md
```diff
@@ -242,4 +242,4 @@ Generated by [AVA](https://avajs.dev).
     > node -p process.env.GATSBY_LOGGER␊
     ␊
     yurnalist␊
-    ◈ "npm run develop" exited with code *. Shutting down Netlify Dev server`
+    .◈ "npm run develop" exited with code *. Shutting down Netlify Dev server`
```
eduardoboucas, line 245: What is this about?

src/recipes/vscode/index.js
```diff
@@ -27,6 +28,26 @@ const getEdgeFunctionsPath = ({ config, repositoryRoot }) =>
 
 const getSettingsPath = (repositoryRoot) => join(repositoryRoot, '.vscode', 'settings.json')
 
+const hasDenoVSCodeExt = async () => {
+  const { stdout: extensions } = await execa('code', ['--list-extensions'], { stderr: 'inherit' })
```
eduardoboucas, line 32: I'm curious why you've used `inherit` here and on L37?

JWhist: I want to save the data from the extensions if the stdout is ok, but log the error to the screen if it is going to stderr.

tests/integration/snapshots/600.framework-detection.test.js.md
JWhist: A flaky test that I was trying to get around, there is a helper function that concats a series of loading dots .... down to a single dot.  But occasionally the load happens fast enough to not have any dots at all (at least that is my theory).  I was toying with the regex in the helper but gave up on it.  This change was actually an error, since it's an autogenerated file anyway, and isn't going to help me pass the test at all.

JWhist: Not necessary, I have another PR going to fix that test

eduardoboucas: Okay, but why would we merge this change? We should revert and let the other PR fix the test?

JWhist: Ok, that makes sense.  Sorry, I'm still in weekend mode!

eduardoboucas (approved):
Nice one! It'd be great to add some tests, even if that happens on a follow up PR. Have you given any thought to how you'd approach that?

PR #4699, opened 2022-06-13
feat: Allow user to configure an editor to open file for Graph

sgrove:
#### Summary

When generating a handler / source file via Graph, the user may indicate they want to open it. This change enables that (see preview video in the appropriate channel)

Incidentally fixes some issues with the data structure of Graph cli session events as well

---


src/lib/one-graph/cli-client.js
```diff
@@ -282,24 +283,68 @@ const updateGraphQLOperationsFileFromPersistedDoc = async (input) => {
...
+        return
+      }
+
+      const config = loadNetlifyGraphConfig()
+      for (const file of files) {
+        const fileWrittenEvent = {
+          __typename: 'OneGraphNetlifyCliSessionFileWrittenEvent',
+          cliSessionId: sessionId,
+          payload: {
+            editor: config.editor,
```
anmonteiro, line 316: I suggest we use `config.editor || process.env.EDITOR`

src/lib/one-graph/cli-client.js
```diff
@@ -282,24 +283,68 @@ const updateGraphQLOperationsFileFromPersistedDoc = async (input) => {
...
+    }
+    case 'OneGraphNetlifyCliSessionOpenFileEvent': {
+      if (!payload.filePath) {
+        warn(`No filePath found in payload, ${JSON.stringify(payload, null, 2)}`)
+        return
+      }
+      const config = loadNetlifyGraphConfig()
+      if (config.editor) {
+        log(`Opening ${config.editor} for ${payload.filePath}`)
+        execa(config.editor, [payload.filePath], {
```
anmonteiro, line 338: Is it important to set some environment variables for the editor, or is it fine to inherit from the current process's env?

anmonteiro (approved):
Couple minor comments

src/lib/one-graph/cli-netlify-graph.js
```diff
@@ -520,6 +529,21 @@ const getGraphEditUrlBySiteId = ({ oneGraphSessionId, siteId }) => {
   return url
 }
 
+/**
+ * Load `netlifyGraph.json` from the appropriate location
+ * @returns {import('netlify-onegraph-internal').NetlifyGraphJsonConfig.JsonConfig}
+ */
+const loadNetlifyGraphConfig = () => {
+  const configPath = path.join(process.cwd(), 'netlifyGraph.json')
```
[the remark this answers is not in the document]
sgrove: I _think_ `siteRoot` (`site.root`) is probably more appropriate in the case of e.g. nested netlify sites.

src/lib/one-graph/cli-client.js
sgrove: I think we want to inherit the same env as the cli (which is presumably set up to work on the current site)

src/lib/one-graph/cli-client.js
sgrove: I'd like to avoid too much magic for the first iteration, but a good idea for down the road if this turns out to test well with users.

anmonteiro (approved):

PR #4708, opened 2022-06-16
feat: add user assigned mock geo for edge local dev

JWhist:

#### Summary

Adds feature [https://github.com/netlify/pillar-runtime/issues/321](https://github.com/netlify/pillar-runtime/issues/321)

Give users the ability to pass a two letter country code via the `--country` flag to mock geoLocation data in local dev environment, or when it cannot be mocked via VPN.

---

      passes our tests.
- [X] Update or add tests (if any source code was changed or added) 🧪
- **I did not add any new tests, as this should work as is with the existing test which uses the same exported value**
- [X] Update or add documentation (if features were changed or added) 📝


eduardoboucas (commented):
Asking people to place a file with a specific format on a specific location feels like a lot of trouble to go through just to populate the `context.geo` property, in my opinion.

I think we should try to look for a simpler alternative.

For example, we already use https://netlifind.netlify.app/ to retrieve the full geolocation object for the visiting IP. What if we extended it to accept an optional country parameter, and when present it'll return a location somewhere in that country?

So if you hit https://netlifind.netlify.app/?country=us you'd get a response like:

```json
{
  "city": "San Francisco",
  "country": {
    "code": "US",
    "name": "United States"
  },
  "subdivision": {
    "code": "CA",
    "name": "California"
  }
}
```

Then people could run Netlify Dev with `ntl dev --geo=us` or something like that.

I know that this makes this initiative a lot more complex (for us), but I think it provides a much simpler experience for the user.

What do you think?

eduardoboucas:
> it would spiral into a long series of inputs for props, codes and names. And just limiting to country would not meet the requirements of the ask

To be clear, I'm not suggesting we add inputs for anything other than the country. I think in most cases people are going to use geolocation data on the country level, because the accuracy goes down substantially as the granularity increases. Under that assumption, I think starting by adding a way for people to specify the country would be a great step. If we hear from people that they'd like more control, we can always revisit.

> Ideally some combination of using a CLI flag without having to introduce an API call would be ideal. Maybe just adding a list of the codes/subdivisions in a file somewhere, and the user could specify by `--geo=US&CA` etc?

We can always cache the API results so that they're available offline in subsequent invocations (we already do that for the geolocation object). Adding a file with geolocation does indeed guarantee that it'll _always_ work offline, but it also adds an awkward file to the codebase that we'll need to maintain, and which will contribute to the package size.

I'm not trying to push you in one direction or another, just sharing my thoughts on some of the pros and cons I can anticipate. Feel free to explore different options, and you can always tag me if you want to discuss it further!

JWhist:
> > it would spiral into a long series of inputs for props, codes and names. And just limiting to country would not meet the requirements of the ask
> 
> To be clear, I'm not suggesting we add inputs for anything other than the country. I think in most cases people are going to use geolocation data on the country level, because the accuracy goes down substantially as the granularity increases. Under that assumption, I think starting by adding a way for people to specify the country would be a great step. If we hear from people that they'd like more control, we can always revisit.
> 
> > Ideally some combination of using a CLI flag without having to introduce an API call would be ideal. Maybe just adding a list of the codes/subdivisions in a file somewhere, and the user could specify by `--geo=US&CA` etc?
> 
> We can always cache the API results so that they're available offline in subsequent invocations (we already do that for the geolocation object). Adding a file with geolocation does indeed guarantee that it'll _always_ work offline, but it also adds an awkward file to the codebase that we'll need to maintain, and which will contribute to the package size.
> 
> I'm not trying to push you in one direction or another, just sharing my thoughts on some of the pros and cons I can anticipate. Feel free to explore different options, and you can always tag me if you want to discuss it further!

Here's a handy free API out there: https://api.3geonames.org/?randomland=RU&json=1 that grabs a random city within a given country code that you pass.  Wish I knew how their code worked!

minivan:
> Here's a handy free API out there: https://api.3geonames.org/?randomland=RU&json=1 that grabs a random city within a given country code that you pass. Wish I knew how their code worked!

They use a database, sometimes those databases are available commerically. In my past gig, we used to buy those to turn zipcodes and city names into geolocations for use in searches.

I think this might be too much overhead for here. The goal is for people to be able to test personalization options with Edge Functions locally. We can always launch it with only the country and wait for requests from folks about their usage patterns.

JWhist:
> > Here's a handy free API out there: https://api.3geonames.org/?randomland=RU&json=1 that grabs a random city within a given country code that you pass. Wish I knew how their code worked!
> 
> They use a database, sometimes those databases are available commerically. In my past gig, we used to buy those to turn zipcodes and city names into geolocations for use in searches.
> 
> I think this might be too much overhead for here. The goal is for people to be able to test personalization options with Edge Functions locally. We can always launch it with only the country and wait for requests from folks about their usage patterns.

Yes, I was just reading through Stargate to see how those headers were added, and found the maxmind DB connection there.  After doing some local testing, sending just country with empty fields for city/subdivision does not break anything, and is definitely an option to pass as a command line flag.  We could also use something like faker to generate random city/state/province names just to fill the fields, although they might not be regionally accurate.

docs/commands/dev.md
```diff
@@ -24,7 +24,7 @@ netlify dev
 - `framework` (*string*) - framework to use. Defaults to #auto which automatically detects a framework
 - `functions` (*string*) - specify a functions folder to serve
 - `functionsPort` (*string*) - port of functions server
-- `geo` (*cache | mock | update*) - force geolocation data to be updated, use cached data from the last 24h if found, or use a mock location
+- `geo` (*string*) - force geolocation data to be updated, use cached data from the last 24h if found, or use a mock location
```
klavavej, line 27: Could be helpful to add an example of using a custom mock location so that folks don't have to trigger the error message / spelunk the CLI code to figure out what format to use. 

FYI, I filed an issue for mentioning this new feature in docs.netlify.com at https://github.com/netlify/docs/issues/2169

src/commands/dev/dev.js
```diff
@@ -366,6 +366,19 @@ const validateShortFlagArgs = (args) => {
   return args
 }
 
+const validateGeoArgs = (arg) => {
+  if (!['cache', 'mock', 'update'].includes(arg) && arg.length !== 2) {
```
eduardoboucas, line 370: What is the `arg.length` check doing?

docs/commands/dev.md
eduardoboucas: On this note, it's probably best to keep `geo` as an enum with `cache | mock | update` and introduce a new flag (e.g. `country`) to be used with `mock`.

So:

```
# Uses the default mock location
ntl dev --geo=mock

# Uses a custom mock location
ntl dev --geo=mock --country=CA
```

eduardoboucas (changes requested):

src/commands/dev/dev.js
JWhist: Validating that the code is two letters, so random input gives the helper message

eduardoboucas: Ah, I see! Mind adding a comment to make that more explicit?

JWhist: Updated to use a separate CLI flag, so closing out this thread

src/commands/dev/dev.js
```diff
@@ -593,6 +611,7 @@ const createDevCommand = (program) => {
         .choices(['cache', 'mock', 'update'])
         .default('cache'),
     )
+    .addOption(new Option('--country <geoCountry>', 'Two letter country code').argParser(validateGeoCountryCode))
```
eduardoboucas, line 614: I think this gives a bit more information about what the flag is for.

```suggestion
    .addOption(new Option('--country <geoCountry>', 'Two-letter country code to use as mock geolocation (to be used with --mock)').argParser(validateGeoCountryCode))
```

src/lib/geo-location.js
```diff
@@ -56,11 +55,14 @@ const getGeoLocation = async ({ mode, offline, state }) => {
...
-  if (mode === 'mock' || offline) {
-    return mockLocation
+  // If the `--geo` option is set to `mock`, we use the default mock location.
+  // If the `--offline` option was used, we can't talk to the API, so let's
+  // also use the mock location.  Otherwise, use the country code passed in by
+  // the user.
+  if (!['cache', 'update'].includes(mode) || offline) {
+    return mode === 'mock' && !geoCountry
+      ? mockLocation
+      : { city: '', country: { code: geoCountry, name: '' }, subdivision: { code: '', name: '' } }
```
eduardoboucas, line 65: What do you think about populating these fields with some mock data, like "Netlify City" and "Jamstack Avenue"? 😎 

Just so that the strings aren't left empty.

src/commands/dev/dev.js
eduardoboucas, line 614: I think it'd be good to throw if this flag is used without `--geo=mock`, but looking at [commander's documentation](https://github.com/tj/commander.js/) I couldn't find a good way to add a constraint like that.

Maybe @danez knows?

src/lib/geo-location.js
JWhist: Sounds good.  Could also use faker to generate some random data if you want to go that route.  Do you think putting a random two letter subdivision code might be worth considering?

eduardoboucas: I'd avoid faker just because it's an extra dependency for very little gain. Filling in any data should be fine. And yeah, I think I'd try to fill in the entire object. Maybe something like:

```json
{
  "city": "Netlicity",
  "country": <whatever they input>
  "subdivision": {
    "code": "JS",
    "name": "Jamstack"
  }
}
```

src/lib/geo-location.js
```diff
@@ -56,11 +55,18 @@ const getGeoLocation = async ({ mode, offline, state }) => {
...
-  // if the `--offline` option was used, we can't talk to the API, so let's
-  // also use the mock location.
-  if (mode === 'mock' || offline) {
-    return mockLocation
+  // If the `--geo` option is set to `mock`, we use the default mock location.
+  // If the `--offline` option was used, we can't talk to the API, so let's
+  // also use the mock location.  Otherwise, use the country code passed in by
+  // the user.
+  if (!['cache', 'update'].includes(mode) || offline) {
+    return mode === 'mock' && !geoCountry
```
eduardoboucas, line 63: Personally, I find this a bit hard to parse. Consider moving this to separate return statements.

```js
if (mode === 'mock') {
  if (geoCountry) {
    return { ... }
  }

  return mockLocation
}
```

eduardoboucas (dismissed):

danez (changes requested):

danez:
Some tests would be also nice. From the top of my head these cases would be good

- happy case `--country=ES`
- invalid cases `a2`, `toolong`
- country flag used without `geo=mock`

src/commands/dev/dev.js
```diff
@@ -386,6 +389,20 @@ const validateShortFlagArgs = (args) => {
   return args
 }
 
+const validateGeoCountryCode = (arg) => {
+  // Validate that the arg passed is two letters only for country
+  // See https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
+  if (arg.length !== 2 || !/[a-z]{2}/gi.test(arg)) {
+    throw new Error(
+      `The geo country code must use a two letter abbreviation.
+      ${chalk.red(BANG)}  Supported formats:
```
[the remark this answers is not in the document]
JWhist: Good catch.  That is a leftover from my first version of the change.  I'll clean that up and just show a single example.

src/commands/dev/dev.js
JWhist: If they pass the `--country` flag it is safe to assume they are mocking the geo data, since that is the only use case for that flag, so I think that is a valid approach.

src/commands/dev/dev.js
```diff
@@ -386,6 +389,20 @@ const validateShortFlagArgs = (args) => {
   return args
 }
 
+const validateGeoCountryCode = (arg) => {
+  // Validate that the arg passed is two letters only for country
+  // See https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes
+  if (arg.length !== 2 || !/[a-z]{2}/gi.test(arg)) {
```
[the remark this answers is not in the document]
JWhist: Thx for this.  I tend to avoid regexes and go for the simplest and most obvious condition, but I should have just done it this way from the start.

src/lib/geo-location.js
```diff
@@ -56,10 +54,18 @@ const getGeoLocation = async ({ mode, offline, state }) => {
...
+  // If the `--geo` option is set to `mock`, we use the default mock location.
+  // If the `--offline` option was used, we can't talk to the API, so let's
+  // also use the mock location.  Otherwise, use the country code passed in by
+  // the user.
+  if (mode === 'mock' || offline || geoCountry) {
+    if (geoCountry) {
+      return {
+        city: 'Netlitown',
+        country: { code: geoCountry, name: 'Mock Country' },
+        subdivision: { code: 'JS', name: 'Jamstackvania' },
```
klavavej, line 66: For the sake of clarity, especially for folks who learned / are learning English as a 2nd+ language, and to avoid potential translation troubles, I suggest avoiding fun pun names like "Netlitown" and "Jamstackvania" in favor of explicit placeholder names like "Fake city" or "Mock city" / "Fake subdivision" or "Mock subdivision"

src/lib/geo-location.js
```diff
@@ -56,10 +54,18 @@ const getGeoLocation = async ({ mode, offline, state }) => {
...
-  if (mode === 'mock' || offline) {
+  // If the `--geo` option is set to `mock`, we use the default mock location.
+  // If the `--offline` option was used, we can't talk to the API, so let's
+  // also use the mock location.  Otherwise, use the country code passed in by
+  // the user.
+  if (mode === 'mock' || offline || geoCountry) {
+    if (geoCountry) {
+      return {
+        city: 'Netlitown',
+        country: { code: geoCountry, name: 'Mock Country' },
```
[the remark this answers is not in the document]
JWhist: This was discussed in the issue earlier, and would require a geo database or API call, and I was advised not to do that at this time.

tests/unit/lib/geo-location.test.js
```diff
@@ -116,3 +116,37 @@ test('`getGeoLocation` returns mock geolocation data if `mode: "mock"`', async (
   t.false(hasCalledStateSet)
   t.deepEqual(geo, mockLocation)
 })
+
+test('`getGeoLocation` returns mock geolocation data if valid country code set', async (t) => {
+  const returnedLocation = {
+    city: 'Netlitown',
+    country: { code: 'CA', name: 'Mock Country' },
```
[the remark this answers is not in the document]
JWhist: It would, but I was advised not to create the necessary file to do this lookup as the data could change, and there are numerous subdivisions in each country etc.  Any testing that needed to be done could be done by country code.  If there is a demand for it, I think we could easily add a file with a map of codes => names though.

src/commands/dev/dev.js
```diff
@@ -599,6 +616,12 @@ const createDevCommand = (program) => {
         .choices(['cache', 'mock', 'update'])
         .default('cache'),
     )
+    .addOption(
+      new Option(
+        '--country <geoCountry>',
+        'Two-letter country code ([ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)) to use as mock geolocation (to be used with --geo=mock)',
```
danez, line 622: We do not have markdown here.

```suggestion
        'Two-letter country code (ISO 3166-1 alpha-2, https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) to use as mock geolocation (enables --geo=mock autmatically)',
```

danez (commented):
Just two small doc things, to make it clear that --geo=mock is activated automatically.

danez (approved):

PR #4711, opened 2022-06-16
feat: add graph:init command

sgrove:
#### Summary

Adds `graph:init` to bootstrap all of the resources for a Netlify site to work with Netlify Graph
---


src/commands/graph/graph-init.js
```diff
@@ -0,0 +1,114 @@
...
+    result.data.oneGraph.createApiToken.accessToken &&
+    result.data.oneGraph.createApiToken.accessToken.token
+
+  if (token) {
+    newEnv.NETLIFY_GRAPH_PERSIST_QUERY_TOKEN = token
+  } else {
+    error(`Unable to create Netlify Graph persist query token: ${JSON.stringify(result.errors, null, 2)}`)
+  }
+
+  if (envChanged) {
```
anmonteiro, line 85: This seems to always be true.

anmonteiro (approved):

src/commands/graph/graph-init.js
```diff
@@ -0,0 +1,114 @@
...
+ * Creates the `netlify graph:init` command
+ * @param {import('commander').OptionValues} options
+ * @param {import('../base-command').BaseCommand} command
+ * @returns
+ */
+const graphInit = async (options, command) => {
+  const { api, site, state } = command.netlify
+  const siteId = site.id
+
+  if (!site.id) {
```
eduardoboucas, line 21: [dust] You're creating `siteId` but still using `site.id` here. Perhaps we could get rid of `siteId` or use it consistently instead of `site.id`?

src/commands/graph/graph-init.js
```diff
@@ -0,0 +1,114 @@
...
+    error(
+      `${NETLIFYDEVERR} Warning: no siteId defined, unable to start Netlify Graph. To enable, run ${chalk.yellow(
```
eduardoboucas, line 23: We have a [`warn` helper](https://github.com/netlify/cli/blob/main/src/utils/command-helpers.js#L168-L171) that sets the appropriate colour scheme and formatting for warning messages, as well as the _Warning:_ prefix. Could we use that here?

If it definitely needs to be an error and not a warning, perhaps we can remove the _Warning:_ prefix?

src/commands/graph/graph-init.js
```diff
@@ -0,0 +1,114 @@
...
+        'netlify init',
+      )} or ${chalk.yellow('netlify link')}`,
+    )
+  }
+
+  const netlifyToken = await command.authenticate()
+  // @ts-ignore
```
eduardoboucas, line 30: Can we add a comment explaining why we're ignoring type checking here?

anmonteiro (approved):

KyleBlankRollins (commented):
Sorry it took so long to get around to looking at this from a docs/copy standpoint. I have some questions and suggestions - mostly around error logs.

src/commands/graph/graph-init.js
```diff
@@ -0,0 +1,116 @@
...
+      result.data.oneGraph &&
+      result.data.oneGraph.createApiToken &&
+      result.data.oneGraph.createApiToken.accessToken &&
+      result.data.oneGraph.createApiToken.accessToken.token
+
+    if (token) {
+      envChanged = true
+      newEnv.NETLIFY_GRAPH_PERSIST_QUERY_TOKEN = token
+    } else {
+      error(`Unable to create Netlify Graph persist query token: ${JSON.stringify(result.errors, null, 2)}`)
```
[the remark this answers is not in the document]
anmonteiro: The errors we're printing should give us more information about what happened.

anmonteiro (approved):

PR #4731, opened 2022-06-23
feat: Add edge function templates to functions:create command

jenae-janzen:

#### Summary

Summary
Fixes #<https://github.com/netlify/pod-compute/issues/99>

Adds edge function templates for JS and TS for the functions:create command.

Opened a new PR because my https://github.com/netlify/cli/pull/4689was getting messy, especially after trying to merge changes from main 😫

---

      passes our tests.


src/commands/functions/functions-create.js
```diff
@@ -128,8 +132,9 @@ const formatRegistryArrayForInquirer = function (lang) {
 /**
  * pick template from our existing templates
  * @param {import('commander').OptionValues} config
+ *
  */
-const pickTemplate = async function ({ language: languageFromFlag }) {
+const pickTemplate = async function ({ language: languageFromFlag }, isEdgeFunc) {
```
eduardoboucas, line 137: Rather than a `isEdgeFunc` boolean, do you think we could we have a `type` enum variable containing `serverless` or `edge`? I feel that could be extended to contain other types in the future.

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,42 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
 
 const DEFAULT_PRIORITY = 999
 
+const selectTypeOfFunc = async () => {
+  const { functionType } = await inquirer.prompt([
+    {
+      name: 'functionType',
+      message: "Select the type of function you'd like to create: ",
```
eduardoboucas, line 205: I think we can leave this type of formatting to the prompt library.

```suggestion
      message: "Select the type of function you'd like to create",
```

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,42 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+}
+
+const ensureEdgeFuncDirExists = function (command) {
+  const siteId = command.netlify.site.id
+
+  if (!siteId) {
+    error(`${NETLIFYDEVERR} No site id found, please run inside a site directory o \`netlify link\``)
+  }
+
+  const functionsDirHolder = 'netlify/edge-functions'
```
eduardoboucas, line 220: While `netlify/edge-functions` is the default directory for edge functions, it's possible for people to specify their own. If they do, we should respect it and not mandate our own.

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,42 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+  if (!fs.existsSync(functionsDirHolder)) {
+    log(
+      `${NETLIFYDEVLOG} functions directory ${chalk.magenta.inverse(
+        functionsDirHolder,
+      )} does not exist yet, creating it...`,
+    )
+
+    fs.mkdirSync(functionsDirHolder, { recursive: true })
+
+    log(`${NETLIFYDEVLOG} functions directory ${chalk.magenta.inverse(functionsDirHolder)} created`)
```
eduardoboucas, line 231: [dust]

```suggestion
    log(`${NETLIFYDEVLOG} Functions directory ${chalk.magenta.inverse(functionsDirHolder)} created`)
```

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,42 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+  }
+  console.log(functionsDirHolder)
```
eduardoboucas, line 233: Did you mean to remove this?

```suggestion
```

src/commands/functions/functions-create.js
```diff
@@ -258,8 +302,17 @@ const ensureFunctionDirExists = async function (command) {
  * @param {import('commander').OptionValues} options
  * @param {string} argumentName
  * @param {string} functionsDir
+ * @param {boolean} isEdgeFunc
  */
-const downloadFromURL = async function (command, options, argumentName, functionsDir) {
+// eslint-disable-next-line max-params
+const downloadFromURL = async function (command, options, argumentName, functionsDir, isEdgeFunc) {
```
eduardoboucas, line 308: Rather than disabling the `max-params` rule, you could look into using named parameters.

```suggestion
const downloadFromURL = async function ({ command, options, argumentName, functionsDir, isEdgeFunc }) {
```

src/commands/functions/functions-create.js
```diff
@@ -258,8 +302,17 @@ const ensureFunctionDirExists = async function (command) {
...
+  if (isEdgeFunc) {
+    log(
+      `${NETLIFYDEVWARN}: Downloading an edge function from a URL is currently not supported. Terminating without further action.`,
+    )
+    process.exit(1)
```
eduardoboucas, line 313: I think you could use the `error` helper for this: https://github.com/netlify/cli/blob/main/src/utils/command-helpers.js#L179-L191

src/commands/functions/functions-create.js
```diff
@@ -371,10 +424,12 @@ const installDeps = async ({ functionPackageJson, functionPath, functionsDir })
  * @param {import('commander').OptionValues} options
  * @param {string} argumentName
  * @param {string} functionsDir
+ * @param {boolean} isEdgeFunc
  */
-const scaffoldFromTemplate = async function (command, options, argumentName, functionsDir) {
+// eslint-disable-next-line max-params
+const scaffoldFromTemplate = async function (command, options, argumentName, functionsDir, isEdgeFunc) {
```
eduardoboucas, line 430: Same comment as above, you could use named parameters.

src/commands/functions/functions-create.js
```diff
@@ -565,6 +624,28 @@ const installAddons = async function (command, functionAddons, fnPath) {
...
+  if (!fs.existsSync('netlify.toml')) {
+    log(`${NETLIFYDEVLOG} \`netlify.toml\` file does not exist yet. Creating it...`)
+  }
+
+  // Todo: validate this input
+  const { funcPath } = await inquirer.prompt([
+    {
+      type: 'input',
+      name: 'funcPath',
+      message: `What route do you want your edge function to be invoked on? \nRout must begin with '/'\
```
eduardoboucas, line 637: This is probably what you mean with the `Todo:` comment above, but rather than saying that the route must begin with `/`  we could be more forgiving with the input and add the `/` ourselves.

src/commands/functions/functions-create.js
```diff
@@ -565,6 +624,28 @@ const installAddons = async function (command, functionAddons, fnPath) {
...
+      and can be configured by editing your \`netlify.toml\` file.`,
+      default: '/test',
+    },
+  ])
+
+  const functionRegister = `\n\n[[edge_functions]]\nfunction = "${funcName}"\npath = "${funcPath}"`
+
+  fs.promises.appendFile('netlify.toml', functionRegister)
```
eduardoboucas, line 645: We should be catching any errors from this before saying that the edge function was successfully registered.

src/commands/functions/functions-create.js
```diff
@@ -582,11 +663,12 @@ const ensureFunctionPathIsOk = function (functionsDir, name) {
  * @param {import('../base-command').BaseCommand} command
  */
 const functionsCreate = async (name, options, command) => {
-  const functionsDir = await ensureFunctionDirExists(command)
+  const isEdgeFunc = (await selectTypeOfFunc()) === 'Edge Function (Deno)'
```
eduardoboucas, line 666: With `inquirer`, you can make `choices` an array of objects with a `name` and `value` properties. While `name` is presented to the user, `value` is what'll be added to the answers object. This way you could check for a value like `edgeFunction` and not the human-friendly, which we might want to change.

That was a bit difficult to explain, but hopefully it makes sense?

src/commands/functions/functions-create.js
```diff
@@ -93,13 +93,17 @@ const filterRegistry = function (registry, input) {
...
+  const folderNames = fs.readdirSync(path.join(templatesDir, lang, functionType))
   const registry = folderNames
     // filter out markdown files
     .filter((folderName) => !folderName.endsWith('.md'))
-    // eslint-disable-next-line n/global-require, import/no-dynamic-require
-    .map((folderName) => require(path.join(templatesDir, lang, folderName, '.netlify-function-template.js')))
+
+    .map((folderName) =>
+      // eslint-disable-next-line n/global-require, import/no-dynamic-require
+      require(path.join(templatesDir, lang, functionType, folderName, '.netlify-function-template.js')),
```
eduardoboucas, line 105: I wonder if you've considered keeping the templates organised by language and using the existing `.netlify-function-template.js` file to store the type (i.e. `serverless` or `edge`)? Feels like it could simplify the logic a little bit, because you'd still load files from the same location but rather than filtering templates by `language` you'd also look at `type`?

eduardoboucas (commented):
Nice work! I did a first pass and left some comments, let me know what you think.

src/commands/functions/functions-create.js
jenae-janzen: Oh interesting. I know you can specify a directory for regular serverless functions in the UI or in the netlify.toml file, and was wondering about this, but couldn't find anything about setting a custom ef directory in the docs. Is this something only possible in the toml file? Should this be added to the docs somewhere?

src/commands/functions/functions-create.js
jenae-janzen: 😅Yes!

src/commands/functions/functions-create.js
jenae-janzen: Makes sense. Will do!

src/commands/functions/functions-create.js
jenae-janzen: Makes sense. I kept forgetting the `/` and having to fix it manually to test the functions so I assumed other people would forget it as well. I'll add a check to prepend the `/` if it's missing instead.

src/commands/functions/functions-create.js
jenae-janzen: I think I see what you mean, setting a `value` property to `edgeFunction` is much cleaner than looking for the human-readable version, and if the prompt changes in the future, it'll only have to be changed in one place. 

I'm just trying to wrap my head around this in conjunction with your other comment before:

> Rather than a isEdgeFunc boolean, do you think we could we have a type enum variable containing serverless or edge? I feel that could be extended to contain other types in the future.

Going to try to walk through this to check my understanding:

- If we create an enum with the types of functions (currently `serverless` and `edge`), rather than a boolean, we have more flexibility to add other types of functions in the future
- Rather than passing around `isEdgeFunc`, we'd pass around something like `funcType`, which will be `serverless` or `edge`
- When we prompt users to choose their type of function using `inquirer`, the values should be the same as the enum values, because that's how the variable will get set.

Does that make sense?

src/commands/functions/functions-create.js
jenae-janzen: Hmm... I'll have to think about this one a little more because I don't quite have a clear understanding of how the `.netlify-function-template.js` files work. It would definitely be nice not to change the entire directory structure, if there's a simpler way.

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,41 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+    },
+  ])
+  return functionType
+}
+
+const ensureEdgeFuncDirExists = function (command) {
+  const siteId = command.netlify.site.id
+
+  if (!siteId) {
+    error(`${NETLIFYDEVERR} No site id found, please run inside a site directory o \`netlify link\``)
```
whitep4nth3r, line 217: Is this message incomplete? Referring to the `o`.

src/commands/functions/functions-create.js
whitep4nth3r: Typo on `Rout`.

src/functions-templates/javascript/edge/hello/{{name}}.js
```diff
@@ -0,0 +1,4 @@
+export default async (Request) =>
+  new Response('Hello, World!', {
+    headers: { 'content-type': 'text/html' },
+  })
```
whitep4nth3r, line 4: I'm not sure we want to use arrow functions in the templates, especially given the associated examples on the [edge functions example site](https://edge-functions-examples.netlify.app/example/hello) doesn't use arrow functions. Plus, I know a lot of people that this syntax still confuses.

I would recommend 

```
export default async (Request) => {
  return new Response("Hello, World!", {
    headers: { "content-type": "text/html" },
  });
};
```

src/functions-templates/javascript/edge/image-external/{{name}}.js
```diff
@@ -0,0 +1,9 @@
+// Let's serve an image of a kitten from the internet
+
+export default async () => {
+  // fetch() is supported natively by Deno!
+  // Returning the awaited response automatically sets the
+  // content-type headers!
+  const kitten = await fetch('https://placekitten.com/g/300/300')
+  return kitten
+}
```
whitep4nth3r, line 9: As above, I think it would be really beneficial to match the examples to the example site and not use arrow functions. I won't add any further comments on this, but they apply to all files.

src/functions-templates/javascript/edge/hello/{{name}}.js
eduardoboucas: If I may piggyback on this, I don't think we should be using `Request` (with a capital R) for the function argument. By convention, that is typically used for classes (like `Response`), but this argument is an instance of a class (and the class is `Request`).

I think this can generate some confusion, especially if you want to create another request inside of the function, because you've overridden the `Request` identifier, so doing `const req = new Request(...)` will throw.

eduardoboucas: Also, I'm a bit confused about the arrow function question, because your recommendation is also using an arrow function. Are you talking about not using an implicit return?

whitep4nth3r: You're definitely right about `Request`.

> Are you talking about not using an implicit return?

Sorry, yes, exactly. (Sorry I had very little sleep last night!)

src/commands/functions/functions-create.js
eduardoboucas: That totally makes sense! 💯

src/commands/functions/functions-create.js
klavavej: > Should this be added to the docs somewhere?

I'm [tracking this as something to document as part of GA readiness](https://www.notion.so/netlify/Save-for-GA-89e6e2871d09447cbc5da15c18e36605#66d30c114db946319cc03d93e3351aca) but am unlikely to add it to the docs before then

Thanks for thinking about docs @jenae-janzen !!

jenae-janzen: @eduardoboucas, do you have any more information about how a user can set a custom edge function directory? I tried a few different versions of this in my toml file but no dice:
Screen Shot 2022-06-28 at 1 54 58 PM
Screen Shot 2022-06-28 at 1 59 06 PM

eduardoboucas: Oops, sorry for the late response. It should be:

```toml
[build]
edge_functions = "some-directory"
```

This is indeed a flow that we haven't documented and that we need to spend some time testing before GA. In any case, I think it would be good for you to retrieve the path from `config.edge_functions` (in the same way we're doing for serverless functions) as opposed to just hardcoding the default value.

src/commands/functions/functions-create.js
eduardoboucas: I think that file is just a sort of manifest file that tells CLI some information about the template. We can extend it with any data we want.

src/commands/functions/functions-create.js
```diff
@@ -371,11 +419,17 @@ const installDeps = async ({ functionPackageJson, functionPath, functionsDir })
...
+ * @param {string} funcType
  */
-const scaffoldFromTemplate = async function (command, options, argumentName, functionsDir) {
+// eslint-disable-next-line max-params
+const scaffoldFromTemplate = async function (command, options, argumentName, functionsDir, funcType) {
   // pull the rest of the metadata from the template
-  const chosenTemplate = await pickTemplate(options)
+  const chosenTemplate = await pickTemplate(options, funcType)
   if (chosenTemplate === 'url') {
+    if (funcType === 'edge') {
```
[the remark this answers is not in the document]
jackiewmacharia: Can't comment directly on it but [this block](https://github.com/netlify/cli/blob/jj/addEdgeFuncTemplate/src/commands/functions/functions-create.js#L184-L194) could be written as:
```js
source(answersSoFar, input) {
    // if Edge Functions template, don't show url option
    const edgeCommands = specialCommands.filter((val) => val.value !== 'url')
    const parsedSpecialCommands = funcType === 'edge' ? edgeCommands : specialCommands

    if (!input || input === '') {
      // show separators
      return [...templatesForLanguage, ...parsedSpecialCommands]
    }
    // only show filtered results sorted by score
    const answers = [...filterRegistry(templatesForLanguage, input), ...parsedSpecialCommands].sort(
      (answerA, answerB) => answerB.score - answerA.score,
    )
    return answers
},
```

src/commands/functions/functions-create.js
jenae-janzen: Good to know! I added a check to see if they have it configured and to use their custom directory if they do. Right now, I'm still using `netlify/edge-functions` as the default if they don't, but could also prompt them to choose a directory for their edge functions to live in and append their choice to the `toml`.

To me, it feels a bit safer to just use `netlify/edge-functions` so people aren't confused running into an undocumented workflow, but what do you think?

src/functions-templates/javascript/edge/hello/{{name}}.js
jenae-janzen: Okay, I believe I added explicit return statements to all the templates now, although they still use arrow functions ;) At any rate, they match the functions on the example site!

src/commands/functions/functions-create.js
```diff
@@ -189,6 +198,46 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+  if (!fs.existsSync(functionsDirHolder)) {
+    log(
+      `${NETLIFYDEVLOG} Functions directory ${chalk.magenta.inverse(
+        functionsDirHolder,
+      )} does not exist yet, creating it...`,
+    )
+
+    fs.mkdirSync(functionsDirHolder, { recursive: true })
+
+    log(`${NETLIFYDEVLOG} Functions directory ${chalk.magenta.inverse(functionsDirHolder)} created.`)
```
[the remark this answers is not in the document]
jenae-janzen: Updated!

src/commands/functions/functions-create.js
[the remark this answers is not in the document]
jenae-janzen: I like this a lot better! Thanks for the tip!

src/functions-templates/javascript/edge/image-external/{{name}}.js
jenae-janzen: This one matches: https://edge-functions-examples.netlify.app/example/image -- I guess I was looking at an older version of it though, which split the internal and external image retrieval. I can update it to match, but then eslint complains about unreachable code (since there are two return statements). It might also not be necessary to include, since there are plenty of other templates to choose from. Any strong feelings on this one @whitep4nth3r?

jenae-janzen:
I think I've addressed most of the comments, the ones still Todo are:
[] Use named parameters
[] Possibly use the existing .netlify-function-template.js file to store the type
[] Fix conflicts

So close!!!

src/commands/functions/functions-create.js
```diff
@@ -189,6 +203,51 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
 
 const DEFAULT_PRIORITY = 999
 
+const selectTypeOfFunc = async () => {
+  const functionTypes = [
+    { name: 'Edge Function (Deno)', value: 'edge' },
```
eduardoboucas, line 208: [dust] I think we only use the capitalised name when referring to the feature (i.e. Edge Functions), not a function.

```suggestion
    { name: 'Edge function (Deno)', value: 'edge' },
```

src/commands/functions/functions-create.js
```diff
@@ -189,6 +203,51 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+    { name: 'Serverless Function (Node/Go)', value: 'serverless' },
```
eduardoboucas, line 209: [dust] I think we only use the capitalised name when referring to the feature (i.e. Edge Functions), not a function.

```suggestion
    { name: 'Serverless function (Node/Go)', value: 'serverless' },
```

src/commands/functions/functions-create.js
```diff
@@ -565,6 +628,41 @@ const installAddons = async function (command, functionAddons, fnPath) {
...
+const registerEFInToml = async (funcName) => {
+  if (!fs.existsSync('netlify.toml')) {
+    log(`${NETLIFYDEVLOG} \`netlify.toml\` file does not exist yet. Creating it...`)
+  }
+
+  let { funcPath } = await inquirer.prompt([
+    {
+      type: 'input',
+      name: 'funcPath',
+      message: `What route do you want your edge function to be invoked on? \nRoute can be configured by editing your \`netlify.toml\` file.`,
```
eduardoboucas, line 640: Not a big deal, but I find the new-line here a bit unexpected. Since you're telling people that they can change the route after they select it, I'm not sure we need to also do it here. Up to you! If you want to keep it, I'd recommend getting rid of the `\n`.

```suggestion
      message: `What route do you want your edge function to be invoked on?`,
```

src/commands/functions/functions-create.js
```diff
@@ -565,6 +628,41 @@ const installAddons = async function (command, functionAddons, fnPath) {
...
+
+  const functionRegister = `\n\n[[edge_functions]]\nfunction = "${funcName}"\npath = "${funcPath}"`
+
+  try {
+    fs.promises.appendFile('netlify.toml', functionRegister)
+    log(
+      `${NETLIFYDEVLOG} Function '${funcName}' registered for route \`${funcPath}\`. To change, edit your \`netlify.toml\` file.`,
+    )
+  } catch {
+    log(`${NETLIFYDEVERR} Unable to register function. Please check your \`netlify.toml\` file.`)
```
eduardoboucas, line 661: I think we should use the `error` helper here instead of `log`. It also means you don't need to explicitly call `process.exit`.

src/commands/functions/functions-create.js
```diff
@@ -189,6 +203,51 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
...
+  const { config, site } = command.netlify
+  const siteId = site.id
+  let functionsDirHolder = config.build.edge_functions
+
+  if (!siteId) {
+    error(`${NETLIFYDEVERR} No site id found, please run inside a site directory or \`netlify link\``)
+  }
+
+  if (!functionsDirHolder) {
+    log(`${NETLIFYDEVLOG} Edge Functions directory not specified in netlify.toml`)
```
eduardoboucas, line 233: `config.build.edge_functions` being defined and the directory being set in the TOML aren't related. That config property will be populated with the default directory if you're using edge functions in your project. Just calling this out because it might confuse folks who read this code in the future.

Personally, I don't think we need this log line, because I'm not sure it provides useful or actionable information to the user.

eduardoboucas (commented):
Great work! I left a few minor comments, but it's nearly there. ✨

src/commands/functions/functions-create.js
jenae-janzen: Oh yeah, that makes sense. Removed!

src/commands/functions/functions-create.js
```diff
@@ -189,6 +203,50 @@ const pickTemplate = async function ({ language: languageFromFlag }) {
 
 const DEFAULT_PRIORITY = 999
 
+const selectTypeOfFunc = async () => {
+  const functionTypes = [
+    { name: 'Edge function (Deno)', value: 'edge' },
+    { name: 'Serverless Function (Node/Go)', value: 'serverless' },
```
eduardoboucas, line 209: The comment about not capitalising _function_ applied to serverless functions too.

```suggestion
    { name: 'Serverless function (Node/Go)', value: 'serverless' },
```

eduardoboucas (approved):

jenae-janzen:
@eduardoboucas thanks for reviewing! I was also wondering about tests -- I updated the existing tests for the serverless function templates so they match the new workflow (the additional prompt at the beginning) and have tested out the edge function templates a lot manually, but I should probably add a few test cases for the edge function templates too, right?

jenae-janzen:
Added one test, but I wanted to include a case where the user specified a custom edge functions directory as well, something like:

```javascript
test('should use specified edge function directory when found', async (t) => {
  const siteInfo = {
    admin_url: 'https://app.netlify.com/sites/site-name/overview',
    ssl_url: 'https://site-name.netlify.app/',
    id: 'site_id',
    name: 'site-name',
    build_settings: { repo_url: 'https://github.com/owner/repo' },
  }

  const routes = [
    {
      path: 'accounts',
      response: [{ slug: 'test-account' }],
    },
    { path: 'sites/site_id/service-instances', response: [] },
    { path: 'sites/site_id', response: siteInfo },
    {
      path: 'sites',
      response: [siteInfo],
    },
    { path: 'sites/site_id', method: 'patch', response: {} },
  ]

  await withSiteBuilder('site-with-custom-edge-functions-dir', async (builder) => {
    builder.withNetlifyToml({ config: { build: { edge_functions: 'somethingEdgy' } } })

    await builder.buildAsync()

    const createFunctionQuestions = [
      {
        question: "Select the type of function you'd like to create",
        answer: answerWithValue(CONFIRM),
      },
      {
        question: 'Select the language of your function',
        answer: answerWithValue(CONFIRM),
      },
      {
        question: 'Pick a template',
        answer: answerWithValue(CONFIRM),
      },
      {
        question: 'Name your function',
        answer: answerWithValue(CONFIRM),
      },
      {
        question: 'What route do you want your edge function to be invoked on?',
        answer: answerWithValue('/test'),
      },
    ]

    await withMockApi(routes, async ({ apiUrl }) => {
      const childProcess = execa(cliPath, ['functions:create'], {
        env: {
          NETLIFY_API_URL: apiUrl,
          NETLIFY_SITE_ID: 'site_id',
          NETLIFY_AUTH_TOKEN: 'fake-token',
        },
        cwd: builder.directory,
      })

      handleQuestions(childProcess, createFunctionQuestions)

      await childProcess

      t.is(await fs.fileExistsAsync(`${builder.directory}/somethingEdgy/hello/hello.js`, true))
    })
  })
})
```

However, it keeps failing -- I've been trying to debug it for over an hour now and not sure what else to try or how to get more visibility into why it's failing. Manually testing shows if the edge functions directory is configured in the TOML but doesn't exist, it gets created and the templates are added to it. Not sure if I'm missing something here or if this test is necessary 🤔

eduardoboucas (approved):

PR #4732, opened 2022-06-23
fix: move try catch to prepareServer

jackiewmacharia:

#### Summary

Related to  https://github.com/netlify/cli/issues/4621

- Looks like we are going to have to have to move the initializeProxy try/catch to prepareServer after all - I've ran into a situation during testing where prepareServer breaks before it gets to [Promise.all](https://github.com/netlify/cli/blob/main/src/lib/edge-functions/proxy.js#L90) - on linux-aarch64. Because we [don't await on prepareServer](https://github.com/netlify/cli/blob/main/src/lib/edge-functions/proxy.js#L67), it's possible that a binary error could be thrown at a different point in time breaking `netlify dev` before we have a chance to catch it in initializeProxy.

---

      passes our tests.

cute animal

### DIFF
--- a/src/lib/edge-functions/proxy.js
+++ b/src/lib/edge-functions/proxy.js
@@ -77,32 +77,26 @@ const initializeProxy = async ({
     projectDir,
   })
   const hasEdgeFunctions = userFunctionsPath !== undefined || internalFunctions.length !== 0
-  let hasServerError = false
 
   return async (req) => {
-    if (req.headers[headers.Passthrough] !== undefined || !hasEdgeFunctions || hasServerError) {
+    if (req.headers[headers.Passthrough] !== undefined || !hasEdgeFunctions) {
       return
     }
 
-    let promiseResult
+    const [geoLocation, preppedServer] = await Promise.all([
+      getGeoLocation({ mode: geolocationMode, offline, state }),
+      server,
+    ])
 
-    try {
-      promiseResult = await Promise.all([getGeoLocation({ mode: geolocationMode, offline, state }), server])
-    } catch (error_) {
-      error(error_.message, { exit: false })
-      hasServerError = true
-      return
-    }
-
-    const [geoLocation, { registry }] = promiseResult
+    if (!preppedServer) return
 
     // Setting header with geolocation.
     req.headers[headers.Geo] = JSON.stringify(geoLocation)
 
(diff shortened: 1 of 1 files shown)


jackiewmacharia:
@eduardoboucas @danez I may have ran into a use case for the try/catch in prepareServer instead of initializeProxy (details in PR description).

eduardoboucas:
> * Looks like we are going to have to have to move the initializeProxy try/catch to prepareServer after all - I've ran into a situation during testing where prepareServer breaks before it gets to [Promise.all](https://github.com/netlify/cli/blob/main/src/lib/edge-functions/proxy.js#L90)

Can you help me understand what leads to this? If `prepareServer` throws, then the `Promise.all` will reject and our existing `try/catch` will kick in. What am I missing?

jackiewmacharia:
Follow up from the pairing call, I've tested this branch against both scenarios and it works for both:

Screenshot 2022-06-24 at 18 34 27

Screenshot 2022-06-24 at 18 45 16

src/lib/edge-functions/proxy.js
```diff
@@ -152,34 +146,39 @@ const prepareServer = async ({
...
+      configPath,
+      directories,
+      getUpdatedConfig,
+      internalFunctions,
+      projectDir,
+      runIsolate,
+    })
 
-  return { registry, runIsolate }
+    return { registry, runIsolate }
```
eduardoboucas, line 178: Since we're not using `runIsolate` outside of this function, maybe we can `return registry`, which would also simplify things in `initializeProxy` because we could do `const [geoLocation, registry] = await Promise.all(...)` and `await registry.initialize()`?

src/lib/edge-functions/proxy.js
```diff
@@ -152,34 +146,39 @@ const prepareServer = async ({
...
+  } catch (error_) {
+    error(error_.message, { exit: false })
```
eduardoboucas, line 180: I find these variable names really confusing (`error` vs. `error_`). What do you think about renaming the `error` util when we import it (e.g. `const { error: printError } = require('../../utils/command-helpers')`) and then let `error` be the name of the error (e.g. `printError(error)`)?

eduardoboucas (approved):
🚀

Issue #4679, opened 2022-06-06
TypeError: `` is not an Option Object when using plugins.js as a function name


labboy0276:
### Describe the bug

Hello

I am currently using the serverless functions to render api end point for https://api.lando.dev/ (lando-api.netlify.app)

My current toml file redirects the base name of all js files in the functions directory to their respective url via:

```
[build]
  functions = "functions"

# API rewrites
[[redirects]]
  from = "/"
  to = "/v1"
  status = 200
[[redirects]]
  from = "/v1/*"
  to = "/.netlify/functions/:splat"
  status = 200
```
So a file names functions/ping.js will redirect to https://api.lando.dev/v1/ping.

The issue at hand is if I name a file plugins.js, I get this error when using the netlify-cli:

```
 ›   TypeError: `` is not an Option Object
```

The error stems from:

```
TypeError: `` is not an Option Object
    at Object.module.exports (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/merge-options/index.js:164:10)
    at getConfigForFunction (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/@netlify/zip-it-and-ship-it/dist/config.js:30:36)
    at /home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/@netlify/zip-it-and-ship-it/dist/runtimes/index.js:72:65
    at Array.map (<anonymous>)
    at getFunctionsFromPaths (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/@netlify/zip-it-and-ship-it/dist/runtimes/index.js:70:43)
    at async FunctionsRegistry.listFunctions (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/@netlify/zip-it-and-ship-it/dist/main.js:25:26)
    at async FunctionsRegistry.scan (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/src/lib/functions/registry.js:175:23)
TypeError: `` is not an Option Object
    at Object.module.exports (/home/john/.nvm/versions/node/v16.15.0/lib/node_modules/netlify-cli/node_modules/merge-options/index.js:164:10)
[...]

PR #4748, opened 2022-06-28
fix: functions with name matching config prop cause crash

JWhist:

#### Summary

Fixes [https://github.com/netlify/cli/issues/4679](https://github.com/netlify/cli/issues/4679)

Fixes bug by sending appropriate functions config data to zisi for bundling.

---

      passes our tests.
- [X] Update or add tests (if any source code was changed or added) 🧪


tests/integration/20.command.functions.test.js
```diff
@@ -840,4 +840,25 @@ test('should handle content-types with charset', async (t) => {
   })
 })
 
+test('should execute function when name clashes with config key names', async (t) => {
```
eduardoboucas, line 843: The fix itself looks good to me, but I'm not sure this is the right test to add. The fact that a function name was clashing with a config property was not the underlying problem, it was one of the side effects. The underlying problem was that CLI was sending the wrong configuration object to zip-it-and-ship-it, and I think that's what we should assert in the test.

I think the best way to do that would be to stub the call to zip-it-and-ship-it, which isn't very straightforward in this type of integration test (i.e. using `withDevServer` and testing the end-to-end flow).

Instead, you could write a unit test for the functions registry, asserting that it receives the entire configuration in its constructor but then passes just the functions block to zip-it-and-ship-it.

tests/unit/lib/functions/registry.test.js
```diff
@@ -0,0 +1,33 @@
+const { mkdirSync, mkdtempSync, writeFileSync } = require('fs')
+const { tmpdir } = require('os')
+const { join } = require('path')
+
+const test = require('ava')
+const sinon = require('sinon')
+
+const { FunctionsRegistry } = require('../../../../src/lib/functions/registry')
+
+test('registry should only pass functions config to zip it ship it', async (t) => {
```
eduardoboucas, line 10:
```suggestion
test('registry should only pass functions config to zip-it-and-ship-it', async (t) => {
```

tests/unit/lib/functions/registry.test.js
```diff
@@ -0,0 +1,33 @@
+const { mkdirSync, mkdtempSync, writeFileSync } = require('fs')
```
eduardoboucas, line 1: [dust] Consider using the async version of these methods.

tests/unit/lib/functions/registry.test.js
```diff
@@ -0,0 +1,33 @@
...
+const { tmpdir } = require('os')
+const { join } = require('path')
+
+const test = require('ava')
+const sinon = require('sinon')
+
+const { FunctionsRegistry } = require('../../../../src/lib/functions/registry')
+
+test('registry should only pass functions config to zip it ship it', async (t) => {
+  const projectRoot = mkdtempSync(join(tmpdir(), 'functions-project-root'))
```
eduardoboucas, line 11: We should clean up this directory at the end of the test.

eduardoboucas (approved):
Nice! I left a couple of non-blocking comments.

tests/unit/lib/functions/registry.test.js
JWhist: I changed them out.  Some of the syntax is older as the import style changes between Node 12 and Node 14, but the CLI still supports Node 12.

eduardoboucas (approved):

tests/unit/lib/functions/registry.test.js
JWhist: Done.  Can you help me understand why this only works when I await the `rmdir()` call?  Does the javascript process end before the actual async function is run otherwise?

eduardoboucas (changes requested):

tests/unit/lib/functions/registry.test.js
eduardoboucas: I don't follow. Can you share an example of what works and what doesn't?

danez: Yes if you do not await `rmdir` Node.JS will not wait for it and just shut down.

danez (dismissed):

JWhist:
Unable to merge this due to problems with the CI failing (integration test 130 in particular, and other eleventy tests - I've retried numerous times)

danez (approved):

PR #4750, opened 2022-06-29
fix: replace ansi2html with ansi-to-html

### LINKED ISSUE #4583: ansi2html dependabot alert
### Describe the bug

Can I make a PR replacing `ansi2html`? [It exists in only one place](https://github.com/netlify/cli/search?q=ansi2html) so the PR would be pretty low-impact. Possibly replace it with [ansi-to-html](https://github.com/rburns/ansi-to-html)

It's creating (the only) unresolved (and unresolvable) [dependabot alert](https://github.com/advisories/GHSA-c2v2-7rcg-2ch7) in my project.  It's not critical at all, and I could just dismiss it, but I'd like to clear it. 

>  At the time of this writing, February 2018, all versions of ansi2html remain vulnerable, and the package has not been updated for 4 years.
> In order to use this package safely, it is necessary to avoid passing user input to the package, or to limit the size of the input string to a size with a parse time you find acceptable. Unfortunately, the match time grows at quite a small character count, so it is unlikely for it to both allow strings of a useful size while protecting against the denial of service attack.
> In the case that user input of significant length must be parsed by ansi2html, the best mitigation is to use an alternative module that is actively maintained and provides similar functionality. There are multiple modules fitting this criteria [available on npm](https://www.npmjs.com/search?q=ansi+html)

### Steps to reproduce

* `npm install netlify-cli` or `yarn add netlify-cli` 
* `npm audit` or `yarn audit`

See [this alert](https://github.com/advisories/GHSA-c2v2-7rcg-2ch7)

### Configuration

_No response_

### Environment

  System:
    OS: Magic
    CPU: (16) x64 Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
    Memory: 2.45 GB | 15.75 GB
  Binaries:
    Node: 14.19.1 - oo|nodejs|node.EXE
    Yarn: 1.22.18 - oo|workspace|projects|rendall|websites|simple-comment|node_modules|.bin|yarn.CMD
    npm: 6.14.16 - oo|nodejs|npm.CMD

just-toby:

#### Summary

Fixes #4583

fixes the dependabot alert for the vulnerability in ansi2html by replacing it with a different package, as suggested by @rendall in #4583


BenLiuMattr:
@just-toby Thank you for the fix. We really need this also in the security audit of our internal projects! 

Just noticed the CI is failing, update the PR with the correct label to let the CI pass?

CC @danez @eduardoboucas Can you make this happen if it makes sense, Thanks

ivan-mattr:
Thanks @just-toby Looking forward to it.

danez (approved):
Thanks for submitting this. 🎉

`ansi-to-html` looks way better and is also a lot more used according to npm (~65k vs ~3.5M downloads).

I'm waiting for #4774 before merging

just-toby:
thanks @danez for your help

PR #4759, opened 2022-07-01
fix: resolve missing output when using --json

### LINKED ISSUE #4758: `deploy` doesn't output when using `--trigger` & `--json` together
### Describe the bug

When using the `--trigger` & `--json` flags together the `deploy` command doesn't output.

This is because there is no business logic to handle the `--json` argument in `triggerDeploy()`. It results in the output being passed to `log()` which in turn doesn't output anything if the `--json` argument is set.

### Steps to reproduce

1. Run `netlify deploy --trigger --json`

### Configuration

N/A

### Environment

  System:
    OS: macOS 12.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
    Memory: 25.86 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  npmGlobalPackages:
    netlify-cli: 10.7.0

dariodevito:
#### Summary

Fixes #4758

Adds support for `--json` when using with `netlify deploy --trigger`.

---

      passes our tests.
- [] Update or add tests (if any source code was changed or added) 🧪

Image from iOS

---

Just to call out, I had intended to add an integration test in `./tests/integration/210.command.deploy.test.js` along the lines of below. However, because the test site isn't built from/linked to a git codebase, the `--trigger` option (which calls `createSiteBuild`) returns a 404 (doesn't have a codebase to build from).

I think it would make sense for a Netlify developer to specify the repo this should be linked to (and by extension add tests for `--trigger`. (I'm guessing that is why it hasn't been covered before now)

```
  test.serial('should return valid json when both --trigger and --json are passed', async (t) => {
    await withSiteBuilder('site-with-public-folder', async (builder) => {
      const content = '<h1>⊂◉‿◉つ</h1>'
      builder
        .withContentFile({
          path: 'public/index.html',
          content,
        })
        .withNetlifyToml({
          config: {
            build: { publish: 'public' },
          },
        })

      await builder.buildAsync()

      const output = await callCli(['deploy', '--trigger', '--json'], {
        cwd: builder.directory,
        env: { NETLIFY_SITE_ID: t.context.siteId },
      })

      JSON.parse(output)
    })
  })
```


danez:
Thanks for creating the PR.

We will check the test case you mentioned.

src/commands/deploy/deploy.js
```diff
@@ -553,7 +562,7 @@ const deploy = async (options, command) => {
   const deployToProduction = options.prod || (options.prodIfUnlocked && !siteData.published_deploy.locked)
 
   if (options.trigger) {
-    return triggerDeploy({ api, siteId, siteData })
+    return triggerDeploy({ api, options, siteData, siteId })
```
danez, line 565: thanks for ordering 😘

danez (approved):
lgtm
thanks again, I created a separate issue for the test.

PR #4760, opened 2022-07-01
fix: use correct paths when deploying edge functions on Windows

eduardoboucas:
#### Summary

The paths we supply to the `files` hash needs to be in the unix format.

### DIFF
--- a/src/lib/edge-functions/deploy.js
+++ b/src/lib/edge-functions/deploy.js
@@ -1,6 +1,5 @@
 // @ts-check
 const { stat } = require('fs').promises
-const path = require('path')
 
 const { getPathInProject } = require('../settings')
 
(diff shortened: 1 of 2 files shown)


src/lib/edge-functions/deploy.js
```diff
@@ -10,7 +9,7 @@ const distPath = getPathInProject([EDGE_FUNCTIONS_FOLDER])
 
 const deployFileNormalizer = (file) => {
   const isEdgeFunction = file.root === distPath
-  const normalizedPath = isEdgeFunction ? path.join(PUBLIC_URL_PATH, file.normalizedPath) : file.normalizedPath
```
eduardoboucas, line 13: This effectively a URL, not a filesystem path. So it needs to use the UNIX format regardless of the host platform.

danez (approved):

PR #4772, opened 2022-07-04
fix: make git clone work in sites:create-template

danez:
#### Summary

`git_url` from the github api contains the plaintext git:// url which github itself [does not support anymore](https://github.blog/2021-09-01-improving-git-protocol-security-github/). Instead, use the `clone_url` form the API which contains the https:// git URL

Currently when running the command this happens:

```
❯ ../netlify/cli/bin/run sites:create-template nextjs-blog-theme
? Team: Daniel Tschinder's team
Choose a unique site name (e.g. isnt-danez-awesome.netlify.app) or leave it blank for a random name. You can update the site name later.
? Site name (optional): undefined

Site Created

Admin URL: https://app.netlify.com/sites/isnt-danez-awesome
URL:       https://isnt-danez-awesome.netlify.app
Site ID:   429aa7d1-77fb-43ef-b45c-05857d885e46
Repo URL:  https://github.com/danez/isnt-danez-awesome
? Do you want to clone the repository? Yes

 ›   Error: Command failed with exit code 128: git clone git://github.com/danez/isnt-danez-awesome.git isnt-danez-awesome
Cloning into 'isnt-danez-awesome'...
fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Operation timed out


```

#### Additional

Also switch to use `execa` instead of `git-clone` to save one dependency. I checked what git-clone does internally and that is pretty much exactly what I do in this PR. `git-clone` has some advanced features, like custom path to git binary or checkout different branch after clone, but we are not using any of these.

No test was added, because we mock the whole github API, so even if I would mock execa and assert it was called with the mocked git URL, it won't test if the actual URL will work or not

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -30,7 +30,7 @@
         "chokidar": "^3.0.2",
         "ci-info": "^3.0.0",
         "clean-deep": "^3.0.2",
-        "commander": "^9.0.0",
+        "commander": "^9.2.0",
         "concordance": "^5.0.0",
         "configstore": "^5.0.0",
         "content-type": "^1.0.4",
--- a/package.json
+++ b/package.json
@@ -233,7 +233,7 @@
     "chokidar": "^3.0.2",
     "ci-info": "^3.0.0",
     "clean-deep": "^3.0.2",
-    "commander": "^9.0.0",
+    "commander": "^9.2.0",
     "concordance": "^5.0.0",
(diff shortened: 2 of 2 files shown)


danez:
Maybe @netlify/templates could have a look because I guess this falls in your territory.

Skn0tt (approved):
Looks good!

PR #4774, opened 2022-07-05
chore: split package size workflow to make it work for PRs from forks

danez:
This enables the package-size to work also from forks. Currently
it does not work because the workflow gets only a readonly token
if started from a fork, hence the workflow cannot post the comment.

https://github.com/netlify/cli/runs/7194403058?check_suite_focus=true#step:6:27

By splitting the workflow, we run the calculation in a secure environment with
only the readonly token and then run the second workflow after the first
one is finished. The second workflow then has write access and can
post the comment

taken from https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

This upload of a workflow artifact will not count against the storage limit of the netlify github-account, because this is a public repository. https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions

There is no way to test this in a PR: 🤞

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/benchmark-post.yml
+++ b/.github/workflows/benchmark-post.yml
@@ -1,0 +1,27 @@
+name: Post package size
+
+on:
+  workflow_run:
+    workflows: [Calculate package size]
+    types:
+      - completed
+
+jobs:
+  post-package-size:
+    runs-on: ubuntu-latest
+    steps:
+      # This posts the status to the PR/commit
+      - uses: haya14busa/action-workflow_run-status@v1
+      - name: Download deltas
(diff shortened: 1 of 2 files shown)


danez:
> Nice idea!

Sorry I now also added a retention of the artifacts for 7 days.

lukasholzer (approved):

lukasholzer (approved):

PR #4787, opened 2022-07-11
fix(deps): update wait-port to 0.2.13

danez:
#### Summary

Fixes the cli tests

Fixes #4783


danez:
Does not work, as they dropped node 12 support.

PR #4788, opened 2022-07-11
fix(deps): update wait-port to 0.2.14

danez:
#### Summary

Reverts #4783 as now fixed upstream


anmonteiro:
we ended up going with https://github.com/netlify/cli/pull/4783 because it has Node.js 12.x compat

anmonteiro (dismissed):

anmonteiro (approved):

PR #4789, opened 2022-07-11
fix: Update getTemplatesFromGitHub to retrieve more repositories

### LINKED ISSUE #4777: Show all valid templates for sites:create-template
### Describe the bug

When using the Netlify CLI's command `ntl sites:create-template` we identified that we were not seeing all public repos within the list. We came to the conclusion that when we perform the fetch request, the returning repo list is limited to 30 repos, including those that are private, archived, etc. and then the CLI is filtering out the rest dependent on their status. 

Here's the code that is making the fetch request for our `netlify-templates` repositories:
https://github.com/netlify/cli/blob/c0b42f94e63696d437fb4de9fd1f665210a6f5cc/src/utils/sites/utils.js#L3-L9

And referring to the GitHub API docs on [listing all repositories for an org](https://docs.github.com/en/rest/repos/repos#list-organization-repositories), it looks as though we have a default limit of 30.

---

Two good options might be for us to:
- Updating the limit for `per_page`
- Modifying the request so it only finds public repositories

### Steps to reproduce

1. Install the Netlify CLI `npm install -g netlify-cli`
2. Run `ntl sites:create-template`

The way we solved this for now is by deleting outdated repositories but that is a temporary fix.

### Configuration

_No response_

### Environment

  System:
    OS: macOS 12.4
    CPU: (8) arm64 Apple M1
    Memory: 115.25 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.volta/tools/image/node/16.15.1/bin/node
    Yarn: 1.22.17 - ~/.volta/tools/image/yarn/1.22.17/bin/yarn
    npm: 8.11.0 - ~/.volta/tools/image/node/16.15.1/bin/npm

maxcell:
#### Summary

Fixes #4777 

When a user uses the CLI for generating a site with a template, we want to make sure only the appropriate repos show up for us. Right now that's all public, non-archived projects. The default limit is 30, so we also doubled the site request for now.

---

I wasn't quite sure if I should make a separate test for validating that the other repository do not show up but I modified the example at least so it should still pass other tests.

cute little hedgehog peering into your heart

### DIFF
--- a/src/commands/sites/sites-create-template.js
+++ b/src/commands/sites/sites-create-template.js
@@ -16,7 +16,7 @@ const fetchTemplates = async (token) => {
   const templatesFromGithubOrg = await getTemplatesFromGitHub(token)
 
   return templatesFromGithubOrg
-    .filter((repo) => !repo.archived && !repo.private && !repo.disabled)
+    .filter((repo) => !repo.archived && !repo.disabled)
     .map((template) => ({
       name: template.name,
       sourceCodeUrl: template.html_url,
--- a/src/utils/sites/utils.js
+++ b/src/utils/sites/utils.js
@@ -1,7 +1,16 @@
 const fetch = require('node-fetch')
 
 const getTemplatesFromGitHub = async (token) => {
-  const templates = await fetch(`https://api.github.com/orgs/netlify-templates/repos`, {
+  const getPublicGitHubReposFromOrg = new URL(`https://api.github.com/orgs/netlify-templates/repos`)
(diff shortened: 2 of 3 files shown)


danez (approved):
LGTM 🎉

Only one thing, this is not type `chore`, because you are actually fixing a bug for users. So when you merge make sure to use `fix: ...` in the squash commit message, so that it shows up in the changelog.

maxcell:
I can do that! Thank you for the feedback @danez!

PR #4790, opened 2022-07-11
chore: send pr number from one to the other workflow

danez:
#### Summary

This is the final fix to make the benchmark work for PRs from outside contributors.
We store the pr number in a file and upload it.
Then in the other workflow, we download it again and send it to delta-action.

As this is a public repo we do not have to pay for storage, also it is just some bytes.

Ref: #4774

---

      passes our tests.


.github/workflows/benchmark-post.yml
```diff
@@ -20,8 +20,12 @@ jobs:
...
           github_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Get PR number
+        if: github.event.workflow_run.event == 'pull_request'
+        id: pr_number
+        run: echo "::set-output name=pr_number::$(cat pr_number)"
       - name: Post deltas to GitHub
         uses: netlify/delta-action@v4
         with:
           title: '📊 Benchmark results'
-          token: ${{ secrets.GITHUB_TOKEN }}
```
danez, line 27: Token is optional and the default is exactly the same as this now.

eduardoboucas (approved):

PR #4791, opened 2022-07-11
chore: enable the use eszip feature flag for edge functions

EwanValentine:

#### Summary

Updates the CLI to utilise the eszip changes for edge functions.

- [internal] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
      passes our tests.
- [n\a] Update or add tests (if any source code was changed or added) 🧪
- [n\a] Update or add documentation (if features were changed or added) 📝


package.json
```diff
@@ -92,6 +92,7 @@
     "Nicolas Gonzalez",
     "Pankaj Patil <pankaj.patil2099@hotmail.com> (https://twitter.com/pankajpatil16)",
     "Peter",
+    "Prince Wilson (https://twitter.com/maxcell)",
```
EwanValentine, line 96: I... didn't add this, I assume there's some automation happening somewhere 😄

danez: Yes just ignore it :), we need to fix the flow to actually do these changes in the related PRs instead of in the next PR.

package.json
```diff
@@ -44,7 +44,6 @@
     "Erez Rokah (https://www.erezro.com)",
     "Erica Pisani <pisani.erica@gmail.com>",
     "Evans Hauser (https://twitter.com/evanshauser)",
-    "Ewan Valentine <ewan.valentine89@gmail.com> (http://ewanvalentine.io)",
```
EwanValentine: Oh :( 😂

danez:
Are we sure to enable this for all cli users? On LD I see it is only on for 5 users.

src/lib/build.js
```diff
@@ -32,6 +32,7 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
   cwd,
   featureFlags: {
     functionsBundlingManifest: true,
+    edge_functions_produce_eszip: true,
```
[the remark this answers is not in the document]
danez: I think the way it is right now is correct, so the same as [here](https://github.com/netlify/build/blob/9a2571472902b300803f583701c86a941eb3b532/packages/build/src/core/feature_flags.js#L20) . The one above `functionsBundlingManifest` I cannot find anywhere, nor build or edge-bundler or zisi. @eduardoboucas? https://github.com/search?q=org%3Anetlify+functionsBundlingManifest&type=code

EwanValentine:
@danez oh, strange. This has been changed whilst I was away. I'll double-check what's happened here, it was fully on before. Good spot.

src/lib/build.js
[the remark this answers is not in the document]
eduardoboucas: I think we've deprecated that flag a while ago and probably forgot to remove it from the CLI.

src/lib/build.js
```diff
@@ -35,6 +35,7 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
...
```
eduardoboucas, line 38: There's an additional place where CLI talks to Netlify Build where we also need to pass the feature flag: https://github.com/netlify/cli/blob/main/src/commands/deploy/deploy.js#L413.

eduardoboucas (commented):
Also, make sure to rename this to `feat:` otherwise it won't generate a release.

src/lib/build.js
EwanValentine: Ahhhh thanks! Will add this now.

Also TIL regarding `feat:` creating a release 👍🏻

EwanValentine: Sorted, here: https://github.com/netlify/cli/pull/4791/commits/a358964b135a645fa47f1a411fe57e1b27c978f2

src/commands/deploy/deploy.js
```diff
@@ -410,7 +410,10 @@ const bundleEdgeFunctions = async (options) => {
     phase: 'start',
   })
 
-  const { severityCode, success } = await runCoreSteps(['edge_functions_bundling'], { ...options, buffer: true })
+  const { severityCode, success } = await runCoreSteps(['edge_functions_bundling', 'edge_functions_produce_eszip'], {
```
eduardoboucas, line 413: I don't think this is doing what you intend. `runCoreSteps` runs a list of core steps, one of which is `edge_functions_bundling`. `edge_functions_produce_eszip` is a feature flag, not a core step.

I think you want to tap into `options.featureFlags`.

eduardoboucas (changes requested):

src/commands/deploy/deploy.js
EwanValentine: Oh, how embarrassing. Fixing now.

EwanValentine: There we go: https://github.com/netlify/cli/pull/4791/commits/6956b93e33229d0cc6c0cdc16edb2e2daf768253 (formatting fixed in subsequent commit).

eduardoboucas (approved):

PR #4802, opened 2022-07-15
fix(graph): regressions introduced during JWT migration

anmonteiro:

#### Summary

Fixes regressions introduced in https://github.com/netlify/cli/pull/4778

---

      passes our tests.

### DIFF
--- a/src/lib/one-graph/cli-client.js
+++ b/src/lib/one-graph/cli-client.js
@@ -403,17 +403,15 @@ const handleCliSessionEvent = async ({
 /**
  *
  * @param {object} input
- * @param {string} input.netlifyToken The (typically netlify) access token that is used for authentication, if any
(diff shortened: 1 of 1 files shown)


src/lib/one-graph/cli-client.js
```diff
@@ -495,16 +492,16 @@ const persistNewOperationsDocForSession = async ({
   siteRoot,
 }) => {
   const { branch } = gitRepoInfo()
+  const { jwt } = await OneGraphClient.getGraphJwtForSite({ siteId, nfToken: netlifyToken })
   const persistedResult = await executeCreatePersistedQueryMutation(
     {
-      nfToken: netlifyToken,
```
danez, line 500: I checked the query and it is not used anymore

danez (approved):
I'm not exactly sure what the change is about, but codewise it makes sense.

PR #4803, opened 2022-07-15
fix(graph): use graphJWT in more places

anmonteiro:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Follow up to #4802 

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #4806, opened 2022-07-15
fix: support IPv4 and IPv6 in dev command

danez:
#### Summary

Fixes #4803 

This fixes support with node 17+ which defaults to IPv6 when resolving `localhost`

The idea in this PR is to listen for `[::1]:port` aswell as `127.0.0.1:port`. This way we do not rely on the name resolution of `localhost` and support IPv4 and IPv6. That way we now also support frameworks that in theory spin up their dev server with IPv6 (which I haven't seen yet).

I tested with astro (which needed some patches to make ipv6 in their dev server work) and it worked fine.

I have to say though that I wasn't able to reproduce the behavior that node resolves `localhost` to `::1`, on my machine it always resolves to `127.0.0.1`.

Is anyone able to reproduce the behavior?


anmonteiro (dismissed):
I can repro on master with `v18.5.0` but not in this branch. Thank you!

danez:
> I have a couple questions about edge case behavior here:
> 
> - Does `Promise.race` cancel promises that don't win the race?
>   - If it doesn't, we will be leaking e.g. a `::1` waiting promise that just keeps running forever

We should test this. i don't think it will cause any side effects and will simply timeout at some point.

anmonteiro:
I'm worried about this pathological case:

```js
const promise1 = new Promise((resolve, reject) => {
  setTimeout(resolve, 500, 'resolved');
});

const promise2 = new Promise((resolve, reject) => {
  setTimeout(reject, 100, 'threw');
});

Promise.race([promise1, promise2]).then((value) => {
  console.log(value);
}).catch(x => console.log('caught', x));
```

Running the code above will log `caught threw`, which is a problem if one of the promises throws before the other one resolves.

For this specific case it probably doesn't matter (barring any `wait-port` bugs), given that `wait-port` handles errors itself, and, as you mentioned, we can probably just wait until the other (pending) promise times out.

danez:
Yes, you are absolutely right. I mixed two different things here. For one I fixed the issue that people are having with node 17+, but I also added new support for IPv6 which I do not think we really need (yet) as no framework is using that by default.

So I reverted the new IPv6 support and only fixed the problem now.

I also created https://github.com/dwmkerr/wait-port/pull/82, but there is no need to wait for that, maybe it is better not to use name resolution anyway.

Thanks for all the feedback @anmonteiro

anmonteiro (approved):
Confirmed that this fix also works, just like the previous `Promise.race` one.

Issue #4819, opened 2022-07-18
Incorrect behaviour of "dev --country" command


MarcL:
### Describe the bug

When using `ntl dev --country=<2-letter-iso-code>` it fails to set `geo=mock` as [suggested in the documentation](https://cli.netlify.com/commands/dev).

> `country` (string) - Two-letter country code (ISO 3166-1 alpha-2, https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements) to use as mock geolocation (enables --geo=mock automatically)

Instead it reads the cached or real version of the country code if it exists.

### Steps to reproduce

This was reproduced with the [Astro starter kit](https://github.com/sarahetter/astro-netlify-edge-starter).

1. Clone the repository `git clone git@github.com:sarahetter/astro-netlify-edge-starter.git`
2. Install dependencies: `npm install`

### Incorrect behaviour
1. Start the dev server using `ntl dev --country=AU` to choose Australia as the country code
2. Observe the page at http://localhost:8888 - the message displayed will be "Welcome, visitor from United Kingdom!" (this will be shown as your own country name instead of "United Kingdom")
3. This is incorrect behaviour as the geolocation has not been mocked with the Australia country code
4. Using `console.log(context);` to log the edge function context shows my UK-based geolocation:
```js
  geo: {
    city: "Warrington",
    country: { code: "GB", name: "United Kingdom" },
    subdivision: { code: "ENG", name: "England" }
  }
```

### Correct behaviour by adding the geo command
1. Start the dev server using `ntl dev --country=AU --geo=mock` to choose Australia as the country code **AND** to enable geolocation mocking
2. Observe the page at http://localhost:8888 - the message displayed will be "Welcome, visitor from Mock Country!"
3. Using `console.log(context);` to log the edge function context now shows:
```js
  geo: {
    city: "Mock City",
    country: { code: "AU", name: "Mock Country" },
    subdivision: { code: "SD", name: "Mock Subdivision" }
  }
```

### Configuration

```toml
[build]
command = "npm run build"
publish = "dist/"


[[edge_functions]]
path = "/"
function = "geolocation"
```

### Environment

```shell
  System:
    OS: macOS 12.4
    CPU: (10) x64 Apple M1 Pro
    Memory: 267.52 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.19.3 - ~/.nvm/versions/node/v14.19.3/bin/node
[...]

PR #4821, opened 2022-07-18
fix: fix cache error with geo country mock

JWhist:

#### Summary

Fixes [error](https://github.com/netlify/cli/issues/4819) by validating that no `geoCountry` value is set before checking the cache.

---

      passes our tests.


src/lib/geo-location.js
```diff
@@ -43,7 +43,7 @@ const getGeoLocation = async ({ geoCountry, mode, offline, state }) => {
 
   // If we have cached geolocation data and the `--geo` option is set to
   // `cache`, let's try to use it.
-  if (cacheObject !== undefined && mode === 'cache') {
+  if (cacheObject !== undefined && mode === 'cache' && !geoCountry) {
```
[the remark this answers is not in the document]
eduardoboucas: I'm not sure there's ever any situation where you'd rather receive dummy data than an actual location from the country you're trying to emulate. But I'll defer to you.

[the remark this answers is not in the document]
JWhist: No, after thinking about it, I agree with you.  I think I was originally looking down the road a bit more towards mocking subdivisions/states at which point the caching might be counterproductive.  But seeing as that is probably either not going to happen, or a ways off at this point, I think your approach makes sense.  I'll abstract out that logic into a function of its own, and it can be modified later if needed.

src/lib/geo-location.js
```diff
@@ -27,6 +27,9 @@ const mockLocation = {
   subdivision: { code: 'CA', name: 'California' },
 }
 
+const isCacheEligible = (cacheObj, mode, country) =>
+  cacheObj !== undefined && mode === 'cache' && (cacheObj.data.country.code === country || !country)
```
eduardoboucas, line 31: If you run `--country=XX`, shouldn't `mode` become `mock`? That's what the docs say is happening, and I wonder if reflecting that in the code would make this simpler, because then you wouldn't need to change this check at all?

MarcL:
> New caching function allows for the behavior like this in the event that a user mocks their own country:
> 
> 1. If a user passes `--country` alone with no `--geo` flag, then the `mode` will default to `'cache'` and will use the cache object if the country matches that of the argument, otherwise the usual mock country.
> 2. If a user passes `--country --geo=mock` then the cache will always be bypassed.

Does this now need a documentation update @JWhist? The docs suggest that not passing the`--geo` flag will default to `mock` and not `cache`.

src/lib/geo-location.js
```diff
@@ -40,10 +40,11 @@ const mockLocation = {
  */
 const getGeoLocation = async ({ geoCountry, mode, offline, state }) => {
   const cacheObject = state.get(STATE_GEO_PROPERTY)
+  if (geoCountry) mode = 'mock'
```
eduardoboucas, line 43: I think adding a comment here would help.

```suggestion

  // If `--country` was used, we also set `--mode=mock`. 
  if (geoCountry) {
    mode = 'mock'
  }
```

src/lib/geo-location.js
```diff
@@ -40,10 +40,11 @@ const mockLocation = {
...
 
   // If we have cached geolocation data and the `--geo` option is set to
   // `cache`, let's try to use it.
-  if (cacheObject !== undefined && mode === 'cache') {
+  if (cacheObject !== undefined && (mode === 'cache' || cacheObject.data.country.code === geoCountry)) {
```
eduardoboucas, line 47:
```suggestion
  // Or, if the country we're trying to mock is the same one as we have in the
  // cache, let's use the cache instead of the mock.
  if (cacheObject !== undefined && (mode === 'cache' || cacheObject.data.country.code === geoCountry)) {
```

eduardoboucas (approved):

MarcL (approved):
👍🏻 Thanks for the fix.

PR #4839, opened 2022-07-21
feat: read scopes and contexts in `env:get`, `env:list`, and `dev`

jasonbarry:
#### Summary
This PR adds support for environment variable scopes and contexts for the `env:get`, `env:list`, and `dev` commands. 

`env:get` and `env:list` support both the `--scope` and `--context` flags if the site has been configured with the new Environment Variables experience:tm:, aka Envelope. If the site is not configured with Envelope and either of those flags are passed, an error message will print. If this site is not configured with Envelope and neither of those flags are passed, it will behave as it normally has. 

A new Scope column has been added to `env:list` if the site is on Envelope: 
Screenshot 2022-07-27 at 22 29 27

The `--context` flag for `netlify dev` is now supported, which will pull in values from Envelope while respecting the precedence of context values in netlify.toml. 

One caveat is that there will be a duplicate GET request sent to Envelope due to [this change in netlify/config](https://github.com/netlify/build/pull/4329). Since we're adding context and scope options on the CLI level, I thought it would make sense to move the requests to happen in the CLI. Once all the calls to Envelope land in CLI, it should be safe to remove in netlify/config to remove the redundancy. 

Fixes https://github.com/netlify/pillar-workflow/issues/755
Fixes https://github.com/netlify/pillar-workflow/issues/734

---

## Test instructions

- Checkout this branch and install dependencies
- Remove your existing global install -- if installed with npm, `npm uninstall -g netlify-cli`
- `npm link`
- Give `env:get`, `env:list`, and `dev` a whirl. I recommend trying with a site opted-in to the new Env Vars experience in the UI (enable in Labs) 

---

      passes our tests.

Screenshot 2022-07-24 at 21 41 04


src/utils/dev.js
```diff
@@ -34,7 +34,7 @@ const ENV_VAR_SOURCES = {
     printFn: chalk.red,
   },
   ui: {
-    name: 'build settings',
+    name: 'site settings',
```
jasonbarry, line 37: Per [this conversation](https://netlify.slack.com/archives/C02TE7L1QUA/p1658512126111859)

jasonbarry:
@codebyuma I would very much appreciate a copy review in this PR for options outputted by the `--help` flag, and log/error messages 🙏 . In the diff, I suggest clicking the "Show file tree" button and looking at `.js` files in `src/commands/*`. The copy in the `docs/` directory is auto-generated from the commands.

src/commands/env/env-get.js
```diff
@@ -16,9 +19,21 @@ const envGet = async (name, options, command) => {
...
+  const { siteInfo } = cachedConfig
+  let { env } = cachedConfig
+
+  if (siteInfo.use_envelope) {
+    env = await getEnvelopeEnv({ api, context, env, key: name, scope, siteInfo })
+  } else if (context !== 'dev' || scope !== 'any') {
+    error(
+      `To specify a context or scope, please run ${chalk.yellowBright(
+        'netlify open:admin',
+      )} and opt in to the new Environment Variables experience`,
```
codebyuma, line 31: [Sand] I think when Melanie and I talked about this a while back, we thought it would be helpful to be more specific about what happens when you run the command and where to go:

```suggestion
      )} to open the Netlify UI and opt in to the new Environment Variables experience from Site settings`,
```

They won't see the env vars section if they don't enable it in Labs first. But I think it's ok to say Site settings here because they only need to enable it in Labs once and they'll have to go back to Site settings for every site they want to enable afterwards.   What do you think?

src/commands/env/env-get.js
```diff
@@ -16,9 +19,21 @@ const envGet = async (name, options, command) => {
...
   }
 
-  const siteData = await api.getSite({ siteId })
+  const { siteInfo } = cachedConfig
+  let { env } = cachedConfig
+
+  if (siteInfo.use_envelope) {
+    env = await getEnvelopeEnv({ api, context, env, key: name, scope, siteInfo })
+  } else if (context !== 'dev' || scope !== 'any') {
+    error(
```
codebyuma, line 28: Did we add a similar error for when a user tries to use the scope and context flags with the env:set command?

src/commands/env/env-get.js
codebyuma: Also, in the API docs, we lowercased `environment variables experience`, let's do that here to keep it consistent. 

```suggestion
      )} to open the Netlify UI and opt in to the new environment variables experience from Site settings`,
```

src/commands/env/env-list.js
```diff
@@ -33,36 +47,55 @@ const envList = async (options, command) => {
...
+  const isUsingEnvelope = siteInfo.use_envelope
+  let environment = env
+
+  if (isUsingEnvelope) {
+    environment = await getEnvelopeEnv({ api, context, env, scope, siteInfo })
+  } else if (context !== 'dev' || scope !== 'any') {
+    error(
+      `To specify a context or scope, please run ${chalk.yellowBright(
+        'netlify open:admin',
+      )} and opt in to the new Environment Variables experience`,
```
codebyuma, line 60: [Sand] Similar comment above - whatever we decide, we should be consistent. I'll suggest the same change here in case that's the one we want to go with

```suggestion
      )} to open the Netlify UI and opt in to the new environment variables experience from Site settings`,
```

src/commands/env/env-get.js
codebyuma: What happens if they try to use the scope flag but are not on a Pro account or higher?

src/commands/env/env-get.js
```diff
@@ -27,7 +42,11 @@ const envGet = async (name, options, command) => {
   }
 
   if (!value) {
-    log(`Environment variable ${name} not set for site ${siteData.name}`)
+    log(
+      `No value set in the ${chalk.magentaBright(context)} context for environment variable ${chalk.yellowBright(
```
codebyuma, line 46: I haven't tested yet but do we need a separate message or need to add to this message for when the scope flag is used?

src/commands/env/env-get.js
```diff
@@ -43,6 +62,16 @@ const createEnvGetCommand = (program) =>
...
     .command('env:get')
     .argument('<name>', 'Environment variable name')
+    .addOption(
+      new Option('-c, --context <context>', 'Specify a deploy context')
+        .choices(['production', 'deploy-preview', 'branch-deploy', 'dev'])
+        .default('dev'),
+    )
+    .addOption(
+      new Option('-s, --scope <scope>', 'Specify a scope')
+        .choices(['builds', 'functions', 'post_processing', 'runtime', 'any'])
```
codebyuma, line 72: One more functionality question (this will impact the docs) - can they filter for variables that have `all` set as the scope? (which is different than `any`, right?)

src/commands/env/env-get.js
```diff
@@ -43,6 +62,16 @@ const createEnvGetCommand = (program) =>
   program
     .command('env:get')
     .argument('<name>', 'Environment variable name')
+    .addOption(
+      new Option('-c, --context <context>', 'Specify a deploy context')
+        .choices(['production', 'deploy-preview', 'branch-deploy', 'dev'])
```
codebyuma, line 67: Is `all` an option for the context flag? Asking as I'll have to update the docs PR if not

src/commands/env/env-list.js
```diff
@@ -89,6 +122,16 @@ const envList = async (options, command) => {
 const createEnvListCommand = (program) =>
   program
     .command('env:list')
+    .addOption(
+      new Option('-c, --context <context>', 'Specify a deploy context')
```
codebyuma, line 126: Similar questions for the context and scope flags here as I left for the get command

codebyuma (commented):
It's exciting to see this coming together! I haven't finished reviewing/testing but I need to run for an appointment, so I thought I'd submit the comments I have so far. I'll come back to this when I'm back

src/commands/env/env-get.js
jasonbarry: Nice, I prefer mentioning Site settings over Labs too -- the feature might not always be in Labs, so we should avoid a temporary location b/c the user might be living on an old / out-of-date version of the CLI.

src/commands/env/env-get.js
jasonbarry: > Did we add a similar error for when a user tries to use the scope and context flags with the env:set command?

Not yet, but we will! `env:set` doesn't support `--context` or `--scope` flags yet, but it's what I'm working on next. 

> What happens if they try to use the scope flag but are not on a Pro account or higher? 

Plan-specific behavior is managed in the API, so the CLI will output any error that the API provides in this case.

src/commands/env/env-get.js
jasonbarry: It's not -- I think if `all` were an option, I'd expect the output to contain all the values across all contexts to be returned, which could be more than 1. If you have an env var value set to the `all` context, and you provide a specific context like `production` in the `--context` flag, it will still match the `all` value. That is to say, if you pass e.g. `--context dev`, that will match the env var value if it is either `dev` or `all`.

src/commands/env/env-get.js
jasonbarry: I interpret `all` to mean "which env vars are available in builds and functions and post_processing and runtime scopes", vs `any` to mean "which env vars are available in builds or functions or post_processing or runtime scopes". I think `any` is more desirable when getting, and `all` more desirable when setting. 

Note that `any` doesn't exist on context, because context is on the value level (where there could be multiple things returned), whereas scope is on the env var level (where there can only be one thing returned).

src/commands/env/env-get.js
jasonbarry: Great suggestion! I'll add this

src/commands/env/env-get.js
codebyuma: I see what you're saying. Ultimately. If you're using the flag, you're looking for the value from one specific context.

src/commands/env/env-get.js
codebyuma: I agree with the distinction between `all` and `any`, so it makes sense for the default to be `any` here.

I thought of `all` as people might want to filter for only those that are set to `all` 🤔 But I actually don't know which use case is more valuable, let's ping @melanierichards for this one.

codebyuma: > Note that any doesn't exist on context, because context is on the value level (where there could be multiple things returned), whereas scope is on the env var level (where there can only be one thing returned).

This makes sense to me!

melanierichards: Both `any` and `all` seem to me to have their place.

jasonbarry:
@netlify/netlify-dev this is ready for a code review now, if one of y'all can spare a moment! ❤️

lukasholzer (approved):
THIS is an outstanding PR 👏 Very nicely documented and well tested! 💯 
🐳

jasonbarry:
Why thank you @lukasholzer!! I realized a test was failing due to some capitalization of strings in logging 🤦 [here's the change](https://github.com/netlify/cli/pull/4839/commits/7b62b1893d376343fb4251af41f9a17e12d2c05a). Would you mind doing a re-review?

akahn (approved):

PR #4841, opened 2022-07-22
feat: add retry count parameter to function uploads

eduardoboucas:
#### Summary

Adds a retry count parameter to the `uploadDeployFunction` API call on retries.

Part of https://github.com/netlify/pod-compute/issues/140.

### DIFF
--- a/src/utils/deploy/upload-files.js
+++ b/src/utils/deploy/upload-files.js
@@ -75,7 +75,6 @@ const uploadFiles = async (api, deployId, uploadList, { concurrentUpload, maxRet
 const retryUpload = (uploadFn, maxRetry) =>
   new Promise((resolve, reject) => {
     let lastError
-    let retryCount = 0
 
     const fibonacciBackoff = backoff.fibonacci({
       randomisationFactor: UPLOAD_RANDOM_FACTOR,
@@ -83,14 +82,13 @@ const retryUpload = (uploadFn, maxRetry) =>
       maxDelay: UPLOAD_MAX_DELAY,
     })
 
-    const tryUpload = async () => {
+    const tryUpload = async (retryIndex = -1) => {
       try {
-        const results = await uploadFn(retryCount)
+        const results = await uploadFn(retryIndex + 1)
 
         return resolve(results)
       } catch (error) {
         lastError = error
-        retryCount += 1
 
         // observed errors: 408, 401 (4** swallowed), 502
         if (error.status >= 400 || error.name === 'FetchError') {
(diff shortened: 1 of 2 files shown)


src/utils/deploy/upload-files.js
```diff
@@ -79,10 +85,12 @@ const retryUpload = (uploadFn, maxRetry) =>
 
     const tryUpload = async () => {
       try {
-        const results = await uploadFn()
+        const results = await uploadFn(retryCount)
         return resolve(results)
```
danez, line 89: [dust] new line before return

src/utils/deploy/upload-files.js
```diff
@@ -79,10 +85,12 @@ const retryUpload = (uploadFn, maxRetry) =>
 
     const tryUpload = async () => {
```
danez, line 86: The first argument is a counter, see https://github.com/MathieuTurcotte/node-backoff#event-ready could we use that? Only thing is that I think by reading the docs it starts at 0 for the first backoff, so we would need to do +1.

```suggestion
    const tryUpload = async (retryCount = 0) => {
```

danez:
Also a test would be nice.

src/utils/deploy/upload-files.js
eduardoboucas: Hmm maybe I'm missing something, but I'm not seeing any parameters being passed to the function.

src/utils/deploy/upload-files.js
eduardoboucas: Tasting my own poison! 🙈 

Done in 400da760aa283649d2145643f0cfcb11a0b4876c.

src/utils/deploy/upload-files.js
danez: https://github.com/MathieuTurcotte/node-backoff/blob/master/lib/backoff.js#L53 The event is passed some params, and as we use this function for the event we could use them.

eduardoboucas:
> chore: add test

Done in https://github.com/netlify/cli/pull/4841/commits/6bec2f54656cef01b1121b4a8e34acd76c12f193.

src/utils/deploy/upload-files.js
eduardoboucas: Right, but I'm logging the actual parameters that are passed to `tryUpload` and nothing comes up.

eduardoboucas: Ah, my bad. I was looking in the wrong place. Will update!

eduardoboucas: Done in https://github.com/netlify/cli/pull/4841/commits/6fed5241e61fa5f47903e7f25d1cca1ea1fa4e4b.

danez (approved):
🚀

PR #4842, opened 2022-07-22
feat: deprecate `netlify:edge` in function templates

eduardoboucas:
#### Summary

Replaces `netlify:edge` with `https://edge.netlify.com` in Edge Functions templates.

Part of https://github.com/netlify/pillar-runtime/issues/322.

### DIFF
--- a/src/functions-templates/typescript/geolocation/{{name}}.ts
+++ b/src/functions-templates/typescript/geolocation/{{name}}.ts
@@ -1,4 +1,4 @@
-import { Context } from "netlify:edge";
+import { Context } from "https://edge.netlify.com";
 
 export default async (request: Request, context: Context) => {
   // Here's what's available on context.geo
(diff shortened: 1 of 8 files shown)


src/functions-templates/typescript/geolocation/{{name}}.ts
```diff
@@ -1,4 +1,4 @@
-import { Context } from "netlify:edge";
+import { Context } from "https://edge.netlify.com";
```
danez, line 1:
```suggestion
import type { Context } from "https://edge.netlify.com";
```

src/functions-templates/typescript/transform-response/{{name}}.ts
```diff
@@ -1,4 +1,4 @@
-import { Context } from "netlify:edge";
+import { Context } from "https://edge.netlify.com";
```
danez, line 1:
```suggestion
import type { Context } from "https://edge.netlify.com";
```

danez (commented):
LGTM, just to small things while we are looking at these files

eduardoboucas:
> LGTM, just to small things while we are looking at these files

I'd rather do that separately, if you don't mind. These templates come mostly from https://edge-functions-examples.netlify.app/, so if there's something in them we want to change, we should update stream as well.

Let's do a more holistic review of these templates as a separate initiative.

danez (approved):

PR #4857, opened 2022-07-25
feat: pass site information to edge function context in dev command

### LINKED ISSUE #4775: edge functions: supply context.site information to edge functions

In https://github.com/netlify/pillar-runtime/issues/399, we tracked adding a `context.site` object that gives edge functions access to information like Site ID, URL or Name. We should also provide this information in local development.

In https://github.com/netlify/cli/blob/8248c355b3af1060a4a4995614a1e98bfca87270/src/lib/edge-functions/proxy.js#L94, we should also set `X-NF-Site-Info` according to the parsing logic in https://github.com/netlify/edge-functions-bootstrap/blob/346ef35335b6874d133927b6e9432980eebdcfba/src/bootstrap/function_chain.ts#L50 (JSON object `id`, `name` and `url`; base64-encoded).

### Pull request (optional)

- [X] I can submit a pull request.

khendrikse:
#### Summary

Fixes #4775 

Passing site information to the Edge Functions context object when using the dev command. 
---

      passes our tests.

confused chicken


tests/integration/100.command.dev.test.js
```diff
@@ -252,6 +253,69 @@ test('Serves an Edge Function with a rewrite', async (t) => {
   })
 })
 
+test('Serves an Edge Function that includes context with site information', async (t) => {
+  await withSiteBuilder('site-with-fully-qualified-redirect-rule', async (builder) => {
```
eduardoboucas, line 257: Not a big deal, but you could change this name to something a bit more aligned with your test.

eduardoboucas (dismissed):
Great work! ✨

tests/integration/100.command.dev.test.js
khendrikse: Do you mean specifying what type of information it returns? 🤔

khendrikse: Wait, do you mean the test name or the name passed to `withSiteBuilder`? 🤔

eduardoboucas, line 257: I mean something like this, since the site doesn't actually have a fully qualified redirect rule. (I notice we've been copy/pasting that test a lot, so the name is in a bunch of places incorrectly 🙈)

```suggestion
  await withSiteBuilder('site-with-edge-function-printing-site-info', async (builder) => {
```

khendrikse: Ah yes I was wondering about that! Yeah I'd love to change that. Will do! Thanks

eduardoboucas (dismissed):

khendrikse:
@eduardoboucas I renamed the other tests withSiteBuilder as well

eduardoboucas (dismissed):

src/lib/edge-functions/proxy.js
```diff
@@ -91,8 +92,9 @@ const initializeProxy = async ({
 
     if (!registry) return
 
-    // Setting header with geolocation.
+    // Setting header with geolocation and site info.
     req.headers[headers.Geo] = JSON.stringify(geoLocation)
+    req.headers[headers.Site] = btoa(JSON.stringify({ id: siteInfo.id, name: siteInfo.name, url: siteInfo.url }))
```
danez, line 97: Where is `btoa` coming from? It is not imported or anything and I do not think it is a global.

danez: I see https://developer.mozilla.org/en-US/docs/Web/API/btoa, but it is only available as of node 16. So we cannot use it.

khendrikse: Hopefully that is the problem! Thank you so. much. 

Somehow because of the use of `atob()` on the receiving end I thought it was ok but I should have looked at the version support.

Will be diving into it!

khendrikse: 👍 that was it!

eduardoboucas: Ah, good catch!

src/lib/edge-functions/proxy.js
```diff
@@ -91,8 +93,11 @@ const initializeProxy = async ({
 
     if (!registry) return
 
-    // Setting header with geolocation.
+    // Setting header with geolocation and site info.
     req.headers[headers.Geo] = JSON.stringify(geoLocation)
+    req.headers[headers.Site] = Buffer.from(
```
eduardoboucas, line 98: I would maybe extract this into a separate function that returns the value of the header and has these ops in separate stages (i.e. creating the object, JSON-stringifying, then base64-encoding).

But that's totally not a blocker, and we can always revisit that later. Let's ship! 🚀

eduardoboucas (approved):

src/lib/edge-functions/proxy.js
khendrikse: Yeah! And then we'd be able to unit test it ✨

PR #4860, opened 2022-07-25
fix: correctly create edge function directory when running in custom cwd

danez:
#### Summary

Fixes netlify/netlify-plugin-nextjs#1400
Fixes netlify/pod-compute#150

The edge function dist path was incorrect when a custom working directory `--cwd` was supplied through the cli options. We were looking at `/repo/path/.netlify/...` instead of `/repo/path/cwd/.netlify/...`

Added a new test (which failed without the fix) and adjusted the other test to be more clear what failed. Just getting the uppercase/lowercase error, usually made me think this is some unimportant server/proxy conversation, but it actually means the edge function is not being loaded. Should be clearer now.

download

### DIFF
--- a/tests/integration/210.command.deploy.test.js
+++ b/tests/integration/210.command.deploy.test.js
@@ -28,12 +28,12 @@ const validateContent = async ({ content, path, siteUrl, t }) => {
   }
 }
 
-const validateDeploy = async ({ content, deploy, siteName, t }) => {
+const validateDeploy = async ({ content, contentMessage, deploy, siteName, t }) => {
   t.truthy(deploy.site_name)
   t.truthy(deploy.deploy_url)
   t.truthy(deploy.deploy_id)
   t.truthy(deploy.logs)
-  t.is(deploy.site_name, siteName)
+  t.is(deploy.site_name, siteName, contentMessage)
 
   await validateContent({ siteUrl: deploy.deploy_url, path: '', content, t })
 }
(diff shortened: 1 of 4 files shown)


tests/integration/210.command.deploy.test.js
```diff
@@ -124,17 +124,12 @@ if (process.env.NETLIFY_TEST_DISABLE_LIVE !== 'true') {
...
         })
         .withEdgeFunction({
-          handler: async (_, context) => {
-            const resp = await context.next()
-            const text = await resp.text()
-
-            return new Response(text.toUpperCase(), resp)
-          },
-          name: 'yell',
+          handler: async () => new Response('Edge Function works'),
```
eduardoboucas, line 131: Not a blocker at all, but I'm curious about why you changed the test here. The previous test edge function was a bit more comprehensive, in that it was calling the origin and making a transformation, whereas the new one is just delivering a response.

eduardoboucas (approved):
Great!

tests/integration/210.command.deploy.test.js
danez: Isn't that what we are testing here? That the function was deployed correctly and responded?

danez: I guess I can revert it, but I thought this is simpler and does exactly what we are trying to test here. And I guess/hope we have some tests somewhere that asserts that the edge internals work.

eduardoboucas: > I can revert it if you want, but I thought this is simpler and does exactly what we are trying to test here. And I guess/hope we have some tests somewhere that asserts that the edge internals work.

No need to revert.

eduardoboucas (approved):

PR #4870, opened 2022-07-27
test: move site info header creation to function and write unit test for function

khendrikse:
#### Summary

As per request, I moved some logic to its own function and added a unit test.

I'd love to get input! Also on the style that I did it in. For example, would it have been better in its own file? (if so, how do we want to structure it? 🤔) and what do we think of the naming. (is it creating a header? or would it be better to name it `createSiteInfoBase64`?

---

      passes our tests.

### DIFF
--- a/src/lib/edge-functions/proxy.js
+++ b/src/lib/edge-functions/proxy.js
@@ -46,6 +46,13 @@ const handleProxyRequest = (req, proxyReq) => {
   })
 }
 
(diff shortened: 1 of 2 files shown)


eduardoboucas (approved):
✨

eduardoboucas:
Small note: we should prefix all commits with `chore:`, `fix:` or `feat:`. In this case, it could probably be a `chore:`. One thing to note is that only `fix:` and `feat:` commits will generate a release.

PR #4892, opened 2022-08-01
chore: fix vale workflow

### LINKED ISSUE #4831: Fix vale workflow, which is currently broken
We updated the vale-action to v2, but it broke because it is not downloading the styles anymore automatically.

https://github.com/netlify/cli/blob/main/.github/workflows/vale.yml

So in order to fix this now we need to download and unpack the styles ourself in the workflow before running vale.

The download can be similar done as mentioned here: https://github.com/netlify/cli/blob/main/CONTRIBUTING.md#lint-docs-per-netlify-style-guide

Ref: netlify/vale-library#51

danez:
#### Summary

Download styles manually and fix the inputs of the vale-action. v2 has different inputs than v1

Created a test commit and it works now again:

Screenshot 2022-08-01 at 14 58 11

Fixes #4831

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -24,7 +24,6 @@ jobs:
 
       - name: Vale
         uses: errata-ai/vale-action@v2
-        continue-on-error: true
         with:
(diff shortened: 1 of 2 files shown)


anmonteiro (approved):
Thank you!

PR #4918, opened 2022-08-10
chore: print stdout/stderr when dev server fails in tests

danez:
#### Summary

This adds a new log message once the edge function server is up and running. The tests wait for this before doing any requests, so we avoid timing issues when we request edge functions but the edge function server is not yet up.

The other improvement is that we log stdout and stderr when a test fails. This gives a lot more information and usually includes the reason why a request failed.

vector-illustration-little-cute-red-ant-vector-illustration-little-cute-red-ant-smiling-116379979


src/lib/edge-functions/proxy.js
```diff
@@ -175,6 +182,9 @@ const prepareServer = async ({
       inspectSettings,
       port,
     })
+
+    log(`${NETLIFYDEVLOG} Edge function server running`)
```
eduardoboucas, line 186: I'm not sure how useful this information will be to users, and I don't think we should be adding user-facing logs just to serve our tests. How do you feel about only printing this if `debug` mode is on (and adjusting the tests so that they use that flag)? Alternatively, we could print this if `process.env.NODE_ENV === 'test'`?

danez: That makes sense

tests/integration/utils/dev-server.js
```diff
@@ -38,17 +40,26 @@ const startServer = async ({ cwd, offline = true, env = {}, args = [], expectFai
...
+  const errorBuffer = []
   const serverPromise = new Promise((resolve, reject) => {
     let selfKilled = false
+    ps.stderr.on('data', (data) => {
+      errorBuffer.push(data)
+    })
     ps.stdout.on('data', (data) => {
       outputBuffer.push(data)
-      if (!expectFailure && data.includes('Server now ready on')) {
+      if (!expectFailure && data.includes('Edge functions server running')) {
```
eduardoboucas, line 51: My understanding is that we were looking for `Server now ready on` because that's a line we always print in the dev server. We're now changing that to a string that only applies when we use edge functions? Can you add a bit more context here?

danez: The main problem here is that when we print `Server now ready on` the edge function server is not yet up and running, because the start function is [async but we do not await it.](https://github.com/netlify/cli/blob/main/src/lib/edge-functions/proxy.js#L74-L77). This makes total sense for users, not so much the tests where we pretty much instantly start the assertions/requests/etc of the tests once this is printed.

Somehow I had the perception yesterday that this message is printed after the edge function proxy server is started. Looking at the code today that is not the case and this change does not make any sense.

Still the question is how can we wait in tests for edge functions to be up before starting requests? Conditionally awaiting the edge functions seems one possibility.

```ts
const server = prepareServer()

if (NODE_ENV === test) {
 await server;
}
```

eduardoboucas: I don't like the idea of adding test-specific logic to the runtime code. Yes, it would be cool to have a reliable way for the tests to wait for the edge functions server, but if we can't do it without changing the runtime code itself, I'd rather continue with polling and flaky tests, personally.

danez: Yeah, I reverted this change and now this PR is only printing stdout and stderr on dev server failure.

The only other thing I added is to resolve in the next [event loop iteration](https://github.com/netlify/cli/pull/4918/files#diff-b0b780c0433e4bc71b3ff83d29b4232e774ea6ce1ce12abe77eb071d00892640R52), to give the edge function server slightly more time to boot.

tests/integration/600.framework-detection.test.js
```diff
@@ -12,9 +12,6 @@ const { normalize } = require('./utils/snapshots')
 
 const content = 'Hello World!'
 
-// Remove extra '.'s resulting from different load times which lead to flaky snapshots
-const frameworkDetectionNormalizer = (output) => normalize(output.replace(/\.+(?=\.)/, ''))
-
```
danez, line 17: Removed because we do not use the dot reporter in `wait-port` #4730

PR #4919, opened 2022-08-10
fix: Update edge-bundler and ensure only allowed environment variables are supplied to edge functions

### LINKED ISSUE #4614: edge-functions: Deno.env should contain the same environment variables as in production
### Describe the bug

Running edge functions locally with `netlify dev`, `Deno.env` contains all environment variables inherited from the CLI's shell. This is different from production, where it will only contain env vars specified in the Netlify UI, and two Deno-specific ones. This should be replicated in local dev.

One possible solution is to work with downstream `@netlify/edge-bundler`, and pass the `--allow-env=...` flag to only allow access to the environment variables that'd also exist in production:

https://github.com/netlify/edge-bundler/blob/0d2c6af674d30f0322bf376ad66acb0d5b5b0124/src/server/server.ts#L113

### Steps to reproduce

Write an edge function, log `Deno.env.toObject()`. It should contain just placeholders for `DENO_REGION` and `DENO_DEPLOYMENT_ID` (probably not even that), but instead it contains all env vars inherited from the CLI's shell.

### Configuration

_No response_

### Environment

not relevant

danez:
#### Summary

Fixes #4614 

This adds the environment variables from the ui and the two DENO environment variables for edge-functions.


src/lib/edge-functions/registry.js
```diff
@@ -178,6 +184,20 @@ class EdgeFunctionsRegistry {
     return declarations
   }
 
+  static getEnvironmentVariables(envConfig) {
+    const env = Object.create(null)
+    Object.entries(envConfig).forEach(([key, variable]) => {
+      if (variable.sources.includes('ui')) {
```
eduardoboucas, line 190: Are you sure we only want to include variables with source of `ui`? Wouldn't it be better to exclude the ones with only `process`? See https://github.com/netlify/cli/blob/main/src/utils/dev.js#L14-L40.

danez: I added account and addons.
I tested account and they are available in deno.
For addons I would assume they are, but couldn't find an addon that sets environment variables and works with edge functions. split testing sets envs but edge functions are not triggered.
general & configFile & process is not available in deno

eduardoboucas (approved):

src/lib/edge-functions/registry.js
```diff
@@ -178,6 +184,24 @@ class EdgeFunctionsRegistry {
...
+        variable.sources.includes('ui') ||
+        variable.sources.includes('account') ||
+        variable.sources.includes('addons')
+      ) {
+        env[key] = variable.value
+      }
+    })
+
+    env.DENO_REGION = 'local'
+    env.DENO_DEPLOYMENT_ID = 'xxx='
```
eduardoboucas, line 200: I just remembered that this is used in the bootstrap layer to detect whether we're running in production or not (see https://github.com/netlify/edge-functions-bootstrap/blob/main/src/bootstrap/environment.ts#L2).

We might want to remove this variable until we figure out a better way to do this detection.

eduardoboucas (changes requested):

src/lib/edge-functions/registry.js
danez: Good catch. I was wondering something similar too and if I should add `NETLIFY_DEV` but I figured that `DENO_REGION === 'local'` would do the same, could we use it in the bootstrap?

Anyway commenting for now.

eduardoboucas (approved):

mlafeldt:
> I just remembered that this is used in the bootstrap layer to detect whether we're running in production or not (see https://github.com/netlify/edge-functions-bootstrap/blob/main/src/bootstrap/environment.ts#L2).

For non-Netlify employees: https://github.com/mlafeldt/netlify-edge-bootstrap/blob/main/modules/edge-bootstrap.netlify.app/bootstrap/environment.ts#L2 😄 

Are you ever going to make that repo public?

PR #4929, opened 2022-08-12
fix: fix local builds with plugins and build settings from the ui

danez:
#### Summary

This is a long story. I was debugging yesterday for hours.

##### Problem

The initial problem was that I tried deploying a nextjs site from the cli and it did not work. The setup is as following:

- project cloned from [nextra](https://github.com/shuding/nextra) a nextjs documentation application
- No local config (`netlify.toml`)
- in the netlify UI I configured github connection, build command and publish directory.
- `netlify link` done

When I tried deploying this site, the nextjs-plugin would be installed automatically, but during the build at first the `onPreBuild` would run for the nextjs-plugin. Next the `onBuild` for the nextjs-plugin woudl run and this fails because the build command was not triggered at all.

##### Investigation

After hours of trying to understand `@netlify/build` and it's connection to the cli and to `@netlify/config` I found out what is happening. Let me try to summarize:

1. The cli build command is triggered
2. In the cli the [base-command calls](https://github.com/netlify/cli/blob/main/src/commands/base-command.js#L421) `@netlify/config` for the config and gets the correct config back (including ui settings)
3. `@netlify/build` is called with the config supplied from the cli
4. `resolveInitialConfig` is [called](https://github.com/netlify/build/blob/main/packages/build/src/core/config.js#L106) in `@netlify/build` which [respects](https://github.com/netlify/build/blob/1337b1b3d43bc1dce7d1e82a10159cfcd4f72c15/packages/config/src/cached_config.js#L23) the passed in config from the cli. *all fine so far*
5. The `onPreBuild` event is triggered for the build plugin and the plugin makes [changes to the config](https://github.com/netlify/netlify-plugin-nextjs/blob/main/plugin/src/index.ts#L58).
6. After the plugin is finished, `@netlify/build` [computes a config changeset](https://github.com/netlify/build/blob/1337b1b3d43bc1dce7d1e82a10159cfcd4f72c15/packages/build/src/steps/plugin.js#L38) called `configMutation`. This details exactly what the plugin changed in the config, in order to be able to print the changes to the output but also to be always able to reapply these changes again to a config object.
7. We now [call ](https://github.com/netlify/build/blob/main/packages/build/src/steps/update_config.js#L13)`@netlify/config` to get a [fresh copy](https://github.com/netlify/build/blob/1337b1b3d43bc1dce7d1e82a10159cfcd4f72c15/packages/build/src/core/config.js#L126) of the config and with the `configMutations` applied. Because we didn't supplied `siteId` in the first place this now failed to grab anything from the API which is site related. So at this point we basically revert the config to some default values (no build command, publish directory = root directory, etc.) and this is the reason that my build command is never run.

By adding the `siteId` to the input Options it all starts working again, and this is now [inline with buildbot](https://github.com/netlify/buildbot/blob/f4df34fdee5e467fb0a8a2f62e3840ffaba0a9bc/bot/stage_build.go#L101), which also supplies the `siteId`.

Alamy_A14A69-c-29b8790


src/lib/build.js
```diff
@@ -20,6 +22,7 @@ const netlifyBuildPromise = import('@netlify/build')
  */
 const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, json, offline, silent }, token }) => ({
   cachedConfig,
+  siteId: cachedConfig.siteInfo.id,
```
danez, line 25: This, and this only is the fix, the rest is for the tests.

src/lib/build.js
```diff
@@ -42,6 +45,18 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
...
+  // If netlify NETLIFY_API_URL is set we need to pass this information to @netlify/build
+  // TODO don't use testOpts, but add real properties to do this.
+  if (process.env.NETLIFY_API_URL) {
+    const apiUrl = new URL(process.env.NETLIFY_API_URL)
+    const testOpts = {
+      scheme: apiUrl.protocol.slice(0, -1),
+      host: apiUrl.host,
+    }
+    options = { ...options, testOpts }
+  }
```
danez, line 58: This is not nice,.
The `host` could be supplied by [setting](https://github.com/netlify/build/blob/1337b1b3d43bc1dce7d1e82a10159cfcd4f72c15/packages/build/src/core/main.js#L160) `apiHost` but the scheme cannot be changed besides the testOpts.

I'm also not sure if `NETLIFY_API_URL` is only for tests  or if that is meant to be used by users?

Should I add `apiScheme` to `@netlify/build`? Or is it okay this way?

tests/integration/110.command.build.test.js
```diff
@@ -9,7 +9,6 @@ const { withSiteBuilder } = require('./utils/site-builder')
 
 const defaultEnvs = {
   NETLIFY_AUTH_TOKEN: 'fake-token',
-  NETLIFY_SITE_ID: 'site_id',
```
danez, line 12: I opted to not set it for every test here, because it does not really reflect real world scenarios.
If `NETLIFY_SITE_ID` this bug I'm fixing here never surfaces, because `@netlify/build` and `@netlify/config` both also read it.
But most users never set it probably.

ericapisani (approved):
Quite the journey to get this bug sorted. Thanks for the detailed summary and the fix, this LGTM 🚀

PR #4947, opened 2022-08-15
chore: support envelope in graph:init

jasonbarry:
#### Summary

Fixes https://github.com/netlify/pillar-workflow/issues/802 (continuation of https://github.com/netlify/cli/pull/4711)

`netlify graph:init` creates two new environment variables on the site. It currently saves those env vars attached to the site record in Mongo. However, env vars are moving to a separate service (called Envelope) that has its own API and that is detached from the site record. This PR adds logic to detect that if a site has opted into Envelope, to save the newly created env vars to the Envelope service. 

Envelope also supports scoped environment variables, which means you can set env vars to only be present in e.g. builds or only available in functions. I think the ones that Graph sets only need to be in the functions runtime, is that correct? 

---

      passes our tests.

Screenshot 2022-08-15 at 09 32 14


src/commands/graph/graph-init.js
```diff
@@ -29,30 +30,11 @@ const graphInit = async (options, command) => {
...
-  let [netlifyToken, loginLocation] = await getToken()
+  let [netlifyToken] = await getToken()
   if (!netlifyToken) {
     netlifyToken = await command.authenticate()
   }
 
-  let siteData = null
-  try {
-    // @ts-ignore: we need better types for our api object
-    siteData = await api.getSite({ siteId })
```
jasonbarry, line 40: I deleted this block of code because everything we need should already be available in `command.netlify.siteInfo` -- is this an incorrect assumption on my part?

src/commands/graph/graph-init.js
```diff
@@ -65,21 +47,27 @@ const graphInit = async (options, command) => {
...
-  await ensureCLISession({
-    config,
-    metadata: {},
-    netlifyToken,
-    site,
-    state,
-    netlifyGraphConfig,
-  })
+
+  if (process.env.NODE_ENV !== 'test') {
```
jasonbarry, line 51: Added this conditional to make integration testing not barf

tests/integration/220.command.graph.test.js
```diff
@@ -1,8 +1,63 @@
...
+    response: [{ slug: site.account_slug }],
+  },
+  {
+    path: 'sites/site_id/service-instances',
+    response: 'uuid-string',
+  },
+  {
+    path: 'accounts/test-account/env',
+    method: 'GET',
+    response: [envelopeResponse[1]],
```
jasonbarry, line 52: In the test, we set `NETLIFY_GRAPH_PERSIST_QUERY_TOKEN` to already exist, in order to skip over some of the logic that reaches out using the OneGraph client. I didn't know how to mock that out, so skipped over it to at least test some of the function.

src/commands/graph/graph-init.js
```diff
@@ -120,8 +108,32 @@ const graphInit = async (options, command) => {
...
+      siteId,
+      body: [
+        !env.NETLIFY_GRAPH_WEBHOOK_SECRET && {
+          key: 'NETLIFY_GRAPH_WEBHOOK_SECRET',
+          scopes: ['functions'],
+          values: [{ context: 'all', value: newEnv.NETLIFY_GRAPH_WEBHOOK_SECRET }],
+        },
+        !env.NETLIFY_GRAPH_PERSIST_QUERY_TOKEN && {
+          key: 'NETLIFY_GRAPH_PERSIST_QUERY_TOKEN',
+          scopes: ['functions'],
```
sgrove, line 131: Are `scopes` typed? As in, will autocomplete ensure that I'm not putting in spurious scope values?

Also, this token needs to be available in build, please.

sgrove (changes requested):
If you wouldn't mind updating the persist query token scope, I can follow up with a quick approval.

Thanks for taking the time to port this over!

src/commands/graph/graph-init.js
jasonbarry: Scopes are defined [here in the openapi spec](https://github.com/netlify/open-api/blob/master/swagger.yml#L257-L261) for this method -- should anything else be needed to populate autocomplete suggestions in the cli codebase?

sgrove (approved):

PR #4964, opened 2022-08-16
feat: run plugins in Netlify Dev

eduardoboucas:
#### Summary

Changes the `netlify dev` command so that it uses the new dev flow introduced in https://github.com/netlify/build/pull/4431. This will run any `onPreDev` and `onDev` events implemented by build plugins.

The new `startDev` entry point from Netlify Build accepts a callback for the dev command, to which CLI supplies the existing `startFrameworkServer` function. For this reason, we don't expect any change in functionality with existing framework setups.

However, build plugins associated with frameworks will be auto-installed and they will technically run even if they don't implement any `*Dev` events, so the compatibility matrix of build plugins will start to apply in `netlify dev`.

This means that if you're running Gatsby on Node 12, for example, `netlify dev` will start failing because the build plugin requires Node 14 and above. For this reason, **this PR is technically a breaking change and therefore bumps a major version**.


tests/integration/600.framework-detection.test.js
```diff
@@ -293,16 +293,15 @@ test('should pass framework-info env to framework sub process', async (t) => {
     await builder
       .withPackageJson({
         packageJson: {
-          dependencies: { gatsby: '^3.0.0' },
-          scripts: { develop: 'node -p process.env.GATSBY_LOGGER' },
+          dependencies: { nuxt3: '^2.0.0' },
```
eduardoboucas, line 296: I've changed this because the Gatsby plugin (which will now run as part of Netlify Dev) needs Node 14, whereas CLI needs to support Node 12. I've changed it to Nuxt 3, which doesn't have a plugin.

src/commands/dev/dev.js
```diff
@@ -472,7 +495,19 @@ const dev = async (options, command) => {
     capabilities,
     timeouts,
   })
-  await startFrameworkServer({ settings })
+
+  log(`${NETLIFYDEVWARN} Setting up local development server`)
```
eduardoboucas, line 499: Adding this as a progress message because installing build plugins might take a few seconds.

src/commands/dev/dev.js
```diff
@@ -417,7 +419,8 @@ const validateGeoCountryCode = (arg) => {
  */
 const dev = async (options, command) => {
   log(`${NETLIFYDEV}`)
-  const { api, config, repositoryRoot, site, siteInfo, state } = command.netlify
+  const { api, cachedConfig, config, repositoryRoot, site, siteInfo, state } = command.netlify
+  const netlifyBuild = await netlifyBuildPromise
```
lukasholzer, line 423: NIT: maybe add todo here:

```suggestion
  // TODO: once ESModule Migration is done we can remove that`
  const netlifyBuild = await netlifyBuildPromise
```

tests/integration/500.command.dev.test.js
```diff
@@ -1,11 +1,14 @@
 // Handlers are meant to be async outside tests
 /* eslint-disable require-await */
+const { promises: fs } = require('fs')
```
lukasholzer, line 3: NIT: 

```suggestion
const fs = require('fs/promises')
```

tests/integration/utils/snapshots.js
```diff
@@ -13,7 +14,22 @@ const normalizers = [
   { pattern: /It should be one of.+/gm, value: 'It should be one of: *' },
 ]
 
-const normalize = (inputString) =>
-  normalizers.reduce((acc, { pattern, value }) => acc.replace(pattern, value), inputString)
+const optionalNormalizers = {
+  // File paths
+  filePath: { pattern: /(^|[ "'(=])((?:\.{0,2}|([A-Z]:)|file:\/\/)(\/[^ "')\n]+))/gm, value: '/file/path' },
```
lukasholzer, line 19: NIT: would love to see a link to a regex 101 with some samples for this pattern.

Not everybody is capable of reading regexes so well

lukasholzer (approved):
That looks super nice! 🐈

Thx for writing such good tests 🥇

ascorbic (approved):
Thank you!

tests/integration/500.command.dev.test.js
ascorbic: I don't think that syntax is supported in Node 12

ascorbic: Confirmed. fs/promises was added in Node 14

src/commands/dev/dev.js
eduardoboucas: I can add it if you feel strongly about it, but we already have this pattern elsewhere in the application without accompanying comments (e.g. https://github.com/netlify/cli/blob/main/src/lib/build.js#L47). I think these instances will be pretty easy to find once we do the migration.

tests/integration/utils/snapshots.js
eduardoboucas: This was taken from https://github.com/netlify/build/blob/main/packages/build/tests/helpers/normalize.js#L46.

I personally don't like the approach of relying on regular expressions to normalise the output of CLI/Netlify Build in snapshots, but that's our reality right now, and I don't have the bandwidth to add regex101 links to every expression.

PR #4966, opened 2022-08-16
feat: support scopes and contexts in `env:set` and `env:unset`

jasonbarry:
#### Summary

Fixes https://github.com/netlify/pillar-workflow/issues/810

This PR adds support for the `--scope` and `--context` flags for `env:set`, and the `--context` flag for `env:unset`. These flags are only available if the site has opted in to Envelope. If a user tries to use these flags with a site not configured with Envelope, then an error message will be printed to stdout. 

`env:set` and `env:unset` now support [variadic options](https://github.com/tj/commander.js#variadic-option). I've added some examples in the help text of each command. 

Branch values are not currently support (coming soon)

---

      passes our tests.

Screenshot 2022-08-16 at 22 40 51


melanierichards (dismissed):
Looks great, thanks JB! 🚀

src/commands/env/env-set.js
```diff
@@ -10,42 +14,64 @@ const { log, logJson, translateFromEnvelopeToMongo } = require('../../utils')
...
+  const { api, cachedConfig, site } = command.netlify
   const siteId = site.id
 
   if (!siteId) {
     log('No site id found, please run inside a site folder or `netlify link`')
     return false
   }
 
-  const siteData = await api.getSite({ siteId })
+  const { siteInfo } = cachedConfig
```
jasonbarry, line 27: Using `cachedConfig.siteInfo` instead of refetching from `api.getSite()` -- this is safe to do, yes?

docs/commands/env.md
```diff
@@ -170,10 +182,19 @@ netlify env:unset
 
 **Flags**
 
+- `context` (*production | deploy-preview | branch-deploy | dev*) - Specify a deploy context (default: all contexts)
```
codebyuma, line 185: Oh, I'll have to add this to context flag to the unset part in the docs 🎉

src/commands/env/env-set.js
```diff
@@ -10,42 +14,64 @@ const { log, logJson, translateFromEnvelopeToMongo } = require('../../utils')
...
   // Return new environment variables of site if using json flag
   if (options.json) {
     logJson(finalEnv)
     return false
   }
 
-  log(`Set environment variable ${key}=${value} for site ${siteData.name}`)
+  const withScope = scope ? ` scoped to ${chalk.white(scope)}` : ''
+  log(
+    `Set environment variable ${chalk.yellow(`${key}${value ? '=' : ''}${value}`)}${withScope} in ${chalk.magenta(
```
codebyuma, line 56: [sand] I think we're missing a `the` (and this will make it consistent with the phrasing for the other commands you added already)

```suggestion
    `Set environment variable ${chalk.yellow(`${key}${value ? '=' : ''}${value}`)}${withScope} in the ${chalk.magenta(
```

codebyuma (commented):
Looks good to me! I left a few copy suggestions for your consideration.

I'll update the docs to mention the context flag for `unset` :)

docs/commands/env.md
```diff
@@ -149,10 +149,22 @@ netlify env:set
...
```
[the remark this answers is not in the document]
jasonbarry: There is a slight difference -- `all` is no longer in the enum, in order to prevent someone from doing `netlify env:set KEY value --context dev production all` (variadic options are supported, but providing multiple contexts that include `all` doesn't make sense). By saying `all contexts` (general) instead of `"all"` (literal) was my (probably too subtle) way of indicating that.

melanierichards (approved):
:shipit:

PR #4978, opened 2022-08-19
feat: allow multiple operation names in graph:hander

dwwoelfel:
#### Summary

Allows multiple operation names to be specified in `graph:handler`, e.g. `ntl graph:handler --codegen ns/handler OpA OpB`.

Still supports showing a list of operations to select from if no operation name is provided.


src/commands/graph/graph-handler.js
```diff
@@ -66,14 +67,17 @@ const graphHandler = async (args, options, command) => {
...
-      generate: codeGenerator.generateHandler,
-      logger: log,
-      netlifyGraphConfig,
-      schema,
-      operationName,
-      handlerOptions,
-    })
+    /* eslint-disable fp/no-loops */
+    for (const operationName of operationNames) {
+      await generateHandlerByOperationName({
```
eduardoboucas, line 72: I don't see any async behaviour in `generateHandlerByOperationName`, so do you need to `await` here? And if not, you could use a `forEach` and avoid disabling the ESLint rule?

sgrove: Good catch, @eduardoboucas . All of the `generateHandler` need to be async now (since they're no longer pure functions) - @dwwoelfel is this PR missing some of the upstream changes?

sgrove: @dwwoelfel I'm fine to merge as-is if you want to follow up with the other changes separately.

sgrove:
@dwwoelfel Async change PR is at https://github.com/netlify/cli/pull/4984/files

sgrove (approved):

PR #4109, opened 2022-01-21
fix: external redirects not being matched correctly closes #3770

tinfoil-knight:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes #3770

External redirects with status code >= 300 & <=400 were not proxied due to requests getting matched incorrectly. This PR fixes the issue by changing the order of matching so that external redirects get matched first.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [x] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [x] Make sure the status checks below are successful ✅

PR #5005, opened 2022-08-26
fix: correctly allow redirects to external URLs

### LINKED ISSUE #1242: Bug: re-write rules are not applied for non static servers
**Describe the bug**

The following rule is not applied for non static servers (when a framework is detected):

```
/* /index.html 200
```

**To Reproduce**

Steps to reproduce the behaviour:

1. Run `git clone https://github.com/11ty/eleventy-base-blog.git test-project`
2. `cd test-project`
3. `npm install`
4. `rm 404.md` (this is important so 11ty won't redirect to a custom error page)
5. `echo "/* /index.html 200" > _redirects`
6. Run `BROWSER=none netlify dev`
7. Visit `http://localhost:8888/test`

See that you get an error from 11ty `Cannot GET /test/index.htm`

**Configuration**

```bash
  System:
    OS: macOS 10.15.6
    Memory: 1.67 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 14.9.0 - /usr/local/bin/node
    npm: 6.14.8 - /usr/local/bin/npm
  npmGlobalPackages:
    netlify-cli: 2.63.2
```

**Expected behavior**

The `index.html` file content should be returned.

**Additional context**

Related to the discussion here https://github.com/netlify/cli/issues/794#issuecomment-610619793 and to  https://github.com/netlify/cli/pull/810

The following does work:

```
/* /index.html 200!
```

```
/* /index.html 302
```

I believe the test [here](https://github.com/netlify/cli/blob/d76df759c9517191f946bc3d43a14bc3d50b0a7b/tests/cra.test.js#L58) passes since `create-react-app` re-routes unknown routes to the index page. Meaning the redirect is handled in `create-react-app`.

The solution is to request the resource from the framework server - if it returns 404 redirect, otherwise don't redirect.

### LINKED ISSUE #2710: Unnecessary url escaping breaks request paths
**Describe the bug**

I wanted to test out netlify functions using netlify-cli. It seems there's an extra `/` character getting added then escaped.

This project works fine when deployed on netlify and when running without netlify-dev. I'm sure it's a simple fix I need to make in my config but I can't seem to find any information in docs or any threads online.

**Configuration**

```netlify.toml

[build]
  command = 'make'
  publish = 'public'

[dev]
  command = 'make start'
  framework = '#custom'
  publish = 'public'
  targetPort = 3000
  port = 3001

[functions]
  directory = 'functions'

[[headers]]
  for = '/*'
  [headers.values]
    cache-control = 'public,max-age=86400,immutable'

[[headers]]
  for = '/cache/*'
  [headers.values]
    cache-control = 'public,max-age=604800,immutable'

[[headers]]
  for = '/fonts/*'
  [headers.values]
    cache-control = 'public,max-age=31536000,immutable'

[[redirects]]
  from = '/*'
  to = '/index.html'
  status = 200
```

**CLI Output**

image

danez:

#### Summary

#4109 change the behavior so that every redirect to an external URL was actually executed as a rewrite. The tests only checked the response body which is identical no matter if redirect or rewrite.

This is now fixed and the tests adapted to actually check if a redirect is happening.

Fixes #2710
Fixes #1242


src/utils/proxy.js
```diff
@@ -70,7 +70,7 @@ const proxyToExternalUrl = function ({ dest, destURL, req, res }) {
     pathRewrite: () => destURL,
     ...(Buffer.isBuffer(req.originalBody) && { buffer: toReadableStream(req.originalBody) }),
   })
-  return handler(req, res, {})
+  return handler(req, res, () => {})
```
danez, line 73: I only noticed this because a request was not matching and http-proxy-rewrite was actually calling `next` (the third parameter)

```
next is not a function
```

src/utils/proxy.js
```diff
@@ -185,7 +185,7 @@ const serveRedirect = async function ({ match, options, proxy, req, res }) {
 
   const staticFile = await getStatic(decodeURIComponent(reqUrl.pathname), options.publicFolder)
   if (staticFile) {
-    req.url = encodeURIComponent(staticFile) + reqUrl.search
+    req.url = encodeURI(staticFile) + reqUrl.search
```
danez, line 188: This fixes a problem where static URLs are escaped to `%2Findex.html`. Introduced in https://github.com/netlify/cli/pull/2665 and the tests still works.

tests/integration/0.command.dev.test.js
```diff
@@ -226,27 +229,16 @@ test('should redirect POST request if content-type is missing', async (t) => {
...
-        req.write('param=value')
-        req.end()
-      })
+      const error = await t.throwsAsync(
+        async () =>
+          await got.post(`${server.url}/api/echo`, {
+            body: 'param=value',
+            followRedirect: false,
+          }),
+      )
```
danez, line 238: Just replaced with got instead of custom HTTP request.

tests/integration/400.command.dev.test.js
```diff
@@ -363,353 +362,4 @@ test('should handle multipart form data when redirecting', async (t) => {
...
-
-      // TODO: check why this doesn't redirect
-      const response3 = await got(`${server.url}/not-foo/index.html`).text()
-
-      t.is(response1, '<html><h1>foo')
-      t.is(response2, '<html><h1>foo')
-      t.is(response3, '<html><h1>not-foo')
-    })
-  })
-})
```
danez, line 714: These tests are all duplicates and are actually also in `500.command.dev.test.js`

tinfoil-knight (dismissed):
[nit]
The commit message is of type `fix` & mentions `fix` again in the commit body. It can just be `fix: redirects to external URLs`

---

Everything else LGTM.

tests/integration/0.command.dev.test.js
```diff
@@ -206,15 +206,18 @@ test('should follow 301 redirect to an external server', async (t) => {
     await builder.buildAsync()
 
     await withDevServer({ cwd: builder.directory }, async (server) => {
-      const response = await got(`${server.url}/api/ping`).json()
-      t.deepEqual(response, { body: {}, method: 'GET', url: '/ping' })
+      const response = await got(`${server.url}/api/ping`, { followRedirect: false })
+      t.is(response.headers.location, `http://localhost:${port}/ping`)
+
+      const body = await got(`${server.url}/api/ping`).json()
```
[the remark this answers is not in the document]
danez: It is the body, coming from here: https://github.com/netlify/cli/blob/1ecab988698cbad7f46b953e7b80d31629820a6a/tests/integration/utils/external-server.js#L7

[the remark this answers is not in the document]
tinfoil-knight: My bad. Thank you for pointing out.

tests/integration/0.command.dev.test.js
```diff
@@ -226,27 +229,16 @@ test('should redirect POST request if content-type is missing', async (t) => {
...
+        async () =>
+          await got.post(`${server.url}/api/echo`, {
+            body: 'param=value',
+            followRedirect: false,
+          }),
+      )
 
       // we're testing Netlify Dev didn't crash
-      t.is(data, 'Method Not Allowed')
+      t.is(error.message, 'Response code 405 (Method Not Allowed)')
```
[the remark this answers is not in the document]
danez: makes sernse, fixed.

danez:
About the commit message, it is true that the commit message on its own sounds weird with the double `fix`, but in the changelog the semantic type will be removed and would just read `redirects to external URLs`, which is a little undescriptive for a changelog message :)
But I changed the PR title now and reworded it, so it will read better.

tinfoil-knight (approved):
LGTM!

Nice catch on the duplicate tests. Weird that they were in the repo for so long.

khendrikse (approved):
lets go boi

PR #5007, opened 2022-08-28
fix: follow standard pattern for auto-generated site names

### LINKED ISSUE #4630: Auto-generated site names do not follow Netlify's standard pattern
### Describe the bug

If a user creates a new site via `ntl init` and accepts a default site name, the generated name is something like `matt-kane-makes-great-sites`, which is not only debatable, but also does not follow the expected sweet site names format used by sites created from the UI

### Steps to reproduce

1. run `ntl sites:create` in an unlinked repo
3. leave "site name" empty

Expected:
- a deliciously sweet site name

Actual:
- a boring and potentially inaccurate name

### Configuration

_No response_

### Environment

```

  System:
    OS: macOS 12.4
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 6.30 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.5.0 - ~/.nvm/versions/node/v16.5.0/bin/node
    npm: 7.19.1 - ~/.nvm/versions/node/v16.5.0/bin/npm
  npmGlobalPackages:
    netlify-cli: 10.3.3
```

### LINKED ISSUE #4653: Suggested site names may have illegal-length domains
### Describe the bug

The pattern used for generating default site names can lead to it generating domains that cannot be resolved as they are too long. According to RFC 1035, a label may be up to 63 characters long. The longest template is `netlify-thinks-${slug}-is-great${suffix}`, where `suffix` is six characters and `slug` is the slugified username, so can be arbitrarily long. When generating a deploy preview, a 24 character hash is prepended. This means that any username that is over 6 characters will lead to a domain that is too long.

Related: #4630

### Steps to reproduce

- generate a site using `ntl sites:create`
- choose the default site name
- run `ntl deploy`
- observe length of domain, and see that it won't resolve

### Configuration

_No response_

### Environment

```

  System:
    OS: macOS 11.6.4
    CPU: (8) arm64 Apple M1
    Memory: 74.05 MB / 8.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.14.2/bin/yarn
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
    Watchman: 2021.06.07.00 - /opt/homebrew/bin/watchman
  npmGlobalPackages:
    netlify-cli: 10.3.1
```

tinfoil-knight:

#### Summary

Fixes #4630
Fixes #4653

The suggestions for auto-generated site names from the CLI don't follow the same pattern as they do when created from the Netlify App. These suggestions also use a very limited number of combinations which results in a user having a lot of site names being very similar & not being able to discern b/w them.

Additionally, the suggested site names can lead to invalid URLs as described here: https://github.com/netlify/cli/issues/4653#issue-1252491139

This PR fixes both the issue with standard names & site length. Also see https://github.com/netlify/cli/pull/5007#issuecomment-1235208464

---

      passes our tests.


src/utils/sites/utils.js
```diff
@@ -55,4 +57,506 @@ const createRepo = async (templateName, ghToken, siteName) => {
   return data
 }
 
-module.exports = { getTemplatesFromGitHub, createRepo, validateTemplate }
+const SITE_NAME_SUGGESTION_SUFFIX_LENGTH = 6
```
tinfoil-knight, line 60: I'm not sure of the logic behind this length. I've set it to 6 because all the sites created through the Netlify App UI had suffixes of this length.

Nonetheless, this length is valid.

tinfoil-knight:
The longest possible site name suggestion is `extraordinary-croquembouche-${suffix}` where suffix is of length 6 making the maximum length 34.

For deploy previous, a 24 char hash is pre-pended (as per https://github.com/netlify/cli/issues/4653#issue-1252491139) which would make the max length 58.

In all cases, the suggested URLs stay below the 63 char length.

danez:
I have mixed feelings about this. I do not like the code duplication from the server-side. In case we ever decide to change the logic server-side this can create problems again.

What if instead of creating a name on the client we simply do not do this at all? If the API call `createSiteInTeam` is called without `name` then the server generates a name. This makes this way more simple and we do not have any code duplication. I quickly tried this by removing all autogeneration and it seems to work fine.

Do you see any problem with this?

Kapture 2022-08-29 at 14 09 31

tinfoil-knight:
> What if instead of creating a name on the client we simply do not do this at all? If the API call `createSiteInTeam` is called without `name` then the server generates a name. This makes this way more simple and we do not have any code duplication. I quickly tried this by removing all autogeneration and it seems to work fine.
> 
> Do you see any problem with this?

Only issue with this is that it seems like the site name is going to be empty & there's no indication of what it's going to be.

Although, the Netlify UI also does the same thing (assign name without pre-indication).

I'll make the changes if you feel its necessary.

danez:
I personally feel this is okay if we add a message like `leave empty for autogenerated name`. 

Ping @netlify/developer-experience-members What is you take on this?

whitep4nth3r:
Something like `Press enter to {generate a random site name}` where 'Press enter' is the action, rather than 'Leave empty' which is an anti-action (kind of).

I left `generate a random site name` in curly brackets because I'm not sold on it.

tinfoil-knight:
@whitep4nth3r

How about `Skip to let Netlify pick a name for your site` ?

The prompt in the CLI would then be:

`? Site name (you can change it later): Skip to let Netlify pick a name for your site`

---

I don't want to go with "Press enter" because the key is named `return` for MacOS.

whitep4nth3r:
But how does someone `Skip`? That's not clear. I know we're getting deep into semantics and copy-writing and I'm not sure that's my area of expertise... but I'm not sure `Skip` instructs the user to press enter/return.

tinfoil-knight:
> But how does someone `Skip`? That's not clear. I know we're getting deep into semantics and copy-writing and I'm not sure that's my area of expertise... but I'm not sure `Skip` instructs the user to press enter/return.

I didn't want to use `Press enter` because it seems like an action step without much choice. `Skip` seems like it offers a choice to let the user not think of the site name (which is the intended use of auto-generated names).

I do agree that `Skip` might be a bit unclear but pressing enter or return to skip a step is usual behaviour for many CLI utilities (from anecdotal experience) so I don't think this should be too confusing.

whitep4nth3r:
You know, I'm not sure we need a message if the auto-generated name is a suggestion like in the image above from @tinfoil-knight.

I just ran through setting up a new git repo via the GitHub CLI for inspiration. All (lighter contrast) prompts are e.g. `(Y/n)` or`(.origin)` where the default option is chosen when you press enter  _without being told_. 

Maybe we just need to show the auto generated name without an extra message. That feels more intuitive to me. What do you think?

image

tinfoil-knight:
> Maybe we just need to show the auto generated name without an extra message. That feels more intuitive to me. What do you think?

@whitep4nth3r I like the auto-generated name option from a user's PoV but the issue with this is that it uses code that relies on internal implementation (Netlify API) of how the site names are generated which makes it brittle which is why @danez suggested to avoid doing this in https://github.com/netlify/cli/pull/5007#issuecomment-1230202734

whitep4nth3r:
Oh I see, that makes sense.

So the options are along the lines of:

1. `Site name (you can change it later): leave empty for autogenerated name`
1. `Site name (you can change it later): skip to let Netlify pick a name for your site`

I wonder whether we can get a bit of Netlify personality in here. Maybe I'm over-thinking it but there's maybe an opportunity to be a little quirky here given we can't show the auto-generated name. 

Along the lines of....?

1.  `Site name (you can change it later): pick for me`
1.  `Site name (you can change it later): I'm feeling lucky`

Maybe we have an array of quirky strings to show a random one out of a selection each time? Am I going too far? 😅

tinfoil-knight:
`Site name (you can change it later): pick for me` sounds okay. I'd avoid going with an array of strings for consistency on the command line. 

@danez I'll go forward with this & make changes to the flow to rely on the server for generating names if left empty.

@whitep4nth3r If you come up with any other ideas in the meantime, just let me know.

tinfoil-knight:
Change Summary

`sites:create` will no longer auto-suggest names & will just a `Pick for me` text. If left as it is, name will be selected by the server.

image

---

The `sites:create-template` command will now use the name of the template if the site name isn't available to create the Github Repository.

If a repository with the same name is already present, the user will get a well defined error.

The site name will be selected by the server if not provided.

The 
image

tinfoil-knight (commented):
@danez Let me know if any additional tests are required.

src/commands/sites/sites-create.js
```diff
@@ -1,58 +1,31 @@
...
-    ]
-    siteSuggestion = sample(suggestions)
-
     const { name: nameInput } = await inquirer.prompt([
       {
         type: 'input',
         name: 'name',
         message: 'Site name (you can change it later):',
-        default: siteSuggestion,
+        default: 'Pick for me',
```
eduardoboucas, line 18: I don't think we should be using the `default` property like this. We're assigning it a value that we know we won't use, just for the purpose of displaying something.

Instead, can we change `message` so that it reads something like `Site name (leave blank for a random name; you can change it later)`, and we assign a value if the value is empty?

If we absolutely must use the current approach, let's please move the string to a constant.

eduardoboucas (changes requested):

src/commands/sites/sites-create.js
tinfoil-knight: I'll change it. I find the current approach a bit "hacky" too.

tinfoil-knight:
Not sure as to what's happening with the failing tests here: https://github.com/netlify/cli/actions/runs/2993351770

danez:
I'm not working till end of September. So i can look then or someone else does. Sorry

tinfoil-knight:
@danez Can you take a look at the tests here? They just seem to be timing out.

danez (approved):
LGTM!

tinfoil-knight:
> @tinfoil-knight I rebased on the current main branch and fixed the tests. It seems the tests match on the exact input line.

Thank you a lot for fixing the tests @danez.

verythorough:
This is awesome! Here I was going to post a list of the words we use to generate the site names in Netlify, and I see y'all settled on a better solution. ✨

PR #5016, opened 2022-08-30
chore: cleanup eslint

danez:
This cleans up two things with eslint:

1. First commit removes the unused custom rule we have for importing chalk and instead uses a built-in one from eslint to restrict the import of chalk.
2. Second commit sets the `sourceType` for babel to `unambiguous` which makes babel guess the source type. This is done because we do parse node_modules for all the import rules, so that the import rules know what modules actually export. And currently, they do print warnings. With this change, they do not anymore, as babel can parse them correctly.

This is the error:
```
Error while parsing /home/runner/work/cli/cli/node_modules/netlify/src/index.js
Line 1, column 0: 'import' and 'export' may appear only with 'sourceType: "module"' (1:0)
`parseForESLint` from parser `/home/runner/work/cli/cli/node_modules/@babel/eslint-parser/lib/index.cjs` is invalid and will just be ignored

...
```


eduardoboucas (approved):

khendrikse (approved):

.eslintrc.js
```diff
@@ -69,12 +84,14 @@ module.exports = {
     {
       files: ['src/**/*.js'],
       rules: {
-        // once a solution for npm 6 is found add this to package.json
-        // "eslint-plugin-local-rules": "file:tools/eslint-rules",
-        // add it to the plugins on top: `plugins: ['sort-destructure-keys', 'local-rules'],`
-        //
-        // after that uncomment the next line
-        // 'local-rules/no-direct-chalk-import': 2,
+        'no-restricted-modules': [
```
lukasholzer, line 87: nice I like cleanup!

lukasholzer (approved):

PR #5029, opened 2022-09-06
feat: run framework detection and run plugins on netlify build

### LINKED ISSUE #4948: Improved automatic framework support in `ntl build`

Currently running `ntl dev` in an unliked project has a good experience: it detects the framework and runs the correct command. However this is not true for `ntl build`. First, it won't allow the site to build without `--offline`, and it doesn't do framework build command detection or build plugin installation. This means it's not possible to build a site without either creating a toml or linking the site. This adds friction for users trying out builds.

`ntl build` should behave like `ntl dev`. It should run even if the site is unlinked, and it should use framework-info to detect settings - including auto-installing plugins.

### Pull request (optional)

- [ ] I can submit a pull request.

sarahetter:

#### Summary

Fixes #4948

This allows users to run `netlify build` and `netlify deploy --build`, and have framework detection run, enabling build plugins (and the next runtime) to install and run automatically, without requiring a user to run `netlify link`. Running these automatically simplifies the flow for the user, requiring less config, and duplicating recent similar work for `netlify dev`.

New flow/scenarios:
1. User runs `ntl build` on a Next.js app with no extra config/netlify.toml/local plugin installed. They are no longer prompted to run `ntl link`. Framework detection runs, and passes the build directory, build command, and plugins specified by framework-info to the runBuild function. The newest plugin is installed and run. 
2. User runs `ntl build` on a Next.js app with a netlify.toml / package.json local plugin installed, with an older version of the plugin specified. They are no longer prompted to run `ntl link`. Framework detection still runs, but the user's local config overrides the version, and the local older version of the plugin is run.
3. User runs `ntl deploy --build` on a Next.js app. This is as above in # 1, however as this is a deploy, they will still be prompted to link / create a new site. Framework detection is run, and the newest plugin is installed and run. 
4. User runs `ntl build` on a static, vanilla javascript website. Framework detection runs, finds no matching framework, and builds as normal.

---

      passes our tests.

images


src/commands/build/build.js
```diff
@@ -7,11 +7,7 @@ const { error, exit, generateNetlifyGraphJWT, getEnvelopeEnv, getToken, normaliz
 /**
  * @param {import('../../lib/build').BuildConfig} options
  */
-const checkOptions = ({ cachedConfig: { siteInfo = {} }, token }) => {
-  if (!siteInfo.id) {
-    error('Could not find the site ID. Please run netlify link.')
```
ascorbic, line 12: So glad to be getting rid of this!

src/commands/build/build.js
```diff
@@ -7,11 +7,7 @@ const { error, exit, generateNetlifyGraphJWT, getEnvelopeEnv, getToken, normaliz
...
-  }
-
+const checkOptions = ({ token }) => {
   if (!token) {
```
ascorbic, line 11: Does this mean it won't work unless logged-in?

src/lib/build.js
```diff
@@ -41,11 +43,21 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
...
  * @returns
  */
-const runBuild = async (options) => {
+const runBuild = async (buildOptions, command, commandOptions) => {
   const { default: build } = await netlifyBuildPromise
+  const { cachedConfig, config, site } = command.netlify
+  const devConfig = {
+    framework: '#auto',
+    ...(config.functionsDirectory && { functions: config.functionsDirectory }),
+    // ...(config.build.publish && { publish: config.build.publish }),
```
ascorbic, line 57: Is this meant to be commented-out?

src/lib/build.js
```diff
@@ -55,10 +67,54 @@ const runBuild = async (options) => {
...
+      defaultConfig.build.command = settings.buildCommand
+    }
+    if (settings.dist) {
+      buildOptions.cachedConfig.config.build.publish = settings.buildCommand
+      defaultConfig.build.publish = settings.dist
+    }
+
+    buildOptions.defaultConfig = defaultConfig
+
+    console.log(buildOptions.defaultConfig)
```
ascorbic, line 91: I'm guessing this needs to be removed?

ascorbic (commented):
Awesome! I've added a few questions. Could you change the PR name to follow conventional commits? Also a bit more detail in the PR description shopwing step by step exactly what happens (i.e. user runs `ntl build`, if the site exists but has no plugins then x, or if it does have plugins, and if thew site doesn't exist then y and so on. I mean it's pretty clear but I know with the `dev` command, there were quite a few nuances that needed covering.

src/commands/build/build.js
sarahetter: This is only run `if (!options.offline)` - we are still making API requests with a user's account, and filling in config from that. This will grab any plugins they have installed that aren't just the ones from framework-info. 

I think changing this is out of scope for this PR, but we could potentially re-think how we're using offline/online mode, and whether or not we error out, or just warn.

sarahetter:
@ascorbic I've updated the description to be more clear about the flow, and made the rest of the recommended changes.

src/commands/build/build.js
ascorbic: Makes sense

src/lib/build.js
```diff
@@ -41,11 +43,20 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
...
 /**
  * run the build command
- * @param {BuildConfig} options
+ * @param {BuildConfig} buildOptions
+ * @param {import('../commands/base-command').BaseCommand} command
+ * @param {import('commander').OptionValues} commandOptions
  * @returns
  */
-const runBuild = async (options) => {
+const runBuild = async (buildOptions, command, commandOptions) => {
```
eduardoboucas, line 51: Conceptually, I feel like it might be best for this function to receive just the options it needs rather than receiving the entire command. Because this is used by multiple commands, it might not be obvious that the arguments will immediately change as soon as one of the commands change.

Not a blocker, though.

src/lib/build.js
```diff
@@ -55,10 +66,52 @@ const runBuild = async (options) => {
...
+
+  /** @type {Partial<import('../../utils/types').ServerSettings>} */
+  let settings = {}
+  try {
+    settings = await detectServerSettings(devConfig, commandOptions, site.root)
+
+    const defaultConfig = { build: {} }
+
+    if (settings.buildCommand) {
+      buildOptions.cachedConfig.config.build.command = settings.buildCommand
```
eduardoboucas, line 80: This is the sort of question that is tricky to answer without types in place, but is there a chance that some of these intermediate properties will be `undefined`, leaving to a reference error?

src/lib/build.js
```diff
@@ -55,10 +66,52 @@ const runBuild = async (options) => {
...
+
+          return { package: pluginName, origin: 'config', inputs: {} }
+        })
+        .filter(Boolean)
+
+      buildOptions.cachedConfig.config.plugins = [...newPlugins, ...cachedConfig.config.plugins]
+    }
+  } catch (error_) {
+    log(NETLIFYDEVERR, error_.message)
+    exit(1)
```
eduardoboucas, line 111: I believe the `error` util should handle this?

```suggestion
    error(error_)
```

src/lib/build.js
sarahetter: @eduardoboucas 

Did some chasing here, and found that cachedConfig is created by `@netlify/config` - and cachedconfig.config.build is always added. https://github.com/netlify/build/blob/d2c74c02280bd2324cb9b933df4be9ba8c399789/packages/config/src/base.js#L32

I'm not sure what I should do to safeguard against that situation popping up if there's a breaking change in `@netlify/config` though. The node version doesn't support optional chaining, and I don't want to do 4 different `hasOwnProperty`s :/ 

But wow, wouldn't this all be so much nicer with Typescript 😅

src/lib/build.js
```diff
@@ -55,10 +66,52 @@ const runBuild = async (options) => {
...
+            return
+          }
+
+          return { package: pluginName, origin: 'config', inputs: {} }
+        })
+        .filter(Boolean)
+
+      buildOptions.cachedConfig.config.plugins = [...newPlugins, ...cachedConfig.config.plugins]
+    }
+  } catch (error_) {
```
ericapisani, line 109: Not sure if this is a stylistic thing that's prevalent in this code base, but I think this is the first time I've seen a variable ending in an underscore rather than beginning with one 🤔 

[dust] I know it's to avoid the clash with the `error` function, but could we maybe call this `err` instead? Or does that lead to other issues/goes against a convention that I'm not aware of?

ericapisani (approved):
One small nit, but this is otherwise looking great! Excited to see this ship 🚀

ericapisani (approved):

ericapisani (approved):

bronze:
hi @sarahetter 
at https://github.com/netlify/cli/blob/main/src/lib/build.js#L83 we have `buildOptions.cachedConfig.config.build.publish = settings.buildCommand`. should it be `settings.dist`?

```
   if (settings.buildCommand && settings.dist) {
      buildOptions.cachedConfig.config.build.command = settings.buildCommand
      defaultConfig.build.command = settings.buildCommand
      //buildOptions.cachedConfig.config.build.publish = settings.buildCommand
      buildOptions.cachedConfig.config.build.publish = settings.dist
      defaultConfig.build.publish = settings.dist
    }
```

Issue #4655, opened 2022-05-31
Bring netlify-branded functions error page to edge functions


Skn0tt:
### Which problem is this feature request solving?

At the moment, when an edge function throws an uncaught exception, we just show the body of the 500 returned. The fact that an uncaught exception led to this is non-obvious.

### Describe the solution you'd like

Instead, we should use the Netlify-branded error page that was introduced in https://github.com/netlify/cli/pull/4486 also for edge functions. We've implemented a similar thing for production via https://github.com/netlify/stargate/pull/554, which uses the structured errors returned from edge functions via https://github.com/netlify/edge-functions-bootstrap/pull/38.

### Pull request (optional)

- [X] I can submit a pull request.

minivan:
From [Slack](https://netlify.slack.com/archives/C023PC3D08J/p1660561133013079): 

> I am assuming, just to clarify, that we mean setting this up for local development using `netlify/cli`?

The easiest way to reproduce this is to have a repo that has an edge function with an error, deploy the site, and invoke the function.

On a live site, we should see an error page.

If locally, we run `netlify dev` and invoke the function, we don't see the Netlify branded error page.

khendrikse:
I think I was a bit confused because @Skn0tt mentioned 

> We've implemented a similar thing for production via https://github.com/netlify/stargate/pull/554, which uses the structured errors returned from edge functions via https://github.com/netlify/edge-functions-bootstrap/pull/38.

But I guess the difference then is that this issue is not about those structured errors, but about uncaught exceptions?

Skn0tt:
In the first implementation of these error pages, we had full stack traces included in the error page, including for uncaught exceptions inside of user code. These were visible to all visitors, and disclosed both that the site is hosted by Netlify, and potentially leaked some source code internals (stacktraces contain file names, function names etc). So we've changed it so that the page itself shows only a bland error code like "Uncaught exception", and then there's a link at the bottom that leads to the edge functions log page. See https://edge-functions-examples.netlify.app/error for an example.

For the CLI though, I think we'd like to change that up a bit. Since there's no logs page to point towards, and we don't need to hide implementation details in local dev, I think it'd be fine to show the full stacktrace.

Skn0tt:
Here's some pointers towards what code we'd need to change:

[...]

PR #5070, opened 2022-09-19
Feat: add error page to edge functions for local dev

khendrikse:
#### Summary

Fixes https://github.com/netlify/cli/issues/4655.

Whenever someone runs `netlify dev` we want people to have a proper error page available for them with a full stacktrace if an edge function has an unhandled error. 
---

      passes our tests.

cat sneezing


src/utils/proxy.js
```diff
@@ -386,6 +390,25 @@ const initializeProxy = async function ({ configPath, distDir, port, projectDir
         res.setHeader(key, val)
       })
 
+      const isUncaughtError = responseStatus === 500 && proxyRes.headers['x-nf-uncaught-error'] === '1'
+
+      if (edgeFunctions.isEdgeFunctionsRequest(req) && isUncaughtError) {
+        const acceptsHtml = req.headers && req.headers.accept && req.headers.accept.includes('text/html')
+        const decompressedBody = await decompress(responseBody)
```
eduardoboucas, line 397: I find this mix of `async/await` and Promises a bit confusing, especially since the code you're running inside the `.then()` is not even asynchronous.

Have you considered using `async/await` there?

khendrikse: Yeah this was a bit messy, how about now? :)

src/lib/render-error-remplate.js
```diff
@@ -0,0 +1,17 @@
+const { readFile } = require('fs').promises
+const { join } = require('path')
+
+let errorTemplateFile
+
+const renderErrorTemplate = async (errString, templatePath = './templates/function-error.html') => {
```
eduardoboucas, line 6: It feels a bit weird to have this default to the template for serverless functions. I think we shouldn't have a default, and then the consumer always has to specify the path for the template.

This can totally be done in a follow up, though.

src/lib/templates/edge-function-error.html
```diff
@@ -0,0 +1,289 @@
+<!DOCTYPE html>
```
eduardoboucas, line 1: What would it take for us to use the same template for both types of functions? Are they all that different? 🤔

src/utils/proxy.js
```diff
@@ -20,14 +22,28 @@ const toReadableStream = require('to-readable-stream')
...
 const { headersForPath, parseHeaders } = require('./headers')
 const { createRewriter, onChanges } = require('./rules-proxy')
 
+const decompress = util.promisify(zlib.gunzip)
 const shouldGenerateETag = Symbol('Internal: response should generate ETag')
 
+const formatEdgeFunctionError = (errorBuffer, acceptsHtml) => {
+  const parsedError = JSON.parse(errorBuffer.toString())
+
+  return acceptsHtml
```
eduardoboucas, line 38: nit: I think doing an `if` with an early return would make this a lot easier than this ternary expression.

src/utils/proxy.js
```diff
@@ -386,6 +402,22 @@ const initializeProxy = async function ({ configPath, distDir, port, projectDir
         res.setHeader(key, val)
       })
 
+      const isUncaughtError = responseStatus === 500 && proxyRes.headers['x-nf-uncaught-error'] === '1'
```
eduardoboucas, line 405: Do we need to check for the status code? I would think that the presence of the header alone should indicate an uncaught error.

src/utils/proxy.js
eduardoboucas: Much better!

eduardoboucas (dismissed):
Left a few comments, but nothing blocking. Great work! ✨

src/lib/templates/edge-function-error.html
khendrikse: There are some differences right now in some of the texts, and some of the links... I also saw some things in the normal functions template that I'm pretty sure we won't use for the edge functions template, but that I'm not sure about if we use it still for normal functions. So I felt unsure as to combine the two. 

I could pass the differences as arguments, but something about that feels off a bit as well because the amount of arguments will grow quite a bit.

src/lib/render-error-remplate.js
khendrikse: Nah that's an easy change! Will do!

src/utils/proxy.js
khendrikse: I do think it's a good check either way, if only from the perspective of readability, as we do set the header ourselves and a 500 is a clear message to the reader that we're dealing with an error.

src/lib/templates/edge-function-error.html
khendrikse: I've asked some input from the team, if it's all good to clear out some old stuff we don't need anymore I'll merge the two :)

eduardoboucas (approved):

src/utils/proxy.js
eduardoboucas: My concern is that the check is not broad enough. If the bootstrap starts producing an uncaught error with a different status code, the CLI will need to adjust to that.

khendrikse: O yeah that makes sense :) In that case I will actually change it

eduardoboucas (approved):

eduardoboucas (approved):

PR #5089, opened 2022-09-26
chore!: remove node 12 support

lukasholzer:

#### Summary

Preparation for the Node 12 support drop we are going to to on the 27th of September 

https://answers.netlify.com/t/netlify-cli-dropping-support-for-node-js-version-12/75130

---

      passes our tests.


jobala (approved):

khendrikse (approved):

docs/README.md
```diff
@@ -5,12 +5,13 @@ description: All Netlify CLI commands
...
+Welcome to the Netlify CLI! This site provides online access to all help strings in the Netlify CLI. For a more in-depth
+guide, please see our [Getting Started](https://docs.netlify.com/cli/get-started/) guide on our main docs site.
 
-If you have questions, ideas, or would like to contribute, check out the [repository on GitHub](https://github.com/netlify/cli/).
+If you have questions, ideas, or would like to contribute, check out the
+[repository on GitHub](https://github.com/netlify/cli/).
 
-**Before you begin**
-Make sure you have [Node.js](https://nodejs.org/en/download/) version 12.20.0, 14.14.0, 16.0.0, or later.
+**Before you begin** Make sure you have [Node.js](https://nodejs.org/en/download/) version 16.0.0, or later.
```
danez, line 14: Also removed 14 here?

danez (commented):
Is there a special reason to do ^14.16.0?  Because 14.15.0 was the first official 14 lts release according to the [changelog](https://github.com/nodejs/node/blob/v14.x/doc/changelogs/CHANGELOG_V14.md)

If we decide to not do 14.15 but something newer, I would suggest 14.18, because then we can already start to use the [`node:` protocol](https://nodejs.org/api/esm.html#node-imports) in our code. Although this is not really important for our code but it might become tricky if we update dependencies that support node 14 (but >=14.18) and use `node:`.

docs/README.md
lukasholzer: Yea this was intentional as we want folks to install at least the current LTS 16 :) 14 will be EOL in March 2023 so there is no reason to keep install instructions on instructing for old versions.

always go with LTS.

Regarding the 14.18.x version I was thinking the same but as this is a rather new version I doubt that all folks have it installed. I would love to do that but 14.16 is already pretty old, released on 2021-02-23. 

As said this should be a minimum required version. Furthermore this version was communicated in the forums post: https://answers.netlify.com/t/netlify-cli-dropping-support-for-node-js-version-12/75130

danez: Yeah I guess we can wait for that for one more year :) Maybe we can change the workflow to use 14.16 so that in case we update dependencies that use `node:` the workflow would catch that.

jobala (approved):

PR #5093, opened 2022-09-26
fix(dev): allow passing mock country for geo-based redirects

### LINKED ISSUE #5051: Redirect rule with US country condition blocks every country
### Describe the bug

When applying a redirect/rewrite rule that targets the US, the rule always triggers when even when accessing the site from outside the US locally.

This makes it difficult to locally when a site-wide rule is applied to block the US for example and the `--country` cli option does not override this behaviour.

### Steps to reproduce

1. Create a Netlify project from one of the templates (e.g Next.js Starter) 
2. Add the following configuration in netlify.toml

```
[[redirects]]
  from = "/"
  to = "/blocked.html"
  status = 200
  force = true
  conditions = {Country = ["US"]}
```
3. Run `netlify dev` 
4. Attempt to access `http://localhost:8888`

An example repo you can clone and run to replicate:  https://github.com/hosainnet/netlify-geo

Expected: you should be able to view the homepage

Result: you'll be served `/blocked.html` (assuming that file exists in the `public` folder)

A quick poke around suggests the rules-proxy https://github.com/netlify/cli/blob/main/src/utils/rules-proxy.js#L83 always fallsback to getCountry (hardcoded to US).

### Configuration

[[redirects]]
  from = "/"
  to = "/blocked.html"
  status = 200
  force = true
  conditions = {Country = ["US"]}

### Environment

N/A

tinfoil-knight:

#### Summary

Fixes #5051 

The `dev` command doesn't allow a way for users to pass any flags to mock the current country to test out country based redirects which hinders local dev workflow.

Also, the "country" value used for redirects just defaults to `US` with no way to over-ride it which causes all requests for a route to be redirected in case a redirect rule with country value as `US` exists & prevents users from checking how the page without redirect looks.

This PR uses the value from pre-existing `country` flag for redirects. This allows the user to pass country values easily through the command line flag & test out country-based redirects properly.

---

      passes our tests.

🦆


src/utils/rules-proxy.js
```diff
@@ -33,11 +33,11 @@ const getLanguage = function (headers) {
   return 'en'
 }
 
-const getCountry = function () {
-  return 'us'
+const getCountry = function (geoCountry) {
+  return geoCountry || 'us'
 }
```
tinfoil-knight, line 38: Discussed w/ @kitop on Slack. Not sure why this fallback to `us` exists but we decided to preserve it for backwards compatibility.

tinfoil-knight:
@JWhist Just tagging you here since the `country` flag was originally added by you in https://github.com/netlify/cli/pull/4708 & I'm re-using it for redirects.

eduardoboucas (approved):

danez (dismissed):

src/utils/rules-proxy.js
```diff
@@ -80,7 +80,7 @@ const createRewriter = async function ({ configPath, distDir, jwtRoleClaim, jwtS
     const cookieValues = cookie.parse(req.headers.cookie || '')
     const headers = {
       'x-language': cookieValues.nf_lang || getLanguage(req.headers),
-      'x-country': cookieValues.nf_country || getCountry(),
+      'x-country': cookieValues.nf_country || getCountry(geoCountry),
```
lukasholzer, line 83: Can we just simplify that instead of having a function for that?

```suggestion
      'x-country': cookieValues.nf_country || geoCountry || 'us',
```

lukasholzer (changes requested):
Thanks for your contribution but can we keep it simple? I think we don't need a function for that one

src/utils/rules-proxy.js
tinfoil-knight: @lukasholzer This is done.

tinfoil-knight:
@danez @eduardoboucas Not sure why Github dismissed the reviews. Only change has been removing a function as suggested here: https://github.com/netlify/cli/pull/5093#discussion_r980952945

lukasholzer (approved):

PR #5140, opened 2022-10-06
chore: start migration to ESM

danez:
This PR is an attempt to get the ball rolling for the ESM migration.

- rename all `*.js` files to `*.cjs` and fix requires
- make the package an esm module by adding `type:module` to package.json
- migrate `bin/run` to an esm module
- remove `main` entry in package.json as it is pointing to an non existant file

I left templates untouched, the only change in templates is that the file extension were added to the `require()` calls.

This way now, we can migrate piece by piece. One command at a time maybe. All that needs to be done is migrate `src/commands/main.cjs`, `src/commands/index.cjs` and `src/commands/base-command.cjs` first an then we can start migrating command by command.

I thought this might be a good approach instead of having a huge diff all at once.


.eslintrc.cjs
```diff
@@ -75,21 +76,35 @@ module.exports = {
       },
     },
     {
-      files: ['e2e/**/*.js'],
+      files: ['src/**/*.cjs'],
```
lukasholzer, line 79:
```suggestion
      files: ['src/**/*.{js,mjs,cjs}'],
```

lukasholzer (approved):
Nice big step forward!

.eslintrc.cjs
danez: `no-restircted-modules` only applies to commonjs. for esm there is `no-restricted-imports`

PR #5167, opened 2022-10-14
chore: migrate BaseCommand  and postinstall to esm

danez:
#### Summary

This migrates the BaseCommand and related files to ESM.
Also migrates everything that directly uses the BaseCommand (postinstall, autocompletion-generation) to ESM (except tests)

---

      passes our tests.


src/commands/dev/dev.cjs
```diff
@@ -222,10 +222,11 @@ const startFrameworkServer = async function ({ settings }) {
     }
 
     stopSpinner({ error: false, spinner })
-  } catch {
+  } catch (error_) {
     stopSpinner({ error: true, spinner })
-    log(NETLIFYDEVERR, `Netlify Dev could not connect to localhost:${settings.frameworkPort}.`)
+    log(NETLIFYDEVERR, `Netlify Dev could not start or connect to localhost:${settings.frameworkPort}.`)
     log(NETLIFYDEVERR, `Please make sure your framework server is running on port ${settings.frameworkPort}`)
+    error(error_)
```
danez, line 229: This is actually unrelated and was for some debugging I did, but it actually makes sense. We usually land here when `wait-port` throws, so it is good to actually see the error that happened.

tests/unit/lib/completion/completion.test.cjs
```diff
@@ -33,11 +33,12 @@ test.afterEach(() => {
   sinon.restore()
 })
 
-test('should generate a completion file', (t) => {
+test('should generate a completion file', async (t) => {
   const stub = sinon.stub(fs, 'writeFileSync').callsFake(() => {})
-  const program = createTestCommand()
   // eslint-disable-next-line n/global-require
   const { createAutocompletion } = require('../../../../src/lib/completion/generate-autocompletion.cjs')
+  const program = await createTestCommand()
```
danez, line 40: I had to move this down, because otherwise the mocking would fail 🙄
My guess is that if the mocked module is imported ina esm module, somehow it resets the mock again, but not sure.

src/lib/completion/index.cjs
```diff
@@ -1,5 +1,5 @@
-const generateAutocompletion = require('./generate-autocompletion.cjs')
+const { createAutocompletion } = require('./generate-autocompletion.cjs')
 
 module.exports = {
-  ...generateAutocompletion,
+  createAutocompletion,
```
danez, line 4: esm cannot import cjs modules that use this pattern

PR #5191, opened 2022-10-27
chore: print env var scope during ntl dev

jasonbarry:
#### Summary

Screenshot 2022-10-27 at 10 06 29

This PR prints a message for sites that are using Envelope during `netlify dev` that environment variable values are being used for all scopes. This local behavior is slightly different than what occurs on Netlify, as env vars can be scoped to e.g. builds _or_ functions. Locally, this scoping doesn't happen -- everything is available in the same `process.env`. This log hopefully brings attention to that difference! 

Fixes https://github.com/netlify/pillar-workflow/issues/812

---

      passes our tests.


codebyuma (approved):
Looks good to me 🎉

PR #5194, opened 2022-10-27
chore: remove beta tags from CLI and docs

pvorozhe:

#### Summary

Fixes [pillar-workflow/191](https://github.com/netlify/pillar-workflow/issues/919) (1/2)

---

      passes our tests.


package.json
```diff
@@ -344,6 +344,7 @@
     "husky": "^8.0.0",
     "ini": "^2.0.0",
     "mock-fs": "^5.1.2",
+    "netlify-cli": "^12.0.11",
```
danez, line 347: I don't think we should have the CLI as dependency in the CLI :D

danez (changes requested):

package.json
lukasholzer: oof :D

pvorozhe: Good catch :my-bad:

akahn (dismissed):

codebyuma (commented):
The copy update looks good to me 🎉

danez (changes requested):
The file `npm-shrinkwrap.json` still has changes, could you revert those?

pvorozhe:
> The file `npm-shrinkwrap.json` still has changes, could you revert those?
I reverted the changes to main branch, they still show up in the diff. not sure why

danez (approved):

Skn0tt:
this PR is failing on main, due to a snapshot not being updated: https://github.com/netlify/cli/actions/runs/3378626827/jobs/5609083160#step:11:136. I wonder why the CI here didn't catch it 🤔 

I'll fix this as part of https://github.com/netlify/cli/pull/5200, but we should ensure that our CI catches these kinds of things in the future.

Skn0tt:
I have a hunch that  the `chore: update contributors field` commits are resulting in green runs, because the `test:affected`  strategy is resulting in untouched files not being tested. Maybe we could adapt the `test:affected` strategy to not only test files touched by recent commits, but files touched in `HEAD..PR HEAD`

Skn0tt:
@lukasholzer maybe you have an idea

PR #5197, opened 2022-10-28
feat: support in-source configuration in edge functions

eduardoboucas:
#### Summary

Adds support for https://github.com/netlify/edge-bundler/pull/133 to the CLI.


src/lib/build.cjs
```diff
@@ -34,6 +34,7 @@ const getBuildOptions = ({ cachedConfig, options: { context, cwd, debug, dry, js
   offline,
   cwd,
   featureFlags: {
+    edge_functions_config_export: true,
```
danez, line 37: I don't see the FF being added in the change. Is that on purpose?

eduardoboucas: It was added in https://github.com/netlify/edge-bundler/pull/133.

danez (approved):
Lgtm

PR #5290, opened 2022-12-03
fix: detect requests without body correctly

### LINKED ISSUE #5030: Netlify Functions not receiving a body in local dev but works once deployed to Netlify
### Describe the bug

When I submit a fetch in from my Next.js app in local development using netlify dev (local node version v17.4.0), the Netlify Function runs but the body in the event is always `undefined`. I expect it to be `{ url: "websiteUrl" }`. When I deployed the function to Netlify it correctly found and parsed the body.

I noticed this occur on both background and regular functions.

I submitted more details about this issue here: https://answers.netlify.com/t/why-arent-my-netlify-functions-finding-a-body/74661 

### Steps to reproduce

1. Run `git clone git@github.com:mclimb/jamhack-personalized-demo.git` (private repo, ask Mike Chirokas for access)
2. Install npm packages 
3. netlify dev
4. Go localhost:8888
5. Submit to any function endpoint, for example, enter the website url as `https://miro.com/` and click the color pallet button (this will try to get the colors on https://miro.com/) and should call the function getBrandColors which should take the url body

### Configuration

_No response_

### Environment

```
System:
    OS: macOS 12.5.1
    CPU: (8) arm64 Apple M1
    Memory: 756.27 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 17.4.0 - ~/.nvm/versions/node/v17.4.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v17.4.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v17.4.0/bin/npm
  npmGlobalPackages:
    netlify-cli: 11.5.1
```

danez:
if a request goes through an edge function the `content-length` header will be removed and instead the `transfer-enconding` is set to chunked. Our detection in the functions (non-edge) server was detecting these passthrough requests as requests without body, because of the missing `content-length`. `content-length` should be ignored anyways when `transfer-encoding` is used according to [the HTTP/1.1 spec](https://httpwg.org/specs/rfc9112.html#rfc.section.6.1).

Fixes #5030

This also changes `request.get(...)` and `request.headers[...]` to `request.header(...)` for readability. 

red-fox-in-winter-forest-royalty-free-image-1610492147

### DIFF
--- a/src/lib/functions/server.mjs
+++ b/src/lib/functions/server.mjs
@@ -41,6 +41,13 @@ const buildClientContext = function (headers) {
   }
 }
 
+const hasBody = (req) =>
+  // copied from is-type package
+  (req.get('transfer-encoding') !== undefined || !Number.isNaN(req.headers['content-length'])) &&
+  // we expect a string or a buffer, because we use the two bodyParsers(text, raw) from express
+  // eslint-disable-next-line n/prefer-global/buffer
+  (typeof req.body === 'string' || Buffer.isBuffer(req.body))
+
 export const createHandler = function (options) {
   const { config, functionsRegistry } = options
 
@@ -62,8 +69,11 @@ export const createHandler = function (options) {
       return
     }
 
-    const isBase64Encoded = shouldBase64Encode(request.headers['content-type'])
-    const body = request.get('content-length') ? request.body.toString(isBase64Encoded ? 'base64' : 'utf8') : undefined
+    const isBase64Encoded = shouldBase64Encode(request.get('content-type'))
+    let body
+    if (hasBody(request)) {
+      body = request.body.toString(isBase64Encoded ? 'base64' : 'utf8')
+    }
 
     let remoteAddress = request.get('x-forwarded-for') || request.connection.remoteAddress || ''
     remoteAddress = remoteAddress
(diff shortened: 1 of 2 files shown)


src/lib/functions/server.mjs
```diff
@@ -41,6 +41,13 @@ const buildClientContext = function (headers) {
   }
 }
 
+const hasBody = (req) =>
+  // copied from is-type package
+  (req.header('transfer-encoding') !== undefined || !Number.isNaN(req.header('content-length'))) &&
```
tinfoil-knight, line 46: Why're specifically checking for the `NaN` value here?

In the `type-is` package the global isNaN function is used which behaves differently from the `Number.isNaN` function since the global isNaN is true for any non-numeric value (strings, undefined, null etc.) while the `Number.isNaN` isn't.


Ref: https://github.com/jshttp/type-is/blob/7d19b7aab1ad671f59ba157ae0640cd4b1302ca5/index.js#L92-L95

danez: Interesting, I wasn't aware of that and I  just changed it because some eslint rule recommended it.

change to the global version

PR #5298, opened 2022-12-06
chore: migrate more to ESM

danez:
We are getting close

I know this is huge, but it will soon be done.


tests/integration/20.command.functions.test.cjs
```diff
@@ -8,35 +8,19 @@ const getPort = require('get-port')
...
-  const functionUrl = `http://localhost:${port}/.netlify/functions/${functionName}`
-
-  const ntlFunction = new NetlifyFunction({
-    name: functionName,
-    settings: { functionsPort: port },
-    config: { functions: { [functionName]: {} } },
-  })
-
-  t.is(ntlFunction.url, functionUrl)
-})
```
danez, line 38: moved to a unit test

src/utils/get-global-config.cjs
```diff
@@ -1,47 +0,0 @@
...
-  for (let retry = 1; retry <= retries; retry++) {
-    try {
-      return await getGlobalConfigOnce()
-    } catch (error) {
-      if (retry === retries) {
-        throw error
-      }
-    }
-  }
-}
```
danez, line 44: Removed retry logic as it was only introduced to fix tests. If this is still a problem, we need to mock this file.
We probably should mock this anyway, as tests should not write global config file.

src/utils/get-global-config.cjs
```diff
@@ -1,47 +0,0 @@
...
-
-// Memoise config result so that we only load it once
-module.exports = memoizeOne(getGlobalConfig)
```
danez, line 47: `memoizeOne` removed as it was completely unneccessary. We never call `getGlobalConfig()` with any arguments so it would anyway only cache it once.

package.json
```diff
@@ -305,7 +303,6 @@
     "netlify-redirector": "^0.3.1",
     "node-fetch": "^2.6.0",
     "node-version-alias": "^1.0.1",
-    "omit.js": "^2.0.2",
```
danez, line 308: Used destructuring instead of omit.js

```diff
-x = omit(config.build, ['publish', 'publishOrigin']) }
+// eslint-disable-next-line no-unused-vars
+const { publish, publishOrigin, ...x } = config.build
```

khendrikse (approved):

PR #5302, opened 2022-12-09
fix: always look for internal edge functions

eduardoboucas:
#### Summary

Currently, when CLI doesn't find a valid `manifest.json` file in the internal edge functions directory, it doesn't even look for edge functions in that directory. This means that declaring an internal edge function using in-source configuration without an accompanying manifest file doesn't work.

This PR addresses that by ensuring we always look for functions in that directory, regardless of whether or not there's a manifest with declarations.

### DIFF
--- a/src/lib/edge-functions/internal.cjs
+++ b/src/lib/edge-functions/internal.cjs
@@ -1,6 +1,6 @@
 // @ts-check
 const { promises: fs } = require('fs')
-const path = require('path')
+const { dirname, join, resolve } = require('path')
 const { cwd } = require('process')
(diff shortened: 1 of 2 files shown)


src/lib/edge-functions/internal.cjs
```diff
@@ -71,7 +71,7 @@ const getInternalFunctions = async () => {
   } catch {
     return {
       functions: [],
-      path: null,
```
eduardoboucas, line 74: This was the actual problem.

danez (approved):

PR #5307, opened 2022-12-13
fix: check for any functions when using EF proxy

khendrikse:

#### Summary

Should fix an issue where having any internal functions with in-source-config, but no other edge functions, would prevent the proxy from running the internal EF.

---

      passes our tests.

### DIFF
--- a/src/lib/edge-functions/proxy.mjs
+++ b/src/lib/edge-functions/proxy.mjs
@@ -89,16 +89,22 @@ export const initializeProxy = async ({
     projectDir,
   })
 
+  if (!userFunctionsPath && !internalFunctionsPath) {
+    return
+  }
+
   return async (req) => {
     const registry = await server
-    if (!registry) return
-    const hasEdgeFunctions = registry.functions.length !== 0
 
-    await registry.initialize()
+    if (!registry) return
+
+    const hasEdgeFunctions = registry.functions.length !== 0
 
     if (req.headers[headers.Passthrough] !== undefined || !hasEdgeFunctions) {
       return
     }
+
+    await registry.initialize()
 
     const geoLocation = await getGeoLocation({ mode: geolocationMode, geoCountry, offline, state })
 
--- a/src/lib/edge-functions/proxy.mjs
+++ b/src/lib/edge-functions/proxy.mjs
@@ -88,25 +88,23 @@ export const initializeProxy = async ({
     port: isolatePort,
     projectDir,
   })
-  const hasEdgeFunctions = userFunctionsPath !== undefined || internalFunctions.length !== 0
 
   return async (req) => {
+    const registry = await server
+    if (!registry) return
+    const hasEdgeFunctions = registry.functions.length !== 0
+
+    await registry.initialize()
+
     if (req.headers[headers.Passthrough] !== undefined || !hasEdgeFunctions) {
       return
     }
 
-    const [geoLocation, registry] = await Promise.all([
-      getGeoLocation({ mode: geolocationMode, geoCountry, offline, state }),
-      server,
-    ])
-
-    if (!registry) return
+    const geoLocation = await getGeoLocation({ mode: geolocationMode, geoCountry, offline, state })
 
     // Setting header with geolocation and site info.
     req.headers[headers.Geo] = JSON.stringify(geoLocation)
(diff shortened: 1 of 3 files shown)


khendrikse:
I still need to add tests

src/lib/edge-functions/proxy.cjs
```diff
@@ -88,26 +88,24 @@ const initializeProxy = async ({
     port: isolatePort,
     projectDir,
   })
-  const hasEdgeFunctions = userFunctionsPath !== undefined || internalFunctions.length !== 0
 
   return async (req) => {
+    const registry = await server
+    if (!registry) return
+    const hasEdgeFunctions = registry.functions.length !== 0
```
eduardoboucas, line 95: The current approach prevents us from initialising the Deno server unnecessarily when edge functions aren't being used. Initialising the Deno server involves potentially downloading the Deno CLI, which is a big deal. We'll lose that with your change.

Rather than checking that we have declarations for internal functions, could we perhaps check that the internal functions directory exists, and proceed with initialising the server if so? That way, if someone is not using edge functions, there's no reason for that directory to exist and therefore we can keep the optimisation for those users.

Does that make sense?

eduardoboucas (changes requested):

src/lib/edge-functions/proxy.cjs
khendrikse: Yeah this makes sense!

tests/integration/100.command.dev.test.cjs
```diff
@@ -781,6 +770,53 @@ test('should respect in-source configuration from edge functions', async (t) =>
...
+
+      const res3 = await got(`http://localhost:${port}/hello-2`, { throwHttpErrors: false })
+
+      t.is(res3.statusCode, 200)
+      t.is(res3.body, 'Hello world')
+    })
+  })
+})
+
+test('should respect in-source configuration from internal edge functions', async (t) => {
```
khendrikse, line 789: @eduardoboucas do you have an idea what I might have done wrong in these two tests? I keep getting failed tests in the pipeline (running it locally is useless, they are incredibly flaky) and I'm having a hard time figuring out the bug I introduced.

src/lib/edge-functions/proxy.mjs
```diff
@@ -88,26 +88,30 @@ export const initializeProxy = async ({
...
     projectDir,
   })
-  const hasEdgeFunctions = userFunctionsPath !== undefined || internalFunctions.length !== 0
+
+  if (!userFunctionsPath && !internalFunctionsPath) {
+    return
+  }
 
   return async (req) => {
+    const registry = await server
```
eduardoboucas, line 97: You're still initialising the server and therefore downloading the Deno CLI, regardless of whether or not the site uses edge functions, which I thought you were going to change. Or am I missing something?

khendrikse: @eduardoboucas I'm doing an empty return on line 92 if there is no userfunctionspath or internalfunctionspath. I thought this would ensure that line 97 would not be run if we didn't need a server?

danez: How I see this I think this is a little bit of a egg-hen problem, without the server we do not know at this point if there a functions. Unless we extract the logic to find edge-functions out from the server/registry.

tests/integration/100.command.dev.test.cjs
```diff
@@ -781,6 +770,53 @@ test('should respect in-source configuration from edge functions', async (t) =>
...
+            publish: publicDir,
+          },
+        },
+      })
+      .withEdgeFunction({
+        config: () => ({ path: '/internal-1' }),
+        handler: () => new Response('Hello from an internal function'),
+        internal: true,
+        name: 'internal',
+      })
```
danez, line 805: About the test failures: It might be that we do not update the actual routing for edge functions on reload in the edge-functions-proxy. That is what I would check first if that actually works. 
I'm not even sure if that works with a `netlify.toml`, so updating the routing in `netlify.toml` while `netlify dev` is running.

danez: Okay here is the watch logic from edge-functions and it looks like we do some special logic for the config file and we probably need to adapt that for isc-edge-functions.

PR #5311, opened 2022-12-14
fix: allow for in-source-config internal edge functions in proxy

khendrikse:

#### Summary

Should fix an issue where having any internal functions with in-source-config, but no other edge functions, would prevent the proxy from running the internal EF.

---

      passes our tests.

### DIFF
--- a/src/lib/edge-functions/internal.mjs
+++ b/src/lib/edge-functions/internal.mjs
@@ -36,9 +36,16 @@ export const getInternalFunctions = async () => {
     const stats = await stat(path)
 
     if (!stats.isDirectory()) {
-      throw new Error('Path is not a directory')
+      throw new Error('Internal edge functions directory expected')
     }
+  } catch {
+    return {
+      functions: [],
+      path: null,
+    }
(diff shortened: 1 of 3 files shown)


src/lib/edge-functions/internal.mjs
```diff
@@ -36,9 +36,16 @@ export const getInternalFunctions = async () => {
     const stats = await stat(path)
 
     if (!stats.isDirectory()) {
-      throw new Error('Path is not a directory')
+      throw new Error('Internal edge functions directory expected')
     }
+  } catch {
+    return {
+      functions: [],
+      path: null,
```
eduardoboucas, line 44: If we wanted to be a bit more consistent, we could make `path: undefined` here and then in `src/lib/edge-functions/proxy.mjs` make the check:

```js
const hasEdgeFunctions = userFunctionsPath !== undefined || internalFunctionsPath !== undefined
```

Not a huge deal, though.

eduardoboucas (approved):
✨

PR #5315, opened 2022-12-14
feat: load edge function import maps from Deno config file

eduardoboucas:
#### Summary

Loads import maps for edge functions from the Deno config file. CLI counterpart of https://github.com/netlify/edge-bundler/pull/256.

### DIFF
--- a/src/commands/deploy/deploy.mjs
+++ b/src/commands/deploy/deploy.mjs
@@ -410,7 +410,7 @@ const bundleEdgeFunctions = async (options) => {
     ...options,
     buffer: true,
     featureFlags: {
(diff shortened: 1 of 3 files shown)


src/commands/deploy/deploy.mjs
```diff
@@ -410,7 +410,7 @@ const bundleEdgeFunctions = async (options) => {
     ...options,
     buffer: true,
     featureFlags: {
-      edge_functions_produce_eszip: true,
```
eduardoboucas, line 413: Flag no longer in use.

danez (approved):
🎉

PR #5316, opened 2022-12-15
fix: centralize edge functions feature flags

eduardoboucas:
#### Summary

We have two different places where we need to send edge functions feature flags to Netlify Build. This PR centralises them.


danez (approved):
Good catch

PR #5319, opened 2022-12-16
chore: introduce test debugging for cli tests

danez:
#### Summary

Introduces a new `DEBUG_TESTS` env variable that enables piping of `stdout` and `stderr` from the subprocess to the main process. It also disables the default reporter in vitests and instead enables the `tap` reporter, because the default fancy reporters hide parts of the piped output because the delete and overwrite console output.

For now this only works for the dev-server, but can be later extended to other cli commands.

In the recording first a normal run, then a run with debug enabled

https://user-images.githubusercontent.com/231804/208127033-b24709fb-64e1-4ff2-a4c4-5106793684b4.mp4

istockphoto-1278559947-612x612


khendrikse (approved):
Making dreams come true

PR #5320, opened 2022-12-16
chore: remove unnecessary abstractions of filesystem functions

danez:
#### Summary

Both `readFileAsyncCatchError` and `rmdirRecursiveAsync` from `src/lib/fs.cjs` can be replaced with node builtin functionality now that we dropped node 12.


tools/e2e/setup.mjs
```diff
@@ -1,22 +1,15 @@
...
 import { join, sep } from 'path'
 import { cwd, env } from 'process'
 import { fileURLToPath } from 'url'
 
-import del from 'del'
 import execa from 'execa'
 import getPort from 'get-port'
 import verdaccio from 'verdaccio'
 
-// TODO: remove this once `../../src/lib/fs.js` is an esm module as well
```
danez, line 12: Not why we should wait, ESM can import commonjs

tools/e2e/setup.mjs
```diff
@@ -1,22 +1,15 @@
-import { appendFileSync, existsSync, promises, readFileSync, writeFileSync } from 'fs'
+import { appendFile, mkdtemp, readFile, rm, writeFile } from 'fs/promises'
```
danez, line 1: I updated this file to use top-level await and async version of the fs functions.

khendrikse (approved):
awesome!

PR #5322, opened 2022-12-17
feat(command-env): add support for plaintext output in env:list

### LINKED ISSUE #3262: Add function like env:export to make .env files from site config

When moving from one dev machine to an other i want to be able to quickly setup my .env file with secrets.

I'm already using the env:import command to move all the .env content into netlify, that same system could be reversed to setup the new dev environment

At this time i can use env:list and write the file manually, but having an env:export which automatically formats the output would be nice

**Additional context**

No. I don't really trust my own skills, but will look into it based on the env:list when i get some time. 

tinfoil-knight:

#### Summary

Fixes #3262 

The `env:list` command currently outputs an ASCII table by default which isn't helpful if someone intends to write the environment variables to a file.

The `--plain` flag introduced in this PR would allow a user to setup their env files easily on local systems since it outputs the environment variables in the plaintext format used in .env files.

Example Output here:
```
$ netlify env:list --plain
NODE_ENV=development
REACT_APP_API_URL=https://api.netlify.com
REACT_APP_SECRET=UILH0lQphn
``` 

Also see: https://github.com/netlify/cli/issues/3262#issuecomment-959746250

---

      passes our tests.

🦥


docs/commands/env.md
```diff
@@ -135,6 +135,7 @@ netlify env:list
 **Flags**
 
 - `context` (*string*) - Specify a deploy context or branch (contexts: "production", "deploy-preview", "branch-deploy", "dev")
+- `plain` (*boolean*) - Output environment variables as plaintext
```
jasonbarry, line 138: Could new entries to the help be put below `scope`? Would be nice to keep context and scope next to each other. Thanks!

tinfoil-knight:
> I have a request that shouldn't be too much lift: considering that we also have the undocumented `--json` flag, and that `--json --plain` doesn't make much sense, could this change to `--format <format>`?
> * `--format` would have an array of choices of `table`, `plain`, and `json`, with `table` being the default

@jasonbarry This seems great. I'll implement this right away.

docs/commands/env.md
tinfoil-knight: Will keep this in mind while implementing the `format` flag.

tinfoil-knight:
> I have a request that shouldn't be too much lift: considering that we also have the undocumented `--json` flag, and that `--json --plain` doesn't make much sense, could this change to `--format <format>`?
> 
> * `--format` would have an array of choices of `table`, `plain`, and `json`, with `table` being the default
> * `options.json` should still be supported on line 71, alongside `options.format === 'json'`, to preserve backwards-compatibility

@jasonbarry I'm slightly opposed to making the suggested changes.

- The default flow first prints the table with env values hidden (key, scope visible) & then asks if the user wants to print the actual env values or not. The `--json` & `--plain` flags just print the env key & values directly which makes their behaviour different than the default one. I don't think values passed to the `--format` flag should have different behaviour (apart from the output format).

- The `--json` flag is used across a lot of commands across the CLI so users familiar with the CLI would expect the JSON output for relevant commands to be accessible through the same flag. Adding the new `--format` flag to `env:list` would make this experience inconsistent with rest of the CLI commands.

WDYT?

docs/commands/env.md
tinfoil-knight: The `plain` flag in between seemed weird to me too but the options are sorted here to produced the documentation:

https://github.com/netlify/cli/blob/d3dfa172383a21396294973a666d9665582864c4/src/utils/command-helpers.mjs#L74-L80

We can rename the option to `text` so it appear after `scope`.
Other alternatives like `raw`, `plaintext` would appear in between `context` & `scope` in the sorted order.

Let me know if you've a better name.

jasonbarry: Gotcha -- let's keep it as `plain`, and let the docs sort them as they currently do. But in `--help`, it would be great if context and scope could appear next to each other.

jasonbarry:
Sounds like a plan @tinfoil-knight, thanks for the research and explanation. I'd be happy to approve with this proposal if a `conflicts` property is added.

docs/commands/env.md
tinfoil-knight: Unfortunately, it seems like we're using the same function to order the help output too so I don't think the re-ordering would be possible unless we rename the option. The current order does appear a bit weird to me too but I don't think there's any way to override this specific instance.

Ref:

https://github.com/netlify/cli/blob/3ff4a57a95c445a185f9a210aced06bfc2b117be/src/commands/base-command.mjs#L249-L256

tinfoil-knight:
@jasonbarry The `conflicts` property has been added now to prevent the `--json` & `--plain` flags from being used together.

jasonbarry (approved):
Awesome work @tinfoil-knight, thanks for adding this feature!

PR #5328, opened 2022-12-20
feat: add support for edge function bypass logic

Skn0tt:
#### Summary

Works towards https://github.com/netlify/edge-functions-bootstrap/issues/78.

This PR teaches the CLI how to deal with the `x-nf-edge-function-bypass` header that EF bootstrap returns on early returns. It will run a second request that skips edge functions, similar to how the CDN works.

---

      passes our tests.

### DIFF
--- a/src/utils/proxy.mjs
+++ b/src/utils/proxy.mjs
@@ -20,6 +20,7 @@ import locatePath from 'locate-path'
 import pFilter from 'p-filter'
 import toReadableStream from 'to-readable-stream'
 
+import edgeFunctionHeaders from '../lib/edge-functions/headers.mjs'
 import {
   handleProxyRequest,
   initializeProxy as initializeEdgeFunctionsProxy,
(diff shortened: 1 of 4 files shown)


src/utils/proxy.mjs
```diff
@@ -363,6 +364,11 @@ const initializeProxy = async function ({ configPath, distDir, host, port, proje
     }
   })
   proxy.on('proxyRes', (proxyRes, req, res) => {
+    if (proxyRes.headers[edgeFunctionHeaders.EdgeFunctionBypass]) {
+      req.headers[edgeFunctionHeaders.Passthrough] = 'true'
+      return proxy.web(req, res, { target: entrypoint })
```
Skn0tt, line 369: this feels very hacky, as it basically makes a full second request to the dev server. I'd prefer if we could continue the existing request chain in some way, but I don't see how.

Skn0tt:
Closed in favour of approach in https://github.com/netlify/edge-functions-bootstrap/pull/144

PR #5335, opened 2022-12-22
chore: add retry to CI workflows

eduardoboucas:
#### Summary

Adds a retry mechanism to the test runs in the CI.


danez:
We can also use the retry option from vitest. Currently not all tests are running with vitest, but soon will.

PR #5338, opened 2022-12-23
feat: add `--prod` flag to `dev` command

### LINKED ISSUE #3905: Add `netlify serve` command

Currently the `dev` command is used to run the framework's dev command, but this is usually not the same as the production site. 

A `serve` command that does the following:

- By default, runs the build command and uses the config modified by the build plugins. This can be disabled with `--no-build`
- Runs the static server and functions server, in the same way as `--framework=#static`

Running `ntl deploy --build`, killing the process once the modified toml has been created, then running `ntl dev --framework=#static`

**Additional context**

Would work well alongside #3906, in which case it could skip building by default and instead have a `--build` flag like `deploy`.

Yes

eduardoboucas:
#### Summary

Needs https://github.com/netlify/build/pull/4792. Closes #3905.


src/commands/dev/dev.mjs
```diff
@@ -710,6 +807,7 @@ export const createDevCommand = (program) => {
     .option('-o ,--offline', 'disables any features that require network access')
     .option('-l, --live', 'start a public live session', false)
     .option('--functionsPort <port>', 'port of functions server', (value) => Number.parseInt(value))
+    .option('-s, --serve', 'run in "serve" mode', false)
```
eduardoboucas, line 810: This needs a better description.

src/lib/functions/registry.mjs
```diff
@@ -145,11 +148,12 @@ export class FunctionsRegistry {
       )
     }
 
-    // This fixes the bug described here https://github.com/netlify/zip-it-and-ship-it/issues/637
-    // If the current function's file is a zip bundle, we ignore it and log a helpful message.
+    // If the function file is a ZIP, we extract it and rewire its main file to
```
eduardoboucas, line 151: We need to do this so that the functions have all its dependencies available, otherwise we'd need to load modules from `.netlify/plugins/node_modules`, which isn't part of the module resolution path. We know we've already bundled the functions as part of the build process, so we can extract their ZIP files and use the exact files that will be used in Lambda, in the spirit of simulation the production environment as closely as possible.

src/commands/dev/dev.mjs
ascorbic: How about `--prod` instead? We're running it in a prod-like mode

eduardoboucas: Yeah, I like that!

src/commands/dev/dev.mjs
```diff
@@ -494,10 +505,20 @@ const dev = async (options, command) => {
     startPollingForAPIAuthentication({ api, command, config, site, siteInfo })
   }
 
+  log(`${NETLIFYDEVWARN} Setting up local development server`)
```
ascorbic, line 508: I think it needs a different message for serve. I can see people being confused that it doesn't watch the source or config files, for example.

eduardoboucas: Any suggestions?

src/commands/dev/dev.mjs
```diff
@@ -494,10 +505,20 @@ const dev = async (options, command) => {
...
+
+  const { configPath: configPathOverride } = await runBuild({ cachedConfig, options, settings, site })
+
+  // When using the `serve` flag, we want to use the production functions built
+  // by Netlify Build rather than building them from source.
+  const dist = Boolean(options.serve)
```
ericapisani, line 514: Is it possible to change the name of this variable to something more along of the lines of what the comment above is suggesting? I worry that since `dist` is often used as a name for where built files are output to during a site build that this will be confusing for a future dev.

eduardoboucas: Totally. Done in https://github.com/netlify/cli/pull/5338/commits/d782283cd00af0cb6b0a5cec751ff1c4a901d349.

src/commands/dev/dev.mjs
eduardoboucas: @ticachica any thoughts/concerns here?

eduardoboucas: Renamed to `--prod` in https://github.com/netlify/cli/pull/5338/commits/78c35b68bf54712bc6230c0c7d874f5a9ac5c780.

src/commands/dev/dev.mjs
eduardoboucas: Had a go at this in https://github.com/netlify/cli/pull/5338/commits/0bcc9a26965ca1c84525f17ad429c5607e0063dd.

src/commands/dev/dev.mjs
```diff
@@ -681,9 +690,92 @@ const dev = async (options, command) => {
...
+    // Run Netlify Build using the main entry point.
+    await buildSite(buildSiteOptions)
+
+    // Start the dev server, forcing the usage of a static server as opposed to
+    // the framework server.
+    await devCommand({
+      command: undefined,
+      useStaticServer: true,
+    })
+
```
ascorbic, line 749:
```suggestion
log(`${NETLIFYDEVLOG} Serving production build \n${NETLIFYDEVWARN} Changed files will not be rebuilt until you restart the server`)
```

src/commands/dev/dev.mjs
```diff
@@ -723,6 +827,7 @@ export const createDevCommand = (program) => {
     .option('-o ,--offline', 'disables any features that require network access')
     .option('-l, --live', 'start a public live session', false)
     .option('--functionsPort <port>', 'port of functions server', (value) => Number.parseInt(value))
+    .option('-P, --prod', 'build the site for production', false)
```
ascorbic, line 830:
```suggestion
    .option('-P, --prod', 'build the site for production and serve locally', false)
```

ascorbic:
Also needs changing for edge functions: https://github.com/netlify/cli/blob/feat/serve-command/src/lib/edge-functions/registry.mjs#L219

eduardoboucas:
> Also needs changing for edge functions: https://github.com/netlify/cli/blob/feat/serve-command/src/lib/edge-functions/registry.mjs#L223

I'd rather do that in a follow-up PR. I don't like the idea of making the edge functions registry aware of the `prod` flag, so I want to shift things around a bit, but that would make this PR too large.

eduardoboucas:
Actually, I managed to do it in a concise way in https://github.com/netlify/cli/pull/5338/commits/29342333651a437275e85bfb28d991e190e7d92d.

docs/commands/dev.md
```diff
@@ -30,6 +30,7 @@ netlify dev
 - `live` (*boolean*) - start a public live session
 - `offline` (*boolean*) - disables any features that require network access
 - `port` (*string*) - port of netlify dev
+- `prod` (*boolean*) - build the site for production and serve locally
```
codebyuma, line 33: Just creeping on your PR 😄  How does this work with or in contrast to the context flag (documented above), that you can also set to `production` ? 

ie `netlify dev --context production`

codebyuma: (the CLI doc above mentions only environment variables, but our [docs site ](https://docs.netlify.com/cli/get-started/#get-started-with-netlify-dev) makes it sound like using the context flag builds your site for production - so this would need to be updated if that's not correct)

eduardoboucas: Your creeping is very much welcome and that's a great point! The CLI docs are correct: the `context` flag only affects environment variables. This makes me wonder whether we should repurpose `context` and introduce this new behaviour (i.e. actually building the site for production) when `context` is `production`. And we could make `--prod` an alias for `--context=production`.

I think that would make things more consistency, with just two concerns worth considering:

1. That would make this a breaking change, since we'd be changing an existing flag rather than introducing a new one
2. This new behaviour means that watching for files won't work, and people will have to restart the CLI for the build to be run again and their changes to be picked up. I worry that it might be confusing that a specific value of `context` triggers this behaviour, rather than an entirely new flag.

But if we decide not to reuse `context`, it'll be a bit difficult to explain why you have `--prod` and `--context=production`.

@ascorbic What are your thoughts on this one?

eduardoboucas: I thought about this a bit more and I feel that a more manageable step would be to keep `context` as is (i.e. just about environment variables) and make `prod` automatically set `--context=production`, so that you're building the production site and using the production env vars.

orinokai: Now that we have named the flag `prod`, I agree that making it automatically set `--context=production` feels most intuitive

ascorbic: This is why I sugegsted having it as a separate command. Using `ntl dev` when it's building for production is a little unintuitive

eduardoboucas: Building a separate command is a bigger lift (if done properly) that unfortunately I don't have the capacity for right now. I've pushed https://github.com/netlify/cli/pull/5338/commits/ae74853c17f5113dae181fe57230fef02fd3c7ea which sets `--context=production` when `--prod` is set, and in https://github.com/netlify/cli/pull/5338/commits/f3cc600d13d8aea1ac8efc75a7093ca99d2b502e added a "beta" label to the flag in the help menu. We do this for other commands.

I suggest that we ship this as is and then iterate on it. If we come to the conclusion that this makes more sense as a separate command, we can break it out and deprecate `--prod`.

src/commands/dev/dev.mjs
```diff
@@ -713,7 +822,6 @@ export const createDevCommand = (program) => {
       '--context <context>',
       'Specify a deploy context or branch for environment variables (contexts: "production", "deploy-preview", "branch-deploy", "dev")',
       normalizeContext,
-      'dev',
```
eduardoboucas, line 716: By setting a default value here, we don't give commands the opportunity to set their own custom values. I've removed the default and added it in https://github.com/netlify/cli/pull/5338/files#diff-67082469659283bc342bd13f7e20570cb3d1bf5096aad88545a9ac4809d6083bR520.

docs/commands/dev.md
ascorbic: Yeah, that is a good compromise. Does our CLI lib support aliases? If so, that would probably be the simplest solution.

eduardoboucas: https://github.com/netlify/cli/pull/5351 turns the flag into a separate command. We can get it merged after this one.

src/commands/dev/dev.mjs
```diff
@@ -681,9 +695,92 @@ const dev = async (options, command) => {
...
+    // location of the config file for the duration of the command.
+    const tempConfigPath = await copyConfig({ configPath: cachedConfig.configPath, siteRoot: site.root })
+    const buildSiteOptions = {
+      ...sharedOptions,
+      outputConfigPath: tempConfigPath,
+      saveConfig: true,
+    }
+
+    // Run Netlify Build using the main entry point.
+    await buildSite(buildSiteOptions)
```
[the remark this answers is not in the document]
eduardoboucas: Agreed. I've done it in https://github.com/netlify/cli/pull/5351/commits/f4f31fb973e07d942454b625d34dd3022425081d to avoid merge conflicts.

Once you're happy with this PR, I can merge that one in.

src/commands/dev/dev.mjs
```diff
@@ -455,6 +460,23 @@ const dev = async (options, command) => {
 
   let { env } = cachedConfig
 
+  if (!options.prod) {
+    // Add `NETLIFY_DEV` to the environment variables.
+    env.NETLIFY_DEV = { sources: ['internal'], value: 'true' }
+
+    // Ensure the internal functions directory exists so that any functions
```
eduardoboucas, line 467: Will move this to a more suitable place in #5351.

ascorbic (approved):
🎉

PR #5340, opened 2022-12-30
feat: use console.log for logging in ef template

Skn0tt:
part of https://github.com/netlify/pod-compute/issues/316

### DIFF
--- a/src/functions-templates/typescript/log/{{name}}.ts
+++ b/src/functions-templates/typescript/log/{{name}}.ts
@@ -1,7 +1,7 @@
 import type { Context } from "https://edge.netlify.com";
 
 export default async (request: Request, context: Context) => {
-  console.log("Hello from the logging service");
+  console.log(`There was a request from ${context.geo.city} to ${request.url}`);
 
   return new Response("The request to this URL was logged", {
     headers: { "content-type": "text/html" },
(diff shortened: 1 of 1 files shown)


src/functions-templates/typescript/log/{{name}}.ts
```diff
@@ -1,9 +1,9 @@
 import type { Context } from "https://edge.netlify.com";
```
Skn0tt, line 1: I think we can leave the type import here. `context` is unused in the template itself, but it's good to have around - this is a template, and devs will either use `context` (and be happy to have the types!) or remove it themselves :)

eduardoboucas: Agreed. Another option would be to change the response to include something from `context`, so that the variable is actually used. Something like `context.ip` or `context.geo.city`?

Skn0tt: I like that idea, a good way for product discovery! Implemented in https://github.com/netlify/cli/pull/5340/commits/30d8ffc0a76bcd7eed52af5201bea221bf7403c1. I chose `context.geo.city`, geolocation is probably more of a mind-blowing feature than the IP address :D

eduardoboucas (approved):

PR #5342, opened 2023-01-02
feat(ef templates): replace context.json with Response.json

Skn0tt:
As per https://github.com/netlify/edge-functions-bootstrap/pull/142#issuecomment-1368897194, let's change the ef template code to use `Response.json` instead of the soon-to-be-deprecated `context.json`.

### DIFF
--- a/src/functions-templates/typescript/json/{{name}}.ts
+++ b/src/functions-templates/typescript/json/{{name}}.ts
@@ -1,5 +1,5 @@
 import type { Context } from "https://edge.netlify.com";
 
 export default async (request: Request, context: Context) => {
-  return context.json({ hello: "world" });
+  return Response.json({ hello: "world" });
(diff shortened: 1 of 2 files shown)


src/functions-templates/typescript/json/{{name}}.ts
```diff
@@ -1,5 +1,5 @@
 import type { Context } from "https://edge.netlify.com";
 
 export default async (request: Request, context: Context) => {
-  return context.json({ hello: "world" });
+  return Response.json({ hello: "world", location: context.geo.city });
```
Skn0tt, line 4: I added the `location` field here to be in line with https://github.com/netlify/cli/pull/5340#discussion_r1059948889.

eduardoboucas (approved):
✨

PR #5345, opened 2023-01-03
chore: migrate everything to ESM

danez:
This is the last ESM PR for the `src` codebase. There are no commonjs files anymore in in the `src` folder after this PR. (except templates for functions/sites, which I did not touch)

There are still some cjs test files, which I want to convert separately, as I also have some Ideas on how to improve them.


tests/integration/520.command.link.test.mjs
```diff
@@ -0,0 +1,47 @@
...
+        await callCli(['link'], getCLIOptions({ builder, apiUrl }))
+
+        expect(await isFileAsync(join(builder.directory, '.gitignore'))).toBe(true)
+      },
+      true,
+    )
+  })
+})
+
+test.skipIf(process.platform === 'win32')(
```
danez, line 28: This was skipped before too, but from what I looked at seems like a problem wither in how we use netlify/config here or a problem inside netlify/config on windows.

khendrikse (approved):
Thanks for doing this!

PR #5359, opened 2023-01-11
fix: publish shrinkwrap with prod dependencies only

### LINKED ISSUE #5323: `EBADPLATFORM` from `npm ci` when running on netlify with >= `12.2.9`
### Describe the bug

`npm ci` on netlify fails with

```
7:41:18 PM: npm ERR! code EBADPLATFORM
7:41:18 PM: npm ERR! notsup Unsupported platform for @esbuild/android-arm@0.15.15: wanted {"os":"android","arch":"arm"} (current: {"os":"linux","arch":"x64"})
7:41:18 PM: npm ERR! notsup Valid OS:    android
7:41:18 PM: npm ERR! notsup Valid Arch:  arm
7:41:18 PM: npm ERR! notsup Actual OS:   linux
7:41:18 PM: npm ERR! notsup Actual Arch: x64
7:41:18 PM: npm ERR! A complete log of this run can be found in:
7:41:18 PM: npm ERR!     /opt/buildhome/.npm/_logs/2022-12-18T19_41_17_238Z-debug-0.log
```

[See build run here](https://app.netlify.com/sites/venerable-puffpuff-36b86b/deploys/639f6cb3c3e60700093128bc)

### Steps to reproduce

View [this test repo](https://github.com/tjenkinson/netlify-cli-install-error).

It does not contain a package-lock, meaning the first `npm i` generates a package-lock on the instance. It then does an `npm ci`, which fails with the error above.

This did not used to happen with versions <= 12.2.8.

Ref https://github.com/video-dev/hls.js/pull/5114
Ref https://github.com/video-dev/hls.js/issues/5115

### Configuration

https://github.com/tjenkinson/netlify-cli-install-error/blob/main/netlify.toml

### Environment

My local environment shouldn't be relevant for this. Please see the repro.

danez:
#### Summary

Fixes #5323

It removes the shrinkwrap from the repo and instead uses the normal package-lock.

On publish we create the shrinkwrap, but only for prod dependencies.

Also split up the workflow into 2 jobs so we do not have to repeat the `if: ${{ steps.release.outputs.release_created }}` for every single step.


package.json
```diff
@@ -38,7 +38,7 @@
     "url": "https://github.com/netlify/cli/issues"
   },
   "scripts": {
-    "prepare": "husky install node_modules/@netlify/eslint-config-node/.husky/",
+    "prepare": "is-ci || husky install node_modules/@netlify/eslint-config-node/.husky/",
```
danez, line 41: We don't need husky in CI I think, it just slows done publish by running lint and stuff, which at that point should all be good.

.github/workflows/benchmark.yml
```diff
@@ -15,10 +15,8 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: '*'
-          cache: 'npm'
-          cache-dependency-path: 'npm-shrinkwrap.json'
-          check-latest: true
+          node-version: lts/*
+          cache: npm
```
danez, line 19: All non-test workflows I changed to use the latest lts version instead of the latest version and not do extra requests to check for the latest. The nodejs version here does not really matter, and I think it is good to use the most stable one which is lts.

lukasholzer (approved):
LGTM! Awesome work :)

PR #5395, opened 2023-01-16
feat: add kebab-case options to cli flags

khendrikse:

#### Summary

Fixes https://github.com/netlify/team-dev/issues/55

This is an example of the way we could work on implementing new kebab-case options as CLI flags, as described in the above ticket. 

In this example you can see how we could achieve the following:

-  Add kebab-case options to flags that are currently camelCase.
-  Keep supporting camelCase, but hide the option from the help command.

---

      passes our tests.


src/commands/base-command.mjs
```diff
@@ -99,13 +104,19 @@ export default class BaseCommand extends Command {
         .addOption(new Option('--cwd <cwd>').hideHelp(true))
         .addOption(new Option('-o, --offline').hideHelp(true))
         .addOption(new Option('--auth <token>', 'Netlify auth token').hideHelp(true))
+        .addOption(
+          new Option('--httpProxy [address]', '[DEPRECATED IN V13] Proxy server address to route requests through.')
```
eduardoboucas, line 108: Am I right in thinking that this string will not be visible to anyone since we're setting `hideHelp(true)`?

src/commands/base-command.mjs
```diff
@@ -420,7 +436,7 @@ export default class BaseCommand extends Command {
     const { buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
     const normalizedConfig = normalizeConfig(config)
     const agent = await getAgent({
-      httpProxy: options.httpProxy,
+      httpProxy: returnDeprecatedOptionValue(options.httpProxy),
```
eduardoboucas, line 439: Rather than adding this helper to the code that retrieves the options, do you think we could change this to `options['http-proxy']` and then have the `preAction` hook translate `options.httpProxy` into `options['http-proxy']` until we fully deprecate the camelCase version?

src/utils/option-deprecation.mjs
```diff
@@ -0,0 +1,31 @@
+import { chalk, log, NETLIFYDEVWARN } from './command-helpers.mjs'
+
+// This file contains logic for flags in camelcase that will be deprecated in v13 and replaced in kebab-case.
+// Flags to be deprecated in v13
+const DEPRECATED_OPTIONS = {
+  httpProxy: { option: '--httpProxy', newOption: '--http-proxy', deprecatedVersion: 'v13' },
```
eduardoboucas, line 6: I'm not sure we should commit to a specific version of the CLI where we'll deprecate these flags. I think we'd want to leave a certain period of time for people to migrate, and it's totally possible that we have to fit in a breaking change and release v13 next week.

Could we just say we'll deprecate this in an upcoming version of the CLI?

src/commands/base-command.mjs
khendrikse: correct!

src/utils/option-deprecation.mjs
khendrikse: Yeah 100%!

src/commands/base-command.mjs
khendrikse: I'd rather not mutate the options, as commander by default changes `http-proxy` to `httpProxy`. It's the default for the library to change the option keys to camelcase and I feel like we'd be working against the commander logic 🤔

eduardoboucas: Then I'm not sure why we need to add this helper to every command? If commander translates kebab-case to camelCase, we could continue to use `options.httpProxy` here without any changes? If the idea is to print a deprecation warning, could we use a single `preAction` hook that looks for camelCase options and prints a warning?

khendrikse: Because I need to somehow find out if someone is using the old flag or not, which is impossible because of the fact that commander will turn `http-proxy` into `options.httpProxy` I added [this logic](https://github.com/netlify/cli/pull/5395/files#diff-67082469659283bc342bd13f7e20570cb3d1bf5096aad88545a9ac4809d6083bR111) that will _flag_ the old option. 

Like I tried to explain in the summary:

> Why this way? because wether or not the user uses --httpProxy or --http-proxy. The name of the option will always be options.httpProxy 😅. So we need a way to distinguish the old vs the new.

It parses the argument and turns it into an object. That is why we need that helper.

For as far as I was able to figure out, it is not possible for us to check in the `preAction` hook if something is camelCase or not, because commander turns _everything_ into camelcase. Using `--http-proxy` will not create `options[http-proxy]` but `options.httpProxy` :) I hope I was able to explain it a bit better.

khendrikse: Oh also: it seems to actually not run `preAction` before running the `init` function here https://github.com/netlify/cli/pull/5395/files#diff-67082469659283bc342bd13f7e20570cb3d1bf5096aad88545a9ac4809d6083bR414

So if we wanted to perform anything in the `preAction` hook, it would already be too late

khendrikse:
@eduardoboucas mentioned that it might be worth not deprecating the old options and just keeping them but hidden. This way we don't have to add all the deprecation logic and it should not be a big maintenance overhead.

Planning to reduce the scope of the issue to not include warnings about deprecating older options.

src/commands/dev/dev.mjs
```diff
@@ -247,10 +257,25 @@ export const createDevCommand = (program) => {
...
+        .conflicts('edgeInspectBrk')
+        .argParser(validateShortFlagArgs)
+        .hideHelp(true),
+    )
+    .addOption(
+      new Option(
+        '-e, --edge-inspect [address]',
         'enable the V8 Inspector Protocol for Edge Functions, with an optional address in the host:port format',
       )
         .conflicts('edgeInspectBrk')
```
eduardoboucas, line 281: I think this needs to change?

```suggestion
        .conflicts('edge-inspect-brk')
```

src/commands/dev/dev.mjs
```diff
@@ -259,6 +284,15 @@ export const createDevCommand = (program) => {
...
+        .conflicts('edgeInspect')
+        .hideHelp(true)
+        .argParser(validateShortFlagArgs),
+    )
+    .addOption(
+      new Option(
+        '-E, --edge-inspect-brk [address]',
         'enable the V8 Inspector Protocol for Edge Functions and pause execution on the first line of code, with an optional address in the host:port format',
       )
         .conflicts('edgeInspect')
```
eduardoboucas, line 298: I think this needs to change?

```suggestion
        .conflicts('edge-inspect')
```

eduardoboucas (commented):
❤️ Nice! Left a couple of minor comments, but this is looking great.

src/commands/dev/dev.mjs
khendrikse: 😅 one would think! But I tested it. This seems to work the same as the options. We need to pass the camelCased option. Example:

I created:
```
  .addOption(new Option('--json', 'Output return values as JSON').hideHelp(true).conflicts('test-karin'))
  .addOption(new Option('--test-karin', 'bleep bloop'))
```

When used like this, I don't get a conflict warning:
`☁  testjes [main] ⚡  ../../cli/bin/run.mjs dev --json --test-karin`

Changed it to camelcase:
```
  .addOption(new Option('--json', 'Output return values as JSON').hideHelp(true).conflicts('testKarin'))
  .addOption(new Option('--test-karin', 'bleep bloop'))
```

Then I got a warning:
```
☁  testjes [main] ⚡  ../../cli/bin/run.mjs dev --json --test-karin
 ›   Error: option '--json' cannot be used with option '--test-karin'
 ›   See more help with --help
```

khendrikse: And others kind of made the same assumption: https://github.com/tj/commander.js/issues/1796

eduardoboucas: Interesting!

eduardoboucas (approved):

PR #5398, opened 2023-01-17
feat: rename x-deno- to x-nf-

Skn0tt:
Part of https://github.com/netlify/partner-deno/issues/11. Renames the `x-deno-` headers to `x-nf-`. Will fail right now, but work once https://github.com/netlify/edge-bundler/pull/290 lands.


Skn0tt:
https://github.com/netlify/cli/pull/5398/commits/ab1da2cf7c6762c122e41dedf79e14bf8f532a3b should've failed some tests 🤔 why is it not?

Skn0tt:
> why is it not?

it should've failed tests in https://github.com/netlify/cli/blob/main/tests/integration/100.command.dev.test.cjs, but these tests were never executed. It looks like that's because we use Vitest's `--changed` option, which only executes tests that are dependent on the changed file. That dependency is detected by tracing imports, but since integration tests don't import the source code but run the CLI via execa instead, it can't detect how these fit together.

Here's the imports for the test file that should've been run, but wasn't: 

https://github.com/netlify/cli/blob/25270706cb3b004a28ea9e260505895442747af0/tests/integration/100.command.dev.test.cjs#L1-L20

As you can see, there's no import of the source files in there, so Vitest's `--change` dependency tracing doesn't pick it up.

I'll open a separate issue to track this bug in our CI. Let's go ahead with this PR.

khendrikse (approved):

PR #5411, opened 2023-01-20
fix: gracefully handle errors when setting headers

eduardoboucas:
#### Summary

When we finish processing a serverless function and transfer any header it sets into the final response, we're doing [a `setHeader call`](https://github.com/netlify/cli/blob/main/src/lib/functions/synchronous.mjs#L15) that may throw. When that happens, the entire app crashes.

This PR addresses that by gracefully handling the error, showing a log in the terminal and displaying the function error page.

Screenshot 2023-01-20 at 10 27 31

### DIFF
--- a/src/lib/functions/server.mjs
+++ b/src/lib/functions/server.mjs
@@ -168,7 +168,7 @@ export const createHandler = function (options) {
         return
       }
 
-      handleSynchronousFunction(error, result, request, response)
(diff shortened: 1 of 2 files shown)


khendrikse (approved):
Very stylish and graceful 💅

danez (approved):

PR #5414, opened 2023-01-20
feat: add support for signed redirects

eduardoboucas:
#### Summary

Adds support for [signed redirects](https://docs.netlify.com/routing/redirects/rewrites-proxies/#signed-proxy-redirects) to the local development flow.


mraerino (approved):
awesome stuff!

PR #5415, opened 2023-01-20
fix: sign redirect with secret from env var

eduardoboucas:
#### Summary

Follow-up to #5414. The `signed` property holds the name of the environment variable that holds the secret for signing redirects. It does not hold the secret itself.

### DIFF
--- a/src/utils/proxy.mjs
+++ b/src/utils/proxy.mjs
@@ -29,7 +29,7 @@ import {
 import { fileExistsAsync, isFileAsync } from '../lib/fs.mjs'
 import renderErrorTemplate from '../lib/render-error-template.mjs'
 
-import { NETLIFYDEVLOG, NETLIFYDEVWARN } from './command-helpers.mjs'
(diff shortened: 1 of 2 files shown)


src/utils/proxy.mjs
```diff
@@ -157,12 +157,21 @@ const serveRedirect = async function ({ match, options, proxy, req, res, siteInf
...
-      secret: match.signingSecret,
-      siteID: siteInfo.id,
-      siteURL: siteInfo.url,
-    })
+    const signingSecretVar = env[match.signingSecret]
+
+    if (signingSecretVar) {
+      req.headers['x-nf-sign'] = signRedirect({
+        deployContext: 'dev',
+        secret: signingSecretVar.value,
```
charliegroll, line 165: is there a chance `signingSecretVar` will ever be falsy or not contain a `value`?

charliegroll (approved):
oops misread - LGTM :shipit:

PR #5428, opened 2023-01-24
feat: send functions_config during deploy

khendrikse:

#### Summary

---

      passes our tests.


src/utils/deploy/hash-fns.mjs
```diff
@@ -110,7 +117,12 @@ const hashFns = async (
     assetType: 'function',
     normalizedPath: path.basename(functionPath, path.extname(functionPath)),
     runtime,
+    displayName,
   }))
+  const fnConfig = functionZips.reduce(
+    (funcs, curr) => (curr.displayName ? { ...funcs, [curr.name]: { display_name: curr.displayName } } : funcs),
+    {},
+  )
```
Skn0tt, line 125: this reads very complicated to me. would this have the same effect?
```suggestion
  const fnConfig = Object.fromEntries(
     functionZips
       .filter(func => !!func.displayName)
       .map(func => [func.name, { display_name: func.displayName}])
   )
```

Skn0tt (approved):
looks generally good to me. left a tiny comment.

do we have tests for this behaviour? if yes, we should probably update one to assert on functions config. If no, we should make an issue to write tests.

src/utils/deploy/hash-fns.mjs
khendrikse: It would not be exactly the same effect, as the result of the reduce would be an object, and the map would be an array. But I did add a filter like you did now, in front of the reduce, maybe that already makes it a bit easier to read?

khendrikse:
> looks generally good to me. left a tiny comment.

do we have tests for this behaviour? if yes, we should probably update one to assert on functions config. If no, we should make an issue to write tests.

@Skn0tt I tried to add a test, but honestly it was... not as easy as I'd like. I'll make an issue to do this so I can spend a bit more time on it 👍

src/utils/deploy/hash-fns.mjs
Skn0tt: > and the map would be an array.

that's what the Object.fromEntries would do, wouldn't it?

khendrikse: Oooh sorry I totally forgot that 😆 in that case I think I would still prefer it as a reduce, for me that is a bit more readable, as it shows how the object will look. But I do understand that it might be personal preference as well 🤔

Skn0tt: sounds good 👍

Skn0tt (approved):

src/utils/deploy/hash-fns.mjs
```diff
@@ -110,7 +117,11 @@ const hashFns = async (
     assetType: 'function',
     normalizedPath: path.basename(functionPath, path.extname(functionPath)),
     runtime,
+    displayName,
   }))
+  const fnConfig = functionZips
```
eduardoboucas, line 122: I think in the future, and especially once we add more properties to this object, it would be nice to abstract this into a separate method and have a proper validation mechanism. For example, in here you're ensuring that you're not passing any unknown properties, but you're not doing anything to ensure that `display_name` has the right type.

src/utils/deploy/hash-fns.mjs
```diff
@@ -65,7 +67,12 @@ const getFunctionZips = async ({
 
   const { zipFunctions } = await import('@netlify/zip-it-and-ship-it')
 
-  return await zipFunctions(directories, tmpDir, { basePath: rootDir, config: functionsConfig })
+  return await zipFunctions(directories, tmpDir, {
+    featureFlags: { project_deploy_configuration_api_use_per_function_configuration_files: true },
```
eduardoboucas, line 71: Another note for future reference: in the same way we have a single place for [all edge functions feature flags](https://github.com/netlify/cli/blob/63d415b4220c0988f23507ec966240754a04d59b/src/lib/edge-functions/consts.mjs#L8), it would be good to do the same for serverless functions. This will ensure the different entry points into zip-it-and-ship-it use the same set of flags, and thus avoiding a situation where two commands have different behaviour leading to a harder time troubleshooting problems.

PR #5429, opened 2023-01-24
fix: load internal environment variables in Envelope

eduardoboucas:
#### Summary

https://github.com/netlify/cli/pull/5338 has changed the way the `NETLIFY_DEV` environment variable is set. Rather than mutating `process.env` immediately, we set a variable in the `env` object with the type `internal`, which is processed as a type of environment variable that can't be overridden.

Unfortunately, this wasn't added to the Envelope methods, and so the variable was being removed for sites with Envelope enabled.

This PR addresses that and adds some tests (to both legacy and new environment variable systems) to ensure we don't introduce this regression again.

### DIFF
--- a/src/commands/dev/dev.mjs
+++ b/src/commands/dev/dev.mjs
@@ -88,7 +88,8 @@ const dev = async (options, command) => {
 
   let { env } = cachedConfig
 
-  env.NETLIFY_DEV = { sources: ['internal'], value: 'true' }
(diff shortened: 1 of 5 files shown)


jasonbarry:
Thanks @eduardoboucas!! Once all sites are migrated to Envelope in Feb/March, we can delete a lot of code from the CLI that juggles env vars between the two systems. I'm looking forward to it 🔥

PR #5436, opened 2023-01-28
feat: support comments in vscode settings

### LINKED ISSUE #5407: Crash when reading `.vscode/settings.json` that contains comments
### Describe the bug

CLI fails to read `.vscode/settings.json` when it contains comments.

```
 ›   Error: Could not open VS Code settings file: Unexpected token / in JSON at position 4
```

I'd recommend using the [comment-json](https://www.npmjs.com/package/comment-json) npm package instead of native `JSON` to parse, patch, and stringify settings.json while preserving comments.

### Steps to reproduce

1. Run `netlify start`
2. Answer "yes" to the question "Would you like to configure VS Code to use Edge Functions?"

### Configuration

_No response_

### Environment

```
  System:
    OS: macOS 13.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.86 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  npmPackages:
    netlify-cli: 12.5.0 => 12.5.0
```

tinfoil-knight:

#### Summary

Fixes #5407 

This PR adds support for having comments in the `settings.json` file used by VSCode while using the `netlify recipes vscode` command.

---

      passes our tests.


src/recipes/vscode/settings.mjs
```diff
@@ -1,17 +1,17 @@
 import { mkdir, readFile, stat, writeFile } from 'fs/promises'
 import { dirname, relative } from 'path'
 
+import CommentJSON from 'comment-json'
```
danez, line 4: Looking at the types of comment-json, it looks like it does not have a default export: https://github.com/kaelzhang/node-comment-json/blob/master/index.d.ts#L106

could we instead do:
```js
import { assign, parse, stringify } from 'comment-json'
```

Or does that not work?

danez (changes requested):

src/recipes/vscode/settings.mjs
tinfoil-knight: It works. I've updated it.

kaelig:
Thank you so much for taking that on! I'd love to see some tests to prevent any future regressions 🙏🏻

tinfoil-knight:
> I'd love to see some tests to prevent any future regressions 🙏🏻

Will add a test for JSON with comments.

tests/integration/640.command.recipes.test.cjs
```diff
@@ -96,3 +97,46 @@ test('Does not generate a new VS Code settings file if the user does not confirm
...
+
+    const settingsText = await readFile(`${builder.directory}/.vscode/settings.json`, { encoding: 'utf8' })
+    t.log(settingsText)
+    t.true(settingsText.includes(comment))
+
+    const settings = parse(settingsText, null, true)
+    t.is(settings.someSetting, 'value')
+    t.is(settings['deno.enable'], true)
+    t.is(settings['deno.importMap'], '.netlify/edge-functions-import-map.json')
+    t.deepEqual([...settings['deno.enablePaths']], ['/some/path', 'netlify/edge-functions'])
```
tinfoil-knight, line 140: Spreading the array here since that converts it from the [CommentArray](https://www.npmjs.com/package/comment-json#commentarray) type to a native JS array for comparison.

tests/integration/640.command.recipes.test.cjs
```diff
@@ -96,3 +97,46 @@ test('Does not generate a new VS Code settings file if the user does not confirm
...
+    handleQuestions(childProcess, [
+      {
+        question: `There is a VS Code settings file at ${settingsPath}. Can we update it?`,
+        answer: answerWithValue(CONFIRM),
+      },
+      {
+        question: 'The Deno VS Code extension is recommended. Would you like to install it now?',
+        answer: answerWithValue(NO),
+      },
+    ])
```
tinfoil-knight, line 128: @danez Note here: 'The Deno VS Code extension is recommended. Would you like to install it now?' is asked on my local system but it seems like this question is not asked during the CI run. The test just ends up stuck & didn't run without handling this question.

(This is applicable for other tests in this file too.)

tinfoil-knight:
@kaelig added the test.

tests/integration/640.command.recipes.test.cjs
```diff
@@ -96,3 +97,45 @@ test('Does not generate a new VS Code settings file if the user does not confirm
     t.is(error.code, 'ENOENT')
   })
 })
+
+test.only('Handles JSON with comments', async (t) => {
```
eduardoboucas, line 101: Don't forget to remove `.only`.

```suggestion
test('Handles JSON with comments', async (t) => {
```

src/recipes/vscode/settings.mjs
```diff
@@ -1,17 +1,17 @@
 import { mkdir, readFile, stat, writeFile } from 'fs/promises'
 import { dirname, relative } from 'path'
 
+import { parse, assign, stringify } from 'comment-json'
```
eduardoboucas, line 4: This is totally down to personal preference, but I would suggest having a single `JSONC` import here and then use it throughout the file as `JSONC.parse`, `JSONC.stringify`, etc.

Seeing a function called `stringify` or `parse` doesn't give a lot of context and makes the code a little bit harder to parse, IMO.

tests/integration/640.command.recipes.test.cjs
tinfoil-knight: Thank you for spotting.

src/recipes/vscode/settings.mjs
tinfoil-knight: It was a single `CommentJSON` import earlier which is what I prefer. @danez suggested otherwise.

The separate functions seem slightly confusing to me too without any context. WDYT @danez?

src/recipes/vscode/settings.mjs
```diff
@@ -42,7 +42,7 @@ export const getSettings = async (settingsPath) => {
 
     return {
       fileExists: true,
-      settings: JSON.parse(file),
+      settings: parse(file.toString()),
```
eduardoboucas, line 45: Rather than doing a `toString()` here, I think you can do `await readFile(settingsPath, "utf8")` above?

tinfoil-knight: Yeah. That seems better.

src/recipes/vscode/settings.mjs
danez: The problem is that the TS types say that these are named exports. https://github.com/kaelzhang/node-comment-json/blob/master/index.d.ts#L83 So as soon as we switch to TS this will not work, because TS would say `no default export`.

eduardoboucas: Could we do `import * as JSSONC from "json-comments"`?

danez: Yeah I guess that should work.

tinfoil-knight:
@danez I think we can merge this now. Also see: https://github.com/netlify/cli/pull/5436/files#r1096563844

danez (approved):

PR #5461, opened 2023-02-10
chore: new fixtures and new setup for dev-server tests

danez:
This is a Proposal for a new setup on how we do tests in the CLI.

- Have the fixtures on the filesystem instead of using the SiteBuilder. This allows reusing fixtures in different tests and makes the codelines of a test shorter.
- Group similar dev-server tests into one test-file or describe()-block. With this we only start the dev-server once for multiple tests. This works for most tests, but not for all. Tests that do modify the file-system during the runtime of the dev-server or forcefully crash it should stay separate. This removes the clear separation of these tests, but the speed of the tests improves a lot.
- With the above we can now also run the tests inside a group concurrently. This further increases the speed.

So in the included test, the runtime of the test went from 20s to 3s.

https://user-images.githubusercontent.com/231804/218103670-8290abf4-983d-42fd-9508-bf13d6344554.mp4

I would love to hear what you think about this.

What do you think about the APIs created in this PR?
What about fixtures vs site-builder? I know that site-builder is nice because one can instantly see in the test what the files look like, but I personally do not think it is too hard to find the fixture and open the files.


tinfoil-knight (commented):
This seems really great! 20 seconds to 3 is a really big improvement.

---

The only issue I do have is that with the SiteBuilder, it was really clear what the minimal setup (of files & `netlify.toml` configuration) would be for a repository to test out or use a specific feature.

A single site fixture for multiple tests makes things slightly vague now.

I couldn't really think of any alternatives to this though.

A preset fixture for every test would be wasteful. We'd only get a slight performance improvement in that case which wouldn't really be worth the migration from SiteBuilder to fixtures.

---

Aside:

When I started contributing to Netlify a few years back, I was initially working on [Netlify CMS](https://github.com/netlify/netlify-cms) but stopped solely because the tests took too much time to run & would crash my laptop.

Currently, when fixing a bug on Netlify CLI or adding a feature, I often end up running a single test locally (which isn't the right thing to do) instead of an entire suite since it takes too much time & then I've to wait & pray that all tests pass on the CI for 10-15 minutes. This causes a lot of delay in case something needs to be fixed & really increases the feedback loop.

An entire test suite running in a few seconds is really awesome. Thank you!

P.S. Do tag me if I can help pick up some of the work here in migrating the tests to using fixtures.

danez:
> A single site fixture for multiple tests makes things slightly vague now.
> 
> I couldn't really think of any alternatives to this though.
> 
> A preset fixture for every test would be wasteful. We'd only get a slight performance improvement in that case which wouldn't really be worth the migration from SiteBuilder to fixtures.

Yeah, that is the compromise. Either each test has its own fixture, is completely isolated, and takes time to run or we group the fixtures, and lose some of the isolation between the tests but get performance improvements.

The change from SiteBuilder vs FileSystemFixture is debatable and has a minimal effect on performance. We could still use the SiteBuilder even in the scenario of only running on dev server for multiple tests.

lukasholzer (approved):
LGTM Awesome improvement 👍

khendrikse (approved):
Looks good :) excited to start working with this

PR #5480, opened 2023-02-16
fix: support multiple path values in edge functions config

eduardoboucas:
#### Summary

Fixes an issue where an error is thrown for edge functions defining a `path` property with multiple values via in-source configuration.


khendrikse (approved):
Looks good :) let's get this merged

PR #5490, opened 2023-02-19
feat!: bring dev server headers to parity with production

eduardoboucas:
#### Summary

Some of the headers returned by the CLI's dev server are different from what we return in production. This PR attempts to bring them closer.


src/lib/edge-functions/proxy.mjs
```diff
@@ -133,7 +132,6 @@ export const initializeProxy = async ({
       [headers.Functions]: functionNames.join(','),
       [headers.ForwardedHost]: `localhost:${passthroughPort}`,
       [headers.Passthrough]: 'passthrough',
-      [headers.RequestID]: generateUUID(),
```
eduardoboucas, line 136: Moved to the proxy so that all requests have it.

khendrikse (dismissed):

eduardoboucas:
@khendrikse I added and updated some tests so your review got stale. Could you give it another look, please?

khendrikse (approved):

src/utils/request-id.mjs
```diff
@@ -0,0 +1,6 @@
+// @ts-check
+import { v4 as generateUUID } from 'uuid'
+
+// Transform a v4 UUID to match the format used in production — alphanumeric
+// characters, all uppercase, 26 characters.
+export const generateRequestID = () => generateUUID().replace(/-/g, '').toUpperCase().slice(0, 26)
```
Skn0tt, line 6: if we wanted, we *could* use https://github.com/ulid/javascript to get the very same format as is being used in production, so if any of our users rely on the characteristics of ULIDs (e.g. the embedded timestamp), that'd make it easier to test in dev. I don't think anybody does that, nor that anybody *should*, so it's probably right to use UUID here. And it sames some NPM dead weight.

src/utils/static-server.mjs
```diff
@@ -21,7 +21,8 @@ export const startStaticServer = async ({ settings }) => {
   })
 
   server.addHook('onRequest', (req, reply, done) => {
-    reply.header('X-Powered-by', 'netlify-dev')
```
Skn0tt, line 24: should we add the `Server: Netlify` header? or `Server: Netlify Dev`, as we call it in https://docs.netlify.com/cli/get-started/#get-started-with-netlify-dev?

Skn0tt (dismissed):
looks good, left two minor non-blocking comments.

src/utils/request-id.mjs
eduardoboucas: Yeah I was torn on this myself. But if the idea is to bring parity with production as closely as possible, we might as well use the same format. Updated in https://github.com/netlify/cli/pull/5490/commits/d0227f71455c0aa8291cd3717760ebb590fc510f.

src/utils/static-server.mjs
eduardoboucas: Unless there's a reason to differentiate them, I'd rather use the same value in both environments.

eduardoboucas:
Changed to `feat!` as this is technically a breaking change.

src/utils/static-server.mjs
Skn0tt: okay, but right now we're not setting `Server` anywhere, right? should we?

eduardoboucas: We're not. Production gives you `server: Netlify` and CLI gives you `x-powered-by: netlify-dev`.

Skn0tt: that's what I figured. This PR is about moving the CLI closer to prod together, so let's add `reply.header("Server", "Netlify")`

eduardoboucas: I'm confused. That's what we're doing [here](https://github.com/netlify/cli/pull/5490/files#diff-5c5309633069037040a13b82d7abcfffe577c7516e6d356e46f407d729aa98a8R392), except we're doing it for all responses and not just the ones served by the static server.

Skn0tt: I was confused. Didn't see that, sorry.

Skn0tt (approved):

PR #5538, opened 2023-03-08
fix: wrong env values could be used for non-dev context in `netlify deploy --build --context`

jasonbarry:
#### Summary

If a site has environment variables with different values for each deploy context, and they deploy to prod using `netlify deploy --build --prod --context production`, then the dev values will be used instead. This is because`options.context` wasn't making its way to the Envelope call (unlike how we do it in `netlify build`, `netlify dev`, and `netlify env`). This PR passes the context from the `--context` flag over to Envelope so the correct values are pulled. 

In doing so, I noticed that env vars from `build_settings.env` were getting included in the functionsConfig. While these values are the site's env vars [from Envelope](https://github.com/netlify/build/blob/main/packages/config/src/api/site_info.js#L25-L29), they are the ones that apply to any scope and any context. Instead, we should pass in only the ones that match the given `--context`, and only the `functions` scope. 

Customer report Slack thread: https://netlify.slack.com/archives/C02TE7L1QUA/p1678283162339439

---

      passes our tests.


src/utils/env/index.mjs
```diff
@@ -145,6 +146,18 @@ export const getEnvelopeEnv = async ({ api, context = 'dev', env, key = '', scop
 
   const accountEnv = formatEnvelopeData({ context, envelopeItems: accountEnvelopeItems, scope, source: 'account' })
   const siteEnv = formatEnvelopeData({ context, envelopeItems: siteEnvelopeItems, scope, source: 'ui' })
+
+  if (raw) {
+    const entries = Object.entries({ ...accountEnv, ...siteEnv })
```
jasonbarry, line 151: Is it correct that only account and site env vars, and not general / addons / configFile env vars, need to be included as siteEnv for the functionsConfig? That's how it seems to be before -- only `get(siteData, 'build_settings.env')` was getting passed to functionsConfig.

eduardoboucas: Yes, I think that's right.

eduardoboucas (approved):

src/commands/deploy/deploy.mjs
```diff
@@ -567,6 +568,16 @@ const deploy = async (options, command) => {
     return triggerDeploy({ api, options, siteData, siteId })
   }
 
+  const isUsingEnvelope = siteData && siteData.use_envelope
+  // if a context is passed besides dev, we need to pull env vars from that specific context
+  if (isUsingEnvelope && options.context && options.context !== 'dev') {
```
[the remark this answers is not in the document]
jasonbarry: `options.context` could be `undefined` if the user doesn't pass the `--context` flag, but the Envelope call defaults to use the `dev` context in that case. So if context isn't passed, or if context is `dev`, then we can skip the call. We could remove this bit if we gave the option a default value besides undefined, but I'm unsure if that would be a breaking change!

src/utils/env/index.mjs
jasonbarry: OK cool! then I think this would fix a bug where account-level env vars weren't getting applied to functions if a user deployed with this method.

Issue #5544, opened 2022-11-17
onPreDev and onDev not not applying netlifyConfig changes


sean-roberts:
### Describe the bug

Modifying the netlify config during onDev/onPreDev are not being reflected on the dev output. So, if I update the headers during the event it will show that the headers are updated via the output, but the changes aren't reflecting.

When doing this process within onPreBuild, the functionality works as expected

### Steps to reproduce

`plugin.js`
```
function pluginWork({netlifyConfig}){
  const headers = netlifyConfig.headers;
  headers.push({
    for: '/*',
    values: {
      'x-test': 'value'
    }
  });
  netlifyConfig.headers = headers;
}

export const onPreDev = pluginWork;
```

### CLI command and flags

`ntl dev`

### CLI output

during `ntl dev` the plugin output includes:
```
Netlify configuration property "headers" value changed to [ { for: '/*', values: { 'x-test': 'value' } } ].
```

However, no requests are returning with that header. There are no other headers defined elsewhere in the product and it's not 

### Environment

```
System:
    OS: macOS 12.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 600.77 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - /opt/homebrew/opt/nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/opt/nvm/versions/node/v16.15.1/bin/yarn
    npm: 8.11.0 - /opt/homebrew/opt/nvm/versions/node/v16.15.1/bin/npm
  npmGlobalPackages:
    netlify-cli: 12.1.1
```

PR #5546, opened 2023-03-09
fix: fix use headers that are set via the onPreDev of a plugin

lukasholzer:

#### Summary

Depends on: https://github.com/netlify/build/pull/4932

Fixes https://github.com/netlify/cli/issues/5544

This PR only parses the headers but we might want to do the same for redirects in a follow up or other config mutations not sure what frameworks needs here? 

cc @kitop @ascorbic 

---

      passes our tests.


eduardoboucas:
How does this handle live updates to the config/headers/redirects files? For example, let's say that I have a plugin with a `onPreDev` event that reads redirect rules and mutates them somehow. On the first run, the plugin will read the input, do the mutations, and produce the right output. But then I make a change to the redirects file and the plugin is not running again, so will we get the right output?

lukasholzer:
> How does this handle live updates to the config/headers/redirects files? For example, let's say that I have a plugin with a `onPreDev` event that reads redirect rules and mutates them somehow. On the first run, the plugin will read the input, do the mutations, and produce the right output. But then I make a change to the redirects file and the plugin is not running again, so will we get the right output?

So if you have headers in a `_headers` or `netlify.toml` file they get picked up and merged with the plugin config:

```
/** @type {import('@netlify/build').Om} */
export const onPreDev = async ({ netlifyConfig }) => {
  console.log('inside onPreDev');
  const headers = netlifyConfig.headers;
  headers.push({
    for: '/*',
    values: {
      'x-test': 'value',
    },
  });
  
  netlifyConfig.headers = headers;
};
```

If you change the toml file and update the header the header from the plugin is still in there:
CleanShot 2023-03-09 at 13 45 49

The `onPreDev` hook i only run before the command and has no watch mode so changes to that would need a restart of `ntl dev` 


Does this answer your question?

eduardoboucas:
The use case I'm thinking is something like this:

```js
export const onPreDev = async ({ netlifyConfig }) => {
  const { headers } = netlifyConfig;

  headers[0].values['x-test'] += " modified"
  
  netlifyConfig.headers = headers;
};
```

So the plugin appends something to the value of a header. If I re-save the file and change `foo` to `bar`, will `x-test` change from `foo modified` to `bar modified`?

(I tried checking out the branch and trying this myself, but I'm getting the same error with `@netlify/config` that we're seeing in the CI.)

lukasholzer:
@eduardoboucas should work now, the newer version of @netlify/config was not in. Now it's merged into it: https://github.com/netlify/cli/commit/bd9823e5ae9994cadc39e1c9571047b340515811

THat was the dependent PR from the description (had it linked locally when developing)

lukasholzer:
> ```js
> 
>   headers[0].values['x-test'] += " modified"
>   
> ```

@eduardoboucas you are right this behaviour is not working as the `onPreDev` is only executed before starting the devServer and NOT during reloads.

TBH I would see this as "expected" behaviour as no build hook is run on file updates.

eduardoboucas:
I'm not sure I agree. It's true that no hooks currently run on file updates, but that's because all hooks used to be tailored for the build process, and they didn't run in Netlify Dev at all. That changed with `onPreDev` and `onDev`.

If we say that these plugin hooks will run in Netlify Dev, and if we say that Netlify Dev has live reloading for any platform configuration files, I think we need to make these two concepts co-exist, otherwise we're creating the precedent that the state of your application may differ based on whether you've updated any files. I think that somewhat defeats the purpose of supporting live reloading in the first place.

The way I was imagining we'd approach this is by running the `*Dev` hooks on file reloads. Netlify Build already has the plumbing for running just the dev pipeline, we'd just need to accommodate the fact that on subsequent runs we'll leave the initial dev command alone and run just the hooks.

ascorbic:
I agree with Eduardo. It would be unexpected if saving a file meant the headers suddenly changed

sean-roberts:
Chiming in from the future...

for onPreDev.. I expect them to run once - I see it as on pre dev command. For onDev, I could see them running multiple times based on throttled file changes.

Can we get this shipped and sort out onDev later? Even if we want to sort out the watching + invoking of onPreDev (I'd disagree with the need) but the initial pass should be simply allowing them to work _at all_

billwatson017:
bumping this issue as we are having the same experience with pushing dynamic headers 

``` JS
module.exports = {
  onPreDev({ netlifyConfig }) {
    netlifyConfig.redirects.push({
      from: '/*',
      to: `${netlifyConfig.build.environment.APP_BASE_URL}/:splat`,
      status: 200,
    });
  },
};
```

sean-roberts:
due to _almost_ a year, creating a new PR was easier than sorting the merge conflicts... so, let's carry the convo over here: https://github.com/netlify/cli/pull/6365

PR #5547, opened 2023-03-09
fix!: publish shrinkwrap with prod dependencies only

danez:
#### Summary

Ref #5323 

This is a reattempt of #5359 which fixes the problems that occurred last time.

It removes the shrinkwrap from the repo and instead uses the normal package-lock.

On publish we remove devDependencies, scripts & ava properties from package.json and then create the shrinkwrap. This new way ensures that all optional dependencies are present in the shrinkwrap file. (Last time the optional dependencies were missing)

Also split up the release workflow into 2 jobs so we do not have to repeat the if: `${{ steps.release.outputs.release_created }}` for every single step.


scripts/prepare-for-publish.mjs
```diff
@@ -0,0 +1,43 @@
...
+}).start()
+await execa('npm', ['install', '--no-audit'])
+spinner.succeed()
+
+spinner = ora({
+  spinner: 'star',
+  text: 'Running `npm shrinkwrap`',
+}).start()
+await execa('npm', ['shrinkwrap'])
+spinner.succeed()
```
danez, line 43: This is the main change. This script will run before publishing (see package.json).

EverlastingBugstopper:
Thanks for the work on this!

lukasholzer (approved):
Should we mark it as a breaking change?

PR #5572, opened 2023-03-15
feat: add canonical redirect for static html files

danez:
#### Summary

Fixes https://github.com/netlify/pod-compute/issues/432

This adds canonical redirect for HTML files, the same as in production.

This is done on every request before proxying to the static file server or the frameworks dev server. Although if a framework uses an in-memory dev server we cannot mimic the behavior, as there are no static files to check against.

Most of the new files is restructuring so that I was able to write unit tests for each file.

---

      passes our tests.

### DIFF
--- a/src/utils/proxy.mjs
+++ b/src/utils/proxy.mjs
@@ -34,8 +34,8 @@ import { NETLIFYDEVLOG, NETLIFYDEVWARN, log, chalk } from './command-helpers.mjs
 import createStreamPromise from './create-stream-promise.mjs'
 import { headersForPath, parseHeaders, NFRequestID } from './headers.mjs'
 import alternativePathsFor from './proxy/alternative-paths-for.mjs'
+import findStaticFileForURLPath from './proxy/find-static-file-for-url-path.mjs'
 import getCanonicalURLPath from './proxy/get-canonical-url-path.mjs'
-import getStatic from './proxy/get-static.mjs'
 import isFunction from './proxy/is-function.mjs'
 import { generateRequestID } from './request-id.mjs'
 import { createRewriter, onChanges } from './rules-proxy.mjs'
@@ -195,7 +195,7 @@ const serveRedirect = async function ({ env, match, options, proxy, req, res, si
 
   const reqUrl = reqToURL(req)
 
-  const staticFile = await getStatic(decodeURIComponent(reqUrl.pathname), options.publicFolder)
+  const staticFile = await findStaticFileForURLPath(decodeURIComponent(reqUrl.pathname), options.publicFolder)
   if (staticFile) {
     req.url = encodeURI(staticFile) + reqUrl.search
     // if there is an existing static file and it is not a forced redirect, return the file
@@ -258,7 +258,7 @@ const serveRedirect = async function ({ env, match, options, proxy, req, res, si
       return proxy.web(req, res, { target: options.functionsServer })
     }
 
-    const destStaticFile = await getStatic(dest.pathname, options.publicFolder)
+    const destStaticFile = await findStaticFileForURLPath(dest.pathname, options.publicFolder)
     let statusValue
     if (match.force || (!staticFile && ((!options.framework && destStaticFile) || isInternal(destURL)))) {
       req.url = destStaticFile ? destStaticFile + dest.search : destURL
@@ -502,7 +502,7 @@ const onRequest = async (
   if (req.method === 'GET') {
     const url = reqToURL(req)
     const decodedUrl = decodeURIComponent(url.pathname)
-    const staticFile = await getStatic(decodedUrl, settings.dist)
+    const staticFile = await findStaticFileForURLPath(decodedUrl, settings.dist)
 
     if (staticFile) {
       const unixStaticFile = unixify(staticFile)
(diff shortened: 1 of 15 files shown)


src/utils/proxy/get-canonical-url-path.mjs
```diff
@@ -0,0 +1,25 @@
+import { basename, extname } from 'path'
+
+/**
```
eduardoboucas, line 3: The JSDoc comment doesn't match the signature.

src/utils/proxy/alternative-paths-for.mjs
```diff
@@ -0,0 +1,33 @@
+import isFunction from './is-function.mjs'
+
+// Used as an optimization to avoid dual lookups for missing assets
+const assetExtensionRegExp = /\.(html?|png|jpg|js|css|svg|gif|ico|woff|woff2)$/
+
+/**
+ * For an URL get alternative static file paths
```
eduardoboucas, line 7: Can you add a bit more detail about what "alternative paths" means in this context?

src/utils/proxy/get-static.mjs
```diff
@@ -0,0 +1,26 @@
...
+
+import alternativePathsFor from './alternative-paths-for.mjs'
+
+/**
+ *
+ * @param {string} pathname
+ * @param {string} publicFolder
+ * @returns {Promise<string | undefined>}
+ */
+const getStatic = async function (pathname, publicFolder) {
```
eduardoboucas, line 13: Can we find a more descriptive name for this? Or failing that, can we add a description to the JSDoc comment?

src/utils/proxy/is-function.mjs
```diff
@@ -0,0 +1,12 @@
+const FUNCTION_PATH_REGEX = /^\/.netlify\/(functions|builders)\/.+/
+
+/**
+ * Is the URL a function url
```
eduardoboucas, line 4:
```suggestion
 * Checks whether the URL targets a serverless function
```

src/utils/proxy.mjs
```diff
@@ -513,11 +484,46 @@ const initializeProxy = async function ({ configPath, distDir, env, host, port,
...
+    const staticFile = await getStatic(decodedUrl, settings.dist)
+
+    if (staticFile) {
+      const unixStaticFile = unixify(staticFile)
+      if (decodedUrl !== unixStaticFile && (unixStaticFile.endsWith('.html') || unixStaticFile.endsWith('.htm'))) {
+        const canonicalPath = getCanonicalURLPath(unixStaticFile)
+
+        if (canonicalPath !== decodedUrl) {
+          console.log(`${NETLIFYDEVLOG} Redirecting to canonical URL ${decodedUrl} -> ${canonicalPath}`)
+          res.setHeader('age', '0')
```
eduardoboucas, line 514: Feels a bit odd to be repeating the code where we set these headers. Do you think it would make sense to centralise that somehow?

src/utils/proxy.mjs
```diff
@@ -513,11 +484,46 @@ const initializeProxy = async function ({ configPath, distDir, env, host, port,
...
+    const decodedUrl = decodeURIComponent(url.pathname)
+    const staticFile = await getStatic(decodedUrl, settings.dist)
+
+    if (staticFile) {
+      const unixStaticFile = unixify(staticFile)
+      if (decodedUrl !== unixStaticFile && (unixStaticFile.endsWith('.html') || unixStaticFile.endsWith('.htm'))) {
+        const canonicalPath = getCanonicalURLPath(unixStaticFile)
+
+        if (canonicalPath !== decodedUrl) {
+          console.log(`${NETLIFYDEVLOG} Redirecting to canonical URL ${decodedUrl} -> ${canonicalPath}`)
```
eduardoboucas, line 513: Do we need this log message?

src/utils/proxy/alternative-paths-for.mjs
```diff
@@ -0,0 +1,33 @@
+import isFunction from './is-function.mjs'
+
+// Used as an optimization to avoid dual lookups for missing assets
```
eduardoboucas, line 3: Could you explain this? Why are we associating these specific extensions with missing assets?

eduardoboucas:
We should turn this into a `feat:` or `fix:` before merging.

src/utils/proxy/alternative-paths-for.mjs
danez: I do not know what this is, I just copied this from `proxy.mjs`

src/utils/proxy.mjs
danez: I can remove it.

src/utils/proxy.mjs
danez: Problem is that this is the default Response object which uses `setHeader()`. In the static server it uses a FastifyResposne object which uses `header()`.

src/utils/proxy.mjs
```diff
@@ -2,7 +2,8 @@
 import { Buffer } from 'buffer'
 import { once } from 'events'
 import { readFile } from 'fs/promises'
-import http from 'http'
+// eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
danez: Yes this is only for the types as long as we do not have TypeScript

danez:
Okay updated.

eduardoboucas:
> This is done on every request before proxying to the static file server or the frameworks dev server. Although if a framework uses an in-memory dev server we cannot mimic the behavior, as there are no static files to check against.

I've been thinking about this, and I think it is a pretty big limitation. If we can't get around it, I wonder about the value of introducing this change, because we'll be incurring the cost of changing existing behaviour without achieving the goal of true parity with production.

I wonder if there's a world in which rather than handling the request before proxying to the static server, we do it on the way out, perhaps with a combination of [the `selfHandleResponse` option](https://www.npmjs.com/package/http-proxy#options)  and the `proxyRes` event? (When we redirect, we still return the data in the body.)

Issue #5573, opened 2023-03-16
CLI is not showing an update banner


lukasholzer:
### Describe the bug

See this issue: https://github.com/netlify/next-runtime/issues/1996

I was using a old CLI version without getting the update banner (therefore I assumed I'm on the latest version and it was crashing)

### Steps to reproduce

hard to reproduce. I thinnk just installing an outdated version and run `ntl dev`

### Environment

 netlify-cli

  System:
    OS: macOS 13.2.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 11.66 GB / 64.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.14.2 - ~/.nvm/versions/node/v18.14.2/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.14.2/bin/yarn
    npm: 9.5.0 - ~/.nvm/versions/node/v18.14.2/bin/npm
  npmGlobalPackages:
    netlify-cli: 13.1.1

danez:
We only check every 12 hours, but I assume the new version was older than 12 hours?

PR #5577, opened 2023-03-17
chore: add unit test for getPackageJson

danez:
#### Summary

While looking at #5573 I created this unit test

### DIFF
--- a/tests/unit/utils/get-package-json.test.mjs
+++ b/tests/unit/utils/get-package-json.test.mjs
@@ -1,0 +1,38 @@
+import { readFile } from 'fs/promises'
+
+import { beforeEach, describe, expect, test, vi } from 'vitest'
+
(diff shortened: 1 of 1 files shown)


tests/unit/utils/get-package-json.test.mjs
```diff
@@ -0,0 +1,38 @@
...
+  test('should return the package.json of netlify-cli', async () => {
+    const packageJson = await getPackageJson()
+
+    expect(packageJson).not.toBeUndefined()
+    expect(packageJson.name).toBe('netlify-cli')
+
+    expect(readFile).toHaveBeenCalledOnce()
+  })
+
+  test('should not reread package.json', async () => {
```
eduardoboucas, line 30: nit: It might be nice to somehow indicate that this test is coupled to the previous one, as the only reason it doesn't re-read `package.json` is that the previous test triggers the first read.

PR #5590, opened 2023-03-23
chore: migrate e2e to vitest

danez:
#### Summary

I created this Pr because I tried to disable telemetry in CI just to notice we already have a check for this in the code. Here are the other changes I made :)

- update vitest and use ts config
- ensure that the global config folder is correctly backup and restored before modifying the config in tests
- use vitest for e2e tests

The backup & restore part came because I noticed that the tests were not correctly restoring the config file locally and overwriting my config :(

The telemetry tests also fail on main, so I fixed them and migrated to vitest.


tests/integration/utils/mock-api-vitest.mjs
```diff
@@ -0,0 +1,94 @@
...
+    context.requests = mockApi.requests
+    mockApi.clearRequests()
+  })
+
+  afterAll(() => {
+    mockApi.server.close()
+  })
+
+  factory()
+}
```
danez, line 82: I created this so that we only start the mockAPI once for multiple tests and it works in the vitest API.

lukasholzer (approved):
LGTM Vitest for the win :)

PR #5244, opened 2022-11-19
fix: prevent cli install for incompatible node versions

tinfoil-knight:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes #4489 

Netlify CLI currently installs for incompatible NodeJS versions (may show a warning) but fails when Netlify CLI runs due to newer features/syntax not being supported by older versions of NodeJS.

This PR adds a preinstall script to report an error & prevent installing Netlify CLI if the NodeJS version on the host machine is  incompatible.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [x] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #5602, opened 2023-03-30
feat!: drop node 14.16 and 14.17

### LINKED ISSUE #5575: Not accurate node engine version range - it should be `^14.18.0 || >=16.0.0`
### Describe the bug

In `package.json` there is following metadata:

```
  "engines": {
    "node": "^14.16.0 || >=16.0.0"
  },
```

But package actually fails to install on Node 14.16 when running `postinstall` script:

```
Error: Cannot find module 'node:fs'
Require stack:
- /Users/misiek/dev/netlify-cli/node_modules/@fastify/send/lib/SendStream.js
- /Users/misiek/dev/netlify-cli/node_modules/@fastify/send/index.js
- /Users/misiek/dev/netlify-cli/node_modules/@fastify/static/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/misiek/dev/netlify-cli/node_modules/@fastify/send/lib/SendStream.js:13:12)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/misiek/dev/netlify-cli/node_modules/@fastify/send/lib/SendStream.js',
    '/Users/misiek/dev/netlify-cli/node_modules/@fastify/send/index.js',
    '/Users/misiek/dev/netlify-cli/node_modules/@fastify/static/index.js'
  ]
```

This is because support for `require('node:X')` was only added in Node 14.18 ( expand `history` in https://nodejs.org/api/esm.html#node-imports for reference ).

This is the case for current and previous major of `netlify-cli`.

I think there is no non-hacky way to fix this to make it work with reported minimum node version + the fact that it wasn't working on that min version it might make sense to adjust package.json to say Node@14.18 - which technically is a breaking change, but that is factual state of supported node version, just wrongly reported.

The hacky way is something like this https://github.com/gatsbyjs/gatsby/blob/master/e2e-tests/production-runtime/polyfill-node-protocol-imports.js but we are only doing this for e2e tests, not actually releasing that to npm and I would not suggest to adopt something like that for package that is released to npm.

This issue came up in https://github.com/netlify/gatsby-plugin-netlify/pull/232

### Steps to reproduce

Easiest is to go inside `netlify-cli` repo clone dir and run following

```
nvm install 14.16
npm run postinstall
```

### Configuration

_No response_

### Environment

```
  System:
    OS: macOS 13.2.1
    CPU: (12) x64 Apple M2 Pro
    Memory: 1.22 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.16.0 - ~/.nvm/versions/node/v14.16.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.14.0/bin/yarn
    npm: 6.14.11 - ~/.nvm/versions/node/v14.16.0/bin/npm
```

danez:

#### Summary

Drops node 14.16 and 14.17 as discussed

Fixes #5575

---

      passes our tests.


lukasholzer (approved):
🥳 Yay

tests/integration/510.hugo.test.cjs
```diff
@@ -1,45 +0,0 @@
...
-
-    t.true(response.includes('Home page!'))
-  })
-}
-
-// This feels awkward, but ava will throw if we have a file with no tests, so
-// we need something to run in case the test above is skipped.
-test('dummy test', (t) => {
-  t.true(true)
-})
```
danez, line 45: This test was failing on Node.js 14.18 on windows, after migrating to vitest it works. 🤷

Also the test was not run on latest node version, although it succeeds on these versions now, so removed the restriction where to run the test

tests/integration/utils/fixture.mjs
```diff
@@ -1,8 +1,9 @@
 // @ts-check
 import { rm } from 'fs/promises'
+import { join } from 'path'
 import { fileURLToPath } from 'url'
 
-import cpy from 'cpy'
+import { copy } from 'fs-extra'
```
danez, line 6: `cpy` cannot copy symlinks, so I exchanged it for `fs-extra -> copy`

tests/integration/utils/fixture.mjs
```diff
@@ -29,7 +30,7 @@ export const setupFixtureTests = async function (options, factory) {
 
   beforeAll(async () => {
     if (options.fixture) fixture = await Fixture.create(options.fixture)
-    if (options.devServer) devServer = await startDevServer({ cwd: fixture.directory })
+    if (options.devServer) devServer = await startDevServer({ cwd: fixture.directory, args: ['--offline'] })
```
danez, line 33: when we run fixture tests we should never read settings from the API.

lukasholzer: good point!

tests/integration/utils/fixture.mjs
lukasholzer: whats about `fs.cp` with `recursive` option? I think this uses the file system copy under the hood which should copy symlinks as well

danez: Only available as of Node.js 16.7.0

lukasholzer (commented):
@danez can we update the postinstall script to not break if being installed on a unsupported engine version?

As the engine is only a warning?

Instead exit the postinstall script with a better error messaging?
CleanShot 2023-04-11 at 13 04 27

danez:
I guess what you want is something like this: https://github.com/netlify/cli/pull/5244

We can try to make the postinstall script Node.js version agnostic, but this is unrelated to this PR.

lukasholzer (dismissed):
LGTM

lukasholzer (approved):

tests/integration/__fixtures__/hugo-site/package.json
```diff
@@ -6,6 +6,6 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "hugo-bin": "^0.101.0"
+    "hugo-bin": "^0.101.5"
```
XhmikosR, line 9: @danez I think you could also go with https://github.com/fenneclab/hugo-bin/releases/tag/v0.102.0 :)

PR #5607, opened 2023-04-08
fix(live): make LIVE_URL env var available in --live mode

biilmann:
#### Summary

Before this fix there was no way to access the live URL from functions or build plugins when running `netlify dev --live`.

It might even be expected the `URL` or `DEPLOY_URL` would map to the live url, but I wanted to make sure I didn't introduce any unexpected behavior.

Oddly, there's a `BASE_URL` env var set by the `handleLiveTunnel` function that I'm not sure is documented anywhere and that doesn't seem very descriptively named.

Again, to avoid causing regressions for anything relying on this `BASE_URL` I added an additional `LIVE_URL` instead of changing is.

DALL·E 2023-04-07 18 55 02 - A picture of a cute animal (not mandatory, but encouraged)


eduardoboucas (commented):
@biilmann Can you share some more details about where you've tried to access the URL from? I just created a basic function and I see the `URL` variable pointing to the live URL:

```js
const { env } = require("process");

module.exports.handler = async () => ({
  statusCode: 200,
  body: env.URL,
});

```

```
% curl https://eduardo-tests-de6a54.netlify.live/.netlify/functions/test-js
https://eduardo-tests-de6a54.netlify.live
```

It feels like this is the expected behaviour, and I'd like to get to the bottom of what's working and not working before introducing a new environment variable, because that has implications like figuring out what happens if someone is already using a variable with that name on their site, etc.

biilmann:
Hmm, it might be more of a niche bug than I thought, I was accessing the process.URL environment variable from a build plugin using the onPreDev hook, and there I never get the live url...

eduardoboucas:
Ah, that definitely sounds like a bug.

eduardoboucas:
I've opened an issue for this: https://github.com/netlify/cli/issues/5612

biilmann:
PR updated, this should be a real fix for the issue of URL and DEPLOY_URL behaving weird for builds and build plugins in dev mode.

So far the URl was always set to the main url for the site for build plugins and builds run as part of netlify dev and the DEPLOY_URL was always a broken URL (with 0 as deploy_id).

This moves up the URL definition to before any of the dev servers are started and passes the URL and DEPLOY_URL for the neetlify dev server as env overrides for the build command.

src/commands/dev/dev.mjs
```diff
@@ -124,9 +125,20 @@ const dev = async (options, command) => {
...
+
   log(`${NETLIFYDEVWARN} Setting up local development server`)
 
-  const { configPath: configPathOverride } = await runDevTimeline({ cachedConfig, options, settings, site })
+  const { configPath: configPathOverride } = await runDevTimeline({ cachedConfig, options, settings, site, env: {
+      URL: url,
+      DEPLOY_URL: url,
+  } })
+
+  log(`${NETLIFYDEVWARN} Done with runDevTimeline`)
```
eduardoboucas, line 141: This is a user-facing log line. Did you intend to leave it here?

src/commands/dev/dev.mjs
```diff
@@ -124,9 +125,20 @@ const dev = async (options, command) => {
     startPollingForAPIAuthentication({ api, command, config, site, siteInfo })
   }
 
+  const liveTunnelUrl = await handleLiveTunnel({ options, site, api, settings })
+  const url = liveTunnelUrl || getProxyUrl(settings)
+  process.env.URL = url
```
eduardoboucas, line 130: Note for our future selves: it would be nice if we could use [this mechanism](https://github.com/netlify/cli/blob/fe4210037d9e36e28109dea53425b8889adc052b/src/commands/dev/dev.mjs#L92) for setting internal environment variables rather than setting them on `process.env` directly, but that would mean moving the `injectEnvVariables` method further down and we'd need to look into any side-effects of that.

Not worth the hassle for now.

eduardoboucas (commented):
LGTM! Left a question about a log line, but other than that this is good to go. 🚀

eduardoboucas (approved):
I pushed https://github.com/netlify/cli/pull/5607/commits/df4ab5ed00d605c348032762777840f860ed9b1f with a couple of linting fixes. It's good to go now. 🚢 

Thanks @biilmann! 🙌🏻

PR #5623, opened 2023-04-13
fix: determine duplicate function precedence based on extension

khendrikse:

#### Summary

This again implements function precedence based on extension by reversing the functions array so it behaves similar to how production works. I also had to reverse the way our sourcedirectories are passed.

---

      passes our tests.


src/lib/functions/registry.mjs
```diff
@@ -212,7 +212,10 @@ export class FunctionsRegistry {
     await Promise.all(deletedFunctions.map((func) => this.unregisterFunction(func.name)))
 
     await Promise.all(
-      functions.map(async ({ displayName, mainFile, name, runtime: runtimeName }) => {
+      // zip-it-ship-it returns an array sorted based on which extension should have precedence.
+      // where the last ones precede the previous ones. This is why
+      // we reverse the array so we get the right functions precedence in the cli.
```
eduardoboucas, line 217:
```suggestion
      // we reverse the array so we get the right functions precedence in the CLI.
```

src/lib/functions/server.mjs
```diff
@@ -222,9 +222,9 @@ export const startFunctionsServer = async (options) => {
       functionsDirectories.push(distPath)
     }
   } else {
-    // The order of the function directories matters. Leftmost directories take
+    // The order of the function directories matters. Rightmost directories take
```
eduardoboucas, line 225: Hmm, this feels unrelated to the extension precedence. Is this a separate problem you've identified?

eduardoboucas: But we're reversing the array of the functions that come out, not the list of directories going in?

(Sorry if I'm missing the point.)

eduardoboucas (approved):

PR #5624, opened 2023-04-13
fix: correctly execute user and integrations edge functions in the right order

danez:
#### Summary

This turned out to be a bigger change than anticipated. The main change is splitting up the directories in user and internal, so that we can call `bundler.find` for each of them. We still have to run them together, but the functionConfig returned from `bundler.serve` is again split up, so that we can generate the correct manifest with the correct order.

---

      passes our tests.


src/lib/edge-functions/registry.mjs
```diff
@@ -28,157 +86,153 @@ export class EdgeFunctionsRegistry {
...
-        (func) => func.name === existingFunc.name && func.path === existingFunc.path,
-      )
+  /**
+   * @returns {Promise<void>}
+   */
+  async #checkForAddedOrDeletedFunctions() {
+    const [internalFunctions, userFunctions] = await Promise.all([
+      this.#scanForFunctions(this.#internalDirectories),
+      this.#scanForFunctions(this.#directories),
+    ])
```
danez, line 194: 1. We read files separately for internal vs. user

src/lib/edge-functions/registry.mjs
```diff
@@ -28,157 +86,153 @@ export class EdgeFunctionsRegistry {
...
+
+      this.#internalFunctionConfigs = this.#internalFunctions.reduce(
+        // eslint-disable-next-line no-plusplus
+        (acc, func) => ({ ...acc, [func.name]: functionsConfig[index++] }),
+        {},
+      )
+
+      this.#userFunctionConfigs = this.#userFunctions.reduce(
+        // eslint-disable-next-line no-plusplus
+        (acc, func) => ({ ...acc, [func.name]: functionsConfig[index++] }),
```
danez, line 175: 2. We create separate functionConfigs for user vs internal

khendrikse (commented):
I'm curious why you decided to put in the work to transform the fields to private fields? Is it necessary in order for this fix to work?

danez:
> I'm curious why you decided to put in the work to transform the fields to private fields? Is it necessary in order for this fix to work?

No it wasn't strictly necessary, but as I moved things around inside the registry I was never sure if this method/property is public or private.
At first i started adding TSDoc comments so that I know what is public API and what is not, but then decided to switch to private fields.

I know it makes the PR a little hard to review.

khendrikse:
> > I'm curious why you decided to put in the work to transform the fields to private fields? Is it necessary in order for this fix to work?
> 
> No it wasn't strictly necessary, but as I moved things around inside the registry I was never sure if this method/property is public or private. At first i started adding TSDoc comments so that I know what is public API and what is not, but then decided to switch to private fields.
> 
> I know it makes the PR a little hard to review.

Yeah I think in general it makes the code harder to read, which makes me wonder if it is worth the trade-off.

khendrikse (approved):

PR #5640, opened 2023-04-18
feat: add bugsnag reporting to the CLI application

lukasholzer:

#### Summary

Fixes https://github.com/netlify/pillar-workflow/issues/1224

---

      passes our tests.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         node-version: ['14.18.0', '*']
+      fail-fast: false
     steps:
       # Sets an output parameter if this is a release PR
       - name: Check for release
--- a/functions/error-reporting.ts
+++ b/functions/error-reporting.ts
@@ -1,0 +1,43 @@
+import process from 'process'
+
+import Bugsnag from '@bugsnag/js'
+import { Handler } from '@netlify/functions'
+
+Bugsnag.start({
+  apiKey: `${process.env.NETLIFY_BUGSNAG_API_KEY}`,
+})
+
+export const handler: Handler = async ({ body }) => {
+  try {
+    if (typeof body !== 'string') {
+      return { statusCode: 200 }
+    }
+    const {
+      cause,
+      cliVersion,
+      message,
+      name,
+      nodejsVersion,
+      osName,
+      severity = 'error',
+      stack,
+      user,
+    } = JSON.parse(body)
+    Bugsnag.notify({ name, message, stack, cause }, (event) => {
+      event.setUser(user.id, user.email, user.name)
+      event.severity = severity
+      event.device = {
+        osName,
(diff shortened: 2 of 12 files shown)


jobala (dismissed):

.github/workflows/unit-tests.yml
```diff
@@ -14,6 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         node-version: ['14.18.0', '*']
+      fail-fast: false
```
lukasholzer, line 17: I think the fail fast is not good on flaky tests as it's always running all tests (you cannot run then "only retry failed tests")

src/utils/telemetry/report-error.mjs
```diff
@@ -0,0 +1,52 @@
...
+ * @returns {Promise<void>}
+ */
+export const reportError = async function (error, config = {}) {
+  const globalConfig = await getGlobalConfig()
+
+  if (isTelemetryDisabled(globalConfig)) {
+    return
+  }
+
+  const user = globalConfig.get(`users.${globalConfig.get('userId')}`)
```
danez, line 26: We should handle logged out errors

src/utils/telemetry/report-error.mjs
```diff
@@ -0,0 +1,52 @@
...
+    data: {
+      message: error.message,
+      name: error.name,
+      stack: error.stack,
+      cause: error.cause,
+      severity: config.severity,
+      user: {
+        id: user.id,
+        name: user.name,
+        email: user.email,
```
danez, line 39: IMHO I think this is not okay to track (name, email). Wouldn't userid be enough?

src/utils/telemetry/report-error.mjs
lukasholzer: `user` is undefined if the user is logged out

src/utils/telemetry/report-error.mjs
lukasholzer: talked with legal it would be ok to report it but I've dropped it as it's not super necessary.

functions/error-reporting.ts
```diff
@@ -0,0 +1,43 @@
...
+      message,
+      name,
+      nodejsVersion,
+      osName,
+      severity = 'error',
+      stack,
+      user,
+    } = JSON.parse(body)
+    Bugsnag.notify({ name, message, stack, cause }, (event) => {
+      event.setUser(user.id, user.email, user.name)
```
danez, line 27: I was unsure if we can do this with all params being undefined, but it seems to be okay according to their types: https://github.com/bugsnag/bugsnag-js/blob/2cfdefb93805673871a676ecbe08155675d046fd/packages/core/types/event.d.ts#L42

danez (approved):

functions/error-reporting.ts
lukasholzer: yep they all accept undefined ;)

PR #5642, opened 2023-04-19
fix: correctly reload edge functions on change

danez:

#### Summary

The problem was that when the handling of internal and user functions was split, the detection for added/deleted function broke. All edge functions were always removed on change, because when detecting internal functions, the user functions were marked as deleted, and vice versa.

Ref https://github.com/netlify/pillar-support/issues/379

### DIFF
--- a/tests/integration/utils/dev-server.cjs
+++ b/tests/integration/utils/dev-server.cjs
@@ -119,11 +119,12 @@ const startDevServer = async (options, expectFailure) => {
   // eslint-disable-next-line fp/no-loops
   for (let attempt = 1; attempt <= maxAttempts; attempt++) {
     try {
-      const { timeout, ...server } = await startServer({ ...options, expectFailure })
-      if (timeout) {
-        throw new Error(`Timed out starting dev server.\nServer Output:\n${server.output}`)
+      // do not use destruction, as we use getters which otherwise would be evaluated here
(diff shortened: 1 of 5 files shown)


tests/integration/utils/dev-server.cjs
```diff
@@ -119,11 +119,12 @@ const startDevServer = async (options, expectFailure) => {
...
-      const { timeout, ...server } = await startServer({ ...options, expectFailure })
-      if (timeout) {
-        throw new Error(`Timed out starting dev server.\nServer Output:\n${server.output}`)
+      // do not use destruction, as we use getters which otherwise would be evaluated here
+      const devServer = await startServer({ ...options, expectFailure })
+      if (devServer.timeout) {
+        throw new Error(`Timed out starting dev server.\nServer Output:\n${devServer.output}`)
       }
-      return server
+      return devServer
```
danez, line 127: All the changes here werw done because we were restructuring the response of `startServer` which was cloning the object, which was evaluating the getters in the response instantly.
So it wasn't working as expected when calling `devServer.output` in the test, as it would never update, because it was evaluated as soon as the server was started.

khendrikse (approved):
Thanks for the information in the comments!

danez:
I adapted the test and removed the HTTP calls, as they weren't necessary to reproduce.

khendrikse (approved):

PR #5643, opened 2023-04-19
feat: report differences for new framework detection

lukasholzer:

#### Summary

Reports differences in the new build settings detection compared to the old framework detection to bugsnag.

---

      passes our tests.

### DIFF
--- a/functions/error-reporting.ts
+++ b/functions/error-reporting.ts
@@ -37,7 +37,10 @@ export const handler: Handler = async ({ body }) => {
       event.setUser(user.id, user.email, user.name)
       event.severity = severity
       event.device = {
-        osName: `${osName} - node.js v${nodejsVersion}`,
+        osName,
+        runtimeVersions: {
+          node: nodejsVersion,
+        },
       }
     })
   } catch (error) {
--- a/functions/error-reporting.ts
+++ b/functions/error-reporting.ts
@@ -16,6 +16,7 @@ export const handler: Handler = async ({ body }) => {
       cause,
       cliVersion,
       message,
+      metadata = {},
       name,
       nodejsVersion,
       osName,
@@ -24,6 +25,15 @@ export const handler: Handler = async ({ body }) => {
       user,
     } = JSON.parse(body)
     Bugsnag.notify({ name, message, stack, cause }, (event) => {
+      event.app = {
+        releaseStage: 'production',
+        version: cliVersion,
+        type: 'netlify-cli',
+      }
+      // eslint-disable-next-line fp/no-loops
+      for (const [section, values] of Object.entries(metadata)) {
+        event.addMetadata(section, values as Record<string, any>)
+      }
       event.setUser(user.id, user.email, user.name)
       event.severity = severity
       event.device = {
--- a/functions/error-reporting.ts
+++ b/functions/error-reporting.ts
@@ -37,11 +37,7 @@ export const handler: Handler = async ({ body }) => {
       event.setUser(user.id, user.email, user.name)
       event.severity = severity
       event.device = {
-        osName,
-        runtimeVersions: {
-          nodejs: nodejsVersion,
-          cli: cliVersion,
-        },
+        osName: `${osName} - node.js v${nodejsVersion}`,
       }
     })
   } catch (error) {
--- a/src/utils/detect-server-settings.mjs
+++ b/src/utils/detect-server-settings.mjs
@@ -205,7 +205,7 @@ const detectSettings = async (projectDir) => {
  */
 const detectChangesInNewSettings = (frameworkSettings, newSettings, metadata) => {
   /** @type {string[]} */
-  const message = []
+  const message = ['']
   const [setting] = newSettings
 
   if (frameworkSettings?.framework !== setting?.framework) {
--- a/src/utils/detect-server-settings.mjs
+++ b/src/utils/detect-server-settings.mjs
@@ -4,6 +4,9 @@ import { EOL } from 'os'
 import path from 'path'
 import process from 'process'
 
+import { Project } from '@netlify/build-info'
+// eslint-disable-next-line import/extensions, n/no-missing-import
+import { NodeFS } from '@netlify/build-info/node'
 import { getFramework, listFrameworks } from '@netlify/framework-info'
 import fuzzy from 'fuzzy'
 import getPort from 'get-port'
@@ -12,6 +15,7 @@ import isPlainObject from 'is-plain-obj'
 import { NETLIFYDEVWARN, chalk, log } from './command-helpers.mjs'
 import { acquirePort } from './dev.mjs'
 import { getInternalFunctionsDir } from './functions/functions.mjs'
+import { reportError } from './telemetry/report-error.mjs'
 
 const formatProperty = (str) => chalk.magenta(`'${str}'`)
 const formatValue = (str) => chalk.green(`'${str}'`)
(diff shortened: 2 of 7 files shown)


functions/error-reporting.ts
```diff
@@ -24,6 +25,15 @@ export const handler: Handler = async ({ body }) => {
       user,
     } = JSON.parse(body)
     Bugsnag.notify({ name, message, stack, cause }, (event) => {
+      event.app = {
+        releaseStage: 'production',
+        version: cliVersion,
```
danez, line 30: we are reporting it twice now, should we remove it from device?

functions/error-reporting.ts
```diff
@@ -24,6 +25,15 @@ export const handler: Handler = async ({ body }) => {
...
+        type: 'netlify-cli',
```
danez, line 31: type should be set to `node`

danez, line 31: Instead of setting this manually should we instead use the config in `Bugsnag.start({})`
https://docs.bugsnag.com/platforms/javascript/configuration-options/#appversion

danez (changes requested):

functions/error-reporting.ts
lukasholzer: Does not need to be you can give any name to it (default is `node`)

danez: but why `netlify-cli` what's the meaning of setting this?

functions/error-reporting.ts
lukasholzer: This is required to have it here: 
CleanShot 2023-04-19 at 16 57 31


But I will remove the runtime version does not seemed to get picked up

functions/error-reporting.ts
lukasholzer: we cannot set the CLI version on bugsnag start as this is a long-running function where multiple different CLI versions are reporting to

lukasholzer: CleanShot 2023-04-19 at 17 14 03@2x

build-info can report to bugsnag as well, so we can differentiate from where this was done. 

So we can report to the same project from multiple sub parts:

that's what they outline here: 
CleanShot 2023-04-19 at 17 15 01@2x

functions/error-reporting.ts
danez: The node version should apparently be added automatically, according to the docs.

src/utils/telemetry/report-error.mjs
```diff
@@ -30,15 +31,17 @@ export const reportError = async function (error, config = {}) {
       user: {
         id: globalConfig.get('userId'),
       },
+      metadata: config.metadata,
       osName: process.platform,
```
danez, line 35:
```suggestion
      osName: `${os.platform()} (${os.arch()})`,
```

from https://github.com/bugsnag/bugsnag-js/blob/f220383cb175cc779d5e6f7d015a8f71946f5a28/packages/plugin-node-device/device.js#L9

danez (approved):
I can also add the stuff I mentioned in a separate PR.

src/utils/telemetry/report-error.mjs
```diff
@@ -30,15 +31,17 @@ export const reportError = async function (error, config = {}) {
...
       cliVersion,
       nodejsVersion,
     },
   })
 
-  // spawn detached child process to handle send
-  execa(process.execPath, [join(dirPath, 'request.mjs'), options], {
+  // spawn detached child process to handle send and wait for the http request to finish
+  // otherwise it can get canceled
+  await execa(process.execPath, [join(dirPath, 'request.mjs'), options], {
```
jobala, line 43: Will this block until the http request finishes?

src/utils/telemetry/report-error.mjs
jobala: Thoughts on adding a hyphen to this `${os.platform()}-(${os.arch()})`?

src/utils/detect-server-settings.mjs
```diff
@@ -300,6 +370,19 @@ const detectServerSettings = async (devConfig, options, projectDir) => {
...
+
+    // just report differences in the settings
+    detectChangesInNewSettings(frameworkSettings, newSettings || [], {
+      ...metadata,
+      settings: {
+        projectDir,
+        devConfig,
+        options,
+        old: frameworkSettings,
+        settings: newSettings,
```
jobala, line 383: Thoughts on calling settings `new` here as well like you did in `detectChangesInNewSettings`? 

```
...
old: frameworkSettings
new: newSettings
...
```

src/utils/detect-server-settings.mjs
```diff
@@ -182,6 +186,71 @@ const getSettingsFromFramework = (framework) => {
...
+  return await project.getBuildSettings()
+}
+
+/**
+ *
+ * @param {import('./types.js').BaseServerSettings | undefined} frameworkSettings
+ * @param {import('@netlify/build-info').Settings[]} newSettings
+ * @param {Record<string, Record<string, any>>} [metadata]
+ */
+const detectChangesInNewSettings = (frameworkSettings, newSettings, metadata) => {
```
jobala, line 206: What do you think of mapping through they keys of `frameworkSettings` and check if their values differ with those of `newSettings` instead of manual `ifs`?

jobala (commented):
Looks good in general, I just have a few questions and suggestions.

jobala (changes requested):
Only requesting changes because the PR is approved and I would like to learn more about the questions asked before this goes in.

src/utils/telemetry/report-error.mjs
lukasholzer: yea it needs to because if it's called in the error reporting the process is otherwise exiting immediately which leads to the function crashing as the TLS handshake could not be finished:

https://app.netlify.com/sites/cli/functions/error-reporting
CleanShot 2023-04-20 at 09 27 56

src/utils/detect-server-settings.mjs
lukasholzer: it's just debugging info in bugsnag

src/utils/detect-server-settings.mjs
lukasholzer: not possible as keys are named differently:

```
if (frameworkSettings?.command !== setting?.devCommand) {
```

src/utils/detect-server-settings.mjs
jobala: Cool, although I think the old/new contrast makes it easy to understand the intent

src/utils/telemetry/report-error.mjs
jobala: Interesting, was this happening before removing the `unref` method call?

src/utils/detect-server-settings.mjs
jobala: Since its only one key and the other keys names are consistent between frameworkSettings & settings does it make sense to rename the key then map? ie `frameworkSettings.devCommand = frameworkSettings.command`

src/utils/detect-server-settings.mjs
lukasholzer: it's just for me ^^

src/utils/detect-server-settings.mjs
lukasholzer: that's not even worth the effort (this is JUST internal reporting for me for some days then it will be completely removed)

src/utils/telemetry/report-error.mjs
lukasholzer: yep exactly that's the reason why I've changed it

jobala (approved):

PR #5650, opened 2023-04-23
fix: follow symlinks w/ `netlify deploy`

### LINKED ISSUE #1809: Deploy does not follow symlinks
**Describe the bug**

`netlify deploy` does not follow symlinks out of the publish directory and misses files.

**To Reproduce**

```
cd out/
ln -s ../index.html out/
cd ..
netlify deploy
```

**Configuration**
```
[build]
base = "."
publish = "out/"
command = "wasm-pack build --target no-modules"
```

**Expected behavior**

`index.html` should be included in the deploy. It isn't.

ajschmidt8:
#### Summary

Fixes #1809

There's a discrepancy between deploys that happen via Netlify's auto-deploys and the `netlify deploy` CLI command. Auto-deploys follow symlinks while the `netlify deploy` CLI command does not.

This PR resolves that discrepancy by ensuring the `netlify deploy` command _does_ follow symlinks.

As mentioned by @tinfoil-knight in [this comment](https://github.com/netlify/cli/issues/1809#issuecomment-1058373415), the [fast-glob](https://www.npmjs.com/package/fast-glob) library follows symlinks by default.

Therefore, I've replaced the [folder-walker](https://www.npmjs.com/package/folder-walker) dependency that was previously used with `fast-glob`.

The APIs and return values of these libraries are different, so I've adjusted the code accordingly.

Further, I've replaced the `streams` implementations in `src/utils/deploy/hash-files.mjs` with `async`/`await` promises (and also updated `src/utils/deploy/hash-fns.mjs` for consistency).

This seemed like the preferred approach based on some of the `TODO` comments in `src/utils/deploy/hasher-segments.mjs` (which has now been deleted).

This switch also enabled the removal of the following dependencies:

- `pump`
- `from2-array`
- `flush-write-stream`
- `parallel-transform`

Finally, tests were added/updated according to these changes.

---

      passes our tests.


src/commands/deploy/deploy.mjs
```diff
@@ -178,31 +178,6 @@ const validateFolders = async ({ deployFolder, functionsFolder }) => {
   return { deployFolderStat, functionsFolderStat }
 }
 
-const getDeployFilesFilter = ({ deployFolder, site }) => {
```
ajschmidt8, line 181: Unlike `folder-walker`, `fast-glob` does not accept a file filter function.

Therefore this logic was adapted for the `fast-glob` file queries that exist in `src/utils/deploy/hash-files.mjs`.

src/lib/edge-functions/deploy.mjs
```diff
@@ -8,17 +8,6 @@ import { EDGE_FUNCTIONS_FOLDER, PUBLIC_URL_PATH } from './consts.mjs'
 
 const distPath = getPathInProject([EDGE_FUNCTIONS_FOLDER])
 
-export const deployFileNormalizer = (rootDir, file) => {
```
ajschmidt8, line 11: This logic was moved to the `normalizePath` function in `src/utils/deploy/util.mjs`.

ajschmidt8 (commented):
Leaving some notes for additional context.

Skn0tt:
Hi @ajschmidt8! I appreciate the PR. Properly supporting symlinks for `netlify deploy` is something we should fix! The PR is pretty big, so it'll take me a while to review it properly - bear with me :)

Skn0tt:
Looks like there's a linting error.

From a first cursory look, it's hard for me to differentiate the changes that fix the issue from refactorings. This makes it really hard for me to think about unwanted side-effects of this change. Would it be possible for you to trim this PR down to the minimal change that fixes the bug? I can also open a PR for that instead.

ajschmidt8:
> Looks like there's a linting error.

I'll resolve the linting errors and merge conflicts soon.

> From a first cursory look, it's hard for me to differentiate the changes that fix the issue from refactorings. This makes it really hard for me to think about unwanted side-effects of this change. Would it be possible for you to trim this PR down to the minimal change that fixes the bug? I can also open a PR for that instead.
> 
> Again, i'm really happy you opened the PR, I'm just thinking that splitting things up into a fix PR and a refactoring PR would make it easier to review for me, and easier to merge with confidence.

I started down that route. Unfortunately most of these changes are intertwined. Here's why:

`folder-walker` only supports streams. `fast-glob` has APIs for streams, promises, and synchronous operations.

Since both libraries have `stream` support, you might think that they'd be swappable.

Unfortunately that's not the case since `fast-glob` requires two queries (vs. `folder-walker`'s single query) to mimic the `getDeployFilesFilter` function that was used by `folder-walker`.

It seemed wasteful to devise a way to merge/concatenate these two independent `fast-glob` streams when the processing could be handled more succinctly using Promises. It would've introduced some throwaway work.

I realize there are a lot of changes. I'm happy to elaborate on any of them as needed.

I would recommend working backward and starting with reviewing the tests. If the test logic looks sound, then you can be assured that the functions are working as expected.

Ultimately the majority of my changes were only made to `hashFiles` (`hashFns` was updated for consistency, but it's functionality didn't change). Any other file changes were only made because the `hashFiles` function interacted with them in some way.

During my development and debugging process, it was useful to set up a small demo site with symlinks to test out my changes.

I used [`npm link`](https://docs.npmjs.com/cli/v9/commands/npm-link) to switch back and forth between the `main` branch and my development branch and run `netlify deploy` on my demo site on each branch.

On each branch, I also added the statements below before the `return` statement in `hashFiles` to ensure that its return values were consistent with my expectations:

```js
console.log("files:", files)
console.log("filesShaMap:", filesShaMap)
process.exit(0) # exit early to prevent having to wait for a full deploy to occur
```

ajschmidt8:
> I can also open a PR for that instead.

I'm not opposed to you opening a separate PR.

But I'd prefer not to spend too much more time on these changes myself (aside from minor PR requests) based on the time it took to get these changes working so far.

src/utils/deploy/hasher-segments.mjs
ajschmidt8, line 1: all of the functionality from these functions is replaced by the few lines of code in the `mapper` functions in `src/utils/deploy/hash-fns.mjs` and `src/utils/deploy/hash-files.mjs`

Skn0tt:
Hey @ajschmidt8! Please excuse me for not replying earlier. We've discussed this quite a bit internally, and while we agree that ideally the behaviour around Symlinks in CLI/NF Build should be the same, we are wary of other customers relying on this "buggy" behaviour. So we'd like to take any changes slow, and this might take a while.

I'll go ahead and close this PR for now - we will come back to solving this in the future, but we'll drive the implementation so you don't have to do any more work on this.

If you rely on this fix, I think you should be able to use your forked version of the CLI, right?

ajschmidt8:
> Hey @ajschmidt8! Please excuse me for not replying earlier. We've discussed this quite a bit internally, and while we agree that ideally the behaviour around Symlinks in CLI/NF Build should be the same, we are wary of other customers relying on this "buggy" behaviour. So we'd like to take any changes slow, and this might take a while.
> 
> I'll go ahead and close this PR for now - we will come back to solving this in the future, but we'll drive the implementation so you don't have to do any more work on this.
> 
> If you rely on this fix, I think you should be able to use your forked version of the CLI, right?

Appreciate the follow-up.

I do hope you guys address this quickly.

I'd prefer to abandon my fork ASAP in favor of an upstream fix.

Keep me posted if there's anything I can do to help with this effort.

CandiedCode:
>Please excuse me for not replying earlier. We've discussed this quite a bit internally, and while we agree that ideally the behaviour around Symlinks in CLI/NF Build should be the same, we are wary of other customers relying on this "buggy" behaviour. So we'd like to take any changes slow, and this might take a while.

@ajschmidt8 the way around this is to provide some type of flag that a user can set to follow a symlink.  This is causing me issues right now as well, and for such an [old issue](https://github.com/netlify/cli/issues/1809) has there been no update on providing this support today?

PR #5681, opened 2023-05-05
fix: set Host header on edge functions in HTTPS

eduardoboucas:
#### Summary

Sets the `Host` header to the passthrough URL in edge functions, so that `req.url` contains a valid URL when running the CLI on HTTPS.

### DIFF
--- a/tests/integration/200.command.dev.test.cjs
+++ b/tests/integration/200.command.dev.test.cjs
@@ -214,6 +214,10 @@ export const handler = async function () {
           path: 'public/index.html',
           content: 'index',
         })
+        .withContentFile({
+          path: 'public/origin.html',
+          content: 'origin',
+        })
         .withRedirectsFile({
           redirects: [{ from: `/api/*`, to: `/.netlify/functions/:splat`, status: '200' }],
         })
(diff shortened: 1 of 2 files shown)


tests/integration/200.command.dev.test.cjs
```diff
@@ -226,15 +230,21 @@ export const handler = async function () {
         })
         .withEdgeFunction({
           handler: async (req, { next }) => {
-            if (!req.url.includes('?ef=true')) {
-              return
+            if (req.url.includes('?ef=true')) {
+              // eslint-disable-next-line n/callback-return
```
lukasholzer, line 234: PLS, can we get rid of this lint rule?

tests/integration/200.command.dev.test.cjs
```diff
@@ -226,15 +230,21 @@ export const handler = async function () {
         })
         .withEdgeFunction({
           handler: async (req, { next }) => {
-            if (!req.url.includes('?ef=true')) {
-              return
+            if (req.url.includes('?ef=true')) {
```
lukasholzer, line 233: What if the `ef` query param is not the first one? is this possible?
I would go with:

```typescript
switch(new URL(req.url).searchParams.get('ef')) {
  case 'true':
    const res = await next()
    const text = await res.text()
    return new Response(text.toUpperCase(), res)
  case 'fetch':
    const url = new URL('/origin', req.url)
    const res = await fetch(url)
    const text = await res.text()
    return new Response(text.toUpperCase(), res)
}
```

tests/integration/200.command.dev.test.cjs
```diff
@@ -248,6 +258,7 @@ export const handler = async function () {
         const options = { https: { rejectUnauthorized: false } }
         t.is(await got(`https://localhost:${port}`, options).text(), 'index')
         t.is(await got(`https://localhost:${port}?ef=true`, options).text(), 'INDEX')
+        t.is(await got(`https://localhost:${port}?ef=fetch`, options).text(), 'ORIGIN')
```
lukasholzer, line 261: What about?
```typescript
        t.is(await got(`https://localhost:${port}?some=other&ef=fetch`, options).text(), 'ORIGIN')
```

lukasholzer (approved):
Left a comment not sure if this is the case, but it might be worth addressing. Other than that looks good.

PR #5684, opened 2023-05-05
chore: tell renovate to pin all dependencies

danez:
#### Summary

Tell renovate to pin all dependencies for the cli and the docs site.

this is a preparation for a potential switch of the package manager, but even if not, it makes sense to do anyway.

Ref: https://github.com/netlify/pod-compute/issues/498


renovate.json
```diff
@@ -0,0 +1,15 @@
...
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["github>netlify/renovate-config:esm"],
+
+  "ignorePresets": [":prHourlyLimit2"],
+  "semanticCommits": "enabled",
+  "dependencyDashboard": true,
+
+  "packageRules": [
+    {
+      "matchFiles": ["package.json", "site/package.json"],
```
lukasholzer, line 11: should we include the testing fixtures here as well?

danez: I wasn't sure if there we might use ranges on purpose, but probably not.

danez: Done, it is now active for all npm dependencies

PR #5692, opened 2023-05-10
chore: use pnpm for 11ty test fixture

lukasholzer:

#### Summary

This should fix our CI flakiness from https://github.com/netlify/pod-compute/issues/498. 
We now have one test fixture with npm (hugo) and one with pnpm (11ty site).

This should give better coverage for running with different package managers.

---

      passes our tests.


lukasholzer:
It seems to fix it now: 
CleanShot 2023-05-10 at 11 50 04

.github/workflows/legacy-tests.yml
```diff
@@ -50,6 +50,11 @@ jobs:
           cache: npm
           check-latest: true
         if: '${{!steps.release-check.outputs.IS_RELEASE}}'
+      - name: Install PNPM
+        run: |
+          corepack enable
```
danez, line 55: corepack is only available as of 14.19, if i read the docs right, but we test with 14.18. Why does this work?

danez (approved):

.github/workflows/legacy-tests.yml
lukasholzer: It got backported ;)

lukasholzer: But only pnpm version 7 works with node 14

PR #5702, opened 2023-05-12
fix: remove dependency `is-plain-obj`

danez:
seems unnecessary to check if it is a plain object. As long as the keys we expect are there it should be fine.


lukasholzer (approved):
Yay!

PR #5725, opened 2023-05-15
feat: add support for streaming functions

eduardoboucas:
#### Summary

Add support for streaming functions in Netlify Dev.

Part of https://github.com/netlify/pod-compute/issues/466.


tests/integration/400.command.dev.test.cjs
```diff
@@ -412,3 +413,69 @@ test('should handle multipart form data when redirecting', async (t) => {
...
+  await withSiteBuilder('site-with-streaming-function', async (builder) => {
+    builder
+      .withPackageJson({ packageJson: { dependencies: { '@netlify/functions': 'latest' } } })
+      .withCommand({ command: ['npm', 'install'] })
+      .withContentFile({
+        content: `
+          const { stream } = require("@netlify/functions");
+
+          class TimerSource {
+            #input;
```
Skn0tt, line 427: TIL that private class fields have been supported since Node v12!

tests/integration/400.command.dev.test.cjs
```diff
@@ -412,3 +413,69 @@ test('should handle multipart form data when redirecting', async (t) => {
...
+
+    await withDevServer({ cwd: builder.directory }, async (server) => {
+      const chunks = []
+      const response = got.stream(`${server.url}/.netlify/functions/streamer`)
+
+      response.on('data', (chunk) => {
+        chunks.push(chunk.toString())
+      })
+
+      await pause(500)
```
Skn0tt, line 476: let's add an assertion on the arrival times of the chunks, to ensure that local dev isn't buffering the response. i've been bitten by that before.

Skn0tt (approved):
looks good! left a suggestion on what to assert on, but we can follow on that if we want.

tests/integration/400.command.dev.test.cjs
eduardoboucas: Done in https://github.com/netlify/cli/pull/5725/commits/4bc7c1562fe2a2dfac263b4b45a80af853ac3137.

Skn0tt (approved):

PR #5743, opened 2023-05-17
feat: refactor getSite to base-command

sarahetter:

#### Summary

Closes https://github.com/netlify/pod-compute/issues/503

We have inconsistent usage of the data from `api.getSite` in the cli, and a lot of code repetition. The main things I did were to add a commander.js `preAction` hook which handles errors that were repeated in files. (`.hook('preAction', requiresSiteInfo)` )The data from `api.getSite()` was already in the `command.netlify` object, but we were fetching it in a number of commands to get error handling, as build-info doesn't error if the site doesn't exist. 

I also moved some logic from the `deploy` command to `base-command` as we were using a property for two different purposes, (name and id) to get rid of that duplicate call.

There remain a few instances of extra `api.getSite()` calls throughout the codebase, but I left them because the error check for site happens after some important logging, or in cases where the data is required after some change/user input/property.

---

      passes our tests.

Screenshot 2023-05-17 at 3 18 48 PM

### DIFF
--- a/src/utils/get-site.mjs
+++ b/src/utils/get-site.mjs
@@ -1,0 +1,9 @@
+export const getSiteByName = async (api, siteName) => {
+  try {
+    const sites = await api.listSites({ name: siteName, filter: 'all' })
+    const siteFoundByName = sites.find((filteredSite) => filteredSite.name === siteName)
+    return siteFoundByName
+  } catch (error) {
+    return { error }
+  }
+}
--- a/src/utils/get-site.mjs
+++ b/src/utils/get-site.mjs
@@ -1,4 +1,4 @@
-import { error } from '../../utils/command-helpers.mjs'
+import { error } from './command-helpers.mjs'
 
 export const getSiteByName = async (api, siteName) => {
   try {
--- a/src/utils/get-site.mjs
+++ b/src/utils/get-site.mjs
@@ -4,6 +4,11 @@ export const getSiteByName = async (api, siteName) => {
   try {
     const sites = await api.listSites({ name: siteName, filter: 'all' })
     const siteFoundByName = sites.find((filteredSite) => filteredSite.name === siteName)
+
+    if (!siteFoundByName) {
+      throw Error
+    }
+
     return siteFoundByName
   } catch {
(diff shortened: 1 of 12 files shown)


src/utils/hooks/requires-site-info.mjs
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * A preAction hook that errors out if siteInfo is an empty object
+ * @param {*} command
+ */
+const requiresSiteInfo = async (command) => {
+  const { api, site } = command.netlify
+  const siteId = site.id
+  if (!siteId) {
+    warn('Did you run `netlify link` yet?')
+    error(`You don't appear to be in a folder that is linked to a site`)
```
eduardoboucas, line 11: nit: I always found it counterintuitive to have a function call actually terminating the process. I think having a `return` would make it clearer that the function call actually ends the flow.

```suggestion
    return error(`You don't appear to be in a folder that is linked to a site`)
```

src/utils/get-site.mjs
```diff
@@ -0,0 +1,11 @@
+import { error } from './command-helpers.mjs'
+
+export const getSiteByName = async (api, siteName) => {
+  try {
+    const sites = await api.listSites({ name: siteName, filter: 'all' })
+    const siteFoundByName = sites.find((filteredSite) => filteredSite.name === siteName)
+    return siteFoundByName
+  } catch {
+    error('Site not found. Please rerun "netlify link"')
```
eduardoboucas, line 9: If we don't find the site because the name doesn't match, then `siteFoundByName` will be `undefined` and this message won't be shown. Is that intended?

eduardoboucas (dismissed):
Left a couple of comments, but LGTM!

You might want to rename the PR to `feat:` before merging, as `chore` won't trigger a release.

src/utils/get-site.mjs
sarahetter: Ah good call! I've added a throw Error if siteFoundByName is undefined

eduardoboucas (approved):

PR #5752, opened 2023-05-25
feat: add framework, build system, and package manager to telemetry for every command

sarahetter:

#### Summary

Closes https://github.com/netlify/pod-dev-foundations/issues/504

Adds the framework, build system and package manager to telemetry for further investigation into cli usage. Only successful runs of a command will be added to telemetry (as telemetry is only sent on base-command's `onEnd` function, which does not run if the process is ended with `process.exit(1)`) so we aren't capturing that information, but this is out of scope for this PR. 

---

      passes our tests.

Screenshot 2023-05-25 at 2 03 29 PM


Skn0tt (dismissed):
Looks good! If there's an easy way for us to add a unit test for this, that'd be great. If it's not obvious, then let's open an issue to track this.

src/commands/base-command.mjs
```diff
@@ -456,6 +459,21 @@ export default class BaseCommand extends Command {
...
+    const frameworks = await project.detectFrameworks()
+
+    const frameworkIDs = frameworks?.map((framework) => framework.id)
+
+    if (frameworkIDs?.length !== 0) {
+      this.setAnalyticsPayload({ frameworks: frameworkIDs })
+    }
+
+    const buildSystem = await project.detectBuildSystem()
+    const packageManager = await project.detectPackageManager()
```
lukasholzer, line 474: That's not needed as `detectFrameworks()` depends on it and run those first, so it just retrieves the value from the `project`.

you can do directly `{ packageManager: project.packageManager?.name, buildSystem: project.buildSystems.map(({id}) => id) }`

lukasholzer: CleanShot 2023-05-26 at 08 53 46

lukasholzer (approved):
LGTM!

Issue #5724, opened 2023-05-15
Fix npm deprecations.


danez:
There are 5 deprecation notices in `netlify-cli@15.0.3`:

```
$ npm install netlify-cli@15.0.3 -g
npm WARN deprecated source-map-url@0.4.1: See https://github.com/lydell/source-map-url#deprecated
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated string-similarity@4.0.4: Package no longer supported. Contact Support at https://www.npmjs.com/support for more info.
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated source-map-resolve@0.5.3: See https://github.com/lydell/source-map-resolve#deprecated
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
```

MikeMcC399:
No change in the deprecations in `netlify-cli@16.3.0` compared to `15.8.0`.

kisaragi-hiu:
Of the four remaining deprecations, *resolve-url@0.2.1*, *source-map-url@0.4.1*, *urix@0.1.0* are dependencies of *source-map-resolve@0.5.3*.

The dependency path to *source-map-resolve@0.5.3* looks like:

*source-map-resolve@0.5.3* ← snapdragon@0.8.2 ← (see note) ← micromatch@3.1.10 ← readdirp@2.2.1 ← **copy-template-dir@1.4.0**

No package/version in this list is depended on by anything else. This can be checked eg. [on npmgraph](https://npmgraph.js.org/?q=netlify-cli): the 4 deprecated packages are around the top right corner. Click each package to trace its dependants or dependencies.

I have filed https://github.com/yoshuawuyts/copy-template-dir/pull/17 to fix this by having it update to readdirp@3.

---

Note: snapdragon@0.8.2 is depended on by these packages, which all begin from micromatch@3.1.10:

- braces@2.3.2 ← micromatch@3.1.10
- micromatch@3.1.10
- extglob@2.0.4 ← micromatch@3.1.10
- expand-brackets@2.1.4 ← extglob@2.0.4 ← micromatch@3.1.10
- nanomatch@1.2.13 ← micromatch@3.1.100

waynemaranga:
Almost five months later this issue is still live. I absolutely cannot install `netlify-cli` from npm:
[...]

PR #5759, opened 2023-06-01
fix(deps): replace string-similarity with fastest-levenshtein

kisaragi-hiu:
#### Summary

Replace string-similarity with [fastest-levenshtein](https://github.com/ka-weihe/fastest-levenshtein) as the former is deprecated, resulting in another warning during installation.

While the algorithm is different (Levenshtein distance instead of Dice's coefficient), the difference shouldn't be big enough to matter, as this is only used for suggestion after a typo.

Partially addresses #5724.

---

      passes our tests.


Skn0tt:
@kisaragi-hiu thanks for the PR! The changes look good to me, but i'd like to make sure that changing the algo to Levenshtein doesn't have unintended consequences.

@lukasholzer I see you originally implemented the string similarity 2 years ago. Did you pick a sorensen-dice-based algorithm on purpose? Do you think we should stick with that, and maybe use a library like https://www.npmjs.com/package/dice-coefficient instead? Or is it fine to switch to Levenshtein?

lukasholzer (commented):
@Skn0tt to be honest I cannot remember why I've decided to go along with the `string-similarity` package. 

But using the fastest Levenshtein distance is probably a good decision as it measures the necessary edits so in this case it looks pretty well suited. 

@kisaragi-hiu thanks for putting the effort into this! Looks good to me :)

Do you mind maybe adding a test to it?

Skn0tt (approved):
looks great! thanks so much for this PR @kisaragi-hiu!

lukasholzer (approved):
LGTM!

PR #5782, opened 2023-06-07
fix: expose original URL search to function redirects

eduardoboucas:
#### Summary

Closes https://github.com/netlify/pillar-support/issues/447.

### DIFF
--- a/tests/integration/commands/dev/redirects.test.ts
+++ b/tests/integration/commands/dev/redirects.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, test } from 'vitest'
+
+import { FixtureTestContext, setupFixtureTests } from '../../utils/fixture.js'
+import got from '../../utils/got.cjs'
+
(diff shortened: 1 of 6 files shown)


tests/integration/commands/dev/redirects.test.ts
```diff
@@ -0,0 +1,22 @@
...
+        throwHttpErrors: false,
+        retry: { limit: 0 },
+      })
+
+      expect(response.statusCode).toBe(200)
+
+      const result = JSON.parse(response.body)
+      expect(result.queryStringParameters).not.toHaveProperty('param1')
+      expect(result.queryStringParameters).toHaveProperty('param2', 'world')
+      expect(result.queryStringParameters).toHaveProperty('other', '1')
```
danez, line 19: @paulo is this the expected behavior? Just to double-check. So `param1` even though it is specified in the redirect is not forwarded.

danez (approved):

PR #5792, opened 2023-06-12
feat: add support for feature flags

sarahetter:

#### Summary

Fixes https://github.com/netlify/pod-dev-foundations/issues/519

Along with [bitballoon](https://github.com/netlify/bitballoon/pull/15495) and [open-api](https://github.com/netlify/open-api/pull/466) changes, this completes the addition of feature flag support in the cli. 

Currently feature_flags are enabled on staging, so you can test this PR by running the cli with `NETLIFY_API_URL` set to staging, and using the `isFeatureFlagEnabled` function in any command.

By passing `feature_flags: 'cli'` to the `api.getSite` call, we are requesting a list of feature flags that are prefixed with `cli`. No other feature flags will be available.

---

      passes our tests.

Screenshot 2023-06-12 at 4 01 51 PM


src/utils/feature-flags.mjs
```diff
@@ -0,0 +1,21 @@
...
+ * @param {string} flagName
+ *
+ * @returns {boolean}
+ */
+export const isFeatureFlagEnabled = async (api, siteId, flagName) => {
+  const siteInfo = await api.getSite({ site_id: siteId, feature_flags: 'cli' })
+  if (siteInfo.feature_flags && siteInfo.feature_flags[flagName] !== false) {
+    return true
+  }
+  return false
```
eduardoboucas, line 20: nit: You can simplify this a bit if you want, but feel free to take or leave.

```suggestion
  return Boolean(siteInfo?.feature_flags[flagName] !== false)
```

src/utils/feature-flags.mjs
```diff
@@ -0,0 +1,21 @@
...
+ * the absense of a flag should also evaluate to the flag being enabled.
+ * Instead, we return that the feature flag is enabled if it isn't
+ * specifically set to false in the response
+ * @param {import('netlify').NetlifyAPI} api
+ * @param {*} siteInfo
+ * @param {string} flagName
+ *
+ * @returns {boolean}
+ */
+export const isFeatureFlagEnabled = async (api, siteId, flagName) => {
```
eduardoboucas, line 15: nit: Really down to personal preference, but I think having the flag name as the first parameter would make the code a bit easier to scan, since the name of the flag is probably the most important parameter and the other two are basically supporting metadata.

This is also in line with other feature flag interfaces across our codebases.

eduardoboucas (approved):
Excited to see this landing! Left two comments, but nothing blocking.

eduardoboucas:
> By passing `feature_flags: 'cli'` to the `api.getSite` call, we are requesting a list of feature flags that are prefixed with `cli`.

This isn't currently part of this PR. Will you add it, or does that live somewhere else?

sarahetter:
@eduardoboucas Thanks for the review!

> > By passing `feature_flags: 'cli'` to the `api.getSite` call, we are requesting a list of feature flags that are prefixed with `cli`.
> 
> This isn't currently part of this PR. Will you add it, or does that live somewhere else?

Ah I was referring to what I'm doing on line 16 in feature-flags.mjs - `const siteInfo = await api.getSite({ site_id: siteId, feature_flags: 'cli' })`

I thought the `feature_flags: 'cli'` bit was requesting `cli` prefixed flags from bitballoon? Or am I mistaken?

eduardoboucas:
> Ah I was referring to what I'm doing on line 16 in feature-flags.mjs - `const siteInfo = await api.getSite({ site_id: siteId, feature_flags: 'cli' })`
> 
> I thought the `feature_flags: 'cli'` bit was requesting `cli` prefixed flags from bitballoon? Or am I mistaken?

Doh, you're right! I missed this somehow, sorry.

danez:
> By passing feature_flags: 'cli' to the api.getSite call, we are requesting a list of feature flags that are prefixed with cli

prefixed? in like `cli-my-ff`? Couldn't we use the CLI environment from devcycle? Because if they need to be prefixed they won't help in enabling feature flags that are used in edge-bundler order zip-it-and-ship-it.

why are we doing separate `getSite` call and do not use the ones we already do on every CLI invocation? Wasn't the idea to smuggle feature_flags into `getSite` so we do not have a separate API call?

Sorry I might be a little late with these questions.

sarahetter:
> prefixed? in like cli-my-ff? 

These decisions were made in the bitballoon PR, I'm not sure about the answer. 

> why are we doing a separate getSite call and do not use the ones we already do on every CLI invocation?

I thought about doing this, but it meant making changes in @netlify/config (allowing passing of a `featureFlagPrefix` but felt that it meant adding complication to another repo, as well as more difficult testing. I can switch to this if we think it's the better solution

eduardoboucas:
> why are we doing a separate `getSite` call and do not use the ones we already do on every CLI invocation? Wasn't the idea to smuggle feature_flags into `getSite` so we do not have a separate API call?

Yeah, that's a good point and it was indeed the intention.

@sarahetter if we could make that happen without too much complexity, I think it's worth a try.

eduardoboucas (approved):
LGTM! I think it would be nicer to handle the `@netlify/config` update via Renovate and then rebase, leaving this PR with just the functional changes. But not a big deal.

eduardoboucas (approved):
🚀

Issue #5790, opened 2023-06-12
15.4.1 breaks URL parameters with duplicate names


oyvindmidtbo:
### Describe the bug

After updating from 15.4.0 to 15.4.1, it's no longer possible to use multiple URL parameters with the same name.

### Steps to reproduce

```
import { APIGatewayProxyEvent, Handler } from 'aws-lambda';

const handler: Handler = async (event: APIGatewayProxyEvent) => {
  console.log(event.queryStringParameters);

  return {
    statusCode: 200,
    body: JSON.stringify({ message: 'Hello World' }),
  };
};

export { handler };
```

1. Install netlify-cli 15.4.0 and run a function, for example `localhost:8888/nf-functions/myfunction?param1=foo&param1=bar`
2. In the handler, I console.log `event.queryStringParameters`: `{ param1: 'foo, bar' }`
3. Install netlify-cli 15.4.1 and run the same function with the same parameters.
4. The new console.log: `{ param1: 'foo' }`

### Environment

System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Max
    Memory: 3.57 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
  npmPackages:
    netlify-cli: ^15.4.1 => 15.4.1

hemm1:
I work on the same team and just to avoid initial confusion: 

This behavior only appears when we are accessing the function through our own redirect. We have the following redirect in the `netlify.toml` file:

```
[[redirects]]
  from = "/nf-functions/*"
  to = "/.netlify/functions/:splat"
  status = 200
```

When we access the function without the redirect (`localhost:8888/.netlify/functions/myfunction?param1=foo&param1=bar`), we get access to the parameters in the same way as before.

Also, as you might expect, this does not affect our functions in production in any way. They still work as before.

eduardoboucas:
Thanks for the detailed report. This should be fixed in 15.5.1 — could you give it a try?

If you're still seeing the issue, please let me know and I'll reopen the ticket.

PR #5797, opened 2023-06-13
fix: preserve duplicate query parameters in functions

eduardoboucas:
#### Summary

Fixes a regression introduced in https://github.com/netlify/cli/pull/5782, and reported in https://github.com/netlify/cli/issues/5790, which swallows duplicate query parameters from functions.

Closes https://github.com/netlify/cli/issues/5790.


danez:
Could you check the `200.command.dev › should handle query params in redirects` test?

eduardoboucas:
> chore: fix test

I fixed the test in https://github.com/netlify/cli/pull/5797/commits/669d593fd63e08402f16ebefacbc09870e689a32. I verified in production that the expected behaviour is for both parameters to be received, so the test was wrong.

danez (approved):

PR #5804, opened 2023-06-14
feat: add support for new functions API

danez:

#### Summary

Fixes https://github.com/netlify/pod-dev-foundations/issues/533

This implements a worker pool for function requests. Every lambda function is executed in a separate worker and the worker afterward restarted. This is done because we cannot otherwise invalidate the `import()` cache.

Typescript wasn't tested for v2 functions, because zisi does not yet support TS with NFT parser.

Missing

- [x] console.log in functions only sometimes works
- [x] debugging does not work in v2 functions with workers. funny enough it works in v1 functions with workers.

I also discovered this bug in tinypool: https://github.com/tinylibs/tinypool/pull/60


tests/integration/commands/dev/v2-api.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      const context = await response.json()
+      expect(context.requestId).toEqual(response.headers.get('x-nf-request-id'))
+      expect(context.site.url).toEqual(`http://localhost:${devServer.port}`)
+      expect(context.server.region).toEqual('dev')
+      expect(context.ip).toEqual('::1')
+      expect(context.geo.city).toEqual('San Francisco')
+
+      expect(context).toEqual('')
+
+      expect(context.cookies).toEqual({ foo: 'bar' })
```
Skn0tt, line 58: i'm unsure why this is failing

tests/integration/commands/dev/v2-api.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      // this is also not set for edge-functions, TODO!
+      expect(context.account.id).toEqual('???')
```
Skn0tt, line 60: i'd say we do this in a follow-up, both for edge functions and functions

eduardoboucas (commented):
I need to do another pass, but here's a thought I wanted to share sooner rather than later.

src/commands/dev/dev.mjs
```diff
@@ -161,7 +161,10 @@ const dev = async (options, command) => {
     },
   })
 
+  const geoConfig = { geolocationMode: options.geo, geoCountry: options.country, offline: options.offline, state }
```
eduardoboucas, line 164: Why are `offline` and `state` part of an object called `geoConfig`?

src/lib/functions/runtimes/js/builders/zisi.mjs
```diff
@@ -109,26 +111,33 @@ const getTargetDirectory = async ({ errorExit }) => {
...
   const functionsConfig = netlifyConfigToZisiConfig({ config, projectRoot })
 
   const packageJson = await readPackageUp(func.mainFile)
   const hasTypeModule = packageJson && packageJson.packageJson.type === 'module'
 
-  // We must use esbuild for certain file extensions.
-  const mustTranspile = ['.mjs', '.ts'].includes(path.extname(func.mainFile))
-  const mustUseEsbuild = hasTypeModule || mustTranspile
+  if (metadata.runtimeAPIVersion === 2) {
+    // For typescript we use NFT otherwise we leave the file untouched with the `none` bundler
```
eduardoboucas, line 121: nit

```suggestion
    // For TypeScript we use NFT, otherwise we leave the file untouched with the `none` bundler
```

src/lib/functions/runtimes/js/index.mjs
```diff
@@ -48,12 +51,49 @@ export const getBuildFunction = async ({ config, directory, errorExit, func, pro
...
 
+const workerURL = new URL('worker.mjs', import.meta.url)
+
 export const invokeFunction = async ({ context, event, func, timeout }) => {
+  const workerData = {
+    clientContext: JSON.stringify(context),
+    event,
+    // If a function builder has defined a `buildPath` property, we use it.
+    // Otherwise, we'll invoke the function's main file.
+    // Because we use import() we have to use file:// URLs for windows
```
eduardoboucas, line 66: nit

```suggestion
    // Because we use import() we have to use file:// URLs for Windows.
```

src/lib/functions/runtimes/js/index.mjs
```diff
@@ -48,12 +51,49 @@ export const getBuildFunction = async ({ config, directory, errorExit, func, pro
...
+      } else {
+        resolve(result)
+      }
+    })
+
+    worker.on('error', reject)
+  })
+}
+
+// unused right now
```
eduardoboucas, line 95: My understanding is that we'll hook this up before merging, so that the new method is only used for functions using the v2 API? Or at the very least use a feature flag. I think we should try to avoid moving every single invocation to the new method at once.

src/lib/functions/runtimes/js/worker.mjs
```diff
@@ -0,0 +1,52 @@
+import { createServer } from 'net'
+import { isMainThread, workerData, parentPort } from 'worker_threads'
+
+import lambdaLocal from '@skn0tt/lambda-local'
+import { isStream } from 'is-stream'
+import sourceMapSupport from 'source-map-support'
+
+import { SECONDS_TO_MILLISECONDS } from './constants.mjs'
+
+if (isMainThread) {
```
eduardoboucas, line 10: When would this happen?

src/lib/functions/runtimes/js/worker.mjs
```diff
@@ -0,0 +1,52 @@
...
+  throw new Error(`Do not import "${import.meta.url}" in the main thread.`)
+}
+
+sourceMapSupport.install()
+
+lambdaLocal.getLogger().level = 'warn'
+
+const { clientContext, event, lambdaPath, timeout } = workerData
+
+// eslint-disable-next-line import/no-dynamic-require
```
eduardoboucas, line 20: I think we can disable this globally.

tests/integration/100.command.dev.test.cjs
```diff
@@ -136,9 +136,11 @@ test('background function clientContext,identity should be null', async (t) => {
       })
       .buildAsync()
 
-    await withDevServer({ cwd: builder.directory }, async ({ outputBuffer, url }) => {
+    await withDevServer({ cwd: builder.directory }, async ({ outputBuffer, url, waitForLogMatching }) => {
```
eduardoboucas, line 139: What is this about?

eduardoboucas (commented):
Can you confirm that [debugging functions](https://cli.netlify.com/vscode/) still works, even when they run in a worker?

src/commands/dev/dev.mjs
Skn0tt: Because these fields are all required by the `getGeoLocation` function:

https://github.com/netlify/cli/blob/5ccdc7115427356f91643120949cce58c6c8513a/src/lib/geo-location.mjs#L47

src/lib/functions/runtimes/js/worker.mjs
Skn0tt: Never on purpose, but maybe during development somebody would import it. And then this error makes it obvious that that's not how it's intended to be used. The Node.js docs also always use this: https://nodejs.org/api/worker_threads.html

But we can also remove it and add a big header comment instead, your call.

tests/integration/100.command.dev.test.cjs
Skn0tt: Daniel added this in https://github.com/netlify/cli/commit/6c1a660990887af4c562bcdfba7328de9fd48e98, but i'm also unsure what it's about. Presumably it fixes some race condition?

src/lib/functions/runtimes/js/index.mjs
Skn0tt: done in https://github.com/netlify/cli/pull/5804/commits/71cb338c5347b86d9f97bea49ed45b128b9171e6

src/lib/functions/runtimes/js/worker.mjs
Skn0tt: done in https://github.com/netlify/cli/pull/5804/commits/8273e8a0553aa23b5003db6abe3228848ebe3354

src/lib/functions/runtimes/js/worker.mjs
eduardoboucas: Nah, that makes sense. Just wanted to double-check.

src/commands/dev/dev.mjs
eduardoboucas: Sure, but these are generic properties that are present in most commands, not specific to geolocation. I think we should be grouping parameters based on what they mean for the function being called, not on how they will be used by a child function many levels deep.

If `startFunctionsServer` (or any other methods downstream) start using `offline` or `state` for other purposes than geo-location, would we move them out of this object?

Not a big deal, just think our future selves might find this confusing.

src/lib/functions/runtimes/js/index.mjs
```diff
@@ -48,12 +51,52 @@ export const getBuildFunction = async ({ config, directory, errorExit, func, pro
...
+
 export const invokeFunction = async ({ context, event, func, timeout }) => {
+  if (func.runtimeAPIVersion !== 2) {
+    return await invokeFunctionDirectly({ context, event, func, timeout })
+  }
+
+  const workerData = {
+    clientContext: JSON.stringify(context),
+    event,
+    // If a function builder has defined a `buildPath` property, we use it.
```
eduardoboucas, line 68: nit: Since we're calling this `entryFilename` in zip-it-and-ship-it, might be nice to use the same naming here to make the relationship clear. Since this is the full path and not just the filename, perhaps `entryPath` or `entryFilePath` would be good?

eduardoboucas (approved):

Skn0tt:
> Can you confirm that [debugging functions](https://cli.netlify.com/vscode/) still works, even when they run in a worker?

Debugging currently doesn't work, and it looks like that's because we don't emit a sourcemap for the .mjs file. I'll see what we can do.

Skn0tt:
Okay, so the underlying problem for debugging is that we're using the `none` bundler for the V2 API, which doesn't have support for sourcemaps: 

https://github.com/netlify/cli/blob/71cb338c5347b86d9f97bea49ed45b128b9171e6/src/lib/functions/runtimes/js/builders/zisi.mjs#L124

I tried switching it to `esbuild` or `nft`, but that didn't help. I'd say we ship this without debugging support for now, and address this in a follow-up. Not sure how hard it is to add sourcemap support to the `none` bundler 🤔

src/commands/dev/dev.mjs
Skn0tt: Yeah, fair point. Refactored in https://github.com/netlify/cli/pull/5804/commits/adbb13ec68945fb524cb1573213307873ac78527.

eduardoboucas:
@Skn0tt Just to confirm: debugging will continue to work the same for existing functions, just not for functions using the V2 syntax, because existing functions will continue to use the existing invocation mechanism and only the V2 ones will use worker threads. Is that right?

Skn0tt:
In my test, debugging continued to work for non-V2 functions, since ZISI emits a sourcemap for them. It won't work for V2 syntax, because we're not emitting a sourcemap. I was able to place a breakpoint in the compiled `.netlify/functions/serve` code and debugging worked just fine, so I don't think the Worker Threads are a problem.

Skn0tt:
I'm unsure what's up with the timouting tests. Will take another look next week.

Skn0tt:
The timeouting tests were fixed in https://github.com/netlify/cli/pull/5804/commits/1df39f719d599d3386db21e2d004f9e2663f3fb6.

Skn0tt:
https://github.com/netlify/cli/pull/5804/commits/efacbd6509f7d9d800fd7fb8a2bf26f63d1da1c0 adds support for debugging by using `esbuild` as the bundler. This emits sourcemaps, so debugging works again. It also decreases performance, so I'd like to come up with a simpler solution if possible.

The reason we need sourcemaps in the first place is that ZISI copies over the source files into the `.netlify/function-serve` directory, and then imports them locally:

https://github.com/netlify/zip-it-and-ship-it/blob/6901bb20c94963aaf23266889ede0cf03627526b/src/runtimes/node/utils/entry_file.ts#L34
Generates this `.neltify/functions-serve/ping/__netlify_entry_point.mjs`:

```js
import * as func from "./src/ping.mjs";
import { getLambdaHandler } from "./___netlify-bootstrap.mjs";
export const handler = getLambdaHandler(func);
```

What if we add a "local dev mode" where the generated file is this instead:

```js
import * as func from ".../../netlify/functions/ping.mjs";
import { getLambdaHandler } from "./___netlify-bootstrap.mjs";
export const handler = getLambdaHandler(func);
```

That way, we wouldn't need any source map generation step.

src/lib/functions/runtimes/js/builders/zisi.mjs
```diff
@@ -109,26 +125,44 @@ const getTargetDirectory = async ({ errorExit }) => {
...
+  const featureFlags = {}
+
+  if (metadata.runtimeAPIVersion === 2) {
+    // For TypeScript we use NFT, otherwise we leave the file untouched with the `none` bundler
+    const isTypescript = ['.ts', '.mts', '.cts'].includes(path.extname(func.mainFile))
+
+    if (isTypescript) {
+      functionsConfig['*'].nodeBundler = 'nft'
+    } else {
+      // using esbuild is less performant than `none`, but it emits sourcemaps and thus
```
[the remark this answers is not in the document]
Skn0tt: > We can only use none if we know upfront the list of files that need to be included, which isn't the case here?

Yes, because we want to copy over the source files. But if we didn't need to copy them in the first place (see https://github.com/netlify/cli/pull/5804#issuecomment-1607359866), then we wouldn't need to know the list.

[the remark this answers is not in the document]
eduardoboucas: Ah, got you!

tests/integration/commands/dev/v2-api.test.ts
Skn0tt: Yes, I did! It worked correctly all the time. The problem was that `context.cookies` is an instance of `CookieStore`, which JSON-stringified to `{}` no matter what's in it. I worked around it here: https://github.com/netlify/cli/pull/5804/files#diff-29d0cf31447374101566bed229b51b2a665f6863e41d5b250cf534d3726b9cacR5

tests/integration/commands/dev/v2-api.test.ts
Skn0tt: captured in https://github.com/netlify/cli/issues/5825

PR #5817, opened 2023-06-21
chore: modify github workflow names for easier navigation in actions

sarahetter:
#### Summary

Updates the names of github workflows so that when trying to find a workflow in Actions  (as seen in screenshot below), you can distinguish which "Tests" you're looking for. 

Screenshot 2023-06-21 at 9 36 17 AM


eduardoboucas (approved):
LGTM! We might need to change the branch protection settings in the repo so that PRs are not blocked waiting for workflows with the old names to finish, but we can do that after we merge this.

PR #5821, opened 2023-06-25
refactor: replace dynamic imports with static imports

### LINKED ISSUE #5697: Refactor all dynamic imports of ESM dependencies to static imports
Now as all files are already ECMAScript modules, we can go ahead and migrate all dynamic imports to static ones.

CleanShot 2023-05-12 at 11 04 39

instead of doing:

```typescript
const bundler = await import('@netlify/edge-bundler')
```

we can go ahead and do a static one

```typescript
import bundler from '@netlify/edge-bundler'

...
```

### LINKED ISSUE #5698: Refactor all tests to use ESM and vitest
We still have some tests that use ava and commonjs, we should refactor them to use vitest and ESM instead.

There is also a new fixture runner for tests which we can use, which can make the tests also faster. But can also be done afterwards.
```[tasklist]
### Tasks
- [ ] https://github.com/netlify/cli/pull/5892
- [ ] https://github.com/netlify/cli/pull/5919
- [ ] 200.command.dev.test.cjs
- [ ] 210.command.deploy.test.cjs
- [ ] 300.command.dev.test.cjs
- [ ] 500.command.dev.test.cjs
- [ ] 600.framework-detection.test.cjs
- [ ] 610.command.env.test.cjs
- [ ] 640.command.recipes.test.cjs
```

hereje:
- Replace await import('path-to-module') with static imports at the beginnunounoing of modules

#### Summary

Fixes #5697  

Refactor  dynamic imports with static imports on ESM (*.mjs) files
---

      passes our tests.


src/utils/detect-server-settings.mjs
```diff
@@ -259,9 +261,6 @@ const detectFrameworkSettings = async ({ projectDir }) => {
   }
 
   if (frameworks.length > 1) {
-    // performance optimization, load inquirer on demand
```
Skn0tt, line 262: if this was here as a performance optimisation, maybe it should stay as a `dynamic import`. @lukasholzer what do you think?

src/lib/functions/runtimes/js/builders/zisi.mjs
```diff
@@ -30,9 +31,6 @@ const buildFunction = async ({ cache, config, directory, func, hasTypeModule, pr
   }
   const functionDirectory = path.dirname(func.mainFile)
 
-  // performance
```
Skn0tt, line 33: this mentions performance as the reason we're late-importing it, so we should double-check wether we want to make it static.

src/lib/functions/registry.mjs
```diff
@@ -175,9 +176,6 @@ export class FunctionsRegistry {
   // This function is here so we can mock it in tests
   // eslint-disable-next-line class-methods-use-this
   async listFunctions(...args) {
-    // Performance optimization: load '@netlify/zip-it-and-ship-it' on demand.
-    const { listFunctions } = await import('@netlify/zip-it-and-ship-it')
```
Skn0tt, line 179: same, mentions perf optimisation

Skn0tt (commented):
Looks generally good. Some imports were moved that mentioned performance reasons for having them as late-imports, we should try out if importing them statically makes a difference. @lukasholzer could you take a second look? You opened the original issue.

src/utils/detect-server-settings.mjs
lukasholzer: TBH I would benchmark it and test it locally if it makes a difference. If not than we can move forward with a static one otherwise I would leave it a dynamic one.

Skn0tt:
I ran a very naïve comparison, measuring the time between starting `bin/run.mjs`, and this line: https://github.com/netlify/cli/blob/456a5bb614496aaf64e1aa2ef7791df3a5f764e8/src/lib/functions/runtimes/js/builders/zisi.mjs#L33. I ran on `netlify-react-ui`, with `BROWSER=none ~/dev/netlify/cli/bin/run.mjs dev --offline`. This PR took 3.7s. When I make the imports dynamic, it's the very same. I'm not sure if my comparison is right :/

Skn0tt (approved):
Performance doesn't seem impacted, let's merge it!

Skn0tt:
Thank you so much for the contribution @hereje! And please excuse the long idle time.

PR #5824, opened 2023-06-27
feat: support setting `is_secret` on environment variables

jasonbarry:

#### Summary

---

      passes our tests.

### DIFF
--- a/src/commands/env/env-set.mjs
+++ b/src/commands/env/env-set.mjs
@@ -18,7 +18,7 @@ import {
  * @returns {Promise<boolean>}
  */
 const envSet = async (key, value, options, command) => {
-  const { context, scope } = options
+  const { context, scope, secret } = options
 
   const { api, cachedConfig, site } = command.netlify
   const siteId = site.id
@@ -33,7 +33,7 @@ const envSet = async (key, value, options, command) => {
 
   // Get current environment variables set in the UI
   if (siteInfo.use_envelope) {
-    finalEnv = await setInEnvelope({ api, siteInfo, key, value, context, scope })
+    finalEnv = await setInEnvelope({ api, siteInfo, key, value, context, scope, secret })
   } else if (context || scope) {
     error(
       `To specify a context or scope, please run ${chalk.yellow(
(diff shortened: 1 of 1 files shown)


docs/commands/env.md
```diff
@@ -175,15 +175,18 @@ netlify env:set
 - `http-proxy` (*string*) - Proxy server address to route requests through.
 - `http-proxy-certificate-filename` (*string*) - Certificate file to use when connecting using a proxy server
 - `scope` (*builds | functions | post-processing | runtime*) - Specify a scope (default: all scopes)
+- `secret` (*boolean*) - Indicate whether the environment variable value can be read again.
```
sean-roberts, line 178: Always worried to suggest you can't read the value because depending on if the dev is thinking about reading from a code standpoint vs a UI/API standpoint can cause confusion.

```suggestion
- `secret` (*boolean*) - Indicate whether this value should be treated as write-only when outside of code running on Netlify.
```

src/commands/env/env-set.mjs
```diff
@@ -88,15 +89,29 @@ const setInMongo = async ({ api, key, siteInfo, value }) => {
...
- * @returns {Promise<object>}
+ * @returns {Promise<object | boolean>}
  */
-const setInEnvelope = async ({ api, context, key, scope, siteInfo, value }) => {
+const setInEnvelope = async ({ api, context, key, scope, secret, siteInfo, value }) => {
   const accountId = siteInfo.account_slug
   const siteId = siteInfo.id
+
+  // secret values must specify deploy contexts. `all` or `dev` are not allowed
+  if (secret && value && (!context || context.includes('dev'))) {
```
sean-roberts, line 99: I'm not entirely sure it's a great experience to have a different pattern for dev context but I understand we're trying to be clear about the expectations of secrets. In the future, if devs have an issue with this nuance we can always allow them to set dev with the secret flag (like they can do on the UI) it just won't be an actual secret (also true for those set on the dev context in UI)

tests/integration/650.command.envelope.test.cjs
```diff
@@ -390,6 +395,164 @@ test('env:set --scope should set the scope of an existing env var without needin
...
+        true,
+      )
+
+      t.deepEqual(cliResponse, finalEnv)
+
+      const patchRequest = requests.find(
+        (request) => request.method === 'PATCH' && request.path === '/api/v1/accounts/test-account/env/EXISTING_VAR',
+      )
+
+      t.is(patchRequest.body.context, 'production')
```
sean-roberts, line 463: should we add a is_secret check?

src/commands/env/env-set.mjs
```diff
@@ -88,15 +89,29 @@ const setInMongo = async ({ api, key, siteInfo, value }) => {
...
   // fetch envelope env vars
   const envelopeVariables = await api.getEnvVars({ accountId, siteId })
   const contexts = context || ['all']
-  const scopes = scope || AVAILABLE_SCOPES
+  let scopes = scope || AVAILABLE_SCOPES
+
+  if (secret) {
+    // post_processing (aka post-processing) scope is not allowed with secrets
+    scopes = scopes.filter((sco) => !/post[-_]processing/.test(sco))
+  }
```
sean-roberts, line 114: We should probably error if the developer passed just `post-processing` right? Otherwise, it gets filtered out and looks empty and all of the empty fallbacks happen implicitly, right?

jasonbarry: Yep, if they explicitly pass in post-processing, it will be silently removed. Erroring out in this case sounds like a good solution 👍

sean-roberts (approved):
Code looks solid to me!

PR #5837, opened 2023-07-05
EF: Don't crash CLI when uncompressed request throws uncaught exception 

Skn0tt:

#### Summary

Fixes https://github.com/netlify/pod-dev-foundations/issues/536

---

      passes our tests.

### DIFF
--- a/src/utils/proxy.mjs
+++ b/src/utils/proxy.mjs
@@ -36,8 +36,28 @@ import { generateRequestID } from './request-id.mjs'
 import { createRewriter, onChanges } from './rules-proxy.mjs'
 import { signRedirect } from './sign-redirect.mjs'
 
-const decompress = util.promisify(zlib.gunzip)
+const gunzip = util.promisify(zlib.gunzip)
+const brotliDecompress = util.promisify(zlib.brotliDecompress)
+const deflate = util.promisify(zlib.deflate)
 const shouldGenerateETag = Symbol('Internal: response should generate ETag')
+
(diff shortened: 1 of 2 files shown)


src/utils/proxy.mjs
```diff
@@ -36,9 +36,29 @@ import { generateRequestID } from './request-id.mjs'
...
+const brotliDecompress = util.promisify(zlib.brotliDecompress)
+const deflate = util.promisify(zlib.deflate)
 const shouldGenerateETag = Symbol('Internal: response should generate ETag')
 
+/**
+ * @param {Buffer} body
+ * @param {string | undefined} contentEncoding
+ * @returns {Promise<Buffer>}
+ */
+const decompressResponseBody = async function (body, contentEncoding = '') {
```
eduardoboucas, line 49: Will this handle the case where `content-encoding` contains multiple values, like `deflate, gzip`?

eduardoboucas (approved):
Left just one question, but LGTM!

src/utils/proxy.mjs
Skn0tt: It won't, but I don't think that case exists for uncaught exceptions. This code path only occurs when `x-nf-uncaught-error` is set. The `content-encoding` is added by Deno automatically: https://deno.land/manual@v1.20.1/runtime/http_server_apis#automatic-body-compression
I don't think they ever combine compressions, so not supporting multiple values should be fine.

PR #5838, opened 2023-07-05
fix: ef error page should have 500 status code

Skn0tt:
Currently, we're returning a 200 status code for the error page. This PR changes that to a 500.


eduardoboucas (approved):
🚀

PR #5850, opened 2023-07-07
fix: dont block watching when netlify.toml doesnt exist

Skn0tt:
In Local Dev, when there is no `netlify.toml` (e.g. because in-source-config is being used), `this.#configPath` was undefined. This caused the config file watcher to block indefinitely, so we never got to register the second watcher. This breaks automatic file reloading.

This PR fixes that by skipping the config watcher, if there is no config file. Ideally, we'd have both watchers and chokidar didn't block, so we'd pick up when the config file is added. Let's do that in a follow-up though.


lukasholzer (approved):
LGTM

PR #5880, opened 2023-07-19
fix: fixes a bug inside the error reporting

lukasholzer:

#### Summary

Fixes an error when a non-error class was reported that it was failing to report the issue.

---

      passes our tests.

### DIFF
--- a/src/utils/telemetry/report-error.mjs
+++ b/src/utils/telemetry/report-error.mjs
@@ -25,15 +25,19 @@ export const reportError = async function (error, config = {}) {
     return
   }
 
+  // convert a NotifiableError to an error class
+  // eslint-disable-next-line unicorn/no-nested-ternary
+  const err = error instanceof Error ? error : typeof error === 'string' ? new Error(error) : error
+
   const globalConfig = await getGlobalConfig()
 
   const options = JSON.stringify({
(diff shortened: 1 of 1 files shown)


src/utils/telemetry/report-error.mjs
```diff
@@ -25,15 +25,19 @@ export const reportError = async function (error, config = {}) {
     return
   }
 
+  // convert a NotifiableError to an error class
+  // eslint-disable-next-line unicorn/no-nested-ternary
+  const err = error instanceof Error ? error : typeof error === 'string' ? new Error(error) : error
```
eduardoboucas, line 30: The nested ternary does feel odd to me, but deferring to you since it's not a massively complex statement.

eduardoboucas (approved):

src/utils/telemetry/report-error.mjs
lukasholzer: with the right prettier settings it would look pretty good IMHO.

```typescript  
const err = error instanceof Error 
  ? error 
  : typeof error === 'string' 
    ? new Error(error) 
    : error
```

whereas with if and assignments it's harder to read IMHO but seems like personal preference

PR #5886, opened 2023-07-22
refactor: tests/integration/310.command.dev.exec.test.cjs

hereje:
- replace ava with vitest
- move and rename test file

#### Summary

Convert test file to ESM and replace ava with vitest

Related to #5698 

---

      passes our tests.


lukasholzer:
@hereje are they ready to review?

hereje:
> @hereje are they ready to review?

Yes, I'll change it to Ready to review.

tests/integration/commands/dev/dev.exec.test.mjs
```diff
@@ -0,0 +1,21 @@
+import process from 'process'
+
+import { test } from 'vitest'
+
+import callCli from '../../utils/call-cli.cjs'
+import { withSiteBuilder } from '../../utils/site-builder.cjs'
+
+test('should pass .env variables to exec command', async (t) => {
+  await withSiteBuilder('site-env-file', async (builder) => {
+    builder.withEnvFile({ env: { VITEST_TEST: 'ENV_VAR' } })
```
lukasholzer, line 10: I think you don't have to change the env var, this should be independent of vitest (could be even `MY_SUPER_SECRET`).

I would even try to use a variable that is not used by vitest to avoid having to leak some variable that is getting set by the test runner. I think `VITEST_TEST` is set by the test runner.


```suggestion
    builder.withEnvFile({ env: { TEST: 'ENV_VAR' } })
```

sarahetter (approved):
Thank you!

PR #5909, opened 2023-07-31
feat: add `context.deploy` to edge functions context

eduardoboucas:
#### Summary

Updates the edge functions bootstrap to the latest version and exposes `context.deploy` to Netlify Dev.


src/lib/edge-functions/proxy.mjs
```diff
@@ -115,6 +115,7 @@ export const initializeProxy = async ({
 
     // Setting header with geolocation and site info.
     req.headers[headers.Geo] = Buffer.from(JSON.stringify(geoLocation)).toString('base64')
+    req.headers[headers.DeployID] = '0'
```
eduardoboucas, line 118: This is the value we use for the `DEPLOY_ID` environment variable in Netlify Dev.

Skn0tt (approved):

PR #5922, opened 2023-08-03
feat: error when V2 functions are used with Node <18

eduardoboucas:
#### Summary

The v2 API for functions requires Node 18 or greater, since it uses built-ins like `Request` and `Response`. If people try to use a function with an older version of Node, it will fail and they might be confused as to why.

In this PR we check for that scenario specifically and show a friendly error message. It also tidies up the logic for reloading functions. Finally, it ensures that if there was a build error in the function, it will be shown in the Netlify-branded error page when the function is accessed.

Screenshot 2023-08-04 at 00 03 05

### DIFF
--- a/src/lib/functions/netlify-function.mjs
+++ b/src/lib/functions/netlify-function.mjs
@@ -1,8 +1,13 @@
 // @ts-check
+import { version as nodeVersion } from 'process'
+
 import CronParser from 'cron-parser'
+import semver from 'semver'
 
 import { error as errorExit } from '../../utils/command-helpers.mjs'
 import { BACKGROUND } from '../../utils/functions/get-functions.mjs'
+
+const V2_MIN_NODE_VERSION = '18.0.0'
 
 // Returns a new set with all elements of `setA` that don't exist in `setB`.
 const difference = (setA, setB) => new Set([...setA].filter((item) => !setB.has(item)))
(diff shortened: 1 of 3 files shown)


src/lib/functions/netlify-function.mjs
```diff
@@ -93,11 +103,22 @@ export default class NetlifyFunction {
...
       this.buildData = buildData
+      this.buildError = null
       this.srcFiles = srcFilesSet
       this.schedule = schedule || this.schedule
 
+      if (!this.isSupported()) {
+        throw new Error(
+          `Function requires Node.js version ${V2_MIN_NODE_VERSION} or above, but ${nodeVersion.slice(
+            1,
+          )} is installed. Refer to https://ntl.fyi/functions-node18 for information on how to update.`,
```
eduardoboucas, line 114: This will eventually point to a forums post where we explain why this requirement exists and how to update Node.js.

Skn0tt, line 114: Let's open an issue to make this link clickable. This would probably need changes here: https://github.com/netlify/cli/blob/9629588af0c30e4c07ed5c249c0d1ba5da5bb27c/src/lib/templates/function-error.html#L279-L280

Skn0tt (approved):

PR #5936, opened 2023-08-10
fix: always fetch immutable ef bootstrap version

### LINKED ISSUE #5923: netlify build fails to bundle edge function
### Describe the bug

Current version of the CLI uses `@netlify/build 29.17.3` which fails to bundle any edge function in production mode.

Error log:
```bash
C:\Users\maddsua\Desktop\New folder>ntl build
​
Netlify Build
────────────────────────────────────────────────────────────────
​
> Version
  @netlify/build 29.17.3
​
> Flags
  dry: false
  offline: false
​
> Current directory
  C:\Users\maddsua\Desktop\New folder
​
> Config file
  No config file was defined: using default values.
​
> Context
  production
​
Edge Functions bundling
────────────────────────────────────────────────────────────────
​
Packaging Edge Functions from netlify\edge-functions directory:
 - hello
Could not load configuration for edge function at 'C:\Users\maddsua\Desktop\New folder\netlify\edge-functions\hello.js'
error: Uncaught (in promise) SyntaxError: The requested module '../log/logger.ts' does not provide an export named 'deta
chedLogger' at https://edge.netlify.com/bootstrap/util/redirect.ts:3:10
const { Netlify } = await import(bootstrapURL)
                    ^
    at async file:///C:/Users/maddsua/AppData/Roaming/nodejs/node_modules/netlify-cli/node_modules/@netlify/edge-bundler
/deno/config.ts:4:21
​
(Edge Functions bundling completed in 2.7s)
​
Netlify Build Complete
────────────────────────────────────────────────────────────────
​
(Netlify Build completed in 2.7s)
```

Example edge function:

```javascript
//netlify/edge-functions/hello.js

export default () => new Response("Hello world");

export const config = { path: "/test" };

```

Edge functions seem work fine in `dev` mode, this only happens when trying to build for production.

This is just a reproduction example to keep things simple, the issue can be seen in other projects too.

### Steps to reproduce

1. Download and extract this example project (or just create a new one with an edge function, does not really matter)
2. Try building it with `netlify build`
3. Get an error on edge functions bundle stage

[example.zip](https://github.com/netlify/cli/files/12261438/example.zip)

### Configuration

A totally blank configuration, just a single example edge function in a project.

### Environment

  System:
    OS: Windows 10 10.0.19043
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 22.90 GB / 31.92 GB
  Binaries:
    Node: 18.12.1 - ~\AppData\Roaming\nodejs\node.EXE
    npm: 9.7.2 - ~\AppData\Roaming\nodejs\npm.CMD

Skn0tt:

#### Summary

Fixes #5923. 

We had an issue where Deno cached an old version of edge-functions-bootstrap locally. On the next run, Deno re-fetched some of the assets but not others. Since it assumes URLs to be immutable, this shouldn't have made a difference - but because we were using the `https://edge.netlify.com` endpoint to parse Edge Functions config, the contents changed. This broke some customers trying to deploy edge functions locally. I have a hunch that this only affected some versions of Deno and not others.

Anyways, this PR fixes the issue by injecting a deploy-scoped Bootstrap URL into netlify/build (and by extension, into netlify/edge-bundler). Deploy-scoped URLs are immutable, so they should fix this bug.

---

      passes our tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@bugsnag/js": "7.20.2",
         "@fastify/static": "6.10.2",
-        "@netlify/build": "29.19.0",
+        "@netlify/build": "29.20.0",
         "@netlify/build-info": "7.7.3",
         "@netlify/config": "20.8.0",
         "@netlify/edge-bundler": "8.17.1",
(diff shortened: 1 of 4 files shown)


eduardoboucas (commented):
Do you also need to supply the new property [here](https://github.com/netlify/cli/blob/main/src/commands/deploy/deploy.mjs#L440-L445)?

eduardoboucas:
> I'd say let's ship this PR, and ship the other addition in a follow-up.

As long as you're tracking it, sounds good to me!

eduardoboucas (approved):

PR #5939, opened 2023-08-11
fix: pass bootstrapUrl to runCoreSteps

Skn0tt:
Closes https://github.com/netlify/pod-dev-foundations/issues/574.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
       "dependencies": {
         "@bugsnag/js": "7.20.2",
         "@fastify/static": "6.10.2",
-        "@netlify/build": "29.20.0",
+        "@netlify/build": "29.20.2",
         "@netlify/build-info": "7.7.3",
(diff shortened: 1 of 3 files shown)


eduardoboucas:
I see that you're also updating esbuild as part of this, but the diff of the lock file makes me think we're now having both versions?

lukasholzer (approved):
LGTM!

Skn0tt:
> I see that you're also updating esbuild as part of this, but the diff of the lock file makes me think we're now having both versions?

Looks like it. Maybe we need to also update ZISI? Will take a look.

Skn0tt:
Done in https://github.com/netlify/cli/pull/5939/commits/b52fbe97fd7dc0cc9b0b611f70b612b053a2cbab. Good catch, Eduardo!

eduardoboucas (approved):

PR #5963, opened 2023-08-24
Feat: Don't retry function uploads on status 400 or 422

jenae-janzen:

#### Summary

Fixes CLI portion of https://github.com/netlify/functions-origin/issues/358

We don't need to retry function uploads when the status code is 400 or 422.

WIP: this change should be wrapped in a feature flag for now, which has not been implemented yet.

---


kitop:
Closing in favor of updated version: https://github.com/netlify/cli/pull/6361

PR #5964, opened 2023-08-25
feat: add support for `context.params` in edge functions

eduardoboucas:
#### Summary

Adds support for the new `context.params` property in edge functions.

Tests will fail until https://github.com/netlify/edge-bundler/pull/461 is released and the dependency is updated in the CLI.


sarahetter (approved):
Ty!

Issue #5945, opened 2023-08-12
Using local SSL for `netlify dev` results in the edge function not receiving the correct scheme, hostname, or port


overlordofmu:
### Describe the bug

When not using SSL locally with `netlify dev`, edge functions will see the hostname requested by the HTTP client (`curl` in these examples).

However, when using SSL locally with `netlify dev`, edge functions will see the hostname as `localhost` and the port number used will also be changed to the randomly assigned port that a `netlify dev` HTTP server is listening on (which is not the port the client makes the request to).

This should not occur. For local development, the edge function should correctly receive the URL requested by the client even when SSL is being used.

### Steps to reproduce

Reproduction steps:

1. Provision a local SSL certificate. (I used the Let’s Encrypt `certbot` utility to do so.)
2. Configure `netlify dev` to use that certificate for SSL:
You can define a local SSL configuration with `netlify dev` using the following syntax in `netlify.toml`:

```toml
[dev.https]
  certFile = "ssl/fullchain.pem"  ## these are the filenames used by Let’s Encrypt
  keyFile = "ssl/privkey.pem"     ## these are the filenames used by Let’s Encrypt
```
3. Add the hostname in question to the local `/etc/host` file like so (using `foo.example.com` here):

```text
127.0.0.1	localhost foo.example.com
```
4. Make an edge functions that logs the request URL. I used this:
```typescript
import type {Context, Config } from "https://edge.netlify.com";

export default (request: Request, context: Context) => {
  const requestURL = new URL(request.url);
  console.log("Logging requestURL:");
  console.log(requestURL);
  return context.next();
};

export const config: Config = {
  path: "/*",
  excludedPath: ["/*.css", "/*.js"],
  onError: "bypass"
};
```
5. Start the CLI dev instance with: `netlify dev`
6. Make a request to the site locally using: `curl https://foo.example.com:8888/` (note: the URL is using HTTPS)
* at this point the logging should show this (and it will if HTTP only is used):
```text
[request-logging] Logging requestURL:
[request-logging] URL {
  href: "https://foo.example.com:8888/",
  origin: "https://foo.example.com:8888",
  protocol: "https:",
  username: "",
  password: "",
  host: "foo.example.com:8888",
  hostname: "foo.example.com",
  port: "8888",
  pathname: "/",
  hash: "",
  search: ""
}
```
7. Note that, instead, it shows something like this:
```text
[request-logging] Logging requestURL:
[request-logging] URL {
  href: "http://localhost:49419/",
[...]

PR #5968, opened 2023-08-28
fix: show proper `req.url` in HTTPS mode

Skn0tt:

#### Summary

Fixes https://github.com/netlify/cli/issues/5945.

Together with https://github.com/netlify/edge-functions-bootstrap/pull/294, this PR fixes what `req.url` shows in HTTPS mode.

It passes the original hostname and protocol into the Edge Function via `X-Forwarded-Host` and `X-Forwarded-Proto`. https://github.com/netlify/edge-functions-bootstrap/pull/294 reads this and shows it in `req.url`.

Bootstrap uses this URL for passthrough requests as well. In order for the Deno runtime to accept the self-signed certificate, we pass it into edge-bundler. This also allows us to remove the secondary HTTP server for passthrough requests.

---

      passes our tests.


src/lib/edge-functions/bootstrap.mjs
```diff
@@ -1,5 +1,5 @@
 import { env } from 'process'
 
-const latestBootstrapURL = 'https://64c264287e9cbb0008621df3--edge.netlify.com/bootstrap/index-combined.ts'
```
Skn0tt, line 3: We need to merge the ef-bootstrap PR first, so we can update the URL here.

src/utils/proxy.mjs
```diff
@@ -742,19 +740,6 @@ export const startProxy = async function ({
 
   const eventQueue = [once(primaryServer, 'listening')]
 
-  // If we're running the main server on HTTPS, we need to start a secondary
-  // server on HTTP for receiving passthrough requests from edge functions.
-  // This lets us run the Deno server on HTTP and avoid the complications of
```
Skn0tt, line 747: With this PR, we're passing the "potentially untrusted certificate" into Deno via the `--certs` flag.

This functionality was originally implemented in https://github.com/denoland/deno/pull/3972, and it works by adding the cert as a trusted root certificate to the HTTP client:

https://github.com/denoland/deno/pull/3972/files#diff-5c6fb275a66a1b02f99597f6e9b36297274abd4b16926b9fb2d412f94ee0700eR52

From my perspective, this should be fine. Maybe Eduardo can elaborate on the complications they had in mind when they're back.

sarahetter (approved):
LGTM, but it'd be great to wait for @eduardoboucas 's feedback on the comment above ^

src/utils/proxy.mjs
eduardoboucas: Have you seen the PR that introduced that functionality? https://github.com/netlify/cli/pull/5409

eduardoboucas (changes requested):
Since you're essentially reverting https://github.com/netlify/cli/pull/5409, can you add some colour to why you think we won't reintroduce the problems people were having with untrusted certificates? There's [a whole Slack thread](https://netlify.slack.com/archives/C0359548J07/p1674171123274979?thread_ts=1674141360.666799&cid=C0359548J07) (internal to Netlify only) with a discussion on why running Deno on HTTP would be preferable.

I understand that the behaviour described in https://github.com/netlify/cli/issues/5945 doesn't feel ideal, but is it breaking anything? I'd like to better understand what exactly we're looking to fix before we risk breaking other things.

Skn0tt:
Wasn't aware of the Slack thread and the nuance around Deno requiring the CA cert. Agree that this isn't the right fix, in that case.

The thing that's broken is that `req.url` shows the wrong protocol and hostname. We can fix that using the `X-Forwarded-Proto` and `X-Forwarded-Host` headers, and that's probably the right way forward. Closing!

src/lib/edge-functions/proxy.mjs
```diff
@@ -167,28 +167,22 @@ export const initializeProxy = async ({
     }
 
     const featureFlags = ['edge_functions_bootstrap_failure_mode']
-    const forwardedHost = `localhost:${passthroughPort}`
 
     req[headersSymbol] = {
       [headers.FeatureFlags]: getFeatureFlagsHeader(featureFlags),
-      [headers.ForwardedHost]: forwardedHost,
+      [headers.ForwardedProtocol]: req.socket.encrypted ? 'https:' : 'http:',
```
eduardoboucas, line 173: Can we make [the caller method](https://github.com/netlify/cli/blob/58b40b3af29a015434b99d854a841f146e291125/src/utils/proxy.mjs#L666) pass `settings.https` rather than using `req.socket.encrypted` here? That way we have a unified detection mechanism for whether we're running on HTTPS.

tests/integration/200.command.dev.test.cjs
```diff
@@ -239,10 +239,15 @@ export const handler = async function () {
 
             if (req.url.includes('?ef=fetch')) {
               const url = new URL('/origin', req.url)
-              const res = await fetch(url)
-              const text = await res.text()
+              try {
+                await fetch(url, {})
+              } catch (error) {
+                return new Response(error)
```
eduardoboucas, line 245: nit: I think using an error code here might come in handy.

```suggestion
                return new Response(error, { status: 500 })
```

src/lib/edge-functions/headers.mjs
```diff
@@ -5,10 +5,13 @@ export const headers = {
   DeployID: 'x-nf-deploy-id',
   FeatureFlags: 'x-nf-feature-flags',
   ForwardedHost: 'x-forwarded-host',
+  ForwardedProtocol: 'x-forwarded-proto',
```
eduardoboucas, line 8: Is this one still needed?

Skn0tt: We're still using `x-forwarded-proto` in the code. `x-forwarded-host` could be removed, if we wanted - but I think it's also fine to leave it in, so it's easier to discover the headers that ef-bootstrap takes in the future 🤷

src/lib/edge-functions/proxy.mjs
Skn0tt: done in dfa73c9881fb44405d6fcf04e8ea738b0bf1c451

tests/integration/200.command.dev.test.cjs
Skn0tt: done in 3c9a98c0dd51b778f1cea572b15bee107f2fc6df

src/lib/edge-functions/bootstrap.mjs
```diff
@@ -1,5 +1,5 @@
 import { env } from 'process'
 
-const latestBootstrapURL = 'https://64e7783fce8cfe0008496c72--edge.netlify.com/bootstrap/index-combined.ts'
+const latestBootstrapURL = 'https://deploy-preview-294--edge.netlify.app/bootstrap/index-combined.ts'
```
eduardoboucas, line 3: Let's not forget to change this before merging.

eduardoboucas (changes requested):
LGTM! I haven't approved because we're still using a deploy preview for the bootstrap URL. Once that's changed, we're good to roll.

src/lib/edge-functions/bootstrap.mjs
Skn0tt: done in https://github.com/netlify/cli/pull/5968/commits/bb5f238fcd8ce2926679167ec1daf84844968490

eduardoboucas (approved):

PR #5969, opened 2023-08-28
Revert "chore(deps): update dependency verdaccio to v5.26.2"

ailadas:
Reverts netlify/cli#5891


lukasholzer:
@ailadas can you please add more description why this needs to be reverted?

PR #5970, opened 2023-08-28
feat: support `context.params` + `config.method` in functions + edge functions

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/578.

This PR does a bunch of related things:

- Adds support for `context.params` in local functions to CLI.
- Adds support for `config.methods` in edge functions.
- Adds support for `config.methods` in functions.

I opted for a single PR instead of multiple ones, so we have fewer merge conflicts.

We're still missing `methods` support for TOML-declared methods. I'll do that in a follow-up, once the `@netlify/config` change is released.


src/lib/edge-functions/registry.mjs
```diff
@@ -303,7 +303,7 @@ export class EdgeFunctionsRegistry {
   /**
    * @param {string} urlPath
```
eduardoboucas, line 304:
```suggestion
   * @param {string} urlPath
   * @param {array<string>} method
```

src/lib/functions/netlify-function.mjs
```diff
@@ -177,7 +182,7 @@ export default class NetlifyFunction {
       return false
     })
 
-    return isMatch
+    return match?.pattern
```
eduardoboucas, line 185: nit: Returning just the pattern feels a bit random and too tied to the way we're currently using this method. I'd say that it feels more natural to return the whole route and then use from it whatever you need. Totally not a big deal, though.

eduardoboucas (commented):
LGTM! Let's please add the JSDoc bit, but otherwise it's good to go.

src/lib/functions/netlify-function.mjs
Skn0tt: done in [0666c4b](https://github.com/netlify/cli/pull/5970/commits/0666c4b01b382937ad7f13d9d2ce13edc329abec)

eduardoboucas (approved):
🚀

PR #5985, opened 2023-09-07
added `deploy --trigger` to `COMMAND_WTIHOUT_WORKSPACE_OPTIONS`

BlenderDude:
#### Summary

Includes `netlify deploy --trigger` in the list of commands that don't require a workspace.

If you can, point me in the right direction for updating the tests necessary for this change and I can get it changed.

**3 babies watching the ice cube for me**
IMG_2623-min

### DIFF
--- a/src/commands/base-command.mjs
+++ b/src/commands/base-command.mjs
@@ -54,11 +54,39 @@ const HELP_INDENT_WIDTH = 2
 const HELP_SEPARATOR_WIDTH = 5
 
 /**
- * A list of commands where we don't have to perform the workspace selection at.
- * Those commands work with the system or are not writing any config files that need to be
- * workspace aware.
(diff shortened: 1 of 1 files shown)


sarahetter:
Hi @BlenderDude 

Thanks for the PR! I'm wondering what the use case is for this - in the case of a monorepo you may have many different sites in one repository, and the workspace options are there to specify which site you'd like to deploy with `netlify deploy --trigger`.

If you'd like to pre-select the site you're deploying, you can use the [`--filter` option](https://docs.netlify.com/cli/get-started/#work-with-monorepos), or `--cwd` to set the working directory.

sarahetter:
Closing due to non-response. Please open an issue if you believe there is still a bug here!

PR #5996, opened 2023-09-15
fix: pass feature flags from siteInfo for build

estephinson:

#### Summary

This should not be merged until after `@netlify/build@29.21.1` has been updated - as there is an underlying fault with how it was reading the paths in non-test conditions

---

      passes our tests.


sarahetter (approved):

src/lib/build.mjs
```diff
@@ -45,6 +45,8 @@ export const getBuildOptions = ({
   cwd,
   featureFlags: {
     ...edgeFunctionsFeatureFlags,
+    // eslint-disable-next-line unicorn/no-useless-fallback-in-spread
+    ...(cachedConfig.siteInfo.feature_flags ?? {}),
```
eduardoboucas, line 49: I'm curious: why are you adding the fallback and disabling the ESLint rule when the fallback is unnecessary?

```js
> {foo: "bar", ...null, ...undefined}
{ foo: 'bar' }
>
```

PR #5998, opened 2023-09-15
feat: return 404 on default URL when function has custom routes

eduardoboucas:
#### Summary

Follow-up to #5954. When a function defines custom routes, we should return 404 when trying to access the function on the `/.netlify/functions` default URL.

When this happens, we show a message in the console to help people understand why they're seeing a 404:

```
›   Warning: Function v2-js cannot be invoked on /.netlify/functions/v2-js, because the function has the following URL paths defined: /v2/:foo
```

Part of https://github.com/netlify/pod-dev-foundations/issues/578 and https://github.com/netlify/pod-dev-foundations/issues/585.


khendrikse (approved):

src/lib/functions/registry.mjs
```diff
@@ -123,6 +124,32 @@ export class FunctionsRegistry {
   }
 
   async getFunctionForURLPath(urlPath, method) {
+    const defaultURLMatch = urlPath.match(DEFAULT_URL_EXPRESSION)
```
Skn0tt, line 127: I believe this block should live in `NetlifyFunction#matchUrlPath`. That'd make the logic a little easier to reason about, as well.

It'd probably result in more regex matchings needed to be done, but I don't think that's a performance concern at the orders of magnitude that functions have in local dev.

src/lib/functions/registry.mjs
```diff
@@ -14,6 +14,7 @@ import { getPathInProject } from '../settings.mjs'
 import NetlifyFunction from './netlify-function.mjs'
 import runtimes from './runtimes/index.mjs'
 
+const DEFAULT_URL_EXPRESSION = /^\/.netlify\/(functions|builders)\/([^/]+).*/
```
Skn0tt, line 17: We should be refactoring this so that the regex is shared with `isFunction`

https://github.com/netlify/cli/blob/556d5456d750861ec1b00842ef88bcff6eef6c87/src/utils/proxy.mjs#L85-L91

Skn0tt (approved):
The functionality looks good. Noted two code-level things that I'd love to be refactored before merging this - fine to land in a refactoring as well, though.

Skn0tt:
Out-of-scope for this PR, but we should still address it before shipping this:

`isFunction` is used in a bunch of places to inform redirecting rules etc. We should check if that needs to take custom routes into account:

https://github.com/netlify/cli/blob/556d5456d750861ec1b00842ef88bcff6eef6c87/src/utils/proxy.mjs#L166-L168

https://github.com/netlify/cli/blob/556d5456d750861ec1b00842ef88bcff6eef6c87/src/utils/proxy.mjs#L214-L216

https://github.com/netlify/cli/blob/556d5456d750861ec1b00842ef88bcff6eef6c87/src/utils/proxy.mjs#L345

src/lib/functions/registry.mjs
Skn0tt: okay, scratch that - would make https://github.com/netlify/cli/blob/05c44a6d2c17ef80a860bce90feebe58c1f02198/src/lib/functions/registry.mjs#L137-L147 a lot harder to implement

PR #6010, opened 2023-09-18
fix: decompress body from edge functions errors

eduardoboucas:
#### Summary

Right now, if an edge function throws an unhandled exception, the CLI will crash with a cryptic error message:

```
SyntaxError: Unexpected token  in JSON at position 0
    at JSON.parse (<anonymous>)
    at formatEdgeFunctionError (file:///Users/eduardoboucas/Sites/netlify/cli/src/utils/proxy.mjs:65:12)
    at IncomingMessage.onEnd (file:///Users/eduardoboucas/Sites/netlify/cli/src/utils/proxy.mjs:548:31)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

This is because the edge function is sending a compressed response which the CLI needs to decompress. We do have some logic for checking whether the body needs to be decompressed, which we've added in https://github.com/netlify/cli/pull/5837, but we're reading the `accept-encoding` header from the incoming request, when we should be reading it from the response we get from the Deno server.

### DIFF
--- a/src/utils/proxy.mjs
+++ b/src/utils/proxy.mjs
@@ -544,7 +544,7 @@ const initializeProxy = async function ({ configPath, distDir, env, host, port, 
 
       if (isEdgeFunctionsRequest(req) && isUncaughtError) {
         const acceptsHtml = req.headers && req.headers.accept && req.headers.accept.includes('text/html')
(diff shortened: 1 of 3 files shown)


lukasholzer (approved):
LGTM!

PR #6018, opened 2023-09-21
feat: support npm modules in edge functions locally

eduardoboucas:
#### Summary

This is the CLI counterpart to https://github.com/netlify/edge-bundler/pull/475.

Part of https://github.com/netlify/pod-dev-foundations/issues/578.


Skn0tt (approved):
Looking good!

src/lib/edge-functions/registry.mjs
```diff
@@ -147,10 +167,23 @@ export class EdgeFunctionsRegistry {
...
+      } = await this.#runIsolate(this.#functions, this.#env, {
         getFunctionsConfig: true,
       })
 
+      if (features.npmModules && !this.#hasShownNPMWarning) {
+        log(
+          `${NETLIFYDEVWARN} Support for npm modules in edge functions is an experimental feature. To learn more about the current state of this capability or to report a problem, refer to https://ntl.fyi/edge-functions-npm.`,
+        )
+
+        this.#hasShownNPMWarning = true
```
eduardoboucas, line 184: Because we don't want to show this message on every file save.

src/lib/edge-functions/registry.mjs
```diff
@@ -256,7 +289,9 @@ export class EdgeFunctionsRegistry {
       return
     }
 
-    log(`${NETLIFYDEVLOG} ${chalk.magenta('Reloading')} edge functions...`)
+    const reason = this.#debug ? ` because ${chalk.underline(path)} has changed` : ''
```
eduardoboucas, line 292: This is particularly useful when debugging failures in the CI, but I can also see it being useful when troubleshooting strange behaviour with file watchers.

src/lib/edge-functions/registry.mjs
```diff
@@ -513,7 +548,9 @@ export class EdgeFunctionsRegistry {
    * @returns {Promise<void>}
    */
   async #setupWatcherForDirectory(directory) {
+    const ignored = [new RegExp(this.#servePath)]
```
eduardoboucas, line 551: We don't want to watch node modules for changes, and those now live in `servePath`.

src/utils/command-helpers.mjs
```diff
@@ -227,12 +227,17 @@ const DEBOUNCE_WAIT = 100
...
  * @param {() => any} [opts.onChange]
  * @param {() => any} [opts.onUnlink]
  */
-export const watchDebounced = async (target, { depth, onAdd = () => {}, onChange = () => {}, onUnlink = () => {} }) => {
-  const watcher = chokidar.watch(target, { depth, ignored: /node_modules/, ignoreInitial: true })
+export const watchDebounced = async (
+  target,
+  { depth, ignored = [], onAdd = () => {}, onChange = () => {}, onUnlink = () => {} },
+) => {
+  const baseIgnores = [/\/(node_modules|.git)\//]
```
eduardoboucas, line 239: We were previously ignoring `node_modules` only. I've added `.git` because setting up watchers there is also wasteful.

tests/integration/commands/dev/edge-functions.test.ts
```diff
@@ -1,12 +1,38 @@
+import process from 'process'
+
 import { describe, expect, expectTypeOf, test } from 'vitest'
 
 import { FixtureTestContext, setupFixtureTests } from '../../utils/fixture.js'
 import got from '../../utils/got.cjs'
 import { pause } from '../../utils/pause.cjs'
 
+// Skipping tests on Windows because of an issue with t he Deno CLI throwing IO
```
eduardoboucas, line 9: This makes me sad, but I really tried.

tests/integration/utils/fixture.ts
```diff
@@ -169,6 +169,12 @@ export async function setupFixtureTests(
       }
     })
 
+    afterEach<FixtureTestContext>((context) => {
+      if (devServer && context.meta.result?.state === 'fail') {
+        console.log(devServer.output)
```
eduardoboucas, line 174: This is useful for troubleshooting failures in the CI. When the test fails, we print the stdout we've accumulated.

src/lib/edge-functions/registry.mjs
Skn0tt, line 551: This might match too many files, if `servePath` contains a `.` character or other Regex control sequences. You never know :D
Since this ends up being passed into chokidar, I think this should work better:
```suggestion
    const ignored = [`${this.#servePath}/**`]
```

(see [anymatch docs](https://github.com/micromatch/anymatch#usage) for reference)

tests/integration/commands/dev/edge-functions.test.ts
```diff
@@ -1,12 +1,38 @@
...
+  { path: 'sites/foo', response: siteInfo },
+
+  { path: 'sites/foo/service-instances', response: [] },
+  {
+    path: 'accounts',
+    response: [{ slug: siteInfo.account_slug }],
+  },
+]
+
 describe('edge functions', () => {
```
Skn0tt, line 33: Mega nit: Instead of adding `skipIf` to every single test, it should be on the test suite:

```suggestion
describe.skipIf(isWindows)('edge functions', () => {
```

Skn0tt (approved):
The approach looks good! Left two nits, take them or leave them.

Skn0tt:
Gonna play around with this now and try to break it with weird NPM modules :)

PR #6027, opened 2023-09-28
fix: include `routes` in functions config on deploy

Skn0tt:
We weren't including `routes` in the `functions_config` parameter we're sending to the API, so custom v2 routes didn't work properly. This PR fixes that.

### DIFF
--- a/tests/integration/210.command.deploy.test.cjs
+++ b/tests/integration/210.command.deploy.test.cjs
@@ -437,7 +437,7 @@ if (process.env.NETLIFY_TEST_DISABLE_LIVE !== 'true') {
     })
   })
 
-  test.serial('should deploy functions from internal functions directory', async (t) => {
+  test.only('should deploy functions from internal functions directory', async (t) => {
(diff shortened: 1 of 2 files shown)


tests/integration/210.command.deploy.test.cjs
```diff
@@ -437,7 +437,7 @@ if (process.env.NETLIFY_TEST_DISABLE_LIVE !== 'true') {
     })
   })
 
-  test.serial('should deploy functions from internal functions directory', async (t) => {
+  test.only('should deploy functions from internal functions directory', async (t) => {
```
lemusthelroy, line 440: do you need to remove the `test.only` ?

estephinson: what a catch from thorley

lemusthelroy: I commit `test.only` at least once a day

lemusthelroy (dismissed):

estephinson (approved):

minivan (approved):

PR #6030, opened 2023-10-02
fix: simplify file watching expression

eduardoboucas:
#### Summary

Follow-up to https://github.com/netlify/cli/pull/6018 addressing a couple of items from the code review.

### DIFF
--- a/src/lib/edge-functions/registry.mjs
+++ b/src/lib/edge-functions/registry.mjs
@@ -548,7 +548,7 @@ export class EdgeFunctionsRegistry {
    * @returns {Promise<void>}
    */
   async #setupWatcherForDirectory(directory) {
-    const ignored = [new RegExp(this.#servePath)]
+    const ignored = [`${this.#servePath}/**`]
     const watcher = await watchDebounced(directory, {
       ignored,
       onAdd: () => this.#checkForAddedOrDeletedFunctions(),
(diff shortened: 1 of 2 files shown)


src/lib/edge-functions/registry.mjs
```diff
@@ -548,7 +548,7 @@ export class EdgeFunctionsRegistry {
    * @returns {Promise<void>}
    */
   async #setupWatcherForDirectory(directory) {
-    const ignored = [new RegExp(this.#servePath)]
+    const ignored = [`${this.#servePath}/**`]
```
lukasholzer, line 551: is this using unix paths on windows as well, I guess so but just checking.

tests/integration/commands/dev/edge-functions.test.ts
```diff
@@ -30,9 +30,9 @@ const routes = [
   },
 ]
 
-describe('edge functions', () => {
+describe.skipIf(isWindows)('edge functions', () => {
```
lukasholzer, line 33: why are they skipped on windows? shouldn't it work there as well?

eduardoboucas: https://github.com/netlify/cli/blob/e08a47f2d362fa82a64237260b5bc517ca75a6b3/tests/integration/commands/dev/edge-functions.test.ts#L9-L10

src/lib/edge-functions/registry.mjs
eduardoboucas: Yes: https://github.com/micromatch/micromatch#backslashes.

lukasholzer (approved):
But not blocking it :)

PR #6033, opened 2023-10-03
fix: log function invocation errors

eduardoboucas:
#### Summary

When using functions with the V2 API, CLI was hiding some useful errors from the user, making troubleshooting very difficult.

This PR addresses that by capturing errors and showing them nicely, indicating the function that they're associated with. I ended up cleaning up the logic for formatting errors, which was a bit messy.

It looks like this:

Screenshot 2023-10-03 at 17 45 17
Screenshot 2023-10-03 at 17 45 27


sarahetter (approved):
lovely, ty!

PR #6046, opened 2023-10-09
fix: use only pathname when matching function URLs

eduardoboucas:
#### Summary

Fixes a regression in https://github.com/netlify/cli/pull/5998. When `getFunctionForURLPath` receives a URL, it can contain query parameters. When matching function routes, we want to match only the pathname, so we must parse the URL.


Skn0tt (approved):

eduardoboucas:
Closing in favour of https://github.com/netlify/cli/pull/6043.

PR #6048, opened 2023-10-09
fix: don't run form submission logic on paths matching function

Skn0tt:
Part of https://github.com/netlify/pod-dev-foundations/issues/578. See title.

### DIFF
--- a/src/lib/functions/form-submissions-handler.mjs
+++ b/src/lib/functions/form-submissions-handler.mjs
@@ -29,7 +29,12 @@ const getFormHandler = function ({ functionsRegistry }) {
 
 export const createFormSubmissionHandler = function ({ functionsRegistry, siteUrl }) {
   return async function formSubmissionHandler(req, res, next) {
-    if (req.url.startsWith('/.netlify/') ||req.method !== 'POST' || (await functionsRegistry.getFunctionForURLPath(req.url, req.method))) return next()
+    if (
+      req.url.startsWith('/.netlify/') ||
+      req.method !== 'POST' ||
+      (await functionsRegistry.getFunctionForURLPath(req.url, req.method))
+    )
+      return next()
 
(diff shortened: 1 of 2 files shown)


eduardoboucas (dismissed):

src/lib/functions/form-submissions-handler.mjs
```diff
@@ -29,7 +29,12 @@ const getFormHandler = function ({ functionsRegistry }) {
 
 export const createFormSubmissionHandler = function ({ functionsRegistry, siteUrl }) {
   return async function formSubmissionHandler(req, res, next) {
-    if (req.url.startsWith('/.netlify/') ||req.method !== 'POST' || (await functionsRegistry.getFunctionForURLPath(req.url, req.method))) return next()
+    if (
+      req.url.startsWith('/.netlify/') ||
+      req.method !== 'POST' ||
+      (await functionsRegistry.getFunctionForURLPath(req.url, req.method))
```
eduardoboucas, line 35: Is it possible that `functionsRegistry` isn't defined here? It's a bit hard to tell, but we do have a few checks for it elsewhere in the codebase, which makes me think it may not be there at times.

eduardoboucas (approved):

src/lib/functions/form-submissions-handler.mjs
Skn0tt: From what i'm seeing, this is the only codepath where this is ever going to be called:

https://github.com/netlify/cli/blob/62e0ef553eb17c872099f25b3e139eecf04aba1d/src/lib/functions/server.mjs#L276-L290

So `functionsRegistry` is always defined.

eduardoboucas: 👍🏻 🚀

PR #6069, opened 2023-10-16
feat: show warning for npm modules with extraneous files

eduardoboucas:
#### Summary

Show warning when edge functions import npm modules that contain extraneous files.

### DIFF
--- a/src/lib/edge-functions/proxy.mjs
+++ b/src/lib/edge-functions/proxy.mjs
@@ -8,7 +8,6 @@ import * as bundler from '@netlify/edge-bundler'
 import getAvailablePort from 'get-port'
 
 import { NETLIFYDEVERR, NETLIFYDEVWARN, chalk, error as printError, log } from '../../utils/command-helpers.mjs'
-import { isFeatureFlagEnabled } from '../../utils/feature-flags.mjs'
 import { getGeoLocation } from '../geo-location.mjs'
 import { getPathInProject } from '../settings.mjs'
 import { startSpinner, stopSpinner } from '../spinner.mjs'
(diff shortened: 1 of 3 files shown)


src/lib/edge-functions/proxy.mjs
```diff
@@ -110,7 +109,7 @@ export const initializeProxy = async ({
   const userFunctionsPath = config.build.edge_functions
   const isolatePort = await getAvailablePort()
   const buildFeatureFlags = {
-    edge_functions_npm_modules: isFeatureFlagEnabled('edge_functions_npm_modules', siteInfo),
+    edge_functions_npm_modules: true,
```
Skn0tt, line 112: I think there's another place where we need to set this, so it applies to `ntl build`

src/lib/edge-functions/registry.mjs
```diff
@@ -207,6 +195,14 @@ export class EdgeFunctionsRegistry {
       )
 
       this.#processGraph(graph)
+
+      if (npmSpecifiersWithExtraneousFiles.length !== 0) {
+        const modules = npmSpecifiersWithExtraneousFiles.map((name) => chalk.yellow(name)).join(', ')
+
+        log(
+          `${NETLIFYDEVWARN} The following npm modules, which are directly or indirectly imported by an edge function, may not be supported: ${modules}. For more information, visit https://ntl.fyi/edge-npm.`,
```
Skn0tt, line 203: this isn't pointing anywhere yet. i'm assuming you want to point this at the docs, or at the blog post?

Skn0tt (changes requested):

src/lib/edge-functions/registry.mjs
eduardoboucas: The blog post won't cover this directly. I think we'd point it either at the docs or at the forums post.

Skn0tt (approved):

PR #6075, opened 2023-10-17
fix: don't swallow file changes when a lot of them come in

Skn0tt:
When frameworks perform a build during local dev, they sometimes touch a lot of files at the same time, and chokidar will notify about those. Since we were debouncing them, only one of those file updates came through to the bundling implementation - all others were swallowed. If that means the function file is swallowed, we don't rebuild functions/edge functions when we should.

I thought about removing the debouncing behaviour, but it's useful so we don't perform too many builds (wasting CPU). Instead, I opted for collecting all paths that are changed, and passing the full list down once the debouncer makes the call. This ensures that our build system has all changed files and can correctly determine when to build.


ascorbic (commented):
I can confirm that this fixes the remix issue!

ascorbic (approved):

PR #6092, opened 2023-10-25
refactor: convert the rest of the project to ESM

ericapisani:
Finish the conversion to ESM.

Also includes the changes made in https://github.com/netlify/cli/pull/5921


.commitlintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["@commitlint/config-conventional"]
```
estephinson, line 2: what be this

tests/integration/utils/snapshots.mjs
```diff
@@ -37,4 +37,5 @@ const normalize = (inputString, { duration, filePath } = {}) => {
     .trim()
 }
 
-module.exports = { normalize }
+const exports = { normalize }
+export default exports
```
estephinson, line 41: copilot let you down?

estephinson (changes requested):
One very small `exports` change, but aside from that lfg

.commitlintrc.json
ericapisani: I think this is the config file used by husky for the git hooks

tests/integration/utils/snapshots.mjs
ericapisani: Lol yeah

ericapisani: Actually, I'm realizing this was here because of this lint error that occurs when I just did `export default {normalize}`

Screenshot 2023-10-30 at 8 11 41 AM

.commitlintrc.json
estephinson: should it be in the PR tho? looks a new thing

tests/integration/utils/snapshots.mjs
estephinson: yeah but it's already exported, so we should be fine without those lines

.commitlintrc.json
ericapisani: Yeah, I double checked and this is correct. We had a `commitlint.config.cjs` and because I transitioned it to a JSON file, the file name is a different.

Source: https://github.com/conventional-changelog/commitlint#config

tests/integration/utils/snapshots.mjs
ericapisani: 🤦🏻‍♀️ 

Excuse me while I go get some caffeine ☕ 

Good shout, removing these lines now.

estephinson (approved):
lets gooooooo

PR #6103, opened 2023-10-30
chore: replace ava with vitest

ericapisani:

#### Summary

Fixes [this issue](https://linear.app/netlify/issue/CT-89/finish-conversion-from-ava-to-vitest)

---

      passes our tests.


.github/workflows/legacy-tests.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         with:
           cond: ${{ github.event_name == 'pull_request' }}
           if_true: 'npm run test:affected ${{ github.event.pull_request.base.sha }}' # on pull requests test with the project graph only the affected tests
-          if_false: 'npm run test:ci:ava:integration' # on the base branch run all the tests as security measure
+          if_false: 'npm run test:ci:vitest:integration' # on the base branch run all the tests as security measure
```
sarahetter, line 72: These tests are already run in the `integration-tests.yml` action. I believe we won't need this action anymore after converting?

ericapisani: Possibly but I think I'd rather do this as a separate PR - I want to look at the `legacy-tests` pipeline more in general before I start messing with it some more

sarahetter: This means that those tests will be run twice though, if I'm not mistaken?

khendrikse (approved):

sarahetter (approved):

PR #6105, opened 2023-10-30
chore: upgrade minimum node version to 18

ericapisani:

#### Summary

Node 16 was end-of-lifed recently so we need to upgrade.

---

      passes our tests.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -24,7 +24,7 @@ jobs:
         #
         # os: [ubuntu-latest, macOS-latest, windows-latest]
         os: [ubuntu-latest, macOS-latest]
-        node-version: ['16.16.0', '*']
+        node-version: ['18.18.2', '*']
       fail-fast: false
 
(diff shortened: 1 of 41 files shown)


ericapisani:
Just to give context on why the removal of `node-fetch` was reversed in this PR - some of the integration tests were failing as a result of this change and I wasn't able to determine why in some cases.

For example, [this assertion](https://github.com/netlify/cli/blob/b72aa6c25e746ac68ac6b6edc7a58386b515ac74/tests/integration/commands/dev/dev.config.test.mjs#L391) failed because the `content-type` header was being set differently between implementations, reasons of which I wasn't able to figure out after poking at it a little bit.

To be on the safe side, I'm going to keep the uses of `node-fetch` where they currently are in order to reduce the risk of a regression.

estephinson (approved):
lets go lets go lets go lets go lets go :shipit: 🚢 🚀 🌔

khendrikse (approved):
LGTM

PR #6109, opened 2023-10-31
feat: add Blobs context to edge functions

eduardoboucas:
#### Summary

Follow-up to #6101. Extends Blobs context to edge functions.

Part of https://github.com/netlify/pillar-runtime/issues/754.


src/lib/functions/netlify-function.mjs
```diff
@@ -195,22 +195,14 @@ export default class NetlifyFunction {
     const environment = {}
 
     if (this.blobsContext) {
-      if (this.runtimeAPIVersion === 2) {
-        // For functions using the v2 API, we inject the context object into an
-        // environment variable.
-        environment.NETLIFY_BLOBS_CONTEXT = Buffer.from(JSON.stringify(this.blobsContext)).toString('base64')
```
eduardoboucas, line 201: I realised we don't need this. The bootstrap layer can take the data from the client context and set the environment variable, like it's doing in production.

mraerino (dismissed):

ascorbic (approved):

sarahetter (approved):

PR #6112, opened 2023-11-01
Revert "fix: min version should be lowest version of 18, not the latest"

ericapisani:
Reverts netlify/cli#6110

There looks to be an issue with E2E tests with v18.0.0 so reverting in order to investigate further


estephinson (approved):
excited for these end to end tests to fail

PR #6120, opened 2023-11-03
feat: add MSW and update `telemetry.test.ts` to start using it

ericapisani:
#### Summary

Adds MSW and updates the `telemetry.test.ts` file to start using it.

Some things to note about the changes here:
* The approach of proxying requests through the express app was taken because MSW only intercepts requests that happen in the same thread. This means that we can't remove the use of the express app because just using `execa` to call the CLI without a mock API/server spawns a child thread whose requests won't be intercepted by MSW.
* MSW recommends against inspecting/asserting against the requests themselves. That being said, sending analytics is one of those use cases where the maintainer sees it as being ok to do request assertions which is why they're being done here.
* MSW doesn't have a convenient way of failing tests if an unhandled request is made which is why there is the possibility of throwing an error in the `afterEach` clause of the tests. See [this discussion](https://github.com/mswjs/msw/discussions/821) for the full context but the short of it is that this incorporates one of the suggested workarounds. We will want to look into how we can do these fallback routes automatically so they're not accidentally missed in test suites as if they are the tests will still pass even though there will be messages saying that there was a request made that had no backing handler defined on MSW.
* While MSW suggests setting the server up on the global level, the setup of globals via [`globalSetup`](https://vitest.dev/config/#globalsetup) sets up those globals in a different global scope than what the tests have access to. This meant that in order to set up the MSW server at as high a level as possible while still giving tests access to the server, we needed to set up the MSW server using the [`setups`](https://vitest.dev/config/#setupfiles) property

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -143,6 +143,7 @@
         "ini": "2.0.0",
         "is-ci": "3.0.1",
         "mock-fs": "5.2.0",
+        "msw": "^2.0.2",
         "nock": "13.3.2",
         "p-timeout": "4.1.0",
         "serialize-javascript": "6.0.1",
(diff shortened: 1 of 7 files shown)


ericapisani:
Closing for now while we take another look at our approach to testing in general.

PR #6122, opened 2023-11-06
feat: Support image cdn

kathmbeck:

#### Summary

---

      passes our tests.


src/lib/images/proxy.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+export const initializeProxy = async function ({ config }) {
+  const remoteDomains = await parseRemoteImageDomains({ config })
+
+  const ipx = createIPX({
+    storage: ipxFSStorage({ dir: config?.build?.publish ?? './public' }),
+    httpStorage: ipxHttpStorage({ domains: remoteDomains }),
+  })
+
+  const handler = createIPXWebServer(ipx)
```
ascorbic, line 105: IPX supports generating express-like handlers directly, so it might be easier to use that rather than converting from the web one

src/lib/images/proxy.mjs
```diff
@@ -0,0 +1,124 @@
...
+
+  const { errors, remoteDomains } = await parseAllDomains(config)
+  await handleImageDomainsErrors(errors)
+
+  return remoteDomains
+}
+
+export const isImageRequest = function (req) {
+  const imageUrlPattern = /^\/\.netlify\/images/
+  return imageUrlPattern.test(req.url)
```
mraerino, line 59: nit: this can be solved without regex

```suggestion
const imageUrlPattern = "/.netlify/images/"
export const isImageRequest = function (req) {
  return req.url.startsWith(imageUrlPattern)
```

src/lib/images/proxy.mjs
kathmbeck: Much nicer - 
https://github.com/netlify/cli/pull/6122/files#diff-b31113e48a7d59ac6763fa71104f1bde79106f080daeeb96a3d85e8c889d834cR104-R112

src/lib/images/proxy.mjs
```diff
@@ -0,0 +1,115 @@
...
+    return []
+  }
+
+  const { errors, remoteDomains } = await parseAllDomains(config)
+  await handleImageDomainsErrors(errors)
+
+  return remoteDomains
+}
+
+const imageUrlPattern = '/.netlify/images'
```
sarahetter, line 55: [dust] I'd make this an all caps const (`export const IMAGE_URL_PATTERN = '/.netlify/images'`) at the top of the file as we have this as a convention in the repo - easier to find if we need to extend this and use it across files

src/lib/images/proxy.mjs
```diff
@@ -0,0 +1,115 @@
...
+
+  const ipx = createIPX({
+    storage: ipxFSStorage({ dir: config?.build?.publish ?? './public' }),
+    httpStorage: ipxHttpStorage({ domains: remoteDomains }),
+  })
+
+  const handler = createIPXNodeServer(ipx)
+  const app = express()
+
+  app.use('/.netlify/images', async (req, res) => {
```
sarahetter, line 106: good place to use the const variable

tests/integration/commands/dev/images.test.mjs
```diff
@@ -0,0 +1,89 @@
...
+          </html>
+        `,
+          path: 'index.html',
+        })
+
+      await builder.buildAsync()
+
+      await withDevServer({ cwd: builder.directory }, async (server) => {
+        await fetch(
+          `${server.url}/.netlify/images?url=https://images.unsplash.com/photo-1517849845537-4d257902454a&w=100&h=200&q=80&fm=avif&fit=crop&crop=left`,
```
sarahetter, line 41: same thing here as in my other comments with the `'/.netlify/images` const :)

src/lib/images/proxy.mjs
ascorbic: Nice

sarahetter (dismissed):

sarahetter (approved):

PR #6130, opened 2023-11-07
fix: support npm modules + edge functions in monorepos

eduardoboucas:
#### Summary

To make npm modules work in edge functions when in a monorepo setup, we need to use the repository root (and not the package root) as the base directory where we look for modules.

Closes https://linear.app/netlify/issue/COM-114/npm-modules-in-edge-functions-not-working-in-monorepo-setups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Manage netlify functions
 | [`functions:create`](/docs/commands/functions.md#functionscreate) | Create a new function locally  |
 | [`functions:invoke`](/docs/commands/functions.md#functionsinvoke) | Trigger a function while in netlify dev with simulated data, good for testing function calls including Netlify's Event Triggered Functions  |
 | [`functions:list`](/docs/commands/functions.md#functionslist) | List functions that exist locally  |
-| [`functions:serve`](/docs/commands/functions.md#functionsserve) | (Beta) Serve functions locally  |
+| [`functions:serve`](/docs/commands/functions.md#functionsserve) | Serve functions locally  |
 
 
 ### [init](/docs/commands/init.md)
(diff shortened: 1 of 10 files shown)


src/lib/edge-functions/registry.mjs
```diff
@@ -321,50 +377,69 @@ export class EdgeFunctionsRegistry {
...
+   * @param {object} data
+   * @param {Error} [data.buildError]
+   * @param {string} [data.functionName]
+   * @param {string[]} [data.warnings]
+   * @returns
    */
-  #logAddedFunction(func) {
-    log(`${NETLIFYDEVLOG} ${chalk.green('Loaded')} edge function ${chalk.yellow(this.#getDisplayName(func.name))}`)
-  }
+  #logEvent(event, { buildError, functionName, warnings = [] }) {
```
eduardoboucas, line 389: I have centralised all log messages in this method, similar to what we're doing with the serverless functions registry.

src/lib/edge-functions/registry.mjs
```diff
@@ -321,50 +377,69 @@ export class EdgeFunctionsRegistry {
...
+      log(`${NETLIFYDEVLOG} ${chalk.magenta('Removed')} ${subject}`)
+    }
   }
 
   /**
    * @param {string} urlPath
    * @param {string} method
    */
   matchURLPath(urlPath, method) {
-    const declarations = this.#bundler.mergeDeclarations(
```
eduardoboucas, line 342: Instead of generating a manifest and routes on each request, we generate it every time we build the functions and then cache it in `this.#manifest` and `this.#routes`.

PR #6136, opened 2023-11-08
feat: convert src/ folder to Typescript

ericapisani:
#### Summary
* Converts `.mjs` files to `.mts` files. This ensures that when compiled, the output is `.mjs` to match what we have today.
* Does not convert `src/functions-templates` files. There were issues coming up as a result and figured it was safer to leave the files within a `.mjs` since converting them to `.mts` didn't provide us much benefit
* Include generated type files in the package for the files
* Because of the desire to revisit the testing strategy/test suite as a whole and to reduce the risk of introducing a regression in the tests themselves, I opted to leave the tests as-is (which are currently a mix of Typescript and `.mjs` files)

While I initially wanted to try and do this folder by folder, because TS pulls in imported files and does type-checking on them as well, it basically became a can of worms. After chatting with @kaelig about this, we're opting for a single migration of the `src` folder.

---

      passes our tests.


package.json
```diff
@@ -190,6 +195,7 @@
     "@netlify/eslint-config-node": "7.0.0",
     "@netlify/functions": "2.0.2",
     "@types/fs-extra": "11.0.1",
+    "@types/inquirer": "^9.0.7",
```
kaelig, line 198: Should this be a strict version? (`9.0.7`)

src/commands/api/api.mts
```diff
@@ -1,4 +1,5 @@
-// @ts-check
+ 
+// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'asci... Remove this comment to see the full error message
 import AsciiTable from 'ascii-table'
```
kaelig, line 3: This may do the trick: https://www.npmjs.com/package/@extend-types/ascii-table

src/commands/build/build.mts
```diff
@@ -9,7 +9,9 @@ import { getEnvelopeEnv, normalizeContext } from '../../utils/env/index.mjs'
 /**
  * @param {import('../../lib/build.mjs').BuildConfig} options
  */
```
kaelig, line 11: You can now convert these comment imports to actual imports 🎉

kaelig (commented):
But since we have a lot of JSDoc type definitions, it'd be great to be able to convert them into TypeScript. It'd probably fix most of the ts-expect-error comments.

Consider taking a look at https://github.com/fivetran/typescript-closure-tools (I haven't tested it before)

ericapisani:
@kaelig Thanks for the feedback! 

I've addressed the version pinning for `inquirer` you mentioned in this branch and I'll make sub-issues for the other things you've pointed out.

src/lib/functions/netlify-function.mjs
```diff
@@ -1,265 +0,0 @@
-// @ts-check
```
sarahetter, line 1: Just wondering why this was deleted?

sarahetter: Ugh nvm, just one that didn't get caught as a rename. Sorry!

sarahetter (dismissed):
💃 🥳 🎈 🎉 🌮

sarahetter (approved):

PR #6139, opened 2023-11-09
feat: stream function and build logs

estephinson:

#### Summary

Beta commands for streaming of function and build logs

Closes CT-97

---

      passes our tests.


tests/integration/utils/call-cli.mjs
```diff
@@ -14,6 +14,7 @@ const CLI_TIMEOUT = 3e5
  * @returns {Promise<string|object>}
  */
 export const callCli = async function (args, execOptions = {}, parseJson = false) {
+  console.log({ args, cliPath, execOptions })
```
sarahetter, line 17: is this meant to still be there?

src/commands/main.mjs
```diff
@@ -205,6 +206,9 @@ export const createMainCommand = () => {
   createUnlinkCommand(program)
   createWatchCommand(program)
 
+  createLogsBuildCommand(program)
```
sarahetter, line 209: Can we combine these into a generic `logs` command (like we do with `functions`) so that it shows up nice in the docs?

sarahetter:
You'll need to add the docs stuff - https://github.com/netlify/cli/pull/6173/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R107

sarahetter:
LGTM so far! You're gonna need to convert to TS, so not gonna add my checkmark yet but ping me when you're ready

tests/integration/utils/call-cli.mjs
estephinson: nay!

estephinson: nay

src/commands/logs/build.mts
```diff
@@ -0,0 +1,92 @@
+import { OptionValues } from 'commander'
```
ericapisani, line 1: Given this is only being used for the type you can change this import to:

```suggestion
import type { OptionValues } from 'commander'
```

src/commands/logs/build.mts
```diff
@@ -0,0 +1,92 @@
...
+import inquirer from 'inquirer'
+
+import { log, chalk } from '../../utils/command-helpers.mjs'
+import { getWebSocket } from '../../utils/websockets/index.mjs'
+import BaseCommand from '../base-command.mjs'
```
ericapisani, line 6: Same thing here as for `OptionValues`

```suggestion
import type BaseCommand from '../base-command.mjs'
```

src/commands/logs/build.mts
```diff
@@ -0,0 +1,92 @@
...
+  ws.on('close', () => {
+    log('---')
+  })
+}
+
+export const createLogsBuildCommand = (program: BaseCommand) =>
+  program
+    .command('logs:deploy')
+    .alias('logs:build')
+    .description('(Beta) Stream the logs of active deploys to the console')
```
ericapisani, line 91: I think it would be worth clarifying that 'active deploys' can mean a production or deploy preview. Given we refer to what's on production as 'published' I think it's easy to read 'active' as potentially being understood as what's 'published'

src/commands/logs/build.mts
```diff
@@ -0,0 +1,92 @@
...
+
+import { log, chalk } from '../../utils/command-helpers.mjs'
+import { getWebSocket } from '../../utils/websockets/index.mjs'
+import BaseCommand from '../base-command.mjs'
+
+export function getName({ deploy, userId }: { deploy: any; userId: string }) {
+  let normalisedName = ''
+  const isUserDeploy = deploy.user_id === userId
+
+  switch (deploy.context) {
```
ericapisani, line 12: Nit: It'd be great to have the potential values be an enum or something so that it's easier to keep track of all the potential values this could be as well as ensure that we factor in all possible cases in future code that looks similar to this

src/commands/main.mts
```diff
@@ -212,6 +212,7 @@ export const createMainCommand = () => {
   createSwitchCommand(program)
   createUnlinkCommand(program)
   createWatchCommand(program)
+  createLogsCommand(program)
```
sarahetter, line 215: ohhh right these show up in order in the `netlify help` command. Can you alphabetize it pls?

estephinson: you're joking

estephinson: nah I'm gonna refactor the help command

src/commands/logs/functions.mts
```diff
@@ -0,0 +1,101 @@
+import { Argument, OptionValues } from 'commander'
+import inquirer from 'inquirer'
+
+import { chalk, log } from '../../utils/command-helpers.mjs'
+import { getWebSocket } from '../../utils/websockets/index.mjs'
+import BaseCommand from '../base-command.mjs'
```
ericapisani, line 6: Similar to earlier comments, there's a few imports here that can be imported using `import type` rather than just `import` since they're only being used for their type definitions

src/commands/logs/functions.mts
```diff
@@ -0,0 +1,101 @@
...
+
+function getLog(logData: { level: string; message: string }) {
+  let logString = ''
+  switch (logData.level) {
```
ericapisani, line 10: Similar to comment above regarding potential deploy context values, we can do something similar for the `level`

src/commands/logs/functions.mts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  if (!selectedFunction) {
+    log(`Could not find function ${functionName}`)
+    return
+  }
+
+  const { a: accountId, oid: functionId } = selectedFunction
+
+  const ws = getWebSocket('wss://socketeer.services.netlify.com/function/logs')
```
ericapisani, line 61: Nit: Not sure if we have a file with common URLs that we refer to in the CLI but if we do we should put the base URL (`wss://socketeer.services.netlify.com`) there as a `const` value

src/commands/logs/functions.mts
```diff
@@ -0,0 +1,101 @@
...
+  })
+
+  ws.on('error', (err: any) => {
+    log('Connection error')
+    log(err)
+  })
+}
+
+/**
+ * Creates the `netlify watch` command
```
ericapisani, line 90: This comment should be updated

src/commands/logs/functions.mts
```diff
@@ -0,0 +1,101 @@
...
+ * @param {import('../base-command.mjs').default} program
+ * @returns
+ */
+export const createLogsFunctionCommand = (program: BaseCommand) =>
+  program
+    .command('logs:function')
+    .alias('logs:functions')
+    .addArgument(new Argument('[functionName]', 'Name of the function to stream logs for'))
+    .addExamples(['netlify logs:function my-function', 'netlify logs:function'])
+    .description('(Beta) Stream serverless function logs to the console')
```
ericapisani, line 100: Nit: From a docs perspective, should we be more specific and say that 'serverless' = 'Netlify Function' ? Unless this command could also be used to output logs for edge functions, in which case this comment can be ignored

tests/integration/commands/help/__snapshots__/help.test.ts.snap
```diff
@@ -33,7 +33,8 @@ COMMANDS
   $ status        Print status information
   $ switch        Switch your active Netlify account
   $ unlink        Unlink a local folder from a Netlify site
-  $ watch         Watch for site deploy to finish"
+  $ watch         Watch for site deploy to finish
+  $ logs          Stream logs from your site"
```
ericapisani: We should be more specific here on what this means

src/commands/base-command.mts
```diff
@@ -270,20 +268,21 @@ export default class BaseCommand extends Command {
     /**
      * @param {BaseCommand} command
      */
-    // @ts-expect-error TS(7006) FIXME: Parameter 'command' implicitly has an 'any' type.
-    const getCommands = (command) => {
+    const getCommands = (command: Command) => {
```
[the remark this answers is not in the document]
estephinson: hmmmm yeah you're right, plus that's the existing typing which I should've removed anyway.

I thought it wouldn't be as this is called on values returned from commander but I guess they have everything in already

tests/integration/commands/logs/build.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        state: 'building',
+        context: 'production',
+        id: 'deploy-id',
+      },
+    ],
+  },
+]
+
+describe('logs:deploy command', () => {
+  //   beforeEach(() => {
```
ericapisani, line 49: Does this need to be removed?

src/commands/logs/build.mts
estephinson: Cool - we should make this a ESLint rule if we're enforcing this

src/commands/logs/build.mts
estephinson: yeah good call out

src/commands/logs/build.mts
sarahetter: Agree with eslint rule

src/commands/logs/functions.mts
estephinson: I mean maybe? Not sure there's much value add at this stage

tests/integration/commands/logs/build.test.ts
estephinson: could be left as a reminder of good times

estephinson: And by that I mean the horrible times I had getting these tests setup

src/commands/logs/functions.mts
estephinson: nuking them as it's self explanatory anyway

src/commands/logs/functions.mts
estephinson: I would like them both to be on this method, but it's currently just netlify functions

src/commands/logs/build.mts
estephinson: Yeah a string literal may be appropriate here but I think it's a bit premature atm, when we expand out be worth typing properly

tests/integration/commands/help/__snapshots__/help.test.ts.snap
estephinson: I think this covers it? Don't want to go too in-depth on a top level command imo

sarahetter (approved):

PR #6143, opened 2023-11-09
[do not merge] Clack POC

sarahetter:
Related to: CT-101 (design system) and CT-102 (removing bloat)

This is a POC of using `@clack/prompts` instead of `inquirer` for our prompting package. 

##  Reasons we'd like to do this:
- Inquirer (and `inquirer-autocomplete-prompt`) brings in rxjs which is a huge sub dependency, and we'd like to remove it. (Ends up being around 17MB on filesystem)

| Screenshot 2023-11-09 at 4 24 25 PM | Screenshot 2023-11-09 at 4 24 37 PM |
|------|------|
- Clack is more customizable than inquirer

## Description of clack
Clack has two packages - `@clack/core` and `@clack/prompts`. 

`@clack/core` is a group of plain, unstyled prompts. When instantiating a new prompt, you pass a `render` callback in, and it is updated along with the state when it changes. You can see an example of using `@clack/core` in `src/commands/clack/clack-core.mjs` or by running `./bin/run.mjs clack:core.

Prompts is just a pre-styled use of `@clack/core` with (imo) a pleasant to use interface, and **could save us a lot of design and dev time**. It also contains some non-prompt utilities such as logging methods to match with the prompts. You can see an example of using `@clack/prompts` in `src/commands/clack/clack-prompts.mjs`, or by running `./bin/run.mjs clack:prompts`

## Screenshots / Gifs
I've converted one command `netlify sites:create-template` over to `@clack/prompts` from `inquirer`

| Inquirer (before)  | `@clack/prompts` (after)  |
|---|---|
| withinquirer   |  withclack |
| withinquirer  | withclack  |


samkap:
Paired with Sarah today. Small pieces of feedback:
- Change `dim` to `white` on input lines (is that what they're called?)
- Switch Success and step green to cyan by default to nod to Netlify branding.

PR #6145, opened 2023-11-10
fix: allow v2 functions to serve /favicon.ico

Skn0tt:
We have this weird edge case, where the functions server was catching all requests to /favicon.ico and prevented them from being served by functions. This PR fixes that, by moving the logic to a place where it only impacts routes that don't have a function assigned.


mraerino (approved):
the change itself is ok, but i'd question if we should have this code at all. we don't have it in prod and i don't see what upside there is for having it locally

Skn0tt:
Agree. I tried finding the original PR that added this, hoping there would be some context on this, but it goes way back to this commit from 2019: https://github.com/netlify/cli/commit/9ffd51413536302becf7e8736536ed40c528c5c2
I'm assuming this was copied from some other repository, but I have no idea which.
Let's remove this code, and see if someone screams 🤷

mraerino (approved):
🚀

PR #6147, opened 2023-11-10
Simplify function templates

jenae-janzen:
Simplest update of the function templates, to just include a "hello world" example for the v2 api.

#### Summary

---

      passes our tests.


src/functions-templates/javascript/set-res-header/{{name}}.js
minivan, line 1: I think this is an edge function template that we need to keep around

minivan (changes requested):

PR #6156, opened 2023-11-14
docs: use short urls in CLI messages to users

ericapisani:
There's a few places where we can use short URLs in our messages to users.

I tried to replace the open-api.netlify.com ones as well but because URL shorteners don't appear to support providing a custom value (as a hash or query param), I'd need to replace each one individually which seems like a lot for not necessarily the most gain.

---

      passes our tests.


sarahetter (approved):

eduardoboucas (approved):
Love this!

sarahetter (approved):

PR #6171, opened 2023-11-15
fix: fix generated docs for recipes command, remove command list from README

sarahetter:
#### Summary

Things achieved in this PR:
1. Removed weird page from the `netlify recipes` command - these docs weren't being generated, so we got a strange version of a 404.
2. Changed the links in the autogenerated docs in `/docs/commands` - before they were linking to other pages on github within docs, but because these are used in the actual cli.netlify.com docs, the links were broken. We've removed the links to pages within the repo (and in the README.md), and instead are generating links that actually work properly on the cli.netlify.com site.


codebyuma (approved):
Confirmed all the updated links work as expected, the recipes command shows up in the left menu now, and the page loads actual content now :) 🎉

ericapisani (approved):

PR #6173, opened 2023-11-15
docs: adding more detail for adding a command

sarahetter:
Adding docs to clarify that you need to add an item to the nav manually


eduardoboucas (approved):
❤️

PR #6176, opened 2023-11-16
feat: miscellaneous Blobs improvements

eduardoboucas:
#### Summary

- Uses a default site ID when running Netlify Dev with an unlinked site. This makes it possible to use Blobs even if the site is not linked.
- When the first request is made to the local blobs server, a message is printed to the console telling people that Blobs is running in sandbox mode for local development, including a link to the docs
- Updates `@netlify/blobs` to the latest version


minivan (approved):
Blobsciting!

mraerino (approved):

PR #6179, opened 2023-11-16
refactor: rename .mts to .ts

ericapisani:
#### Summary
Using the `.mjs` extension is no longer needed now that we've completed the transition to ESM. Having these file extensions is redundant since we have `type:module` in the `package.json`.

This change updates files that were previously `.mts` to `.ts`, which means that the compiled files will be `.js`. As a result, import lines needed to be updated as well.

---

      passes our tests.


.prettierignore
```diff
@@ -1,3 +1,2 @@
 node_modules
-src/**/*.mjs
-src/**/*.d.mts
+src/**/*.js
```
sarahetter, line 2: We still want the *.d.ts as prettier will run on the compiled `*.d.ts` files in CI without this ignore directive

CHANGELOG.md
```diff
@@ -4424,7 +4424,7 @@ Generated by [`auto-changelog`](https://github.com/CookPete/auto-changelog).
 
 ### Bug Fixes
 
-* serve .mjs functions with esbuild ([#3253](https://www.github.com/netlify/cli/issues/3253)) ([c4ce183](https://www.github.com/netlify/cli/commit/c4ce183ac5b57025cbec42bb1e1f53c32b850909))
+* serve .js functions with esbuild ([#3253](https://www.github.com/netlify/cli/issues/3253)) ([c4ce183](https://www.github.com/netlify/cli/commit/c4ce183ac5b57025cbec42bb1e1f53c32b850909))
```
sarahetter, line 4427: Probably don't want to rewrite the changelog as this PR will add a new update

site/src/_ui.mjs
sarahetter, line 1: Noticing that we're renaming a few `site/` files from .js -> .mjs. Not sure why?

.prettierignore
ericapisani: Good catch, I missed that when resolving the merge

site/src/_ui.mjs
ericapisani: My mistake - I had tried to convert the rest of the `.mjs` files earlier on to `.js` but ran into issues. Instead of checking each file to see what the extension was before I had bulk converted everything to `.mjs` (not remembering that some were `.js` before I started making these changes)

sarahetter (approved):

PR #6180, opened 2023-11-16
feat: load command's actions async to improve performance

sarahetter:

#### Summary

Thanks to @rjbeers investigation, we've found that loading in individual commands' action functions asynchronously leads to faster load time, as we aren't loading in every single package on each command run. I've done this for each command we have in the CLI.

I've also added Typescript types for commands and their actions in this PR.

|  Before | After  |
|---|---|
| **netlify -v** - around 2 seconds netlify -v on main |  **netlify -v** - just over 0.5 secondsnetlify -v on branch |
|  **netlify status** - around 4 seconds before | **netlify status** - around 2.75 seconds after |

---

      passes our tests.

mothman-alien


src/commands/addons/addons.mts
```diff
@@ -1,32 +1,72 @@
...
+  )
+  // allow for any flags. Handy for variadic configuration options
+  .allowUnknownOption(true)
+  .action(async (addonName: string, options: OptionValues, command: BaseCommand) => {
+    const { addonsCreate } = await import('./addons-create.mjs')
+    await addonsCreate(addonName, options, command)
+  })
+
+  program
+  .command('addons:delete', { hidden: true })
```
eduardoboucas, line 48: The indentation looks a bit off here. I don't think we've changed the Prettier settings to run on `.mts` files?

sarahetter: Ah yeah @ericapisani is this in the plans with what you're doing now?

eduardoboucas: I believe we need to add the new extensions here: https://github.com/netlify/cli/blob/7ab484bb94e0eed680a6e17e1eb34e5aaea9fe8d/package.json#L79.

sarahetter: Added config in https://github.com/netlify/cli/pull/6187, and ran format. Should be prettier now :D

eduardoboucas (approved):
🚀

ericapisani (approved):

pieh:
@sarahetter Is it possible that this change could cause some issues with lambda execution timeouts? Tests in https://github.com/netlify/netlify-plugin-gatsby started failing because of errors like:

```
Function __api has returned an error: Task timed out after 10.00 seconds
    at new TimeoutError (/opt/hostedtoolcache/node/20.9.0/x64/lib/node_modules/netlify-cli/node_modules/lambda-local/build/lib/utils.js:119:28)
    at Context.<anonymous> (/opt/hostedtoolcache/node/20.9.0/x64/lib/node_modules/netlify-cli/node_modules/lambda-local/build/lib/context.js:110:19)
    at listOnTimeout (node:internal/timers:573:17)
    at processTimers (node:internal/timers:514:7)
```

Downgrading CLI to `17.6.0` "fixes" those timeouts.

Using 17.7.0 test run example - https://github.com/netlify/netlify-plugin-gatsby/actions/runs/6983007552/job/19003322129?pr=722 (above snippet copied from it) - it seems to happen every time now with that version of CLI
Using 17.6.0 test run example is fine https://github.com/netlify/netlify-plugin-gatsby/actions/runs/6982743678/job/19002516599?pr=721

I don't quite know if it's possible this would be the case as this is pretty hefty change in this PR - but would async loading part possibly be part of lambda execution time now, meaning that part of 10s lambda execution timeout would be taken by async loading?

sarahetter:
@pieh I just took a look through the `netlify-plugin-gatsby` tests and it looks like they're using `netlify dev` - which didn't actually get the async treatment in this PR because I had failing tests, so it was kept as is. So for now, I'm just not sure.

@eduardoboucas I think I'll need your help trying to debug this. I'm not familiar with the dev/functions handling code within the CLI, and can't guess as to what's causing these timeouts

PR #6181, opened 2023-11-16
feat: speed up `netlify status` by removing extra api call

sarahetter:

#### Summary

Removes a second run of `getSite` from the `netlify status` command as we already have that data when `base-command` runs first.

This makes this command a bit faster at the cost of losing errors based on error code (401, 404) but I think this a reasonable tradeoff.

|  Before |  After  |
|---|---|
| Between 3.6 and 5.1 seconds Screenshot 2023-11-16 at 5 22 42 PM |  Between 3.2 and 3.7 seconds Screenshot 2023-11-16 at 5 23 39 PM |

---

      passes our tests.


eduardoboucas (approved):
Nice!

PR #6183, opened 2023-11-17
fix: also run blobs on `ntl serve`

Skn0tt:
Noticed that we're only running blobs for `ntl dev`, but not for `ntl serve`. That's probably not right! This PR also starts the blob server during `ntl serve`.


src/commands/serve/serve.mts
```diff
@@ -71,7 +71,7 @@ const serve = async (options, command) => {
   // Netlify Build are loaded.
   await getInternalFunctionsDir({ base: site.root, ensureExists: true })
 
-  let settings = /** @type {import('../../utils/types.js').ServerSettings} */ ({})
+  let settings = /** @type {import('../../utils/types.js').ServerSettings} */ {}
```
eduardoboucas, line 74: nit: Since you're touching this line, could you use the actual type instead of the JSDoc now that this is TypeScript?

eduardoboucas (dismissed):

src/commands/serve/serve.mts
Skn0tt: done in https://github.com/netlify/cli/pull/6183/commits/9eea77d37cb74b878be984a7b82a89cac0693cd6

src/commands/serve/serve.mts
```diff
@@ -71,7 +72,8 @@ const serve = async (options, command) => {
   // Netlify Build are loaded.
   await getInternalFunctionsDir({ base: site.root, ensureExists: true })
 
-  let settings = /** @type {import('../../utils/types.js').ServerSettings} */ ({})
+  // @ts-expect-error TS(2571) FIXME: Type '{}' is not assignable to type 'ServerSettings'.
+  let settings: ServerSettings = {}
```
eduardoboucas, line 76: We can get rid of the `ts-expect-error` ugliness if you change L84 to `return exit(1)` (GitHub won't let me add a suggestion for that line).

```suggestion
  let settings: ServerSettings
```

Skn0tt: smart! done in https://github.com/netlify/cli/pull/6183/commits/2719485ae063f5fb2813e1f3c044f5350624a501

eduardoboucas (approved):

eduardoboucas (approved):

PR #6212, opened 2023-11-27
fix: don't mistake paths for protocols

Skn0tt:
Found another small bug in our routing logic for v2 functions. When requesting a path like `//foo`, our routing logic was mistaking the `//` to be a protocol separator, and inferred `foo` to be the hostname instead of the path.

This PR fixes that!


src/lib/functions/registry.ts
```diff
@@ -290,15 +290,11 @@ export class FunctionsRegistry {
...
-   * @param {string} url
+   * @param {string} urlPath
    * @param {string} method
    */
-  // @ts-expect-error TS(7006) FIXME: Parameter 'url' implicitly has an 'any' type.
-  async getFunctionForURLPath(url, method) {
-    // We're constructing a URL object just so that we can extract the path from
-    // the incoming URL. It doesn't really matter that we don't have the actual
-    // local URL with the correct port.
-    const urlPath = new URL(url, 'http://localhost').pathname
```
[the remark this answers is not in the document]
eduardoboucas: There is nothing ensuring that `urlPath` is indeed a URL path and not a full URL, right? I have a feeling that this is precisely why we had this logic here.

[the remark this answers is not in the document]
Skn0tt: I've played around with it a bit more, and it seems like we had this in place to remove query parameters from the path. https://github.com/netlify/cli/pull/6212/commits/146cb80d7e270c44918dcc77f48e564c8f5b40e6 implements a fix that still does that

src/lib/functions/registry.ts
```diff
@@ -297,8 +297,8 @@ export class FunctionsRegistry {
   async getFunctionForURLPath(url, method) {
     // We're constructing a URL object just so that we can extract the path from
     // the incoming URL. It doesn't really matter that we don't have the actual
-    // local URL with the correct port.
-    const urlPath = new URL(url, 'http://localhost').pathname
+    // local URL with the correct port.)
```
eduardoboucas, line 300:
```suggestion
    // local URL with the correct port.
```

src/lib/functions/registry.ts
```diff
@@ -297,8 +297,8 @@ export class FunctionsRegistry {
   async getFunctionForURLPath(url, method) {
```
eduardoboucas, line 297: The parameter should probably be named to `urlPath`, as it will now break if it's a fully qualified URL? Alternatively, we can keep it as `url` and check whether if it's a full URL or a URL path, adjusting accordingly.

Skn0tt: done in https://github.com/netlify/cli/pull/6212/commits/f2056907cc0e01f51127e6c779735b7916a0df97

src/lib/functions/registry.ts
```diff
@@ -51,28 +77,18 @@ export class FunctionsRegistry {
     settings,
     // @ts-expect-error TS(7031) FIXME: Binding element 'timeouts' implicitly has an 'any'... Remove this comment to see the full error message
     timeouts,
-  }) {
+  }: { projecRoot: string; debug?: boolean; isConnected?: boolean; blobsContext: BlobsContext } & object) {
```
eduardoboucas, line 80:
```suggestion
  }: { projectRoot: string; debug?: boolean; isConnected?: boolean; blobsContext: BlobsContext } & object) {
```

eduardoboucas (approved):

PR #6213, opened 2023-11-27
fix: prevent image cdn from being shadowed by v2 functions

Skn0tt:
Image CDN shouldn't be shadowable by V2 Functions.

Depends on https://github.com/netlify/cli/pull/6212.

Haven't checked how prod behaves on this, will do that now.


Skn0tt:
only https://github.com/netlify/cli/pull/6213/commits/50b8f25dbed7af1315036d7f2e99868d31914438 is part of this PR, the other commits are from  https://github.com/netlify/cli/pull/6212 - but they contain fixes that the tests in this PR need as well

Skn0tt:
> Haven't checked how prod behaves on this, will do that now.

This is, indeed, how prod behaves!

eduardoboucas (approved):

PR #6214, opened 2023-11-27
fix: don't show warning when there are no ef

Skn0tt:
We're currently showing this warning when there's dependency graph:

https://github.com/netlify/cli/blob/dbd6066a69bc273846c640b8e5ca090695bde3c9/src/lib/edge-functions/registry.ts#L411

One case where there's no dependency graph is when there's no functions:

https://github.com/netlify/cli/blob/dbd6066a69bc273846c640b8e5ca090695bde3c9/src/lib/edge-functions/registry.ts#L467-L473

We shouldn't be showing the warning in that case. This PR adds an early-return that does that.

### DIFF
--- a/src/lib/edge-functions/registry.ts
+++ b/src/lib/edge-functions/registry.ts
@@ -126,10 +126,6 @@ export class EdgeFunctionsRegistry {
         throw new Error('Build error')
       }
 
-      if (this.functions.length === 0) {
-        return { warnings }
-      }
-
       this.buildError = null
 
       // We use one index to loop over both internal and user function, because we know that this.#functions has internalFunctions first.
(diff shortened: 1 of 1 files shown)


src/lib/edge-functions/registry.ts
```diff
@@ -126,6 +126,10 @@ export class EdgeFunctionsRegistry {
         throw new Error('Build error')
       }
 
+      if (this.functions.length === 0) {
```
eduardoboucas, line 129: If we don't want to print the warning when there are no functions (which I agree we shouldn't do), should we have a check inside `processGraph`, next to the code that prints the warning? That way we'll handle a correct behaviour regardless of where we call `processGraph` from, otherwise it's up to the caller to ensure that.

eduardoboucas (approved):

PR #6221, opened 2023-11-28
fix: image CDN asset url shouldn't have double slash

Skn0tt:
Found one more bug in the Image CDN implementation! The logic to generate the Asset URL had one `/` too much in it, resulting in it matching the wrong resource.

This PR also converts two test files into TypeScript.

Best-reviewed commit by commit - the typescript conversions are pretty big diffs.


kathmbeck (approved):

kathmbeck:
Good catch!

PR #6225, opened 2023-11-29
fix: inject env from envelope for `ntl deploy --build`

Skn0tt:
Addresses https://linear.app/netlify/issue/COM-207/inject-env-vars-from-envelope-on-ntl-deploy-build.

We were fetching env vars from Envelope for `ntl build`, but not for `ntl deploy --build`. This PR also adds the env var injection logic to `ntl deploy --build`.


Skn0tt:
ha, I should've tested this out locally before working on the PR. it looks like it was indeed already working before? will investigate, turning this to draft

Skn0tt:
oh, we're doing that here! https://github.com/netlify/cli/blob/55ff33eaf91fcbfd4a49f9fa5937483ee77b790b/src/commands/deploy/deploy.ts#L629-L639

forget about this PR, then :D

PR #6226, opened 2023-11-29
feat: allow frameworks to detect version of CLI that's being used

Skn0tt:
Addresses https://linear.app/netlify/issue/COM-189/give-frameworks-a-way-to-detect-version-of-netlify-build-cli-thats.

This PR gives frameworks + framework adapters a way to detect the version of the Netlify CLI that it's called through. The Astro adapter, for example, can use this to detect if the CLI is at least on 17.8.x - else it can choose to fail the local dev, because not all features that it depends on are present on that version. It can then prompt the user to update their CLI.


src/commands/base-command.ts
```diff
@@ -606,7 +609,8 @@ export default class BaseCommand extends Command {
       token,
       ...apiUrlOpts,
     })
-    const { buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
+    const { buildDir, config, configPath, env, repositoryRoot, siteInfo } = cachedConfig
+    env.NETLIFY_CLI_VERSION = { sources: ['internal'], value: netlifyCliVersion }
```
Skn0tt, line 613: This is the main functional change in this PR.

I don't have strong feelings around the name of the var - maybe `NETLIFY_CLI` would be better? We already have `NETLIFY_LOCAL` that we set to `true`. We could also reuse that, but I could imagine that some consumers are depending on its value being `"true"`.

sarahetter: I think this var name is just fine - clear is best imo (even if it's wordy :D )

sarahetter (dismissed):

Skn0tt:
Oops, looks like this didn't auto-merge yet! Resolved the conflicts, would love another stamp :)

sarahetter (dismissed):

Skn0tt:
And now the tests are failing :/ I swear they worked locally!

Skn0tt:
Whoohoo, tests are passing! Would appreciate another approval in the new year :D

src/commands/base-command.ts
```diff
@@ -26,6 +26,7 @@ import {
   exit,
   getToken,
   log,
+  netlifyCliVersion,
```
eduardoboucas, line 29: nit: This name feels a bit redundant and overly verbose to me, since everything here pertains to the Netlify CLI and there's no need to make that explicit with a prefix. Could we use something like `version`?

src/commands/env/env-list.ts
```diff
@@ -61,8 +61,10 @@ export const envList = async (options: OptionValues, command: BaseCommand) => {
 
   // filter out general sources
   environment = Object.fromEntries(
-    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
-    Object.entries(environment).filter(([, variable]) => variable.sources[0] !== 'general'),
+    Object.entries(environment).filter(
+      // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
```
eduardoboucas, line 65: What would it take to get rid of this?

eduardoboucas (approved):

PR #6242, opened 2023-12-04
fix: clear `dist` directory before Dev for Nuxt

Skn0tt:
Part of https://linear.app/netlify/issue/COM-184/nuxt-netlify-dev-doesnt-work. Running `ntl build` and then `ntl dev` on a Nuxt project leads to Dev picking up `dist/_redirects` and `dist/_headers`, even though the Nitro dev server will handle all of this on its own. This is breaking local dev currently.

In https://github.com/unjs/nitro/issues/1970, the Nitro team suggests we add `rm -rf` to the dev command. I tried doing that in https://github.com/netlify/build/pull/5429, but since we don't spawn a full shell for the dev process, we can't use command concatenation like `&&`.

This PR contains an alternative fix, where we manually remove the `dist` directory just for Nuxt projects.


Skn0tt:
I couldn't find an easy way of writing a test for this :/

src/utils/framework-server.ts
```diff
@@ -38,12 +41,17 @@ export const startFrameworkServer = async function ({ cwd, settings }) {
     text: `Waiting for framework port ${settings.frameworkPort}. This can be configured using the 'targetPort' property in the netlify.toml`,
   })
 
+  if (settings.clearPublishDirectory && settings.dist) {
+    await rm(settings.dist, { recursive: true, force: true })
```
Skn0tt, line 45: this is the main change, all the rest is typescript conversion stuff

lukasholzer: haha I feel you just opened a just change types PR before performing the actual change

lukasholzer (approved):

PR #6252, opened 2023-12-06
fix: streaming response with content-length breaks local dev

Skn0tt:
Addresses https://linear.app/netlify/issue/COM-222/local-dev-cli-tries-to-compute-etag-on-streaming-response-breaks


Skn0tt:
this is related to some behaviour in the fetch spec, that we won't be able to change anytime soon. closing!

PR #6257, opened 2023-12-07
feat: detect internal functions created after dev server already started

Skn0tt:
in https://github.com/netlify/cli/pull/6244 we've merged a change from build that clears the `.netlify/functions-internal` and `.netlify/edge-functions` folders on dev startup. We need to update some of our code & tests to handle that, and this PR does it!

The main thing is that right now, we're checking if a `.netlify/functions-internal` directory exists once on startup. If it doesn't, we don't watch it, and we don't get notified about functions being added to it.

Instead, we should *always* watch the directory, so any functions added after startup are detected correctly.

This PR fixes all skipped tests one-by-one. It's a rather big refactor, best reviewed commit by commit!


Skn0tt:
looking at that test failure 🤔

Skn0tt:
I can't reproduce this type check failure locally 🤔

Skn0tt:
> I can't reproduce this type check failure locally 🤔

Interesting - the CI is pretending that this PR is already merged into main, and running the tests on that state. I could replicate it locally after merging in main. Should be fixed now!

Skn0tt:
Hmm, it looks like there's a test that times out on windows. Looks like i'll need to fire up my VM!

Skn0tt:
Well well, I can't reproduce it locally on windows. I've restarted CI, hopefully it's "just" flaky?

Skn0tt:
Nope, can't replicate locally 😢

Skn0tt:
While debugging via SSH, I was able to narrow it down to this test, and I have a hunch that the test is fixed by https://github.com/netlify/cli/pull/6257/commits/5f0530030cf6dab1b81f1c01888af426bac1ed72. The problem was that we weren't awaiting the expectation, so the exception was thrown outside of the test execution. As to why this only occurs now, and only on Windows, I don't have a convincing answer - maybe the changes made the stars align subtly so that the event loop now notices the uncaught rejection.

Skn0tt:
Welp, scratch that - the test still fails. But at least it fails equally on all operating systems now. Still can't reproduce locally, though :/

Skn0tt:
Tests pass, Gotcha!

src/commands/types.d.ts
```diff
@@ -7,7 +7,7 @@ import StateConfig from '../utils/state-config.js'
 type $TSFixMe = any;
 
 export type NetlifySite = {
-  root?: string
+  root: string
```
lukasholzer, line 10: I don't think the root is always defined; I believe we need to cope with scenarios where it's undefined. I would propose doing a type guard on the usage instead of making the type required all the time

src/lib/edge-functions/registry.ts
```diff
@@ -33,18 +38,19 @@ interface EdgeFunctionsRegistryOptions {
...
   servePath: string
+  importMapFromTOML?: string
 }
 
 export class EdgeFunctionsRegistry {
   private buildError: Error | null = null
   private bundler: typeof import('@netlify/edge-bundler')
   private configPath: string
-  private declarationsFromDeployConfig: Declaration[]
+  public importMapFromDeployConfig?: string
```
lukasholzer, line 51: NIT: the public keyword is not needed.

```suggestion
 importMapFromDeployConfig?: string
```

src/lib/functions/registry.ts
```diff
@@ -453,34 +447,27 @@ export class FunctionsRegistry {
   /**
    * A proxy to zip-it-and-ship-it's `listFunctions` method. It exists just so
    * that we can mock it in tests.
-   * @param  {Parameters<listFunctions>} args
-   * @returns
    */
-  // @ts-expect-error TS(7019) FIXME: Rest parameter 'args' implicitly has an 'any[]' ty... Remove this comment to see the full error message
   // eslint-disable-next-line class-methods-use-this
```
lukasholzer, line 451: Can we kill this lint rule? 🗡️

src/lib/functions/registry.ts
```diff
@@ -322,6 +320,7 @@ export class FunctionsRegistry {
     if (event === 'buildError') {
       log(
         `${NETLIFYDEVERR} ${chalk.red('Failed to load')} function ${chalk.yellow(func?.displayName)}: ${
+          // @ts-expect-error buildError is unknown
```
lukasholzer, line 323: NIT: wouldn't it make more sense to do a type guard here?

```typescript
if (func?.buildError && typeof func.buildError === 'object' && 'message' in func.buildError) {
 // ...
}

lukasholzer (commented):
Looks great just a small comment about a type where I think it should stay optional

src/lib/edge-functions/registry.ts
Skn0tt: Agree it's not needed, but I prefer to have this here - makes it obvious that it's accessed from other places.

src/lib/functions/registry.ts
Skn0tt: You mean the `class-methods-use-this` rule? Yeah, maybe we should 😅

src/lib/functions/registry.ts
Skn0tt: Happy to do that. Wanted to keep the diff small and easy to reason about, but agree a type guard is better here.

src/commands/types.d.ts
Skn0tt: addressed in https://github.com/netlify/cli/pull/6257/commits/c4136e5b39f746e472ed9ae1c95193a871c1dfd2

src/lib/functions/registry.ts
Skn0tt: done in https://github.com/netlify/cli/pull/6257/commits/4b7cbb6585444ff788629004663bb59b13297a6e

lukasholzer (approved):
LGTM! 🚀

PR #6261, opened 2023-12-12
chore: enhance types for base-command.ts

lukasholzer:

#### Summary

To not have my change and the types refactoring in one PR this is just the PR for the types refactorings.

The actual change will come in a follow-up.

---

      passes our tests.

### DIFF
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -38,6 +38,13 @@ import openBrowser from '../utils/open-browser.js'
 import StateConfig from '../utils/state-config.js'
 import { identify, reportError, track } from '../utils/telemetry/index.js'
 
+import { type NetlifyOptions } from './types.js'
+
+type Analytics = {
+  startTime: bigint
+  payload?: Record<string, unknown>
+}
+
 // load the autocomplete plugin
 inquirer.registerPrompt('autocomplete', inquirerAutocompletePrompt)
 /** Netlify CLI client id. Lives in bot@netlify.com */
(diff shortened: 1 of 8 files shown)


src/commands/base-command.ts
```diff
@@ -229,38 +203,28 @@ export default class BaseCommand extends Command {
...
-      await this.init(actionCommand)
+      await this.init(actionCommand as BaseCommand)
       debug(`${name}:preAction`)('end')
     })
   }
 
-  /** @private */
-  noBaseOptions = false
-
+  #noBaseOptions = false
```
eduardoboucas, line 211: Can we not use `private noBaseOptions`?

src/commands/base-command.ts
```diff
@@ -613,7 +548,8 @@ export default class BaseCommand extends Command {
       certificateFile: flags.httpProxyCertificateFilename,
     })
     const apiOpts = { ...apiUrlOpts, agent }
-    const api = new NetlifyAPI(token || '', apiOpts)
+    // TODO: remove typecast once we have proper types for the API
```
eduardoboucas, line 551: 🙏🏻

src/utils/init/utils.ts
```diff
@@ -1,53 +1,57 @@
...
+ * there our build will automatically install the latest without
+ * user management of the versioning.
+ * @param pluginsInstalled
+ * @param pluginsRecommended
  * @returns
  */
-export const getPluginsToAutoInstall = (pluginsInstalled = [], pluginsRecommended = []) =>
-  pluginsRecommended.reduce(
+export const getPluginsToAutoInstall = (
+  command: BaseCommand,
```
Skn0tt, line 32: I don't see this parameter used anywhere - am I missing something?

eduardoboucas (approved):

src/commands/base-command.ts
lukasholzer: TypeScript private is not a JavaScript private this is truly private and not accessible after transpiling.

TBH I would prefer this notation over using they private keyword as this is just dropped on transpiling

src/utils/init/utils.ts
lukasholzer: Nope it is a prerequisite for the follow up PR just forgot to remove it from this one. In the follow up I need to access the feature flags from the base command.

src/commands/base-command.ts
eduardoboucas: Do we care if it's truly private or not, though? I think we should focus on code readability and type safety, and having to prefix every private property with `#` feels a bit awkward and noisy to me, personally.

PR #6276, opened 2023-12-28
fix(dev): reloading edge functions for child dependencies

JGAntunes:

#### Summary

Fixes COM-209 - https://linear.app/netlify/issue/COM-209/reloading-issues-with-edge-function-files-when-using-cli-netlify-dev.  The way we currently process the dependency graph of edge function is limited to the direct dependencies of it. This PR changes it to perform a depth first search and build a full depedency list for a given function.

Keeping it as draft as I would love to add a test for it but unsure what would be the best way to go about it? 🤔 Any ideas? Maybe @Skn0tt?

---

      passes our tests.


Skn0tt:
The rough problem seems to be that we were only checking one level-deep dependencies. I haven't fully grasped the change you're proposing here, but traversing the full module graph seems to be the right solution.

As for adding a test, I'd add a case similar to https://github.com/netlify/cli/blob/3379f40ff8657cdbf03cc9be7cc08f2dc4e1ce49/tests/integration/commands/dev/edge-functions.test.ts#L135-L152, where a fixture with a nested dependency is created and then updated.

Skn0tt:
I've added a test, refactored the code to make it a bit more concise, and added a couple other small refactorings. @JGAntunes would love your review!

JGAntunes (commented):
Can't approve because I kicked of the PR 😅 but LGTM

Skn0tt (approved):

PR #6279, opened 2024-01-02
fix: support Blobs in unlinked sites

eduardoboucas:
#### Summary

We want to support Blobs in Netlify Dev even if the site is not linked — this was supposed to be addressed by https://github.com/netlify/cli/pull/6176, but it didn't work properly (I've added more context in a comment). This PR fixes that.

Closes https://linear.app/netlify/issue/COM-211/blobs-possibly-not-working-in-framework-generated-functions.

### DIFF
--- a/src/lib/blobs/blobs.ts
+++ b/src/lib/blobs/blobs.ts
@@ -28,7 +28,7 @@ const printLocalBlobsNotice = () => {
 }
 
 const startBlobsServer = async (debug: boolean, projectRoot: string, token: string) => {
-  const directory = path.resolve(projectRoot, getPathInProject(['blobs-serves']))
+  const directory = path.resolve(projectRoot, getPathInProject(['blobs-serve']))
   const server = new BlobsServer({
     debug,
     directory,
--- a/src/lib/functions/server.ts
+++ b/src/lib/functions/server.ts
@@ -157,7 +157,7 @@ export const createHandler = function (options) {
       'client-ip': [remoteAddress],
       'x-nf-client-connection-ip': [remoteAddress],
       'x-nf-account-id': [options.accountId],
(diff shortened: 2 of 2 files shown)


src/lib/functions/server.ts
```diff
@@ -157,7 +157,7 @@ export const createHandler = function (options) {
       'client-ip': [remoteAddress],
       'x-nf-client-connection-ip': [remoteAddress],
       'x-nf-account-id': [options.accountId],
-      'x-nf-site-id': [options?.siteInfo?.id] ?? 'unlinked',
```
eduardoboucas, line 160: This was the problem. `[options?.siteInfo?.id]` is never falsy, even when `options?.siteInfo?.id` is `undefined`, so the fallback string never kicked in.

src/lib/blobs/blobs.ts
```diff
@@ -28,7 +28,7 @@ const printLocalBlobsNotice = () => {
 }
 
 const startBlobsServer = async (debug: boolean, projectRoot: string, token: string) => {
-  const directory = path.resolve(projectRoot, getPathInProject(['blobs-serves']))
+  const directory = path.resolve(projectRoot, getPathInProject(['blobs-serve']))
```
eduardoboucas, line 31: Unrelated to the fix, but `blobs-serves` was a typo. It should be `blobs-serve`, in line with `functions-serve` and `edge-functions-serve`.

Skn0tt (approved):

src/lib/blobs/blobs.ts
ascorbic: I forgot to mention this! I saw the directory and guessed it was a typo.

PR #6288, opened 2024-01-08
refactor: type cachedConfig.env

Skn0tt:
Follow-up to https://github.com/netlify/cli/pull/6226#discussion_r1444364865. Adds typescript types to `cachedConfig.env`!

### DIFF
--- a/src/commands/types.d.ts
+++ b/src/commands/types.d.ts
@@ -24,7 +24,10 @@ type PatchedConfig = NetlifyTOML & {
   }
 }
 
-export type EnvironmentVariables = Record<string, { sources: string[], value: string; scopes?: string[] }>
+type EnvironmentVariableScope = 'builds' | 'functions' | 'runtime' | 'post_processing'
+type EnvironmentVariableSource = 'account' | 'addons' | 'configFile' | 'general' |  'internal' | 'ui'
+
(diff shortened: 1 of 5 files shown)


src/commands/types.d.ts
```diff
@@ -24,6 +24,8 @@ type PatchedConfig = NetlifyTOML & {
   }
 }
 
+export type EnvironmentVariables = Record<string, { sources: string[], value: string; scopes?: string[] }>
```
eduardoboucas, line 27: Might be nice to get more specific about the types here. How do you feel about something like:

```suggestion
type EnvironmentVariableScope = 'builds' | 'functions' | 'runtime' | 'post_processing'
type EnvironmentVariableSource = 'account' | 'addons' | 'configFile' | 'general' |  'internal' | 'ui'

export type EnvironmentVariables = Record<string, { sources: EnvironmentVariableSource[], value: string; scopes?: EnvironmentVariableScope[] }>
```

eduardoboucas (commented):
Nice! One small comment.

src/commands/types.d.ts
Skn0tt: Looks good!

eduardoboucas (approved):
🚀

PR #6289, opened 2024-01-08
test: separate windows tests into another job

ericapisani:
this will allow us to up the timeout time for that alone without affecting the ubuntu and macos tests

#### Summary

Fixes CT-133

---

      passes our tests.


.github/workflows/e2e-tests.yml
```diff
@@ -10,19 +10,64 @@ on:
...
+      - name: Cache verdaccio storage
+        uses: actions/cache@v3
+        with:
+          path: ./.verdaccio-storage
+          key: verdaccio-e2e-cli-${{ hashFiles('./package-lock.json') }}
+      - name: Install dependencies
+        run: npm ci --no-audit
+      - name: Install pnpm
+        run: npm install -g pnpm
+      - run: npm run e2e
```
lemusthelroy, line 45: we're installing pnpm but then using npm to run stuff - just checking this is intended

ericapisani: Oh good catch - I just copied and pasted this from the other job so the same thing exists there, will fix

lemusthelroy (approved):

PR #6293, opened 2024-01-09
feat: implement clack DO NOT MERGE

khendrikse:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

This is a big branch, containing all the work for https://linear.app/netlify/issue/CT-101/cli-design-system. 

### BEFORE MERGING, REMOVE THE GITHUB ACTIONS LOGIC THAT RUNS TESTS ON PR'S TO THIS BRANCH.

We're working on aligning the styles in the CLI. This PR contains all the work that has been done for that. But due to the enormity of the review work that would otherwise be necessary, smaller PR's will be merged into this bigger PR after being reviewed by the team, keeping the workload manageable. 

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #6295, opened 2024-01-10
Cp 476/convert api command

khendrikse:

#### Summary

This adds some changes to converting the api command into using the new NetlifyLog. It is part of a bigger piece of work that can be seen in https://github.com/netlify/cli/pull/6293. This is also the reason why this PR is merged into the branch [CP-101/design-system-clack-implementation](https://github.com/netlify/cli/tree/CP-101/design-system-clack-implementation). It's a way in which I'm trying to keep possible failing tests, but also the review work to a minimum amount of work for those who are reviewing.

## Before

CleanShot 2024-01-10 at 10 28 07

## After
CleanShot 2024-01-10 at 12 28 19

---

      passes our tests.


samkap:
@khendrikse I understand. When there isn't a message will it end with the last line of the table above it? I think that is fine.

khendrikse:
> @khendrikse I understand. When there isn't a message will it end with the last line of the table above it? I think that is fine.

Yeah that's the problem @samkap, right now we don't have easy logic that will allow that. 

We could default to 'Done' or 'Finished' if someone does not pass a message?

ericapisani (approved):

PR #6299, opened 2024-01-10
fix: upload file blobs before serving by providing build with blobs context

Skn0tt:
Resolves https://linear.app/netlify/issue/COM-282/fix-blobs-in-netlifyblobsdeploy-on-ntl-serve.

Right now, `ntl serve` fails on sites with files in `.netlify/blobs/deploy`, because it tries to upload the files to `api.netlify.com`. But because it doesn't pass a valid deploy id, that fails.

The fix is to populate the blobs context, so that the client writes to the local blobs server instead.

Keeping as draft until I have a test for this.


src/lib/functions/server.ts
```diff
@@ -157,7 +158,7 @@ export const createHandler = function (options) {
       'client-ip': [remoteAddress],
       'x-nf-client-connection-ip': [remoteAddress],
       'x-nf-account-id': [options.accountId],
-      'x-nf-site-id': [options?.siteInfo?.id ?? 'unlinked'],
+      'x-nf-site-id': [options?.siteInfo?.id ?? UNLINKED_SITE_MOCK_ID],
```
Skn0tt, line 161: It's important we use the same ID here as in the blobs context, because the blobs server uses `siteID` as part of its internal keys.

eduardoboucas (approved):
LGTM

PR #6323, opened 2024-01-16
feat: convert status command to clack

lemusthelroy:

#### Summary

Converts status command to clack

Before:
image

After:
image

---

      passes our tests.

### DIFF
--- a/src/commands/status/status.ts
+++ b/src/commands/status/status.ts
@@ -2,9 +2,10 @@ import clean from 'clean-deep'
 import { OptionValues } from 'commander'
 import prettyjson from 'prettyjson'
 
-import { chalk, error, exit, getToken, log, logJson, warn } from '../../utils/command-helpers.js'
+import { chalk, getToken, logJson } from '../../utils/command-helpers.js'
 import BaseCommand from '../base-command.js'
 import { NetlifyLog, intro, outro } from '../../utils/styles/index.js'
+import { exit } from 'process'
 
 export const status = async (options: OptionValues, command: BaseCommand) => {
   intro('status')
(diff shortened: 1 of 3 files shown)


src/commands/status/status.ts
```diff
@@ -4,24 +4,22 @@ import prettyjson from 'prettyjson'
...
   const { api, globalConfig, site, siteInfo } = command.netlify
   const current = globalConfig.get('userId')
   const [accessToken] = await getToken()
 
   if (!accessToken) {
-    log(`Not logged in. Please log in to see site status.`)
-    log()
-    log('Login with "netlify login" command')
-    exit()
+    NetlifyLog.error(`Not logged in. Please log in to see site status.`)
```
khendrikse, line 16: For this one, you might want to add a second parameter with `{exit: false}` so that line 16 doesn't immediately exit the process. Otherwise line 17 and 18 won't be shown. OR you can change it into a `NetlifyLog.warn` because it is more informative than actually going wrong :) 

I also think we don't need the additional empty one on line 17. I've so far removed all the empty `log()` ones I've seen.

Then on line 18 you can just use `NetlifyLog.message()` and then after that you could do `outro()`.

src/commands/status/status.ts
```diff
@@ -93,5 +91,6 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
       'Site Id': chalk.yellowBright(siteInfo.id),
     }),
   )
-  log()
+  NetlifyLog.message()
```
khendrikse, line 94: I'd just remove this :)

khendrikse (changes requested):
Some small feedback!

src/commands/status/status.ts
lemusthelroy: Awesome, I needed this thank you

lemusthelroy: image

lemusthelroy: done

PR #6324, opened 2024-01-17
chore: convert log command

khendrikse:

#### Summary

Fixes: https://linear.app/netlify/issue/CT-479/convert-logs-command-to-clack

This converts the `logs` command into using the new NetlifyLog. It is part of a bigger piece of work that can be seen in https://github.com/netlify/cli/pull/6293. This is also the reason why this PR is merged into the branch [CP-101/design-system-clack-implementation](https://github.com/netlify/cli/tree/CP-101/design-system-clack-implementation). It's a way in which I'm trying to keep possible failing tests, but also the review work to a minimum amount of work for those who are reviewing.
CleanShot 2024-01-17 at 16 39 44

https://github.com/netlify/cli/pull/6325 was also merged into this branch which converts the open command to clack

---

      passes our tests.

### DIFF
--- a/src/commands/logs/functions.ts
+++ b/src/commands/logs/functions.ts
@@ -100,7 +100,6 @@ export const logsFunction = async (functionName: string | undefined, options: Op
   })
 
   process.on('SIGINT', () => {
-    outro({ message: 'Closing connection' })
-    process.exit()
+    outro({ message: 'Closing connection', exit: true })
   })
 }
--- a/src/commands/logs/functions.ts
+++ b/src/commands/logs/functions.ts
@@ -1,7 +1,9 @@
-import { OptionValues } from 'commander'
-import inquirer from 'inquirer'
+import process from 'process'
 
-import { chalk, log } from '../../utils/command-helpers.js'
+import { OptionValues } from 'commander'
+
+import { chalk } from '../../utils/command-helpers.js'
+import { NetlifyLog, intro, outro, select } from '../../utils/styles/index.js'
 import { getWebSocket } from '../../utils/websockets/index.js'
 import type BaseCommand from '../base-command.js'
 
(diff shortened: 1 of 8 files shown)


lemusthelroy:
Just a note to say this branch now includes code to convert the open command to clack

src/commands/logs/functions.ts
```diff
@@ -28,39 +30,40 @@ function getLog(logData: { level: string; message: string }) {
...
   }
 
   const levelsToPrint = options.level || LOG_LEVELS_LIST
 
   const { functions = [] } = await client.searchSiteFunctions({ siteId })
 
   if (functions.length === 0) {
-    log(`No functions found for the site`)
-    return
+    NetlifyLog.error(`No functions found for the site`)
```
lemusthelroy, line 47: Think we need to exit here

src/commands/logs/functions.ts
```diff
@@ -28,39 +30,40 @@ function getLog(logData: { level: string; message: string }) {
...
 
   let selectedFunction
   if (functionName) {
-    selectedFunction = functions.find((fn: any) => fn.n === functionName)
+    selectedFunction = functions.find((fn: { n: string }) => fn.n === functionName)
   } else {
-    const { result } = await inquirer.prompt({
-      name: 'result',
-      type: 'list',
+    const result = await select({
```
lemusthelroy, line 54: I've managed to narrow down the failing tests to this select prompt but I have no idea why that would cause the timeouts. Digging further...

lemusthelroy: Ah it's because we previously mocked the interaction for inquirer but we need to do the same here for the clack prompt

lemusthelroy (approved):

PR #6331, opened 2024-01-18
feat: expose site feature flags to EF registry

eduardoboucas:
#### Summary

Feature flags coming from the site info are not being surfaced to the edge functions registry. This PR fixes that.

### DIFF
--- a/src/lib/edge-functions/proxy.ts
+++ b/src/lib/edge-functions/proxy.ts
@@ -7,6 +7,7 @@ import * as bundler from '@netlify/edge-bundler'
 import getAvailablePort from 'get-port'
 
 import { NETLIFYDEVERR, chalk, error as printError } from '../../utils/command-helpers.js'
+import { getFeatureFlagsFromSiteInfo } from '../../utils/feature-flags.js'
 import { getGeoLocation } from '../geo-location.js'
 import { getPathInProject } from '../settings.js'
 import { startSpinner, stopSpinner } from '../spinner.js'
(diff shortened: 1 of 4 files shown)


src/lib/edge-functions/proxy.ts
```diff
@@ -136,11 +137,9 @@ export const initializeProxy = async ({
 }) => {
   const userFunctionsPath = config.build.edge_functions
   const isolatePort = await getAvailablePort()
-  const buildFeatureFlags = {
-    edge_functions_npm_modules: true,
-  }
   const runtimeFeatureFlags = ['edge_functions_bootstrap_failure_mode', 'edge_functions_bootstrap_populate_environment']
   const protocol = settings.https ? 'https' : 'http'
+  const buildFeatureFlags = { ...getFeatureFlagsFromSiteInfo(siteInfo), edge_functions_npm_modules: true }
```
eduardoboucas, line 142: This flag is not used anymore. We'll be able to get rid of this special case once we update `@netlify/edge-bundler`.

src/lib/edge-functions/registry.ts
```diff
@@ -80,10 +80,11 @@ function traverseLocalDependencies(
 }
 
 export class EdgeFunctionsRegistry {
+  public importMapFromDeployConfig?: string
```
Skn0tt, line 83: I don't see this referenced any where else in this diff. Assuming it's a type addition, no runtime effect.

Skn0tt (dismissed):

src/lib/edge-functions/registry.ts
eduardoboucas: I moved it from L86 to separate private properties from public ones.

src/lib/edge-functions/registry.ts
```diff
@@ -31,15 +32,14 @@ type RunIsolate = Awaited<ReturnType<typeof import('@netlify/edge-bundler').serv
 
 type ModuleJson = ModuleGraph['modules'][number]
 
-const featureFlags = { edge_functions_correct_order: true }
```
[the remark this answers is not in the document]
eduardoboucas: No. https://github.com/netlify/edge-bundler/blob/fd07443937e6b1d3db4fca6a425e581bc791e5df/node/feature_flags.ts#L1-L3

Skn0tt (approved):

PR #6346, opened 2024-01-24
chore: converts watch and init commands to clack

lemusthelroy:

#### Summary

Converts watch and init commands to clack. Also, adjusts styling for the outro of unlinking.

Watch, where you link to existing site:
image

Watch, where site already linked:
image

Watch, where site gets created:
image

Link:
image

Also contains guidance so we avoid having multiple intros and outros like this: 
image

      passes our tests.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -57,7 +57,7 @@ visual end for your command already. Outro must be called with an object:
 
 ### Child commands
 
-As some of our commands call other CLI commands, we will sometimes we want to conditionally call `outro` depending
+As some of our commands call other CLI commands, we will sometimes want to conditionally call `outro` depending
 on whether or not the command is a child command. To do this, we use add a property to the `OptionValues` object
 called `isChildCommand`. This property should be set to `true` when the command is a child command. This means that we can
 conditionally call `outro` like this:
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -25,7 +25,7 @@ whether or not the command is a child command. To do this, we use add a property
 `isChildCommand`. This property is should be set to `true` when the command is a child command. This means that we can
 conditionally call `intro` like this:
 
-```js
+```ts
 import { intro } from '../../utils/styles/index.js'
 
 export const basicCommand = (options: OptionValues) => {
@@ -60,7 +60,7 @@ on whether or not the command is a child command. To do this, we use add a prope
 called `isChildCommand`. This property should be set to `true` when the command is a child command. This means that we can
 conditionally call `outro` like this:
 
-```js
+```ts
 import { intro, outro } from '../../utils/styles/index.js'
 
 export const basicCommand = (options: OptionValues) => {
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -34,6 +34,8 @@ export const basicCommand = (options) => {
 }
 ```
 
+Doing this ensures the CLI groups the messages together correctly and ensures two `intro` messages are not shown.
+
 ## Finishing your command
 
 Each command should end by using the `outro` method to create a clear visual end for your command. You can import it
@@ -75,8 +77,7 @@ export const basicCommand = (options) => {
 }
 ```
 
-Doing this ensures the CLI groups the messages and does
-
+Doing this ensures the CLI groups the messages together correctly and ensures two `outro` messages are not shown.
 
 ## Other content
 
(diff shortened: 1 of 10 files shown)


lemusthelroy:
Only TODO left on this is to ensure the site creation path for the init route is converted to clack

STYLEGUIDE.md
```diff
@@ -18,6 +18,24 @@ export const basicCommand = () => {
 
 Pass the name of the command to the `intro` method.
 
+### Child commands
+
+As some of our commands call other CLI commands, we will sometimes want to conditionally call `intro` depending on
+whether or not the command is a child command. To do this, we use add a property to the `OptionValues` object called
+`isChildCommand`. This property is should be set to `true` when the command is a child command. This means that we can
```
ericapisani, line 25:
```suggestion
`isChildCommand`. This property should be set to `true` when the command is a child command. This means that we can
```

STYLEGUIDE.md
```diff
@@ -29,13 +47,37 @@ import { intro, outro } from '../../utils/styles/index.js'
 export const basicCommand = () => {
   intro('basic command')
   // do stuff
-  outro()
+  outro({ message: 'Your message', exit: true })
```
ericapisani, line 50: If `exit: true` were not passed, what would happen?

STYLEGUIDE.md
```diff
@@ -29,13 +47,37 @@ import { intro, outro } from '../../utils/styles/index.js'
...
 
 You don't need to call `outro` if your command throws an error, as `NetlifyLog.error()` automatically creates a clear
-visual end for your command already. It's possible to also exit the process by calling outro with an object:
-`outro({ exit: true })` and you can pass a message to it with `outro({ message: 'Your message' })`.
+visual end for your command already. Outro must be called with an object:
+`outro({ exit: true })` and you can pass a message to it with `outro({ exit: true, message: 'Your message' })`.
+
+### Child commands
+
+As some of our commands call other CLI commands, we will sometimes we want to conditionally call `outro` depending
```
ericapisani, line 60:
```suggestion
As some of our commands call other CLI commands, we will sometimes want to conditionally call `outro` depending
```

STYLEGUIDE.md
```diff
@@ -29,13 +47,37 @@ import { intro, outro } from '../../utils/styles/index.js'
...
+on whether or not the command is a child command. To do this, we use add a property to the `OptionValues` object
```
ericapisani, line 61:
```suggestion
on whether or not the command is a child command. To do this, we add a property to the `OptionValues` object
```

src/commands/init/init.ts
```diff
@@ -26,19 +26,20 @@ const getRepoUrl = (siteInfo) => siteInfo?.build_settings?.repo_url
...
+  NetlifyLog.info('')
+  NetlifyLog.message(`This site has been initialized`)
+  NetlifyLog.info('')
+  NetlifyLog.info(`Site Name:  ${chalk.cyan(siteInfo.name)}`)
+  NetlifyLog.info(`Site Url:   ${chalk.cyan(siteInfo.ssl_url || siteInfo.url)}`)
+  NetlifyLog.info(`Site Repo:  ${chalk.cyan(getRepoUrl({ siteInfo }))}`)
+  NetlifyLog.info(`Site Id:    ${chalk.cyan(siteInfo.id)}`)
+  NetlifyLog.info(`Admin URL:  ${chalk.cyan(siteInfo.admin_url)}`)
+  NetlifyLog.info('')
+  NetlifyLog.info(`To disconnect this directory and create a new site (or link to another siteId)`)
```
ericapisani, line 38: Should this maybe be this instead?

```suggestion
  NetlifyLog.info(`To disconnect this directory and create a new site (or link to another site)`)
```

ericapisani (approved):
Small grammatical things here and there but otherwise good.

One thing that I'm not sure about though is the need for `exit: true` to be passed in all over the place with the use of the `outro` method.

Is this something that we'd always want to have happen? If so, could we expose a helper method that allows a user to call outro with the message and we automatically pass in `exit: true`?

STYLEGUIDE.md
lemusthelroy: It would just print the outro message. I think the signature of this was updated recently so this just updates it to the current signature. I am intrigued as to whether exit should always just happen otherwise clack will just look weird, but will probably tackle that separately.

lemusthelroy: Agree with your other comment on the outro bit. Will circle back to that as I agree

PR #6365, opened 2024-01-31
fix: fix use headers/redirects that are set via the onPreDev of a plugin

sean-roberts:

#### Summary

Reimplements @lukasholzer 's code from https://github.com/netlify/cli/pull/5546 which went stale and out of date. That part covered headers. 

My changes ported that, added support for redirects, and added additional checks for this behavior in a *Dev test.

Fixes https://github.com/netlify/cli/issues/5544

> This PR only parses the headers but we might want to do the same for redirects in a follow up or other config mutations not sure what frameworks needs here?

Currently, if you use the *Dev events, config changes within the callbacks are not applied. 

This PR does not include the ability to watch for these file changes and reapply the new config on file changes while dev is running. Meaning the config will get included on the first pass but it will not apply new config file changes unless the dev command is invoked again. For now, this is fine, we want to be able to unblock this issue for a [customer](https://github.com/netlify/cli/pull/5546#issuecomment-1919620467) and we can follow up with another PR to add file watching.

I verified locally that headers and redirects set in onPreDev work as expected.

---

      passes our tests.

### DIFF
--- a/src/commands/dev/dev.ts
+++ b/src/commands/dev/dev.ts
@@ -1,6 +1,8 @@
 import process from 'process'
 
 import { OptionValues, Option } from 'commander'
+// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module '@net... Remove this comment to see the full error message
+import { applyMutations } from '@netlify/config'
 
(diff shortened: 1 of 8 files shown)


billwatson017 (approved):
code looks good to me but want to ensure this also applies to `redirects` as well as the `headers` which what initially opened this issue

lemusthelroy (approved):

PR #6366, opened 2024-02-01
chore: convert dev command to clack

lemusthelroy:

#### Summary

Converts `dev` command to Clack. This is a difficult command to test as there is a lot of different routes to cover, based on what type of project the `dev` command is being run on. I have tested, and include screenshots below, for the following scenarios:

Static site (with edge function):
image

Astro site:
image

Next site:
image

For reference, this is what this command looked like previously:
image

#### What this PR does not include
- A more comprehensive check on other `dev` scenarios
- We have an ongoing discussion on the impact of removing the `boxen` that present the server as being ready. We know this is a visual cue for many users so want to make sure we get this right.

#### Other notes to reviewers
- We use some libraries, including `@netlify/build` that log out messages that break the clack styling which is why I have opted to transport any `console.log` into `NetlifyLog`. This is worth taking some time to review but I think it is necessary.

---

      passes our tests.


khendrikse (approved):

lemusthelroy:
I ran into issues with this branch locally when trying to switch user. Will investigate further.

PR #6381, opened 2024-02-13
fix: only intercept form submissions if form handler is configured

Skn0tt:

#### Summary

Resolves https://linear.app/netlify/issue/COM-366/ntl-dev-intercepts-all-form-submissions.

We're currently sending all form submissions (POST requests with the right content-type) towards the functions server, which will call the `submission-created` handler if it exists. That breaks frameworks like Remix that make heavy use of form submissions, because those requests aren't passed through to the framework server!

This PR fixes this, by only sending form submissions to the functions server if there's a submission handler registered.

---

      passes our tests.

### DIFF
--- a/src/lib/functions/form-submissions-handler.ts
+++ b/src/lib/functions/form-submissions-handler.ts
@@ -11,22 +11,31 @@ import { warn } from '../../utils/command-helpers.js'
 import { BACKGROUND } from '../../utils/functions/index.js'
 import { capitalize } from '../string.js'
 
+import type NetlifyFunction from './netlify-function.js'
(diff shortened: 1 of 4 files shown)


ascorbic (dismissed):
Speedy work! Nice

Skn0tt:
Looks like there was a failing test, that I had touched a couple weeks ago in https://github.com/netlify/cli/pull/6257. That PR looks like it introduced this regression, and somehow I just changed the test instead of changing the behaviour 😮 I've changed the test back to what it was before https://github.com/netlify/cli/pull/6257, and now it's passing.

ascorbic (approved):

PR #6387, opened 2024-02-14
chore: increase timeout for streaming test

Skn0tt:
We have an integration test that's a little flaky on CI. I'm not sure what's causing that, but I have a hunch that we're sending the two chunks fast enough that a busy CI runner will see them batched together. Let's see if this CI runs better!

Resolves https://linear.app/netlify/issue/COM-373/unflake-cli-test-for-streaming


Skn0tt:
I've run the CI two times in a row now, and there wasn't a single flake. Looks like it's good now!

TylerBarnes (approved):

PR #6388, opened 2024-02-16
fix: support remote_images regexes

pieh:

#### Summary

Currently CLI tries to parse `images.remote_images` as urls and extract domains. Because those are regular expressions and not URL patterns this lead to issues with local dev / serve. This adjust CLI to support regexes just like platform does. IPX doesn't have 1:1 support for that, we can however have custom checking before handing request to IPX and just let IPX "allow every domain"

---

      passes our tests.

Fixes ADN-343


package.json
```diff
@@ -128,7 +128,7 @@
     "https-proxy-agent": "5.0.1",
     "inquirer": "6.5.2",
     "inquirer-autocomplete-prompt": "1.4.0",
-    "ipx": "2.0.1",
+    "ipx": "2.1.0",
```
pieh, line 131: Needed to have https://github.com/unjs/ipx/pull/189 and `allowAllDomains` option for IPX http storage actually working (otherwise it would need setting env var).

There is also new major of `ipx available overall, however I wanted to limit scope of this PR not to have to handle potential breaking changes

src/lib/images/proxy.ts
```diff
@@ -152,7 +150,12 @@ export const initializeProxy = async function ({
       console.log(`fullImageUrl: ${fullImageUrl}`)
       req.url = `/${modifiers}/${fullImageUrl}`
     } else {
-      // If the image is remote, we can just pass the URL as is
+      // If the image is remote, we first check if it's allowed by any of patterns
+      if (!remoteImages.some((remoteImage) => remoteImage.test(sourceImagePath))) {
+        res.status(403).send('Not allowed Remote Image URL.')
```
pieh, line 155: Not sure what platform does for not allowed - so probably this should match platform. For now status matches one from `ipx` and message is a bit custom to match our config field name

pieh: adjusted it to 400 status without body which seems to be behavior on the platform

src/lib/images/proxy.ts
```diff
@@ -152,7 +152,32 @@ export const initializeProxy = async function ({
...
+                ? 'Currently no remote images are allowed.'
+                : `Currently allowed remote images configuration details:\n${remoteImages
+                    .map((pattern) => ` - ${pattern}`)
+                    .join('\n')}`
+            }\n\nSee https://docs.netlify.com/image-cdn/overview/#remote-path how to configure allowed remote images.`,
+          )
+          lastTimeRemoteImagesConfigurationDetailsMessageWasLogged = Date.now()
+        } else {
+          console.error(remoteImageNotAllowedLogMessage)
+        }
```
pieh, line 175: This is what this does in practice (this was refreshing same page multiple times to trigger the not allowed requests): 
image

The more verbose variant try to help guide user why things happen to allow faster diagnose and fixing time by pointing to relevant resource (not sure if it should be path like that or some link shortener could be used, or even if we are fine with adding links that might get out of sync all together). The more verbose variant is ... verbose so that's why I only display it at most once per 30 seconds in current shape of things, otherwise log reading (when combined with logs produced by other things) becomes very messy frustrating for users

src/lib/images/proxy.ts
```diff
@@ -152,7 +152,32 @@ export const initializeProxy = async function ({
...
+              remoteImages.length === 0
+                ? 'Currently no remote images are allowed.'
+                : `Currently allowed remote images configuration details:\n${remoteImages
+                    .map((pattern) => ` - ${pattern}`)
+                    .join('\n')}`
+            }\n\nSee https://docs.netlify.com/image-cdn/overview/#remote-path how to configure allowed remote images.`,
+          )
+          lastTimeRemoteImagesConfigurationDetailsMessageWasLogged = Date.now()
+        } else {
+          console.error(remoteImageNotAllowedLogMessage)
```
eduardoboucas, line 174: Can you use the `log()` method instead, which respects the global verbosity settings?

https://github.com/netlify/cli/blob/e1372a3c228246c7feb9bf4f4e63dad547b2bfa9/src/utils/command-helpers.ts#L163

src/lib/images/proxy.ts
```diff
@@ -31,33 +31,29 @@ interface IpxParams {
...
+      const urlRegex = new RegExp(patternString)
+      remotePatterns.push(urlRegex)
     } catch (error) {
       if (error instanceof Error) {
-        errors.push({ message: `Invalid URL '${patternString}': ${error.message}` })
+        errors.push({ message: `Invalid URL pattern '${patternString}': ${error.message}` })
       } else {
-        errors.push({ message: `Invalid URL '${patternString}': An unknown error occurred` })
+        errors.push({ message: `Invalid URL pattern '${patternString}': An unknown error occurred` })
       }
```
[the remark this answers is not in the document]
pieh: Done in https://github.com/netlify/cli/pull/6388/commits/b3a3a0d381428c7f584ea4bae4dea4232a9df288

src/lib/images/proxy.ts
```diff
@@ -152,7 +152,32 @@ export const initializeProxy = async function ({
...
+        // Each not allowed remote image will still be logged, just without configuration details
+        if (Date.now() - lastTimeRemoteImagesConfigurationDetailsMessageWasLogged > 1000 * 30) {
+          console.error(
+            `${remoteImageNotAllowedLogMessage}\n\n${
+              remoteImages.length === 0
+                ? 'Currently no remote images are allowed.'
+                : `Currently allowed remote images configuration details:\n${remoteImages
+                    .map((pattern) => ` - ${pattern}`)
+                    .join('\n')}`
+            }\n\nSee https://docs.netlify.com/image-cdn/overview/#remote-path how to configure allowed remote images.`,
```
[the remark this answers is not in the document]
pieh: Committed suggestion and used shortend link in https://github.com/netlify/cli/pull/6388/commits/1e74c87c72da6adbe2b7bf96adfe01c0eb71964b

src/lib/images/proxy.ts
```diff
@@ -152,7 +152,32 @@ export const initializeProxy = async function ({
...
-      // If the image is remote, we can just pass the URL as is
+      // If the image is remote, we first check if it's allowed by any of patterns
+      if (!remoteImages.some((remoteImage) => remoteImage.test(sourceImagePath))) {
+        const remoteImageNotAllowedLogMessage = `Remote image "${sourceImagePath}" source for Image CDN is not allowed.`
+
+        // Contextual information about the remote image configuration is throttled
+        // to avoid spamming the console as it's quite verbose
+        // Each not allowed remote image will still be logged, just without configuration details
+        if (Date.now() - lastTimeRemoteImagesConfigurationDetailsMessageWasLogged > 1000 * 30) {
+          console.error(
```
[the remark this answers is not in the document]
pieh: Done in https://github.com/netlify/cli/pull/6388/commits/6e40766605a8e34054cf7b0918dfe9fb43138e5d

tests/integration/commands/deploy/deploy.test.js
```diff
@@ -27,9 +27,10 @@ const validateContent = async ({ content, path, siteUrl, t }) => {
     }
     t.expect(body).toEqual(content)
   } catch (error) {
-    const {
-      response: { statusMessage },
-    } = error
+    let statusMessage = 'Unable to retrieve status message'
+    if (typeof error.response !== 'undefined') {
+      ;({ statusMessage } = error.response)
+    }
```
pieh, line 33: I've been getting failures here to extract `statusMessage` because `error.response` was undefined. After this change now I'm seeing:

https://github.com/netlify/cli/actions/runs/7958720805/job/21724155516#step:11:1309
```
 FAIL  tests/integration/commands/deploy/deploy.test.js > commands/deploy > should deploy Edge Functions when directory exists
Error: Failed getting content: 504 - Unable to retrieve status message - 
 ❯ validateContent tests/integration/commands/deploy/deploy.test.js:35:11
```

https://github.com/netlify/cli/actions/runs/7958720805/job/21724155516#step:11:1320
```
 FAIL  tests/integration/commands/deploy/deploy.test.js > commands/deploy > should deploy Edge Functions with custom cwd when directory exists
Error: Failed getting content: 502 - Unable to retrieve status message - The subhoster origin returned a failure status for the boot RPC call. Status: 502 Bad Gateway
 ❯ validateContent tests/integration/commands/deploy/deploy.test.js:35:11
```

Is this at all related to changes I'm attepting in this PR? It wouldn't seem so to me because those tests don't seem related to images and I don't think they even use code paths I'm changing, but I'm not 100% sure here

pieh: Those failures do seem like "flakes" - but at least this change exposes underlying problems instead of "cannot read statusMessage of undefined" (to make it easier for future)

src/lib/images/proxy.ts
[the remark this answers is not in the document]
pieh: This is result after changes - there still is "Remote images" in header so it should be clear it applies to Image CDN configuration

image

tests/integration/commands/deploy/deploy.test.js
pieh: This was changed a bit in https://github.com/netlify/cli/pull/6385 so I just used that when resolving conflicts

pieh: Yeah it seems fine now :)

kathmbeck (approved):

PR #6410, opened 2024-03-04
feat: introduce env var to include dev server plugin

davbree:

#### Summary

---

      passes our tests.


src/commands/dev/dev.ts
```diff
@@ -137,6 +137,11 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
   try {
     settings = await detectServerSettings(devConfig, options, command)
 
+    if (process.env.NETLIFY_INCLUDE_DEV_SERVER_PLUGIN) {
+      log(`${NETLIFYDEVLOG} Including dev server plugin`)
```
eduardoboucas, line 141: Is this the customer-facing name we're going with? And is there value in showing this message? I believe we'll still show a generic message from Netlify Build indicating that this plugin is being loaded? I would consider printing this only if `debug` is true.

eduardoboucas (approved):

PR #6418, opened 2024-03-07
feat: add `blobs-migrate` recipe

eduardoboucas:
#### Summary

Adds a new `blobs-migrate` recipe that lets folks migrate a Netlify Blobs store as they update to version 7.0.0 of `@netlify/blobs`.

Keeping this as draft until https://github.com/netlify/blobs/pull/148 is merged and the `@netlify/blobs` dependency is updated.

Part of ADN-370.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   "dependencies": {
     "@bugsnag/js": "7.20.2",
     "@fastify/static": "6.10.2",
-    "@netlify/blobs": "6.5.0",
+    "@netlify/blobs": "^7.0.0",
     "@netlify/build": "29.36.1",
     "@netlify/build-info": "7.13.0",
     "@netlify/config": "20.12.1",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   "dependencies": {
     "@bugsnag/js": "7.20.2",
     "@fastify/static": "6.10.2",
-    "@netlify/blobs": "^7.0.0",
+    "@netlify/blobs": "7.0.0",
     "@netlify/build": "29.36.1",
     "@netlify/build-info": "7.13.0",
     "@netlify/config": "20.12.1",
--- a/src/commands/recipes/recipes.ts
+++ b/src/commands/recipes/recipes.ts
@@ -11,11 +11,18 @@ import { getRecipe, listRecipes } from './common.js'
 
 const SUGGESTION_TIMEOUT = 1e4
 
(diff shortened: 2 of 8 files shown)


src/commands/recipes/recipes.ts
```diff
@@ -26,8 +33,10 @@ export const recipesCommand = async (recipeName: string, options: OptionValues,
     return command.help()
   }
 
+  const args = command.args.slice(1)
+
   try {
-    return await runRecipe({ config, recipeName: sanitizedRecipeName, repositoryRoot })
+    return await runRecipe({ args, command, config, recipeName: sanitizedRecipeName, repositoryRoot })
```
eduardoboucas, line 39: We were previously not passing the command arguments to recipes, but it's useful when a recipe needs to take input from the user.

package.json
```diff
@@ -72,7 +72,7 @@
   "dependencies": {
     "@bugsnag/js": "7.20.2",
     "@fastify/static": "6.10.2",
-    "@netlify/blobs": "6.5.0",
+    "@netlify/blobs": "^7.0.0",
```
mraerino, line 75: why is this the only dependency with a caret?

src/recipes/blobs-migrate/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    ...clientOptions,
+    name: `netlify-internal/legacy-namespace/${storeName}`,
+  })
+
+  // The store we'll write to.
+  const newStore = getStore({
+    ...clientOptions,
+    name: storeName,
+  })
+  const { blobs } = await oldStore.list()
```
mraerino, line 41: does this really return all blobs? no pagination?

package.json
eduardoboucas: Good catch! We're pinning the exact versions of dependencies to avoid issues with transitive dependencies. It shouldn't be an issue here, since `@netlify/blobs` doesn't have any dependencies, but I've pinned it anyway in https://github.com/netlify/cli/pull/6418/commits/b0572b1a55f1ef38016480a339ddc32d2c038076.

src/recipes/blobs-migrate/index.ts
eduardoboucas: Yes, see https://docs.netlify.com/blobs/overview/#list.

mraerino (approved):
i haven't tested this locally. this review is just to say the code looks fine

PR #6420, opened 2024-03-07
fix: handle empty store on `blobs:list`

eduardoboucas:
#### Summary

Gracefully handles the case of an empty store in the `blobs:list` command.

_Before_

Screenshot 2024-03-07 at 16 07 46

_After_

Screenshot 2024-03-07 at 16 07 51

Closes COM-302


ascorbic (approved):
The Q of my L is increasing

charliegroll:
I personally prefer the `Before` image, but to each their own

PR #6424, opened 2024-03-10
fix: pass api host to @netlify/build

davbree:
If API host is configured to staging, currently CLI will use that but `@netlify/build` will use production.
To fix this we need to pass in the configured host to `@netlify/build`.


src/utils/run-build.ts
```diff
@@ -66,6 +66,7 @@ export const runNetlifyBuild = async ({ command, env = {}, options, settings, ti
     configPath: cachedConfig.configPath,
     siteId: cachedConfig.siteInfo.id,
     token: cachedConfig.token,
+    host: apiOpts.host,
```
eduardoboucas, line 69: Isn't the property called `apiHost` instead of `host`? https://github.com/netlify/build/blob/2d10996ee91e7640be0e4b53105ae30cb42752fe/packages/build/src/core/normalize_flags.ts#L94

Also, we're already doing something similar [here](https://github.com/netlify/cli/blob/37ac46e2bc1e169733bd7d1c63a6e80ea00628ba/src/lib/build.ts#L103). I understand these are currently different entry points into `@netlify/build`, but it'd be nice to at least standardise the way we pass these properties so it's easier for us to consolidate things into a single entry point in the future.

davbree: good catch, not sure what happened there!

PR #6433, opened 2024-03-11
fix: broken deploy with --skip-functions-cache

lukasholzer:

#### Summary

Fixes broken deploys when run with `ntl deploy --build --skip-functions-cache` inside monorepos.

As the `getPathInProject` function is not monorepo aware, this breaks on monorepos.

There are more areas that depend on this function, but I'll do this in follow-up PRs to keep the PR size small.

Fixes https://linear.app/netlify/issue/FRA-385/could-not-resolve-vartaskappsdocsnetlifydistrunhandlerstracingjs

---

      passes our tests.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -43,6 +43,8 @@ module.exports = {
     // TODO: harmonize with filename snake_case in other Netlify Dev projects
     'unicorn/filename-case': [2, { case: 'kebabCase' }],
     'max-params': 'off',
+    'no-empty-function': 'off',
+    '@typescript-eslint/no-empty-function': 'off',
   },
   overrides: [
     ...overrides,
@@ -137,6 +139,7 @@ module.exports = {
       rules: {
         '@typescript-eslint/no-non-null-assertion': 'off',
         '@typescript-eslint/no-explicit-any': 'off',
+        '@typescript-eslint/no-empty-function': 'off',
       },
     },
   ],
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -42,6 +42,7 @@ module.exports = {
     'unicorn/consistent-destructuring': 0,
     // TODO: harmonize with filename snake_case in other Netlify Dev projects
     'unicorn/filename-case': [2, { case: 'kebabCase' }],
+    'max-params': 'off',
   },
   overrides: [
     ...overrides,
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -1,3 +1,4 @@
+import { Stats } from 'fs'
 import { stat } from 'fs/promises'
 import { basename, resolve } from 'path'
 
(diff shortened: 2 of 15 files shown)


.eslintrc.cjs
```diff
@@ -42,6 +42,7 @@ module.exports = {
     'unicorn/consistent-destructuring': 0,
     // TODO: harmonize with filename snake_case in other Netlify Dev projects
     'unicorn/filename-case': [2, { case: 'kebabCase' }],
+    'max-params': 'off',
```
lukasholzer, line 45: just useless this linting rule

src/commands/deploy/deploy.ts
```diff
@@ -102,31 +106,26 @@ const getDeployFolder = async ({ command, config, options, site, siteData }) =>
...
   try {
     stats = await stat(deployFolder)
   } catch (error_) {
-    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
-    if (error_.code === 'ENOENT') {
-      return error(`No such directory ${deployFolder}! Did you forget to run a build?`)
-    }
+    if (error_ && typeof error_ === 'object' && 'code' in error_) {
+      if (error_.code === 'ENOENT') {
+        return error(`No such directory ${deployFolder}! Did you forget to run a build?`)
```
eduardoboucas, line 122: This is not related to your PR, but I think this message could be improved.

```suggestion
        return error(`The deploy directory "${deployFolder}" has not been found. Did you forget to run 'netlify build'?`)
```

src/commands/deploy/deploy.ts
```diff
@@ -353,13 +354,14 @@ const uploadDeployBlobs = async ({
...
 }: {
-  cachedConfig: any
+  cachedConfig: $TSFixMe
   deployId: string
   options: OptionValues
   packagePath?: string
   silent: boolean
   siteId: string
 }) => {
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
```
eduardoboucas, line 364: You could use this: https://github.com/netlify/cli/blob/e1372a3c228246c7feb9bf4f4e63dad547b2bfa9/src/utils/command-helpers.ts#L300

src/commands/deploy/deploy.ts
```diff
@@ -572,6 +575,7 @@ const bundleEdgeFunctions = async (options, command: BaseCommand) => {
   // eslint-disable-next-line n/prefer-global/process, unicorn/prefer-set-has
   const argv = process.argv.slice(2)
   const statusCb =
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
```
eduardoboucas, line 578: You could use this: https://github.com/netlify/cli/blob/e1372a3c228246c7feb9bf4f4e63dad547b2bfa9/src/utils/command-helpers.ts#L300

eduardoboucas (approved):

src/commands/deploy/deploy.ts
lukasholzer: I fear this will break now all the snapshots 😂

src/commands/deploy/deploy.ts
lukasholzer: TBH It feels to me awkward to have a noop function definition to bypass a useless lint rule

eduardoboucas: Fair, but in that case we should disable the lint rule as a whole rather than adding exceptions on every instance?

lukasholzer: Done

eduardoboucas (dismissed):

eduardoboucas (approved):

PR #6440, opened 2024-03-14
fix: cache edge function dependency tracing

eduardoboucas:
#### Summary

In #6276, we have started to trace transitive dependencies of edge function files for the purpose of hot reloading. This has introduced a regression where, for non-trivial dependency trees, we'll get into an infinite loop trying to recursively trace the dependencies of interdependent files.

This PR addresses that by ensuring we don't trace the same file twice.

Fixes COM-506.


lukasholzer (approved):
LGTM!

PR #6495, opened 2024-04-05
feat: add `--internal-disable-edge-functions` flag to skip Deno setup

Skn0tt:
Resolves https://linear.app/netlify/issue/COM-579/run-ntl-dev-without-the-need-for-deno.


docs/commands/dev.md
```diff
@@ -21,6 +21,7 @@ netlify dev
 - `context` (*string*) - Specify a deploy context or branch for environment variables (contexts: "production", "deploy-preview", "branch-deploy", "dev")
 - `country` (*string*) - Two-letter country code (https://ntl.fyi/country-codes) to use as mock geolocation (enables --geo=mock automatically)
 - `dir` (*string*) - dir with static files
+- `disable-edge-functions` (*boolean*) - disables edge functions. use this if your environment doesnt support Deno
```
seancdavis, line 24:
```suggestion
- `disable-edge-functions` (*boolean*) - disables edge functions. use this if your environment doesn't support Deno
```

docs/commands/serve.md
```diff
@@ -19,6 +19,7 @@ netlify serve
 - `context` (*string*) - Specify a deploy context or branch for environment variables (contexts: "production", "deploy-preview", "branch-deploy", "dev")
 - `country` (*string*) - Two-letter country code (https://ntl.fyi/country-codes) to use as mock geolocation (enables --geo=mock automatically)
 - `dir` (*string*) - dir with static files
+- `disable-edge-functions` (*boolean*) - disables edge functions. use this if your environment doesnt support Deno
```
seancdavis, line 22:
```suggestion
- `disable-edge-functions` (*boolean*) - disables edge functions. use this if your environment doesn't support Deno
```

src/commands/dev/dev.ts
```diff
@@ -271,6 +272,7 @@ export const createDevCommand = (program: BaseCommand) => {
     .option('-d ,--dir <path>', 'dir with static files')
     .option('-f ,--functions <folder>', 'specify a functions folder to serve')
     .option('-o ,--offline', 'disables any features that require network access')
+    .option('--disable-edge-functions', 'disables edge functions. use this if your environment doesnt support Deno')
```
seancdavis, line 275:
```suggestion
    .option('--disable-edge-functions', "disables edge functions. use this if your environment doesn't support Deno")
```

src/commands/serve/index.ts
```diff
@@ -19,6 +19,7 @@ export const createServeCommand = (program: BaseCommand) =>
     .option('-d ,--dir <path>', 'dir with static files')
     .option('-f ,--functions <folder>', 'specify a functions folder to serve')
     .option('-o ,--offline', 'disables any features that require network access')
+    .option('--disable-edge-functions', 'disables edge functions. use this if your environment doesnt support Deno')
```
seancdavis, line 22:
```suggestion
    .option('--disable-edge-functions', "disables edge functions. use this if your environment doesn't support Deno")
```

seancdavis (changes requested):
Added a few suggestions to change "doesnt" to "doesn't". Didn't review the functionality or the code in depth.

src/commands/dev/dev.ts
```diff
@@ -271,6 +272,7 @@ export const createDevCommand = (program: BaseCommand) => {
     .option('-d ,--dir <path>', 'dir with static files')
     .option('-f ,--functions <folder>', 'specify a functions folder to serve')
     .option('-o ,--offline', 'disables any features that require network access')
+    .option('--disable-edge-functions', "disables edge functions. use this if your environment doesn't support Deno")
```
eduardoboucas, line 275: I don't think we should treat this as a customer-facing flag. The only use case we have for it is purely internal, and we haven't fully considered the experience people will get if they disable edge functions. I suggest we use `.hideHelp()` here ([example](https://github.com/netlify/cli/blob/edbfbb452de28e7aa31d27c4dd125112e64ece17/src/commands/dev/dev.ts#L265)) so it's removed from the docs, and I would even consider renaming it to something like `--internal-disable-edge-functions`. That way we're not committing to indefinitely supporting something we're just exploring at this point.

eduardoboucas (changes requested):

src/commands/dev/dev.ts
Skn0tt: that's a very fair point. implemented in https://github.com/netlify/cli/pull/6495/commits/6540b878e458e26d15aac1c2c0e7c1edf6d5cab8

eduardoboucas (approved):

PR #6496, opened 2024-04-05
feat: load edge functions bootstrap from module

eduardoboucas:
#### Summary

Rather than having to manually update the version of the edge functions bootstrap every time there's a new release, this PR loads the bootstrap URL from the `@netlify/edge-functions` module. When there's a new version of the bootstrap, we'll release a new version of the module and Renovate will kick the usual update routine, meaning we'll have a semi-automated way of keeping the bootstrap up-to-date.

### DIFF
--- a/src/lib/build.ts
+++ b/src/lib/build.ts
@@ -30,7 +30,7 @@ import { featureFlags as edgeFunctionsFeatureFlags } from './edge-functions/cons
  * @param {*} config.deployHandler
  * @returns {BuildConfig}
  */
-export const getBuildOptions = ({
+export const getBuildOptions = async ({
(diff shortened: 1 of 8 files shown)


src/lib/build.ts
```diff
@@ -30,7 +30,7 @@ import { featureFlags as edgeFunctionsFeatureFlags } from './edge-functions/cons
  * @param {*} config.deployHandler
  * @returns {BuildConfig}
  */
-export const getBuildOptions = ({
+export const getBuildOptions = async ({
```
eduardoboucas, line 33: Every call to this method is already using `await`, so this is fine.

Skn0tt (approved):

Skn0tt (approved):

PR #6503, opened 2024-04-08
feat: custom rate limit rules payload

paulo:
#### Summary

Related to ADN-294, sends the same payload to the API we're sending now on normal builds.


src/utils/deploy/hash-fns.ts
```diff
@@ -136,8 +136,26 @@ const hashFns = async (
     tmpDir,
   })
   const fileObjs = functionZips.map(
-    // @ts-expect-error TS(7031) FIXME: Binding element 'buildData' implicitly has an 'any... Remove this comment to see the full error message
-    ({ buildData, displayName, generator, invocationMode, path: functionPath, priority, runtime, runtimeVersion }) => ({
+    ({
+      // @ts-expect-error TS(7031) FIXME: Binding element 'buildData' implicitly has an 'any... Remove this comment to see the full error message
```
paulo, line 140: If someone knows a way to apply this to the whole block, please do share 😅

src/utils/deploy/hash-fns.ts
```diff
@@ -195,4 +219,25 @@ const hashFns = async (
   return { functionSchedules, functions, functionsWithNativeModules, fnShaMap, fnConfig }
 }
 
+const trafficRulesConfig = (trafficRules: $TSFixMe) => {
```
eduardoboucas, line 222: I would be nice to import the type that we'll expose to users and use it here, but we can do that as a follow-up.

eduardoboucas (dismissed):

tests/integration/commands/deploy/deploy.test.js
```diff
@@ -509,7 +509,14 @@ describe.skipIf(process.env.NETLIFY_TEST_DISABLE_LIVE === 'true').concurrent('co
         .withContentFile({
           content: `
           export default async () => new Response("Internal V2 API")
-          export const config = { path: "/internal-v2-func" }
+          export const config = {
+            path: "/internal-v2-func",
+            rateLimit: {
```
Skn0tt, line 514: I don't see any assertions added. I'm assuming you used this for testing things locally? Fine by me, i'm wondering if there's a good way of asserting that the ratelimit config was actually applied 🤔

src/utils/deploy/hash-fns.ts
Skn0tt: in https://github.com/netlify/cli/pull/6503/commits/8f4a538e1b497b94d6c5cf575d1d29141836abd2, i've imported the TrafficRules type from the ZISI internals. Is that what you had in mind? Also, do we consider importing those internals criminal?

eduardoboucas: Yes, exactly! I don't think it's criminal, but I think it would be good to properly export them so we're not relying on internal file paths.

src/utils/deploy/hash-fns.ts
paulo: woah, thank you so much!

src/utils/deploy/hash-fns.ts
Skn0tt: agreed. i'll make the zisi PR and replace the import in a followup.

tests/integration/commands/deploy/deploy.test.js
paulo: yes I tested it out locally

src/utils/deploy/hash-fns.ts
Skn0tt: here's said PR: https://github.com/netlify/zip-it-and-ship-it/pull/1724

Skn0tt (approved):

PR #6506, opened 2024-04-10
fix: use the right Blobs context in `serve`

eduardoboucas:
#### Summary

`netlify serve` is a bit special in that it first runs a build and then starts a server. In production, these steps have different Blobs contexts — API access should be used at build time and edge access should be used at runtime. This PR makes the `serve` command respect that.

Part of COM-580.

### DIFF
--- a/src/commands/serve/serve.ts
+++ b/src/commands/serve/serve.ts
@@ -2,7 +2,12 @@ import process from 'process'
 
 import { OptionValues } from 'commander'
 
-import { BLOBS_CONTEXT_VARIABLE, encodeBlobsContext, getBlobsContext } from '../../lib/blobs/blobs.js'
+import {
+  BLOBS_CONTEXT_VARIABLE,
+  encodeBlobsContext,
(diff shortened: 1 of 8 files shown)


src/commands/serve/serve.ts
```diff
@@ -143,6 +157,7 @@ export const serve = async (options: OptionValues, command: BaseCommand) => {
   const inspectSettings = generateInspectSettings(options.edgeInspect, options.edgeInspectBrk)
   const url = await startProxyServer({
     addonsUrls,
+    blobsContext: runtimeBlobsContext,
```
eduardoboucas, line 160: This was missing entirely, so edge functions were not getting a Blobs context.

Skn0tt (dismissed):

Skn0tt (approved):

JGAntunes (approved):
Please GitHub let us merge this PR 🙏

eduardoboucas:
Superseded by #6507.

PR #6533, opened 2024-04-26
fix: respect `excludedPattern` defined in netlify.toml

Skn0tt:
Title says it all! We were only looking at the per-function setting for this, not at the per-route setting. This PR fixes that.


Skn0tt:
The test failures seem unrelated to me.

eduardoboucas (approved):

lemusthelroy:
We are working on the test failures. Will update you when this is resolved.

PR #6557, opened 2024-05-06
fix: bug around v1 functions with `nodeModuleFormat: "esm"`

Skn0tt:
Resolves https://linear.app/netlify/issue/COM-434/internal-v1-functions-with-esm-nodemoduleformat-are-broken-in-local

### DIFF
--- a/src/lib/functions/registry.ts
+++ b/src/lib/functions/registry.ts
@@ -1,4 +1,4 @@
-import { mkdir, readdir, stat } from 'fs/promises'
+import { mkdir, stat } from 'fs/promises'
 import { createRequire } from 'module'
 import { basename, extname, isAbsolute, join, resolve } from 'path'
 import { env } from 'process'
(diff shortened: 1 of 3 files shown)


src/lib/functions/registry.ts
```diff
@@ -434,7 +434,7 @@ export class FunctionsRegistry {
 
         func.mainFile = v2EntryPointPath
       } catch {
-        func.mainFile = join(unzippedDirectory, `${func.name}.js`)
+        func.mainFile = join(unzippedDirectory, basename(manifestEntry.mainFile))
```
Skn0tt, line 437: the problem here was that esm files end in `.mjs`, not `.js`. By using the basename from the manifest, we'll have the right extension

src/lib/functions/runtimes/js/index.ts
```diff
@@ -118,6 +130,7 @@ export const invokeFunctionDirectly = async ({ context, event, func, timeout })
     lambdaPath,
     timeoutMs: timeout * SECONDS_TO_MILLISECONDS,
     verboseLevel: 3,
+    esm: lambdaPath.endsWith('.mjs'),
```
eduardoboucas, line 133: We're already running ESM functions locally today, so I'm not sure I follow why we're adding this now. What does it do exactly?

eduardoboucas: We respect the module format specified by the file extension and the `module` property. So we're keeping ESM files as ESM, at least in production. I'd be surprised if this wasn't the case locally. The CLI itself is ESM, so by importing `lambda-local` as ESM it should be able to import both ESM and CJS files without a dynamic import?

Skn0tt: But lambda-local is a CJS module, and if we don't set the `esm` property then it will always `require` the file at `lambdaPath`. See here: https://github.com/ashiina/lambda-local/blob/8914e6804533450fa68c56fe6c34858b645735d0/src/lambdalocal.ts#L337-L344

I'm not sure why this hasn't been a problem before, but if I remove this line locally I get:

Screenshot 2024-05-07 at 13 19 48

PR #6566, opened 2024-05-08
chore: infer tmp directory name from test name everywhere

Skn0tt:
We have some flaky tests on main, that are being flaky because they're reusing the same directory across multiple, concurrent tests. In bad race conditions, this means the directory is being cleaned up by one test, while a different test still uses it 💥 

fa5de3b866f20a7814d52229c98d09be2ec1b156 contains a fix, where we're not specifying the directory name by hand, but we use the test name as a basis for it. And because i'm a little annoyed this happened in the first place, i'm also replacing all other usage of `withSiteBuilder(string)` in f62520320b8840b988768b69a2b8534bba8ce920.


ericapisani (approved):

mrstork (dismissed):

mrstork (approved):
Awesome! We might just change the title before merging is all

PR #6662, opened 2024-05-24
fix: crash on `envelopItems.filter`

Skn0tt:
Resolves https://linear.app/netlify/issue/CT-1215/%5Bbug%5D-cli-throwing-envelopeitems-error.


khendrikse (approved):
Let's go!

khendrikse (approved):

PR #6676, opened 2024-05-29
feat: propagate per-function timeout in deploy

Skn0tt:
Resolves https://linear.app/netlify/issue/COM-772/update-cli-to-send-per-function-timeout-parameter.

Depends on ZISI update to be merged first.


biruwon (approved):

Skn0tt:
The required ZISI update landed on main, so tests should pass now.

PR #6686, opened 2024-06-02
fix: adds html injections to dev proxy [CRE-1203]

smnh:
This enables visual-editor plugin to inject snippet into html in dev mode.

Related PRs:
- https://github.com/netlify/build/pull/5692


src/commands/base-command.ts
```diff
@@ -7,7 +7,6 @@ import { format } from 'util'
 
 import { DefaultLogger, Project } from '@netlify/build-info'
 import { NodeFS, NoopLogger } from '@netlify/build-info/node'
-// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module '@net... Remove this comment to see the full error message
 import { resolveConfig } from '@netlify/config'
```
smnh, line 10: This is fixed by this:
https://github.com/netlify/build/pull/5692/commits/cf459316c28d49444330593fcc393c0febc80cb0

src/commands/dev/dev.ts
```diff
@@ -1,6 +1,5 @@
 import process from 'process'
 
-// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module '@net... Remove this comment to see the full error message
 import { applyMutations } from '@netlify/config'
```
smnh, line 3: This is fixed by this:
https://github.com/netlify/build/pull/5692/commits/cf459316c28d49444330593fcc393c0febc80cb0

src/utils/proxy.ts
```diff
@@ -58,12 +61,48 @@ const decompressResponseBody = async function (body: Buffer, contentEncoding = '
     case 'br':
       return await brotliDecompress(body)
     case 'deflate':
-      return await deflate(body)
+      return await inflate(body)
```
smnh, line 64: `inflate` is for decompressing.
`deflate` is for compressing.

TylerBarnes: Why is the case `deflate`?

TylerBarnes (commented):
@smnh I think a small test is a good idea if possible

src/utils/proxy.ts
TylerBarnes: @smnh I think the PR is ready to merge, just curious about this last point. If `case "deflate": return inflate` is intentional I think adding a code comment about why we handle the "deflate" case with "inflate" is a good idea here

PR #6691, opened 2024-06-05
feat: make feature flags available to `@netlify/config`

lemusthelroy:

#### Summary
By constructing our API client earlier, this PR feeds the feature flags into `@netlify/config` where the flags are required

https://linear.app/netlify/issue/CT-1291/ensure-cli-passes-feature-flags-to-netlifyconfig

---

      passes our tests.


DanielSLew (approved):

khendrikse (dismissed):
Awesome!!

eduardoboucas:
This is touching production code, so it shouldn't be a `chore:`. Can we please rename before merging?

src/commands/base-command.ts
```diff
@@ -551,6 +559,34 @@ export default class BaseCommand extends Command {
...
+
+    // If a user passes a site name as an option instead of a site ID to options.site, the siteInfo object
+    // will only have the property siteInfo.id. Checking for one of the other properties ensures that we can do
+    // a re-call of the api.getSite() that is done in @netlify/config so we have the proper site object in all
+    // commands.
+    // options.site as a site name (and not just site id) was introduced for the deploy command, so users could
+    // deploy by name along with by id
+    let siteIdByName = ''
+    if (flags.site) {
+      const site = await getSiteByName(api, flags.site)
```
lemusthelroy, line 578: I might be wrong about what `getSiteByName` does, but isn't this going to lead to more than one request to the API to get the site given the logic on line 582?

src/commands/serve/serve.ts
```diff
@@ -148,7 +148,7 @@ export const serve = async (options: OptionValues, command: BaseCommand) => {
 
   // TODO: We should consolidate this with the existing config watcher.
   const getUpdatedConfig = async () => {
-    const { config: newConfig } = await command.getConfig({ cwd: command.workingDir, offline: true, state })
```
lemusthelroy, line 151: Not sure if it was me that committed this, but should we keep state here? And in the dev command too?

Skn0tt: That change was me! I've removed the `state` argument, because its only purpose was to be used in determining the site ID. I've moved that logic into an instance variable now, so it's not needed anymore.

khendrikse (approved):
I am loving this :)

PR #6723, opened 2024-06-23
feat: add flag to skip env var fetching

davbree:

#### Summary

Introduce internal option `--offline-env` to disable using any env vars fetched from the API.
This offers better support for the case of receiving all env vars externally (via devbot).

ref CRE-1078

---

      passes our tests.


smnh (approved):

lukasholzer:
why cannot we use the existing `--offline` flag for it?

davbree:
We still want to fetch the site config (for the dev command and target port). I could provide that from bitballoon as well in the future and then we can use `--offline` (and remove `--offline-env`)

lukasholzer:
> We still want to fetch the site config (for the dev command and target port). I could provide that from bitballoon as well in the future (CRE-1431) and then we can use `--offline` (and remove `--offline-env`)

As this is public API on the CLI we should provide that probably first to be able to leverage the --offline flag.

davbree:
I'm using `hideHelp(true)` to keep things internal (like we do for other parameters). Makes it easier to do this not all together..

PR #6729, opened 2024-06-24
fix: use heuristics for build and deploy command as well

lukasholzer:

#### Summary

**Depends on: https://github.com/netlify/build/pull/5735**

Fixes https://linear.app/netlify/issue/CPLA-738/sites-created-in-the-cli-should-use-framework-detection

Run the framework detection before `ntl build` and `ntl deploy --build` as well to not need a `netlify.toml` in simple cases. (`ntl dev` already supported this)

---

      passes our tests.


src/utils/build-info.ts
```diff
@@ -122,3 +123,34 @@ command = "${chosenSettings.devCommand}"
...
+    cachedConfig.config.build.commandOrigin = 'heuristics'
+  }
+
+  if (cachedConfig.config.build.publishOrigin === 'default') {
+    cachedConfig.config.build.publish = settings.dist
+    cachedConfig.config.build.publishOrigin = 'heuristics'
+  }
+
+  settings.plugins_recommended.forEach((plugin) => {
+    if (!cachedConfig.config.plugins.includes(plugin)) {
```
eduardoboucas, line 150: How does this check work? We're pushing objects such as `{ package: "package-name" }` to `cachedConfig.config.plugins`, so how can we use `includes` to check if a plugin is already there?

lukasholzer: that happens if you don't have types and everything is any 😬

eduardoboucas (approved):
It'd be nice to have some tests, but LGTM.

src/utils/build-info.ts
```diff
@@ -122,3 +123,32 @@ command = "${chosenSettings.devCommand}"
     return chosenSettings
   }
 }
+
+/**
+ * Merges the settings from the heuristics with the cached config.
+ * Returns the updated cached config
+ * @param cachedConfig The cached config
```
eduardoboucas, line 130: I think this needs updating, since I don't see this param?

lukasholzer: oh good catch!

lukasholzer: fixed here: https://github.com/netlify/cli/pull/6729/commits/049f4a9c2819f2326e2289c593613defca3ea828

eduardoboucas (approved):
🚀

lukasholzer:
@eduardoboucas I'm going to add an integration test to it as well :)

eduardoboucas (approved):

ascorbic:
This is great! Should it be working for `serve` as well? I ask because I tried creating a new astro site and then ran `ntl serve` and it gave me a 404.

flore2003:
Is there a way to overwrite or turn off this heuristic framework detection for the build and deploy commands? This new behavior breaks our current monorepo setup, because we have both Next.js and vite based React apps in it, and the heuristic detects next as the framework, even for the vite based React app. The `dev` command has an option to specify the framework, but neither build nor deploy seem to have it, if I see it correctly.

PR #6733, opened 2024-06-25
feat: pass accountId to resolveConfig

lemusthelroy:

#### Summary

Passes `accountId` to `resolveConfig()` which is going to be used when fetching build events

---

      passes our tests.


DanielSLew (dismissed):

DanielSLew (dismissed):

src/commands/base-command.ts
```diff
@@ -685,6 +687,7 @@ export default class BaseCommand extends Command {
 
     try {
       return await resolveConfig({
+        accountId: this.accountId,
```
khendrikse, line 690: I'm curious how we know we'll have `accountId` on `this`?

lemusthelroy: I share your curiosity because we do the same for siteId, but it seems to work, let me see

lemusthelroy: ah I understand, because we assign it to the `actionCommand` which is of type `BaseCommand`. This means that anything assigned to it will be available on `this` when the command is run. Happy I understand it now.

fwiw I tried just assigning it to `this` in the baseCommand itself but it doesn't persist

khendrikse: ah ok awesome!

khendrikse (approved):

PR #6735, opened 2024-06-27
feat: add support for the Frameworks API

eduardoboucas:
#### Summary

Adds support for the Frameworks API to the `deploy` and `serve` commands.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -33,6 +33,7 @@ import {
 import { DEFAULT_DEPLOY_TIMEOUT } from '../../utils/deploy/constants.js'
 import { deploySite } from '../../utils/deploy/deploy-site.js'
 import { getEnvelopeEnv } from '../../utils/env/index.js'
+import { getFrameworksAPIPaths } from '../../utils/frameworks-api.js'
 import { getFunctionsManifestPath, getInternalFunctionsDir } from '../../utils/functions/index.js'
 import openBrowser from '../../utils/open-browser.js'
 import BaseCommand from '../base-command.js'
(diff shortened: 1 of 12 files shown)


tests/unit/lib/functions/registry.test.js
```diff
@@ -6,6 +6,7 @@ import { describe, expect, test, vi } from 'vitest'
 
 import { FunctionsRegistry } from '../../../../dist/lib/functions/registry.js'
 import { watchDebounced } from '../../../../dist/utils/command-helpers.js'
+import { getFrameworksAPIPaths } from '../../../../dist/utils/frameworks-api.js'
```
serhalp, line 9: 🤔 Huh, why do all the unit tests test the built code while the integration tests test the source? That's odd.

src/lib/functions/server.ts
```diff
@@ -348,7 +350,11 @@ export const startFunctionsServer = async (
   } else {
     // The order of the function directories matters. Rightmost directories take
     // precedence.
```
serhalp, line 352: same here, is it possible to test this?

src/utils/feature-flags.ts
```diff
@@ -23,6 +23,7 @@ export const getFeatureFlagsFromSiteInfo = (siteInfo: {
   ...(siteInfo.feature_flags || {}),
   // see https://github.com/netlify/pod-dev-foundations/issues/581#issuecomment-1731022753
   zisi_golang_use_al2: isFeatureFlagEnabled('cli_golang_use_al2', siteInfo),
+  netlify_build_frameworks_api: true,
```
serhalp, line 26: 🤔 Were you initially thinking of feature flagging this and then you changed your mind and forgot to remove this? I might be missing something otherwise.

serhalp (approved):
LGTM, just some uncertainty around the feature flag and some nits

src/commands/deploy/deploy.ts
```diff
@@ -460,12 +461,15 @@ const runDeploy = async ({
...
+
+    await frameworksAPIPaths.functions.ensureExists()
 
     // The order of the directories matter: zip-it-and-ship-it will prioritize
     // functions from the rightmost directories. In this case, we want user
     // functions to take precedence over internal functions.
-    const functionDirectories = [internalFunctionsFolder, functionsFolder].filter((folder): folder is string =>
-      Boolean(folder),
+    const functionDirectories = [internalFunctionsFolder, frameworksAPIPaths.functions.path, functionsFolder].filter(
+      (folder): folder is string => Boolean(folder),
```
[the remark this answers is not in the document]
eduardoboucas: We're already testing that [here](https://github.com/netlify/cli/blob/f07ecdcc203f4508442b11076e75f7e1e2be7dce/tests/integration/commands/deploy/deploy.test.js#L471), by hitting the deployed site and asserting that the output comes from the function we expect. I can add a few more cases here in a follow-up, though.

tests/unit/lib/functions/registry.test.js
eduardoboucas: Good question. I'm not sure! I know we've gradually introduced vitest over ava, and there was some complexity in that transition. But that doesn't explain this. I'll take a look.

src/lib/functions/server.ts
eduardoboucas: Same answer as above. We end up testing this in the different commands by creating conflicting function and checking the output of the one that wins.

src/utils/feature-flags.ts
eduardoboucas: No, the idea here is to always enable the Frameworks API in Netlify Build, which is still behind a feature flag in our build system.

PR #6777, opened 2024-08-05
fix: load functions timeout from site properly

### LINKED ISSUE #6125: Make function timeout editable OR inherit from linked site

Users who request an increased function timeout applied to their site or account from Netlify staff frequently expect the CLI to behave like production, but it does not by default.

Perhaps a flag to specify a timeout, or automatically checking the timeout set in Netlify's system for the linked site, would be a better UX?

### Pull request (optional)

- [ ] I can submit a pull request.

biilmann:

#### Summary

Fixes #6125

When a site has a custom timelimit for functions, its exposed as `function_timelimit` directly on the site object returned from the API. The current CLI code always tries to read it from the generic function_config hash if one is present.

---

99676c72-5a6c-48b1-891a-373ceac143f1 (1)

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -122,7 +122,7 @@ export const getSiteInformation = async ({ api, offline, site, siteInfo }) => {
         backgroundFunctions: supportsBackgroundFunctions(account),
       },
       timeouts: {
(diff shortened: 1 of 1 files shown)


khendrikse (approved):
Looks good! Also, good panda

PR #6781, opened 2024-08-06
fix: apply default origin to plugins provided by plugin detection

pieh:

#### Summary

Part of https://linear.app/netlify/issue/FRP-1218/plugin-netlifyplugin-nextjs-must-not-be-specified-twice-in-the-app

We are setting `default` origin for build command and publish dir, but not for plugins we set from plugin detection - this makes that all setting provided by plugin detection has corresponding `origin` property set. See https://github.com/netlify/build/pull/5793 on how added `origin` actually fixes the issue

---

      passes our tests.


lukasholzer (approved):
Oh that's a great addition!

PR #6802, opened 2024-08-26
test: disable node@22.7 Module syntax detection for zisi tests

pieh:

#### Summary

https://nodejs.org/en/blog/release/v22.7.0#module-syntax-detection-is-now-enabled-by-default 

Node 22.7 now will automatically fallback to trying to parse `.js` files as ESM when `package.json` doesn't specify type. This makes assumptions that some of our test make no longer true and this just update test setup to continue to behave in same way as it was before.

https://github.com/netlify/cli/actions/runs/10563005609/job/29264301704?pr=6801#step:11:1263 is example how test as-is now are failing with node 22.7


TylerBarnes (approved):
Oh awesome, thanks for this! Didn't know about that change 🎉

PR #6803, opened 2024-08-27
fix: "dev:exec" command description

sean-roberts:
Descriptions don't support \n line breaks so the description is currently just "Exec Command" Removing that prefix and line break makes the description more consistent with our other command description patterns.

shoutout to @rjbeers for catching it


serhalp (approved):
Thanks for fixing!

serhalp (dismissed):

TylerBarnes (dismissed):

codebyuma (dismissed):
Sorry for the late review! 

Looks good to me, though I left a comment to fix some style guide things that Vale is complaining about. I know it's from existing content but since you're touching these lines anyway, I figure we can fix them at the same time 😄 

🚢

serhalp:
bump @sean-roberts 😁

sean-roberts:
> bump @sean-roberts 😁

ooof I did not see this, thanks for the tag!

docs/commands/dev.md
```diff
@@ -57,7 +57,6 @@ BROWSER=none netlify dev # disable browser auto opening
 ---
 ## `dev:exec`
 
-Exec command
 Runs a command within the netlify dev environment, e.g. with env variables from any installed addons
```
serhalp, line 60: Missed one here (I'll push it):
```suggestion
Runs a command within the netlify dev environment. For example, with environment variables from any installed add-ons
```

serhalp (dismissed):

codebyuma (dismissed):

serhalp (approved):

PR #6838, opened 2024-09-25
fix: honor --auth option

pieh:

#### Summary

Fixes https://linear.app/netlify/issue/FRB-1343/internal-server-error-on-netlify-builds-via-bitbucket-pipeline-missing

Our `--auth` CLI option handling seems a bit hit and miss. We had a lot of optional argument and `@ts-expect-error` around code that was dealing with tokens - this removes all of those, adjust types to require token from CLI toggle being passed in to helper functions (do note that this doesn't fully ensure type safety, because `options.auth` has type `any`, but at least it ensures that _something_ is passed in instead of being optional)

---

      passes our tests.


src/commands/deploy/deploy.ts
```diff
@@ -382,7 +382,7 @@ const uploadDeployBlobs = async ({
     packagePath,
     deployId,
     siteId,
-    token,
+    token: token === null ? undefined : token,
```
pieh, line 385: `token` here is of type `string | null`, while `runCoreSteps` expect it to be `string | undefined`

Overall at this stage there should be a token always, otherwise we would bail much earlier, but stuff like this is annoying with TS and probably would need more refactoring to not have to deal with `null / undefined` cases.

src/commands/integration/deploy.ts
```diff
@@ -394,8 +394,7 @@ export const getConfiguration = (workingDir) => {
 export const deploy = async (options: OptionValues, command: BaseCommand) => {
   const { api, cachedConfig, site, siteInfo } = command.netlify
   const { id: siteId } = site
-  // @ts-expect-error TS(2554) FIXME: Expected 1 arguments, but got 0.
-  const [token] = await getToken()
+  const [token] = await command.authenticate(options.auth)
```
pieh, line 397: Note that this one will have a potential change in behavior - by using `command.authenticate` (like all the other commands) - this will start auth flow if user is not logged in (or didn't provide token via env var or `--auth`), while previously it would result in getting no token (and presumably getting failures later down the road?)

After fixing up bunch of related TypeScript elsewhere, using the `getToken` that could be `null` (previously it would be just `any`) it would complain about `token` being possibly `null` in https://github.com/netlify/cli/blob/e64e30c66d4612836926d27d23fe12e97726582e/src/commands/integration/deploy.ts#L421-L428

so I assume this should be required that token is present before executing the rest of this command?

cc @netlify/ecosystem-pod-composable-tooling

tests/integration/commands/deploy/deploy.test.ts
```diff
@@ -953,6 +954,71 @@ describe.skipIf(process.env.NETLIFY_TEST_DISABLE_LIVE === 'true').concurrent('co
     })
   })
 
+  test('should upload blobs when saved into .netlify directory and using --auth flag', async (t) => {
```
pieh, line 957: this test alone was failing (at least in GHA) without code changes done in 2nd commit

src/commands/integration/deploy.ts
pieh: This is probably wrong thing to do, as we do authenticate in BaseCommand already - I will be looking if maybe there is some nice way to avoid token being typed as potentially `null`

pieh: Ops, my bad - we don't actually force authenticate in base command, so this change seems still good here

serhalp:
@pieh is this something we want to get over the finish line?

pieh:
This got really outdated and has so many merge conflicts that is not worth keeping open. There is a workaround with `NETLIFY_AUTH_TOKEN=$NETLIFY_AUTH npx netlify deploy --build`, so I'll close this PR to clean up PR list.

PR #6842, opened 2024-09-27
fix: prioritize redirects ahead of function matches

minivan:

#### Summary

Fixes RUN-974

---

If a function has a `path` specified in the config, that path will be matched ahead of the redirect. This is not the production behavior.

🥘


tests/integration/commands/dev/dev.config.test.js
```diff
@@ -451,6 +451,29 @@ describe.concurrent('commands/dev/config', () => {
     })
   })
 
+  test('will prefer a redirect to a function that matches the path', async (t) => {
```
minivan, line 454: I'm not a fan of this, and I don't know how to reproduce it better.

When I checked locally, the behavior was fixed by the changes in `src/utils/proxy.ts`. However, if I undo the changes in proxy, this test doesn't fail, and I don't understand why.

src/utils/proxy.ts
```diff
@@ -757,6 +757,28 @@ const onRequest = async (
...
+      settings.frameworkHost && isIPv6(settings.frameworkHost) ? `[${settings.frameworkHost}]` : settings.frameworkHost
+    }:${settings.frameworkPort}`,
+    publicFolder: settings.dist,
+    functionsServer,
+    functionsPort: settings.functionsPort,
+    jwtRolePath: settings.jwtRolePath,
+    framework: settings.framework,
+  }
+
+  if (match) {
```
mraerino, line 774: what if the match is not a redirect, but a rewrite? we'll want to match functions on the rewritten path in that case

lukasholzer (approved):

PR #6865, opened 2024-10-08
fix: link --name prefers exact match if exists, first match otherwise

dylanspyer:

#### Summary

 If `ntl link --name <name>` matched multiple site names, it previously linked the first site name listed, even if there was an exact match. This fix makes `ntl link --name <name>` prefer an exact match if there is one. It follows previous behavior if there is not an exact match.

Fixes # From internal ticket system

```In netlify-react-ui, I ran ntl link --name app, and it connected me to next-app-playground. It should’ve connected me to app. The reason is that in https://github.com/netlify/cli/blob/79732adbc486516a2d2e244d0d160bacb94c34af/src/commands/link/link.ts#L303-L334, it’s always selecting the first site in the search result. Instead, it should check if there’s an exact match, and use that if it exists.```

---

      passes our tests.

### CODE REVIEW

**Review by DanielSLew — DISMISSED** [review]

File: src/commands/link/link.ts [hunk 1]
```diff
@@ -322,13 +322,13 @@ export const link = async (options: OptionValues, command: BaseCommand) => {
     if (results.length === 0) {
       error(new Error(`No sites found named ${options.name}`))
     }
-    const [firstSiteData] = results
-    state.set('siteId', firstSiteData.id)
+    const matchingSiteData = results.find((site: any) => site.name === options.name) || results[0]
```
serhalp, line 325: can we avoid introducing new `any`s here?

dylanspyer: Created a new type `SiteInfo` and used that to avoid introducing new `any` types.

src/commands/sites/sites-list.ts
```diff
@@ -40,9 +44,8 @@ export const sitesList = async (options: OptionValues, command: BaseCommand) =>
 
     // Json response for piping commands
     if (options.json) {
-      // @ts-expect-error TS(7006) FIXME: Parameter 'site' implicitly has an 'any' type.
       const redactedSites = sites.map((site) => {
-        if (site && site.build_settings) {
+        if (site && site.build_settings && site.build_settings.env) {
```
DanielSLew, line 48:
```suggestion
        if (site?.build_settings?.env) {
```

src/commands/sites/sites-list.ts
```diff
@@ -21,17 +21,21 @@ export const sitesList = async (options: OptionValues, command: BaseCommand) =>
...
-    // @ts-expect-error TS(7006) FIXME: Parameter 'site' implicitly has an 'any' type.
     const logSites = sites.map((site) => {
-      const siteInfo = {
+      const siteInfo: {
+        id: string
+        name: string
+        ssl_url: string
+        account_name: string
+        repo_url?: string
+      } = {
```
DanielSLew, line 31: Instead of defining a new type here you could do something like
```suggestion
      const siteInfo: Pick<SiteInfo, 'id' | 'name' | etc...> = {
```

DanielSLew (commented):
Left a few comments and suggestions

dylanspyer:
@DanielSLew thanks for the feedback! I added the suggestions.

DanielSLew (dismissed):

DanielSLew (approved):

PR #6878, opened 2024-10-15
feat: add confirmation prompts to unsafe cli commands

wconrad265:
## Problem

Several CLI commands currently allow users to perform actions that may accidentally modify production environments. Some actions, like setting environment variables without specifying context (or scope) can inadvertently affect all contexts (or scopes) leading to unintended consequences.

Worked on this with @tlane25 

## Solution

To address this, we added confirmation prompts to critical commands. These prompts inform users of potentially unsafe changes and give them a chance to cancel the operation. The prompts can be bypassed using the `-f` or `--force` flag for advanced users who understand the risks.

Additionally, we refactored the codebase to improve prompt management by:

- Creating a dedicated `prompts` folder under the `utils` directory to store logic and messages for each command. Each command with prompts now has a corresponding file (e.g., `env-set-prompts.ts` for `env:set`) to keep the code modular and organized.
- Leveraging the existing `inquirer` package for prompts, since it was already installed in the project.

### **`env:set`**
    
    
The current code already checks if the environment variable exists. If it does, and the user does not pass the `--force` or `-f` flag, a prompt now displays, informing the user that they are about to modify the existing variable. 
    
before
    
image
    
after

image

---

### **`env:unset`**
    
The user will be prompted to confirm deletion if they don’t pass the `--force` flag
    
before
    
image
    
after

image
  
--- 
    
### **`env:clone`**
    
The original code already checks, if there are environment variables of the same in name in the target site. If an environment variable with the same name already exists on the target site, a prompt informs the user that variables will be overwritten.
    
The `-f` flag is already used for the `--from` flag, so users can only bypass it with the `--force` flag.
    
before

image

after
    
image    

--- 

### **`blob:set`**
    
Will check to see if a key in the store exists. If the key does exist, the user will be prompted with a warning message. 
    
Also a confirmation message has been added, letting the user know the operation was successful, as there was not one before. 
    
before
    
image
    
after
    
image

--- 
   
### **`blob:delete`**
    
A prompt now displays to the user if they are deleting a key of a store. 
    
Also a confirmation message was added. 
    
before
    
image
    
after

image

---

## Testing and Validation

- **Manually tested each command** to ensure that confirmation prompts appear under the correct conditions.
    - Example: For `env:set`, the prompt only appears if the environment variable already exists.
    - The `-force` flag correctly skips the prompt where applicable.
- **Added tests** in the corresponding test files for each command:
    - The tests validate that the prompt appears under the correct conditions.
    - The tests verify that the `-force` flag bypasses the confirmation prompt as intended.
    - The tests also confirm that if the user opts not to continue, the process exits as expected.

### Testing and Setup Explanation

In the test setups using `setupFixtureTests('empty-project')`, we needed to ensure that commands like `blobs:set` function as expected. However, since `callCli` spins up a separate Node.js process for each test and interacts with the command-line interface, we encountered issues with `inquirer` prompts that require user input.

### Issues

- **CLI prompts**: The commands now include `inquirer` confirmation prompts that expect user interaction. When running tests with `callCli`, these prompts cause the tests to hang and eventually time out, as no input is provided to the prompts.
- **Solution**: We added the `-force` flag to all tests invoking commands with confirmation prompts. This bypasses the interactive prompts, ensuring the tests proceed without waiting for user input.

```jsx
setupFixtureTests('empty-project', { mockApi: { routes } }, () => {
  const expectedSuccessMessage = 'Success: Blob my-key set in store my-store';

  test<FixtureTestContext>('should set, get, list, and delete blobs', async ({ fixture }) => {
    // The --force flag is used to bypass the inquirer prompt and prevent timeouts
    expect(
      await fixture.callCli(['blobs:set', 'my-store', 'my-key', 'Hello world', '--force'], {
        offline: false,
      }),
    ).toBe(expectedSuccessMessage);
  });
});

```

This setup allows us to validate command behavior without requiring manual input or causing test timeouts.

I wanted to mention this here, because I had to add to `--force` flag to several test files. This might be good to add to the documentation somewhere. I wasn’t sure where to add it. 

### Testing the `inquirer` prompts

We used `withMockApi`, which spins up a local Express server to simulate the API for testing. Then, we mocked `inquirer.prompt` to simulate user interaction, ensuring the correct prompts and messages were displayed.

This is consistent with other tests that mock inquirer prompts as well.

Each of these commands has tests in their corresponding test files. If a command did not have a test file that corresponded to it, it was created. 

## Documentation

Documentation has been updated to reflect the new prompts and the option to use the force flag for bypassing.

Looking forward to comments! 

Also here is some dog pictures of my dogs!
Looking forward to comments! 

Also here is some dog pictures of my dogs!

      passes our tests.

image
image


src/commands/env/env-set.ts
```diff
@@ -29,6 +30,8 @@ const setInEnvelope = async ({ api, context, key, scope, secret, siteInfo, value
   }
 
   // fetch envelope env vars
+  // const userData = await api.getAccount({accountId})
+  // log(userData)
```
DanielSLew, line 34:
```suggestion
```

src/utils/prompts/blob-set-prompt.ts
```diff
@@ -0,0 +1,18 @@
...
+import { confirmPrompt } from './confirm-prompt.js'
+
+const generateBlobWarningMessage = (key: string, storeName: string): void => {
+  log()
+  log(`${chalk.redBright('Warning')}: The following blob key already exists in store ${chalk.cyan(storeName)}:`)
+  log()
+  log(`${chalk.bold(key)}`)
+  log()
+  log(`This operation will ${chalk.redBright('overwrite')} the existing value.`)
+  log(`${chalk.yellowBright('Notice')}: To overwrite without this warning, you can use the --force flag.`)
```
DanielSLew, line 12: There's a lot of places where we re-use the same messages. Can we extract all these messages out to a single location to make it easier to change the message for all these commands

DanielSLew: A pattern I like for this type of thing is doing something like 
```
const destructiveCommandMessages = {
  blobs: {
    exist: 'yadda yadda',
    delete: 'yadda yadda'
  },
  clone: {
    something: 'yadda yadda',
    somethingElse: 'yadda yadda'
  }
  general: {
     noWarning: `${chalk.yellowBright('Notice')}: To overwrite without this warning, you can use the --force flag.`
  }
}

log(`${destructiveCommandMessages.blobs.exists} ${chalk.cyan(storeName)}:`)
log(destructiveCommandMessages.general.noWarning)
```

Doesn't have to be exactly like this but basically throwing all your related messages into a single object to make it easier to manage the messaging for a specific feature. Also helps with reusability

DanielSLew: You can also make them functions if you want to pass in a variable. like 
```
const destructiveCommandMessages = {
  blobs: {
    exist: 'yadda yadda',
    delete: (key) => 'yadda yadda ${key}'
  },
```

src/utils/prompts/env-clone-prompt.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type EnvVar = {
+  key: string
+  scopes: string[]
+  values: Record<string, any>[]
+  updated_at: string
+  updated_by: User
+  is_secret: boolean
+}
```
DanielSLew, line 19: let's throw these in the types file

tests/integration/commands/blobs/blobs.test.ts
```diff
@@ -64,12 +64,14 @@ describe('blobs:* commands', () => {
   })
 
   setupFixtureTests('empty-project', { mockApi: { routes } }, () => {
+    const expectedSucusesMessage = 'Success: Blob my-key set in store my-store'
```
DanielSLew, line 67:
```suggestion
    const expectedSuccessMessage = 'Success: Blob my-key set in store my-store'
```

DanielSLew (changes requested):
Awesome, thank you for you doing this! Just a couple suggestions!

tlane25:
Updates
1. Scripted command prompt override:
    - added the property `scriptedCommand` to `BaseCommand` to override prompts when hazardous commands run in non-interactive shell or CI environment.
2. Testing for Scripted command prompt override
3. Centralized Message Storage:
    - Created a new file called `prompt-messages` in the prompts folder
    - This file now serves as a central repository for all prompt-related messages
4. Streamlined Prompt Files:
    - Cleaned up and standardized the format of all prompt files
    - Made them more concise and consistent with each other
5. Test Updates:
    - Modified existing tests to reference messages from the new centralized store
    - This ensures all tests now use the same standardized messages

src/commands/blobs/blobs-delete.ts
```diff
@@ -1,21 +1,35 @@
...
 
 /**
  * The blobs:delete command
  */
 export const blobsDelete = async (storeName: string, key: string, _options: Record<string, unknown>, command: any) => {
+  // Prevents prompts from blocking scripted commands
+  if (command.scriptedCommand) {
+    _options.force = true
+  }
+
```
DanielSLew, line 14: Can we do this top level? Ie can we just add it to every option for every command at when we're instantiating the command rather than doing it for every individual command?

src/utils/prompts/blob-delete-prompts.ts
```diff
@@ -0,0 +1,17 @@
...
+  const { overwriteNoticeMessage } = destructiveCommandMessages
+  const { generateWarningMessage, overwriteConfirmationMessage } = destructiveCommandMessages.blobDelete
+
+  const warningMessage = generateWarningMessage(key, storeName)
+
+  log()
+  log(warningMessage)
+  log()
+  log(overwriteNoticeMessage)
+  await confirmPrompt(overwriteConfirmationMessage)
```
DanielSLew, line 16: This reads a lot easier. The way you have it destructing you have to logically read it backwards to understand what it's doing.

Also have a `somethingMessages.somethingElseMessage` is redundant, you don't need the extra `Message` on the end.

```suggestion
  const warningMessage = destructiveCommandMessages.blobDelete.generateWarning(key, storeName)

  log()
  log(warningMessage)
  log()
  log(destructiveCommandMessages.overwriteNotice)
  await confirmPrompt(destructiveCommandMessages.blobDelete.overwriteConfirmation)
```

DanielSLew: I won't make a comment for each of the `*-prompt.ts` files but it's the same for each

DanielSLew (changes requested):
Looks good! Just have a request to clean up the prompt files a little bit!

src/commands/blobs/blobs-delete.ts
tlane25: We decided on this approach as it's more precise (only changes behavior on the commands we added prompting to). We can update to `--force` all commands within CI/non-interactive shell.

wconrad265:
Updates

- Created central location for prompt messages
- Refactored the individual prompt files to not de-structure messages
- Prompts will no longer show in non-interactive shell sessions or CI/CD environment
  - These environments are automatically detected and the force flag is only added to the appropriate commands
  - extracted this logic to a reusable function that can also be used for testing 
- Created a set for the commands that need the `--force` flag added, commands in this set will have the `--force` flag automatically added as a flag option
- Refactored tests for new prompts files
  - Created a new ENV variable that is added, for testing prompts so they can be tested in CI/CD environments
- Added tests to verify that prompts will not show in non-interactive shell sessions or CI/CD environments
- Fixed a bug where tests that mutated the `proces.env` were not restoring the env variable to the original state, that was causing tests to fail. These tests would pass if you ran the test file by its self, but fail when running multiple tests together such as the integration tests above.  
- Fixed a bug in the `src/commands/intergration/deploy.ts` file.
 - When you import `{ env }` from `'process'`, it takes a snapshot of the `env` variable. If tests before this call `{ env }`, then it would not have the properties that were added by the test, causing the `env.INTEGRATION_URL` to be undefined. This happened  when multiple tests files were run in a certain order together. Test should no longer be flakey

src/commands/addons/addons.ts
wconrad265, line 1: This is the old `--force` option, the `-f` option no longer works. Do we want to add `-f` for just this option. Is this ok?

src/commands/init/index.ts
wconrad265, line 1: This is another spot, where the message will be changed from this message to a more generic message, for `--force`

src/commands/lm/lm.ts
wconrad265, line 1: This is another place as well, the `-f` option no longer works.

src/commands/lm/lm.ts
```diff
@@ -33,7 +33,7 @@ and configures your Git environment with the right credentials.`,
     .command('lm:setup', { hidden: true })
     .description('Configures your site to use Netlify Large Media')
     .option('-s, --skip-install', 'Skip the credentials helper installation check')
-    .option('-f, --force-install', 'Force the credentials helper installation')
```
wconrad265, line 36: This is another location where we use `--force-install`, and `-f`. In the new update, it is a unified command of `--force`

src/commands/lm/lm.ts
```diff
@@ -23,7 +23,7 @@ export const createLmCommand = (program: BaseCommand) => {
 It installs the required credentials helper for Git,
 and configures your Git environment with the right credentials.`,
     )
-    .option('-f, --force', 'Force the credentials helper installation')
```
wconrad265, line 26: This is another place as well, the -f option no longer works. It is now just `--force` with a generic message

src/commands/sites/sites.ts
```diff
@@ -81,7 +81,7 @@ export const createSitesCommand = (program: BaseCommand) => {
     .command('sites:delete')
     .description('Delete a site\nThis command will permanently delete the site on Netlify. Use with caution.')
     .argument('<siteId>', 'Site ID to delete.')
-    .option('-f, --force', 'delete without prompting (useful for CI)')
```
wconrad265, line 84: Another place where just `--force` will now work

src/commands/main.ts
```diff
@@ -230,5 +242,11 @@ export const createMainCommand = () => {
     })
     .action(mainCommand)
 
+  program.commands.forEach((cmd) => {
```
wconrad265, line 245: This is where we add the `--force` option to each command that is in the set above. This allows the `--force` option to be unified across all commands, for adding the `--force` flag in ci/cd environments.

src/commands/addons/addons.ts
```diff
@@ -51,7 +51,7 @@ Add-ons are a way to extend the functionality of your Netlify site`,
     .description(
       `Remove an add-on extension to your site\nAdd-ons are a way to extend the functionality of your Netlify site`,
     )
-    .option('-f, --force', 'delete without prompting (useful for CI)')
+    // .option('-f, --force', 'delete without prompting (useful for CI)')
```
DanielSLew, line 54: Why is this commented out?

tests/integration/commands/blobs/blobs-set.test.ts
```diff
@@ -0,0 +1,284 @@
...
+          const mockSet = vi.fn().mockResolvedValue('true')
+
+          ;(getStore as any).mockReturnValue({
+            get: mockGet,
+            set: mockSet,
+          })
+
+          const promptSpy = spyOnMockPrompt()
+
+          await runMockProgram(['', '', 'blobs:set', storeName, key, newValue, '--force'])
```
DanielSLew, line 274: I might be misunderstanding but don't we want to not include the `--force` arg here so that we test is it gets set automatically?
```suggestion
          await runMockProgram(['', '', 'blobs:set', storeName, key, newValue])
```

tests/integration/commands/dev/dev.exec.test.js
```diff
@@ -5,7 +5,7 @@ import { test } from 'vitest'
 import { callCli } from '../../utils/call-cli.js'
 import { getCLIOptions, withMockApi } from '../../utils/mock-api.js'
 import { withSiteBuilder } from '../../utils/site-builder.ts'
-import routes from '../env/api-routes.ts'
+import { routes } from '../env/api-routes.ts'
```
DanielSLew, line 8: Is this import used anywhere else?

DanielSLew (changes requested):
a couple questions and small things to fix!

src/commands/addons/addons.ts
wconrad265: It was left over, it was there as a placeholder.

tests/integration/commands/blobs/blobs-set.test.ts
wconrad265: Done.

tests/integration/commands/dev/dev.exec.test.js
wconrad265: It is used for all of the `env` test files.

wconrad265:
Hi Daniel,

The above changes have been made. 

Also fixed an issue with an integration test that was added. We also made sure that the force option being added is backwards compatible as well.

DanielSLew (approved):
Really awesome work here!

PR #6880, opened 2024-10-15
chore: fix TypeScript errors in login and logout files

dylanspyer:

#### Summary

Fixes conversion to TypeScript

Fixed TS errors in login and logout.
Created `tokenTuple` type for the `getToken` function's return type
Updated `getToken` function calls to remove TypeScript ignore comments

---

      passes our tests.


src/commands/login/login.ts
```diff
@@ -3,8 +3,7 @@ import { OptionValues } from 'commander'
 import { chalk, exit, getToken, log } from '../../utils/command-helpers.js'
 import BaseCommand from '../base-command.js'
 
-// @ts-expect-error TS(7006) FIXME: Parameter 'location' implicitly has an 'any' type.
-const msg = function (location) {
+const msg = function (location: 'env' | 'flag' | 'config' | 'not found') {
```
DanielSLew, line 6: you can extract this type and move it into the types file as it's used later as well

src/commands/integration/deploy.ts
```diff
@@ -394,8 +394,10 @@ export const getConfiguration = (workingDir) => {
 export const deploy = async (options: OptionValues, command: BaseCommand) => {
   const { api, cachedConfig, site, siteInfo } = command.netlify
   const { id: siteId } = site
-  // @ts-expect-error TS(2554) FIXME: Expected 1 arguments, but got 0.
-  const [token] = await getToken()
+  let [token] = await getToken()
+  if (!token) {
+    token = ''
+  }
```
DanielSLew, line 400: why did this need to change?

src/utils/command-helpers.ts
```diff
@@ -143,7 +151,7 @@ export const getToken = async (tokenFromOptions) => {
   if (tokenFromConfig) {
     return [tokenFromConfig, 'config']
   }
-  return [null, 'not found']
+  return [undefined, 'not found']
```
DanielSLew, line 154: why did this change as well?

DanielSLew (changes requested):

src/commands/integration/deploy.ts
dylanspyer: Prior to our changes, `getToken` returned an `any` type. In reality, this value would be a `string` or `undefined`. The `fetch` request below it (see screenshot) relies on `token` being a `string` type. Because `getToken` can return `undefined`, and we don't want to pass `undefined` to the `fetch` request, we reassigned it to an empty string. 

image
image

It does seem that our change may have unintended consequences as a result of changing `token` from `undefined` to an empty string in cases where there is no token. We are pushing a change that goes about fixing this in a way that does not involve reassigning `token`.

src/utils/command-helpers.ts
dylanspyer: This was originally intended to satisfy the type requirements of a pair of functions that use `getToken`'s return value. Our updated commit restores the original return value and creates a new variable whose value is set based on `getToken`'s return value to preserve type safety for these function calls.

src/utils/types.ts
```diff
@@ -62,3 +62,5 @@ export interface Request extends IncomingMessage {
 }
 
 export type Rewriter = (req: Request) => Match | null
+
+export type Location = 'env' | 'flag' | 'config' | 'not found'
```
DanielSLew, line 66: Location is vague and has lots of meaning, let's be more specific here
```suggestion
export type TokenLocation = 'env' | 'flag' | 'config' | 'not found'
```

DanielSLew (changes requested):

src/utils/types.ts
dylanspyer: Updated!

DanielSLew (approved):

PR #6851, opened 2024-10-02
feat(deploy): print edge function logs url

serhalp:
### Summary

The `netlify deploy` command prints the build logs URL and the function logs URL:

Screenshot 2024-10-02 at 10 59 27

It's inconvenient and inconsistent that this doesn't also include a link to edge function logs. So this PR adds it.

#### bee888c88fd0a7c28aa4b3c5204ce57f2ed669eb: fix issues uncovered by above

- `name` in the JSON output was coming from a field that did not exist, so it was always undefined (and `undefined` values are omitted when JSON stringifying), so I just removed it, as it's effectively never truly existed
  - I'm not sure what this was supposed to be anyway. There's already a `site_name` field which is working as intended.
- `site_id` never worked (or at least hasn't for some time) as it was reading from a field that didn't exist. I fixed this one.

#### 8bee1a9290c7ae481f1985d7ee0acf08df3eea70: add edge function logs URL to pretty output

Example:
Screenshot 2024-10-02 at 11 00 49

#### aed9afed559761f4f6aee5c93fc9dfd2bd0e5889: add `function_logs` and `edge_function_logs` to `--json` output

It wasn't consistent that these weren't included in the JSON output. These also could be useful to users. But mostly it allowed me to more easily add test coverage (none of the printed URLs — neither pretty nor json — had test coverage).

Example before:

```json
{
  "site_name": "hydrogen-remix-template",
  "deploy_id": "66fd5eacf914aa188272ebe8",
  "deploy_url": "https://66fd5eacf914aa188272ebe8--hydrogen-remix-template.netlify.app",
  "logs": "https://app.netlify.com/sites/hydrogen-remix-template/deploys/66fd5eacf914aa188272ebe8"
}
```

Example after:

```json
{
  "site_id": "da7ae002-b3be-4dd5-a294-d44321e2459f",
  "site_name": "hydrogen-remix-template",
  "deploy_id": "66fd5ef32443dc1423ce705b",
  "deploy_url": "https://66fd5ef32443dc1423ce705b--hydrogen-remix-template.netlify.app",
  "logs": "https://app.netlify.com/sites/hydrogen-remix-template/deploys/66fd5ef32443dc1423ce705b",
  "function_logs": "https://app.netlify.com/sites/hydrogen-remix-template/logs/functions?scope=deploy:66fd5ef32443dc1423ce705b",
[...]

PR #6881, opened 2024-10-17
fix(deploy): fix edge function logs url scope key

serhalp:
Follow-up to https://github.com/netlify/cli/pull/6851.

You can see here it says "Show logs for deploy...": https://app.netlify.com/sites/hydrogen-advanced-caching/logs/edge-functions?scope=deployid:6705473738103f2856071dbd.

This doesn't work with `?scope=deploy:`.

This is a bit of a footgun since the Functions log URL takes `deploy:` but the Edge Functions log URL takes `deployid:`. 😓

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -536,7 +536,7 @@ const runDeploy = async ({
 
   if (!deployToProduction) {
     functionLogsUrl += `?scope=deploy:${deployId}`
-    edgeFunctionLogsUrl += `?scope=deploy:${deployId}`
+    edgeFunctionLogsUrl += `?scope=deployid:${deployId}`
   }
 
   return {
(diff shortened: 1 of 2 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -536,7 +536,7 @@ const runDeploy = async ({
 
   if (!deployToProduction) {
     functionLogsUrl += `?scope=deploy:${deployId}`
-    edgeFunctionLogsUrl += `?scope=deploy:${deployId}`
+    edgeFunctionLogsUrl += `?scope=deployid:${deployId}`
```
mrstork, line 539: Maybe we do a follow-up PR to make these two match? The `deploy` naming seems better.

mrstork (approved):

src/commands/deploy/deploy.ts
serhalp: The downside is we'd effectively have to support both keys forever. Well, maybe not forever, but we'd have to do a two-phase change with several months in between.

Issue #6724, opened 2024-06-23
Completions don't work


alexschcom:
### Describe the bug

I ran `netlify completion:install`, with the following output:

```
? Which Shell do you use ? zsh
? We will install completion to ~/.zshrc, is it ok ? Yes
=> Tabtab line already exists in ~/.config/tabtab/__tabtab.zsh file
=> Tabtab line already exists in ~/.zshrc file
=> Wrote completion script to /Users/alex/.config/tabtab/netlify.zsh file

      => Tabtab source line added to ~/.zshrc for netlify package.

      Make sure to reload your SHELL.
    
Completion for netlify successful installed!
```

But no completions when tabbing :(

### Steps to reproduce

1. run `netlify completion:install`
2. type `netlify s`
3. press <kbd>tab</kbd>

Completions for `serve`, `sites`, `status`, `switch` should appear

### Environment

System:
  OS: macOS 14.5
  CPU: (8) arm64 Apple M1
  Memory: 63.78 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.3.0 - /opt/homebrew/bin/node
  npm: 10.8.1 - /opt/homebrew/bin/npm
  pnpm: 8.15.6 - /opt/homebrew/bin/pnpm
npmGlobalPackages:
  netlify-cli: 17.29.0

benhancock:
@dylanspyer and I had this issue too on zsh. There are at least two issues that might cause completion to fail 

1. On Linux/MacOS, the completion script may not have the necessary executable permissions.
- Fix: Run chmod +x with the file path to the completion script, typically `/Users/yourname/projects/netlify/netlify-cli/dist/lib/completion/script.js`
2. Netlify uses [Tabtab](https://github.com/denosaurs/tabtab) for completions, and with zsh, Tabtab (and other) completions do not work until [compinit](https://github.com/zsh-users/zsh/blob/master/Completion/compinit) has been run.
- Fix: Add the line `autoload -U compinit; compinit` to the user’s `~/.zshrc` file above the Tabtab config line to load and then run `compinit`

[This PR](https://github.com/netlify/cli/pull/6882) attempts to address both of these issues by:
1. automatically outputting the `chmod +x `command with the appropriate file path for the user to copy and run if needed after the installation is complete;
2. prompting the user to have the `autoload -U compinit; compinit` line added to the top of their `~/.zshrc` file automatically if not already present.

alexschcom:
[...]

PR #6877, opened 2024-10-15
chore: created new type for API errors, created and implemented type predicates

dylanspyer:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes conversion to TypeScript

Previously, most invocations of the custom `error` function needed to have a TypeScript ignore comment due to the `error_` object being passed to the function being of type `unknown`. Because the `error` function already takes `error_` arguments that are of various different shapes, adding `unknown` to the parameter's type union allowed us to remove these TypeScript ignore comments without sacrificing type safety. Added a new type `APIError` and type predicates for custom type guarding of errors that are a result of an internal API call.

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #6882, opened 2024-10-17
fix: add zsh autocomplete setup and file permissions instructions to completion:install

benhancock:

#### Summary

Fixes [Completions don't work](https://github.com/netlify/cli/issues/6724)

There are at least two issues that might cause completions to fail after running `completion:install`:

1. On Linux/MacOS, the completion script may not have the necessary executable permissions.
- **Fix:** Run `chmod +x` with the file path to the completion script

2. Netlify uses [Tabtab](https://github.com/denosaurs/tabtab) for completions, and with zsh, Tabtab (and other) completions do not work until [compinit](https://github.com/zsh-users/zsh/blob/master/Completion/compinit) has been run. 
- **Fix:** Add the line `autoload -U compinit; compinit` to the user’s `~/.zshrc` file above the Tabtab config line to load and then run `compinit`

This PR addresses both of these issues by:
1. automatically outputting the `chmod +x` command with the appropriate file path for the user to copy and run if needed after the installation is complete;
2. prompting the user to have the `autoload -U compinit; compinit` line added to the top of their .zshrc file automatically if not already present.

Here's what the `netlify completion:install` output looks like before:

Screenshot 2024-10-17 at 11 44 44 AM

Here's what the `netlify completion:install` output looks like after our changes:

Screenshot 2024-10-17 at 11 45 41 AM

This PR also creates a new test file for `completion:install` with tests for the `compinit -> ~/.zshrc` functionality.

---

      passes our tests.


dylanspyer:
We decided not to modify file permissions on the user's behalf and instead output the `chmod` command for the user to run manually. This ensures that the user maintains ownership of their security and configuration settings.

src/commands/completion/completion.ts
```diff
@@ -20,13 +23,53 @@ export const completionGenerate = async (options: OptionValues, command: BaseCom
...
+
+          log('Successfully added compinit line to .zshrc')
+        }
+      })
+  }
+
+  log(`Completion for ${parent.name()} successfully installed!`)
+
+  if (process.platform !== 'win32') {
+    log("\nTo ensure proper functionality, you'll need to set appropriate file permissions.")
```
benhancock, line 67: Because `chmod` is a Unix command, we only output this instruction if the user is not on windows

src/commands/completion/completion.ts
```diff
@@ -20,13 +23,53 @@ export const completionGenerate = async (options: OptionValues, command: BaseCom
...
   generateAutocompletion(parent)
-
   await install({
     name: parent.name(),
     completer,
   })
 
-  console.log(`Completion for ${parent.name()} successful installed!`)
+  const TABTAB_CONFIG_LINE = '[[ -f ~/.config/tabtab/__tabtab.zsh ]] && . ~/.config/tabtab/__tabtab.zsh || true'
+  const AUTOLOAD_COMPINIT = 'autoload -U compinit; compinit'
```
DanielSLew, line 32: Throw these in the constants file in this folder

src/commands/completion/completion.ts
```diff
@@ -20,13 +23,53 @@ export const completionGenerate = async (options: OptionValues, command: BaseCom
...
+    await inquirer
+      .prompt([
+        {
+          type: 'confirm',
+          name: 'compinitAdded',
+          message: `Would you like to add it?`,
+          default: true,
+        },
+      ])
+      .then((answer) => {
```
DanielSLew, line 51: Generally you wouldn't use `await` and `.then()` you would use one or the other. If you want it to be blocking use `await`, if you want it to not be blocking use `.then()`

```suggestion
    const answer = await inquirer
      .prompt([
        {
          type: 'confirm',
          name: 'compinitAdded',
          message: `Would you like to add it?`,
          default: true,
        },
      ])
```

src/commands/completion/completion.ts
```diff
@@ -20,13 +23,53 @@ export const completionGenerate = async (options: OptionValues, command: BaseCom
...
+        if (answer['compinitAdded']) {
+          fs.readFile(zshConfigFilepath, 'utf8', (err, data) => {
```
DanielSLew, line 53: I would await this. So await fs.readFile or use fs.readFileSync.

src/utils/command-helpers.ts
```diff
@@ -300,3 +301,14 @@ export const nonNullable = <T>(value: T): value is NonNullable<T> => value !== n
 export const noOp = () => {
   // no-op
 }
+
+export const checkFileForLine = (filename: string, line: string) => {
+  let filecontent = ''
+  try {
+    filecontent = fs.readFileSync(filename, 'utf8')
+  } catch (error_) {
+    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
```
DanielSLew, line 310: Pull in the main branch to get updates from this PR https://github.com/netlify/cli/pull/6877 to fix this type error

tests/integration/commands/completion/completion-install.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import fs from 'fs'
+import { rm } from 'fs/promises'
+import { temporaryDirectory } from 'tempy'
+import { handleQuestions, CONFIRM, DOWN, NO, answerWithValue } from '../../utils/handle-questions.js'
+import execa from 'execa'
+import { cliPath } from '../../utils/cli-path.js'
+import { join } from 'path'
+
+const TABTAB_CONFIG_LINE = '[[ -f ~/.config/tabtab/__tabtab.zsh ]] && . ~/.config/tabtab/__tabtab.zsh || true'
+const AUTOLOAD_COMPINIT = 'autoload -U compinit; compinit'
```
DanielSLew, line 11: pull these from `constants` when you make the change from above

DanielSLew (changes requested):
Looks great! Just a few changes that I've requested

tests/integration/commands/completion/completion-install.test.ts
DanielSLew, line 11: Use the exports you created in the other file already

DanielSLew (changes requested):

DanielSLew (approved):

Issue #6736, opened 2024-06-27
Misleading error message when an invalid `NETLIFY_AUTH_TOKEN` is provided


zetaraku:
### Describe the bug

A confusing error message is shown when running `netlify deploy` with an invalid `NETLIFY_AUTH_TOKEN`:

```
»   Error: Site not found. Please rerun "netlify link"
```

I did some inspection and found out that a `TextHTTPError: Unauthorized` error is thrown from line 6 instead of expected line 11 in this file:

https://github.com/netlify/cli/blob/73f4eb59f859adfee142b7213e70168a8cec7681/src/utils/get-site.ts#L4-L18

See also:

- https://answers.netlify.com/t/error-site-not-found-please-rerun-netlify-link/117207

### Steps to reproduce

1. Run `netlify deploy` with a valid `NETLIFY_SITE_ID` and an invalid `NETLIFY_AUTH_TOKEN`.

### Configuration

```toml
[build]
  command = "npm run build"
  publish = "./dist/"
```

### Environment

```
System:
  OS: Windows 10 10.0.19045
  CPU: (8) x64 Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz
  Memory: 1.44 GB / 11.84 GB
Binaries:
  Node: 20.12.0 - C:\Program Files\nodejs\node.EXE      
  Yarn: 1.22.19 - ~\.yarn\bin\yarn.CMD
  npm: 10.8.1 - C:\Program Files\nodejs\npm.CMD
  pnpm: 8.15.5 - C:\Program Files\nodejs\pnpm.CMD
```

PR #6884, opened 2024-10-17
fix: updated error message for attempting to deploy a site with a bad…

dylanspyer:
**🎉** Thanks for submitting a pull request! 🎉

#### Summary

Fixes: [6736](https://github.com/netlify/cli/issues/6736)

Previously, `netlify deploy` with a valid site ID and an invalid auth token would log `Error: Site not found. Please rerun "netlify link"`. Updated this error message to be more clear by logging the `message` property from the API response when the response status is 401.

**Before**: 
image

**After**: 
image

---

      passes our tests.

image


dylanspyer:
> Hi @dylanspyer! Thank you so much for your contribution :) could you maybe still log 'site not found' but also render the error message? I think this could inform our users better on what part of the process went wrong and why?

Thanks for the feedback! I pushed an update with a clearer error message:
image

In this failure case, it's possible that the site id is valid and the auth token is invalid. Logging `site not found` in that case might be misleading. Let me know your thoughts :) @khendrikse

src/utils/get-site.ts
```diff
@@ -12,7 +12,13 @@ export const getSiteByName = async (api, siteName) => {
     }
 
     return siteFoundByName
-  } catch {
-    error('Site not found. Please rerun "netlify link"')
+  } catch (error_) {
+    // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
+    if (error_.status === 401) {
+      // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
```
DanielSLew, line 18: Fix these TS errors with pulling in this PR https://github.com/netlify/cli/pull/6877.

DanielSLew (changes requested):
This looks good to me! Just have a request to remove those type errors.

src/utils/get-site.ts
dylanspyer: Fixed!

DanielSLew (approved):

Issue #6840, opened 2024-09-26
Function with serverless app crashes after second request running in `netlify dev`


rlmcneary2:
### Describe the bug

I have a function named "index" that uses serverless to return a Koa app. The function builds successfully and starts using dev. If I make requests against the root endpoint using a tool like postman, the cli will crash after the second request completes. It crashes in the browser too but it's easier to see the behavior in a tool that makes a single request.

### Steps to reproduce

1. `git clone https://codeberg.org/rlmcneary2/funcstache-website.git`
2. Switch to branch "cli-crash"
3. `npm i`
4. `npm run build && npm run dev`
5. Open a tool like postman or rested and send a request to "http://localhost:8888/"
6. Send the request again.

Following the second request others will fail because the cli process has exited.

```sh
◈ Rewrote URL to /.netlify/functions/index/
Request from 127.0.0.1: GET /.netlify/functions/index/
--- index: ENTER[catcher] status=404
--- index: ENTER[funcstache] status=404
funcstache(middleware): url='/'
funcstache(middleware): requestedUrl='site/'
--- index: koa app started; http://localhost:3000
getTemplatePartials: source template='layout', add to pending: { sourceTemplateDir: '', templateFilename: '', templateName: 'main' }
getTemplatePartials: source template='layout', add to pending: {
  sourceTemplateDir: './layouts/default',
  templateFilename: '../../shared/nav/nav',
  templateName: 'navItems'
}
readTemplateFile: templateName='navItems', sourceTemplateDir='./layouts/default'
readTemplateFile: templateName='main', sourceTemplateDir=''
readTemplateFile: templateName='layout', sourceTemplateDir=''
funcstache(middleware): url='/'; exit
--- index: EXIT[funcstache] status=200
--- index: EXIT[catcher] status=200
Response with status 200 in 68 ms.
◈ Rewrote URL to /.netlify/functions/index/
Request from 127.0.0.1: GET /.netlify/functions/index/
--- index: ENTER[catcher] status=404
--- index: ENTER[funcstache] status=404
funcstache(middleware): url='/'
funcstache(middleware): requestedUrl='site/'

 ›   Error: Netlify CLI has terminated unexpectedly
This is a problem with the Netlify CLI, not with your application.
If you recently updated the CLI, consider reverting to an older version by running:

npm install -g netlify-cli@VERSION
You can use any version from https://ntl.fyi/cli-versions.
Please report this problem at https://ntl.fyi/cli-error including the error details below.

[...]

Issue #6805, opened 2024-09-03
Error: Netlify CLI has terminated unexpectedly due to EADDRINUSE on port 5200


ibrahimsaber1:
### Describe the bug

"While deploying my MERN stack project using the Netlify CLI, I encountered an unexpected termination with the error EADDRINUSE. The CLI terminated because the address on port 5200 was already in use. This issue seems to be related to the Netlify CLI itself rather than my application."

Error details:

Error: Netlify CLI has terminated unexpectedly
This is a problem with the Netlify CLI, not with your application.

Error: listen EADDRINUSE: address already in use :::5200
    at Server.setupListenHandle [as _listen2] (node:net:1904:16)
    at listenInCluster (node:net:1961:12)
    at Server.listen (node:net:2063:7)
    at Function.listen (C:\Users\ibRABHIM Saber\Downloads\deploy\mernStackMilestoneProject_ITI-main\backend\node_modules\express\lib\application.js:635:24)
    at Object.<anonymous> (C:\Users\ibRABHIM Saber\Downloads\deploy\mernStackMilestoneProject_ITI-main\backend\netlify\functions\index.js:46:5)
    at Module._compile (node:internal/modules/cjs/loader:1358:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1024:12)
    at Module.require (node:internal/modules/cjs/loader:1233:19)

### Steps to reproduce

Clone the repository using git clone https://github.com/ibrahimsaber1/good-reads
Navigate to the cloned repository directory.
Run the command netlify dev to start the development server.
Observe that the Netlify CLI exits with an error related to the port being in use.

### Configuration

[build]
functions = "netlify/functions"

[[redirects]]
from = "/*"
to = "/.netlify/functions/index"
status = 200

### Environment

npx envinfo --system --binaries --npmPackages netlify-cli --npmGlobalPackages netlify-cli

rlmcneary2:
Seeing this error when running the dev server for a function. First request succeeds, second request causes the error.
netlify-cli: 17.36.2
repo: https://codeberg.org/rlmcneary2/funcstache-website.git
tag to reproduce error: netlify-issue-6805
clone: `git clone --branch netlify-issue-6805 https://codeberg.org/rlmcneary2/funcstache-website.git`

PR #6887, opened 2024-10-18
feat: port in use

wconrad265:

#### Summary

This pull request addresses these issues found below

- https://github.com/netlify/cli/issues/6840
- https://github.com/netlify/cli/issues/6805

We (@tlane25) outline the problem in the comments in https://github.com/netlify/cli/issues/6840.

This pull request adds better error handling for users who open a port for a serverless function. 

---

      passes our tests.


src/commands/main.ts
```diff
@@ -40,27 +40,39 @@ import { createWatchCommand } from './watch/index.js'
 
 const SUGGESTION_TIMEOUT = 1e4
 
-process.on('uncaughtException', async (err) => {
+process.on('uncaughtException', async (err: any) => {
```
DanielSLew, line 43: Let's use a proper type here rather than adding in an `any`

DanielSLew (changes requested):
Looks good, just a small nit about `any` types.

wconrad265:
This has been updated with a new type.

layout.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
```
DanielSLew, line 1: why was this changed?

src/commands/types.d.ts
```diff
@@ -73,3 +73,11 @@ export type NetlifyOptions = {
   state: StateConfig
   frameworksAPIPaths: FrameworksAPIPaths
 }
+
+export interface AddressInUseError extends Error {
```
DanielSLew, line 77: I would just add this into the same error type as optional fields

package.json
```diff
@@ -161,6 +161,8 @@
     "parse-github-url": "1.0.3",
     "parse-gitignore": "2.0.0",
     "path-key": "4.0.0",
+    "pnpm": "^9.12.2",
+    "prettier": "^3.3.3",
```
DanielSLew, line 165: why were these added for this PR?

src/commands/types.d.ts
tlane25: The `Error` type lives in `~external/38d21a27-4378-4895-a56a-d0fa5726a379/lib.es2022.error.d.ts`. I think it's from an dependency/ external source.

layout.html
wconrad265: image

When we were trying to push this commit, we kept having a prettier format error for this file. This is what it recommended that we do, to update to a lowercase `doctype`. From my understanding, there is no functional difference between the two.

package.json
wconrad265: I believe we were working on an new machine, and they were accidently not install locally, they have been removed in the update.

DanielSLew (approved):

wconrad265:
The tests have been fixed. Ready for re-review

wconrad265:
Look like this might be a flakey test. It passed in every single other environment, and passes the test locally. The tests were passing until I merged the latest version of main.

DanielSLew (approved):

PR #6896, opened 2024-10-23
feat: rename next.js adapter package

orinokai:
#### Summary

Update to use new package name for our Next.js Adapter, which is moving to the @opennextjs scope.

#### To do

- [ ] when https://github.com/netlify/build/pull/5897 is released, bump it in this PR (tests won't pass otherwise)

### DIFF
--- a/src/utils/init/utils.ts
+++ b/src/utils/init/utils.ts
@@ -30,7 +30,7 @@ const formatTitle = (title: string) => chalk.cyan(title)
  * @returns
  */
 export const getPluginsToAutoInstall = (
-  command: BaseCommand,
+  _command: BaseCommand,
   pluginsInstalled: string[] = [],
   pluginsRecommended: string[] = [],
 ) => {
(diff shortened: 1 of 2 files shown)


src/utils/init/utils.ts
```diff
@@ -34,7 +34,7 @@ export const getPluginsToAutoInstall = (
   pluginsInstalled: string[] = [],
   pluginsRecommended: string[] = [],
 ) => {
-  const nextRuntime = '@netlify/plugin-nextjs'
+  const nextRuntime = '@opennextjs/netlify'
```
serhalp, line 37: This took me a while to understand — this is part of 'init' i.e. used in `ntl init` and related scenarios, and its purpose is to literally auto-install the plugin on the site via the Netlify API if it isn't already installed and we've detected that it should be. So it's OK to just update this name and not "handle" the old name in any way.

serhalp: ... although, wait, isn't there a case like this?
1. next.js site exists today with @netlify/plugin-nextjs installed
2. we release this cli change
3. user clones their repo, runs `ntl init`, it checks and finds that `@opennextjs/netlify` is **not** auto-installed (even though `@netlify/plugin-nextjs` **is**) and auto-installs `@opennextjs/netlify`.
4. user has both plugins installed

serhalp:
No longer planned

PR #6915, opened 2024-11-05
fix: make outputs more user-friendly for `sites:create-template`

benhancock:

#### Summary

This PR implements two small changes to the `sites:create-template` command to make it more user-friendly.

1. When user validates with GitHub personal access token, the PAT is hidden from the console, so as to protect these sensitive credentials.
2. If creating a GitHub repo fails because of a possible PAT issue, the error output reads "GitHub personal access token" rather than "GitHub PAT"

These are both suggestions from internal Netlify folks. Thanks for the feedback!

---

      passes our tests.

Lion Cub


DanielSLew (approved):
thanks for this quick followup!

PR #6936, opened 2024-11-25
fix: check ip version if settings.useStaticServer

khendrikse:

#### Summary

This hopefully fixes this: https://answers.netlify.com/t/128792

I noticed, while running this on windows in a VM, that the proxy was trying to connect to an IPv4 server, while the static server was actually an IPv6 one. 

With this logic, we explicitly check what version we are running within the `settings.useStaticServer` branch of logic as well. 

---

      passes our tests.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
         # Pinning 20.x version as a temporary workaround due to this https://github.com/nodejs/node/issues/52884
-        node-version: ['18.17.0', '20.12.2', '22']
+        node-version: ['18.17.0', '20.12.2', '22', '23']
         shard: ['1/4', '2/4', '3/4', '4/4']
 
         exclude:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -123,15 +123,12 @@ jobs:
           flags: ${{ steps.test-coverage-flags.outputs.os }},${{ steps.test-coverage-flags.outputs.node }}
           token: ${{ secrets.CODECOV_TOKEN }}
         if: '${{ !steps.release-check.outputs.IS_RELEASE }}'
+  # Specific tests for known test that failed on windows using node 23.
+  # Can be replaced with larger node 23 tests in the future.
   integration-win-node-23:
     name: Integration test windows latest node23 specific
-    runs-on: ${{ matrix.os }}
+    runs-on: windows-latest
     timeout-minutes: 40
-    strategy:
-      matrix:
-        os: [windows-latest]
-        node-version: ['23']
-      fail-fast: false
     steps:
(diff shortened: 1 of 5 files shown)


src/utils/framework-server.ts
```diff
@@ -30,9 +31,11 @@ export const startFrameworkServer = async function ({
     if (settings.command) {
       runCommand(settings.command, { env: settings.env, cwd })
     }
-    await startStaticServer({ settings })
+    const { address } = await startStaticServer({ settings })
```
eduardoboucas, line 34: Rather than getting `address` and then trying to figure out the IP family from the hostname, can you not use the `family` property directly?

https://nodejs.org/api/net.html#serveraddress

eduardoboucas (commented):
Were you able to see the failure in the CI? Any chance we could add a test to verify it's fixed?

src/utils/framework-server.ts
khendrikse: I'm gonna find out!!! yes!

khendrikse:
> Were you able to see the failure in the CI? Any chance we could add a test to verify it's fixed?

No we didn't see it, but I'm going to try out running tests on win + node 23 in the pipelines. Wish me luck lol.

src/utils/framework-server.ts
khendrikse: ✅

DanielSLew (approved):

.github/workflows/integration-tests.yml
```diff
@@ -123,3 +123,105 @@ jobs:
...
           token: ${{ secrets.CODECOV_TOKEN }}
         if: '${{ !steps.release-check.outputs.IS_RELEASE }}'
+  integration-win-node-23:
+    name: Integration test windows latest node23 specific
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 40
+    strategy:
+      matrix:
+        os: [windows-latest]
+        node-version: ['23']
```
eduardoboucas, line 133: Is this not `latest`?

.github/workflows/integration-tests.yml
```diff
@@ -123,3 +123,105 @@ jobs:
           flags: ${{ steps.test-coverage-flags.outputs.os }},${{ steps.test-coverage-flags.outputs.node }}
           token: ${{ secrets.CODECOV_TOKEN }}
         if: '${{ !steps.release-check.outputs.IS_RELEASE }}'
+  integration-win-node-23:
```
eduardoboucas, line 126: What is the long-term plan for this action? I understand we're looking to test this specific version today, but what happens after this goes to LTS and beyond? Can this be removed? Just trying to make sure we're not left with this extra step forever without people understanding why. Perhaps adding a comment here would help?

.github/workflows/integration-tests.yml
khendrikse: I want to be explicit in it being node 23 :)

DanielSLew (approved):

PR #6994, opened 2025-01-10
fix: check if endpoint exists for dynamic pages when checking for red…

VitaliyR:
…irect

#### Summary

Fixes WRFL-2122

Basically we are checking only for static files when trying to determinate if we should do the `non forced` redirect.

In dynamic SSG, when they either run in `dev mode` or having dynamic endpoints (not a static files), we are not detecting it and doing the redirect, which is wrong.

I've added a simple checker via `fetch url, { method: HEAD }` - which checks if the endpoint exists.

#### Example

Assume next `netlify.toml`:

```
[[redirects]]
from = "*"
to = "https://www.netlify.app/:splat"
status = 200
force = false # by default
```

Run `netlify dev`.

Make change to some code file.

Open `http://localhost:8888`. See that change is not applied. See in console `Proxying to https://www.netlify.app/`.

Thats wrong, it should actually detect that there is an index route in current site and not proxy.

---

      passes our tests.


src/utils/proxy.ts
```diff
@@ -149,6 +149,16 @@
   return `/${path.relative(publicFolder, file)}`
 }
 
+const isEndpointExists = async function (endpoint: string, origin: string) {
+  const url = new URL(endpoint, origin)
+  try {
+    const res = await fetch(url, { method: 'HEAD' })
```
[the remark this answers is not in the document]
VitaliyR: Not really sure this is useful, should we dismiss it?

src/utils/proxy.ts
```diff
@@ -149,6 +149,16 @@ const getStatic = async function (pathname: string, publicFolder: string) {
...
```
smnh, line 155: Do we check if the existing endpoint is external (https://...) or some existing dynamic route (localhost:3000/...).

VitaliyR: https://github.com/netlify/cli/pull/6994/files#diff-7c2533ca30d8693b3cecb3406646bc13b86b3969b43a0aa0659bb52c7676caa9R357

`options.target` matches the ssg itself e.g. 3000. If we'd do it to 8888, it will be pointless since we managing routes

Scutua (approved):

Scutua (approved):

Scutua (approved):

smnh (approved):

src/utils/proxy.ts
```diff
@@ -691,9 +703,11 @@ const initializeProxy = async function ({
 
       if (htmlInjections) {
         responseBody = await injectHtml(responseBody, proxyRes, htmlInjections)
-        proxyResHeaders = {
-          ...proxyResHeaders,
-          'content-length': String(responseBody.byteLength),
+        if (!proxyResHeaders['transfer-encoding']) {
```
VitaliyR, line 706: @smnh you added this, but when there is transfer-encoding, you should not send content-length.
https://github.com/sindresorhus/got/discussions/1576

WDYT about this change?

smnh: According to MDN, It looks like it should be ommitted only when `Transfer-Encoding` is `chunked`.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#syntax

And it make sense, because for regular non chunked encodings like gzip and deflate you do want to have content-length.

```
if (proxyResHeaders['transfer-encoding']?.includes('chunked') {
```

smnh (approved):

PR #5816, opened 2023-06-21
feat: Replace framework-info with build-info

Skn0tt:
🎉 Thanks for submitting a pull request! 🎉

#### Summary

Fixes https://github.com/netlify/pod-dev-foundations/issues/540
Fixes https://github.com/netlify/pod-dev-foundations/issues/528

---

For us to review and ship your PR efficiently, please perform the following steps:

- [ ] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻. This ensures we can discuss the changes and get feedback from everyone that should be involved. If you\`re fixing a typo or something that\`s on fire 🔥 (e.g. incident related), you can skip this step.
- [ ] Read the [contribution guidelines](../CONTRIBUTING.md) 📖. This ensures your code follows our style guide and
      passes our tests.
- [ ] Update or add tests (if any source code was changed or added) 🧪
- [ ] Update or add documentation (if features were changed or added) 📝
- [ ] Make sure the status checks below are successful ✅

**A picture of a cute animal (not mandatory, but encouraged)**

PR #7008, opened 2025-01-21
fix!: upgrade to commander@12 and fix some conflicting short option names

serhalp:
#### Summary

This upgrades commander from v10 to v12. (The latest is v13, but we'll cross that bridge later.)

This revealed some errors in some command options, which this PR also fixes.

#### Details

The two main updates in this PR are due to https://github.com/tj/commander.js/pull/2055, which no longer allows redefining the same option twice on the same command:
1. We were doing this intentionally by defining a few options in the `BaseCommand` with the option hidden from help and then redefining it in some specific commands. I decided to bite the bullet here and just refactored these three problematic options:
    - I moved `--json` and `--offline` entirely out of `BaseCommand`. This uncovered that we weren't exhaustively redeclaring this on all the commands that use it. I fixed those.
    - I moved `--auth` entirely _into_ `BaseCommand`. It's intended as an alternative authentication mechanism to `ntl login` and `NETLIFY_AUTH_TOKEN`, so I believe this is reasonable.
2. We were doing this unintentionally in a few places, where we were using the same short flag for two options on the same command, resulting in the last registered one being quietly not registered. I chose new flags for these. You could argue this is a breaking change, but these options seemingly never worked. This will be in a major anyway so it's fine.

This did result in some docs/help changes, but IMO they're good changes. It did result in a handful of commands now having a documented `--auth` option that doesn't make much sense, but I'm not sure it's worth the effort to find some way to hide those.

### DIFF
--- a/docs/commands/completion.md
+++ b/docs/commands/completion.md
@@ -20,6 +20,7 @@ netlify completion
 **Flags**
 
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
 
 | Subcommand | description  |
 |:--------------------------- |:-----|
@@ -47,6 +48,7 @@ netlify completion:install
 
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
 
 ---
 
--- a/docs/commands/login.md
+++ b/docs/commands/login.md
@@ -21,6 +21,7 @@ netlify login
 
 - `new` (*boolean*) - Login to new Netlify account
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
 
 
 <!-- AUTO-GENERATED-CONTENT:END -->
--- a/docs/commands/switch.md
+++ b/docs/commands/switch.md
@@ -19,6 +19,7 @@ netlify switch
 **Flags**
 
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
 
 
 <!-- AUTO-GENERATED-CONTENT:END -->
--- a/docs/commands/unlink.md
+++ b/docs/commands/unlink.md
@@ -20,6 +20,7 @@ netlify unlink
 
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
 
 
 <!-- AUTO-GENERATED-CONTENT:END -->
--- a/site/scripts/util/generate-command-data.js
+++ b/site/scripts/util/generate-command-data.js
@@ -3,13 +3,13 @@ import { sortOptions } from '../../../dist/utils/command-helpers.js'
 
 const program = createMainCommand()
 
-/** @type {Array<import('../../src/commands/base-command.js').default>} */
+/** @type {Array<import('../../../src/commands/base-command.js').default>} */
 // @ts-ignore typecast needed
 const commands = program.commands.sort((cmdA, cmdB) => cmdA.name().localeCompare(cmdB.name()))
 
 /**
  *
- * @param {import('../../src/commands/base-command.js').default} command
+ * @param {import('../../../src/commands/base-command.js').default} command
  */
 const parseCommand = function (command) {
   const args = command._args.map(({ _name: name, description }) => ({
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -188,7 +188,7 @@ export default class BaseCommand extends Command {
       .addOption(new Option('--json', 'Output return values as JSON').hideHelp(true))
       .addOption(new Option('--silent', 'Silence CLI output').hideHelp(true))
       .addOption(new Option('--cwd <cwd>').hideHelp(true))
-      .addOption(new Option('-o, --offline').hideHelp(true))
+      .addOption(new Option('-o, --offline').default(false).hideHelp(true))
       .addOption(new Option('--auth <token>', 'Netlify auth token').hideHelp(true))
       .addOption(
         new Option('--http-proxy [address]', 'Proxy server address to route requests through.')
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -183,13 +183,12 @@ export default class BaseCommand extends Command {
    */
   createCommand(name: string): BaseCommand {
     const base = new BaseCommand(name)
-      // If  --silent or --json flag passed disable logger
       // .addOption(new Option('--force', 'Force command to run. Bypasses prompts for certain destructive commands.'))
-      .addOption(new Option('--json', 'Output return values as JSON').hideHelp(true))
       .addOption(new Option('--silent', 'Silence CLI output').hideHelp(true))
       .addOption(new Option('--cwd <cwd>').hideHelp(true))
-      .addOption(new Option('-o, --offline').default(false).hideHelp(true))
-      .addOption(new Option('--auth <token>', 'Netlify auth token').hideHelp(true))
+      .addOption(
+        new Option('--auth <token>', 'Netlify auth token - can be used to run this command without logging in'),
+      )
       .addOption(
         new Option('--http-proxy [address]', 'Proxy server address to route requests through.')
           .default(process.env.HTTP_PROXY || process.env.HTTPS_PROXY)
--- a/src/commands/blobs/blobs.ts
+++ b/src/commands/blobs/blobs.ts
@@ -1,4 +1,4 @@
-import { OptionValues } from 'commander'
+import { Option, OptionValues } from 'commander'
 
 import requiresSiteInfo from '../../utils/hooks/requires-site-info.js'
 import BaseCommand from '../base-command.js'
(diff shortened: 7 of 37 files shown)


src/commands/base-command.ts
```diff
@@ -192,11 +192,6 @@ export default class BaseCommand extends Command {
       .addOption(new Option('--cwd <cwd>').hideHelp(true))
       .addOption(new Option('-o, --offline').hideHelp(true))
       .addOption(new Option('--auth <token>', 'Netlify auth token').hideHelp(true))
-      .addOption(
-        new Option('--httpProxy [address]', 'Old, prefer --http-proxy. Proxy server address to route requests through.')
-          .default(process.env.HTTP_PROXY || process.env.HTTPS_PROXY)
-          .hideHelp(true),
-      )
```
serhalp, line 199: It looks like there was an `httpProxy` option and an `http-proxy` option but these got accidentally(?) squashed into a duplicated `httpProxy` option in https://github.com/netlify/cli/pull/5816/files#diff-67082469659283bc342bd13f7e20570cb3d1bf5096aad88545a9ac4809d6083bL106-L140. The latest version of commander throws when you declare duplicate options.

What's extra confusing is the copy here implies that `http-proxy` was intended to be the _new_ one but that's not where it landed 🤷🏼.

src/commands/blobs/blobs.ts
```diff
@@ -33,7 +33,7 @@ export const createBlobsCommand = (program: BaseCommand) => {
     )
     .argument('<store>', 'Name of the store')
     .argument('<key>', 'Object key')
-    .option('-o, --output <path>', 'Defines the filesystem path where the blob data should be persisted')
+    .option('-O, --output <path>', 'Defines the filesystem path where the blob data should be persisted')
```
serhalp, line 36: `-o` is already used in the base command for `--offline`, so this was being quietly ignored (I tested it locally).

This is fine to be considered a patch change since it has never worked.

Caught by changes in commander upgrade: https://github.com/tj/commander.js/pull/2055.

src/commands/blobs/blobs.ts
```diff
@@ -53,7 +53,8 @@ export const createBlobsCommand = (program: BaseCommand) => {
       '-p, --prefix <prefix>',
       `A string for filtering down the entries; when specified, only the entries whose key starts with that prefix are returned`,
     )
-    .option('--json', `Output list contents as JSON`)
+    // The BaseCommand defines a `--json` option which is hidden from the help by default
+    .addHelpOption(new Option('--json', 'Output list contents as JSON'))
```
serhalp, line 57: Same as above here, but in this case it's intentional. Just removing this line works, but then `--json` no longer shows up in `ntl blobs:get --help`. This solution seems to cover all the bases.

serhalp: ... and then same for all the command `--offline` cases

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
```diff
@@ -27,6 +27,7 @@ export const detectNetlifyLambda = async function ({ packageJson } = {}) {
       .option('-b, --babelrc [file]')
       .option('-t, --timeout [delay]')
 
+    program.allowExcessArguments()
```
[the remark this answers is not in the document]
serhalp: (To be clear here, I ended up only upgrading to commander@12 after all. This behavioural change is in commander@13, but I left this here so we're ready for it, since it doesn't hurt.)

src/commands/deploy/index.ts
```diff
@@ -100,11 +100,13 @@ Support for package.json's main field, and intrinsic index.js entrypoints are co
       '-b, --branch <name>',
       'Serves the same functionality as --alias. Deprecated and will be removed in future versions',
     )
-    .option('-o, --open', 'Open site after deploy', false)
+    .option('-O, --open', 'Open site after deploy', false)
```
serhalp, line 103: Same here, there's already `-o` for `--offline` in the BaseCommand, so `deploy -o` wasn't working as intended. It just started an offline deploy which of course didn't work.

src/commands/base-command.ts
serhalp: Handled in https://github.com/netlify/cli/pull/7046

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
```diff
@@ -14,19 +14,21 @@ export const detectNetlifyLambda = async function ({ packageJson } = {}) {
...
+
+  program.allowExcessArguments()
+
   // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
   const matchingScripts = Object.entries(scripts).filter(([, script]) => script.match(/netlify-lambda\s+build/))
 
   for (const [key, script] of matchingScripts) {
     // E.g. ["netlify-lambda", "build", "functions/folder"]
     // these are all valid options for netlify-lambda
-    program
```
serhalp, line 23: because this is in a loop it was redeclaring these options over and over, which is no longer allowed

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
```diff
@@ -14,19 +14,21 @@ export const detectNetlifyLambda = async function ({ packageJson } = {}) {
     return false
   }
 
+  const program = new Command()
```
serhalp, line 17: using this constructor is also cleaner since it avoids mutating the exported singleton

docs/commands/login.md
```diff
@@ -21,6 +21,7 @@ netlify login
 
 - `new` (*boolean*) - Login to new Netlify account
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
```
serhalp, line 24: 🤔 I guess this one's pretty weird, huh?

docs/commands/switch.md
```diff
@@ -19,6 +19,7 @@ netlify switch
 **Flags**
 
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
```
serhalp, line 22: 🤔 does `ntl switch` make sense with a token?

docs/commands/unlink.md
```diff
@@ -20,6 +20,7 @@ netlify unlink
 
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
```
serhalp, line 23: arguably slightly weird

site/scripts/util/generate-command-data.js
```diff
@@ -3,13 +3,13 @@ import { sortOptions } from '../../../dist/utils/command-helpers.js'
 
 const program = createMainCommand()
 
-/** @type {Array<import('../../src/commands/base-command.js').default>} */
+/** @type {Array<import('../../../src/commands/base-command.js').default>} */
```
serhalp, line 6: this is totally unrelated, but I happened to notice this was wrong 😬

src/commands/env/env.ts
```diff
@@ -38,10 +39,11 @@ export const createEnvCommand = (program: BaseCommand) => {
     .command('env:import')
     .argument('<fileName>', '.env file to import')
     .option(
-      '-r, --replace-existing',
+      '-R, --replace-existing',
```
serhalp, line 42: undo this one

docs/commands/completion.md
```diff
@@ -20,6 +20,7 @@ netlify completion
 **Flags**
 
 - `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
```
serhalp, line 23: arguably this one's slightly weird too

src/commands/env/env.ts
ndhoule: What does this mean? Is this a note to self or you noting that this reverts changes from a previous PR?

src/commands/blobs/blobs.ts
ndhoule: What does "same as above here" refer to? Can you spell out what you mean?

src/commands/blobs/blobs.ts
ndhoule: Sort of a bummer--I think `-o` for output is a somewhat universal convention. Maybe something for us to change in future release? I think I'd opt for making `--offline` a long option only.

tests/integration/commands/help/__snapshots__/help.test.ts.snap
```diff
@@ -44,6 +44,8 @@ USAGE
 OPTIONS
   -h, --help             display help for command
   --debug                Print debugging information
+  --auth <token>         Netlify auth token - can be used to run this command
+                         without logging in
```
ndhoule, line 48: I'm not sure I fully understand this, but:

If the auth option was previously undocumented, you might consider continuing hiding it. That'd let us kick the can down the road on hiding it from the commands where it doesn't make semantic sense (and I think I agree with all the ones you pointed out).

...Although, when we remove this undocumented flag from those commands, it'll be a breaking change anyway. So... either way is fine, I suppose.

src/commands/blobs/blobs.ts
ndhoule: Oops, got it.

ndhoule (approved):
Looks great! I trust your judgment on things like whether or not to hide the `--auth` flag, but I would err on the side of documenting what already exists. It's the shorter path and we can remove these flags where they make no semantic sense as part of a future breaking change.

tests/integration/commands/help/__snapshots__/help.test.ts.snap
serhalp: The challenge is that `--auth` was documented on only a handful of commands, but existed on all of them and actually did something on some subset of them. Now it will be documented on all of them.

The challenge is that AFAICT there's no built-in way with commander to mutate the config (to change `hideHelp`) of an already registered option on a command, so we'd have to refactor to come up with some solution here to allow for it, or to follow the same pattern I did for `--json` and `--offline`, which requires adding `--auth` explicitly to ~50 commands (i.e. all but 2-3) and then possibly deciding to `.hideHelp()`on the ones where it currently isn't documented (if want want to not change what is currently documented).

My inclination is to merge this as is, since none of the above seems worth the effort.

PR #7026, opened 2025-02-05
fix(deps): replace `ora` and `log-symbols` with tiny dependency `picospinner`

serhalp:
#### Summary

This should reduce our package size a bit, which is huge. See https://github.com/es-tooling/module-replacements/blob/main/docs/modules/ora.md.

I also added some missing types along the way.

### DIFF
--- a/scripts/prepare-for-publish.js
+++ b/scripts/prepare-for-publish.js
@@ -3,7 +3,7 @@ import { dirname, join } from 'path'
 import { fileURLToPath } from 'url'
 
 import execa from 'execa'
-import ora from 'ora'
+import yoctoSpinner from 'yocto-spinner'
 
 // These scripts from package.json need to be preserved on publish
 const preserveScripts = new Set([
@@ -15,8 +15,7 @@ const preserveScripts = new Set([
   'prepublishOnly',
 ])
 
-let spinner = ora({
-  spinner: 'star',
+let spinner = yoctoSpinner({
   text: 'Patching package.json (removing devDependencies, scripts, etc)',
 }).start()
 
@@ -41,18 +40,16 @@ pkgJson.scripts.postinstall = pkgJson.scripts['postinstall-pack']
 delete pkgJson.scripts['postinstall-pack']
 
 await writeFile(packageJsonPath, JSON.stringify(pkgJson, null, 2))
-spinner.succeed()
+spinner.success()
 
-spinner = ora({
-  spinner: 'star',
+spinner = yoctoSpinner({
   text: 'Running `npm install --no-audit`',
 }).start()
 await execa('npm', ['install', '--no-audit'])
-spinner.succeed()
+spinner.success()
 
-spinner = ora({
-  spinner: 'star',
+spinner = yoctoSpinner({
   text: 'Running `npm shrinkwrap`',
 }).start()
 await execa('npm', ['shrinkwrap'])
-spinner.succeed()
+spinner.success()
--- a/scripts/prepare-for-publish.js
+++ b/scripts/prepare-for-publish.js
@@ -3,7 +3,7 @@ import { dirname, join } from 'path'
 import { fileURLToPath } from 'url'
 
 import execa from 'execa'
-import yoctoSpinner from 'yocto-spinner'
+import { createSpinner } from 'nanospinner'
 
 // These scripts from package.json need to be preserved on publish
 const preserveScripts = new Set([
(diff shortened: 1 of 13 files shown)


scripts/prepare-for-publish.js
```diff
@@ -15,8 +15,7 @@ const preserveScripts = new Set([
   'prepublishOnly',
 ])
 
-let spinner = ora({
-  spinner: 'star',
```
serhalp, line 19: this one is a CI script so I didn't bother preserving the exact styling

src/lib/spinner.ts
```diff
@@ -1,34 +1,38 @@
-import logSymbols from 'log-symbols'
-import ora, { Ora } from 'ora'
+import yoctoSpinner, { Spinner } from 'yocto-spinner'
+
+const DOTS_SPINNER = {
+  interval: 80,
+  frames: ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'],
+}
```
serhalp, line 6: This was the default in `ora`, which we were implicitly using here.

src/utils/framework-server.ts
```diff
@@ -44,7 +44,9 @@ export const startFrameworkServer = async function ({
     await rm(settings.dist, { recursive: true, force: true })
   }
 
-  runCommand(settings.command, { env: settings.env, spinner, cwd })
+  if (settings.command) {
+    runCommand(settings.command, { env: settings.env, spinner, cwd })
+  }
```
serhalp, line 49: If you scroll up you'll see we were already doing this in one branch but not the other. Adding some type safety uncovered this.

src/utils/shell.ts
```diff
@@ -62,16 +59,13 @@ export const runCommand = (command, options = {}) => {
 
   // This ensures that an active spinner stays at the bottom of the commandline
   // even though the actual framework command might be outputting stuff
-  // @ts-expect-error TS(7006) FIXME: Parameter 'writeStream' implicitly has an 'any' ty... Remove this comment to see the full error message
-  const pipeDataWithSpinner = (writeStream, chunk) => {
-    if (spinner && spinner.isSpinning) {
+  const pipeDataWithSpinner = (writeStream: NodeJS.WriteStream, chunk: any) => {
```
serhalp, line 62: To be clear, chunk is literally typed as `any` in the NodeJS types. This isn't a lazy type 😄.

src/utils/shell.ts
```diff
@@ -62,16 +59,13 @@ export const runCommand = (command, options = {}) => {
...
     }
     writeStream.write(chunk, () => {
-      if (spinner && spinner.isSpinning) {
-        spinner.isSilent = false
-        spinner.render()
+      if (spinner?.isSpinning) {
+        spinner.start()
       }
     })
   }
```
serhalp, line 71: Feel free to compare the docs for ora and yocto-spinner and double-check me here, but I think this should reproduce the behaviour. Of course I'll poke at it manually before considering merging.

serhalp:
> Dependency count: 1,158 ⬇️ 1.04% decrease
> Package size: 319 MB ⬆️ 1.01% increase

Noooooooo

43081j:
@serhalp i would actually recommend using [nanospinner](https://github.com/usmanyunusov/nanospinner/) or [picospinner](https://github.com/tinylibs/picospinner/) (if you need a little more control of rendering)

particularly nanospinner has been adopted widely elsewhere so we can benefit from npm de-duplication by using it. its also generally faster/smaller (though these are all spinners so they're about as big/fast as each other...)

src/lib/api.ts
```diff
@@ -16,7 +16,7 @@ const MAX_PAGES = 10
 const MAX_PER_PAGE = 100
 
 // @ts-expect-error TS(7023) FIXME: 'listSites' implicitly has return type 'any' becau... Remove this comment to see the full error message
-export const listSites = async ({ api, options }): SiteInfo[] => {
+export const listSites = async ({ api, options }): Promise<SiteInfo[]> => {
```
serhalp, line 19: I have the rest of this fixed in another draft PR. I just needed this to unblock errors elsewhere.

src/commands/sites/sites-list.ts
```diff
@@ -8,16 +8,14 @@ import BaseCommand from '../base-command.js'
...
   let spinner
   if (!options.json) {
     spinner = startSpinner({ text: 'Loading your sites' })
   }
   await command.authenticate()
 
   const sites = await listSites({ api, options: { filter: 'all' } })
-  if (!options.json) {
-    // @ts-expect-error TS(2345) FIXME: Argument of type '{ spinner: Ora | undefined; }' i... Remove this comment to see the full error message
+  if (spinner) {
```
serhalp, line 18: These are equivalent, but this is less leaky and makes TS happy. We only define a spinner (upstream) when `--json` isn't true.

src/commands/functions/functions-create.ts
```diff
@@ -45,6 +45,11 @@ const languages = [
   { name: 'Rust', value: 'rust' },
 ]
 
+const MOON_SPINNER = {
```
khendrikse, line 48: Cute!

khendrikse (approved):
I'm excited about this!

PR #7034, opened 2025-02-12
fix: support for redirects with 200 status. Fix crashing dev server w…

VitaliyR:
…hen have socket error. Follow redirects from proxyRes

#### Summary

Fixes WRFL-2263

One of the customers who are using Gatsby uses a 200 redirect
```
[[redirects]]
from = "/de-DE/*"
to = "/:splat"
status = 200
```

It works on prod but doesn't in `netlify dev`

Root cause for that we are not overriding the url if the file is not static, and in case of dev server running on gatsby (or next as well) - it is never static.

---

      passes our tests.


src/utils/proxy.ts
```diff
@@ -596,24 +605,15 @@ const initializeProxy = async function ({
...
-        req,
-        res,
-        proxy: handlers,
-        imageProxy,
-        match: null,
-        // @ts-expect-error TS(2339) FIXME: Property 'proxyOptions' does not exist on type 'In... Remove this comment to see the full error message
-        options: req.proxyOptions,
-        siteInfo,
-        env,
+    if (isRedirect({ status: proxyRes.statusCode }) && proxyRes.headers.location) {
```
VitaliyR, line 608: here is another change.
User has a redirect to the page, which also has internal redirect.
For instance, gatsby sites likes to do a redirection from `/page` to `/page/` (slash in the end).

So we basically want to follow it:
1. user hits /test/page
2. it matches redirect /test/* -> /:splat
3. cli proxies contents of the `/:splat`, which includes `Location` header. But status is 200 instead of 3xx (and original is 302 for instance, meaning localhost:8000 returns { 302, Location }, but location:8888 returns { 200, Location }
Users see now this page:
image

This is browser built-in contents. Location header is there. Browser `will never do` the redirect since, if you want it to happen, either status should be 3xx, or another header `Refresh` should be there.

So, instead, here we are changing the flow for the netlify dev (which is typically runs SSGs in develop mode, not in the production mode) to this: 

| Target app wants redirect on this route? Follow it

This change can be left intact and for instance, I can enable it only for some of the contexts (e.g. for dev server context). WDYT?

tests/integration/__fixtures__/site-with-redirect/index.js
```diff
@@ -0,0 +1,25 @@
...
+const server = http.createServer((req, res) => {
+  const pathname = new URL(req.url, 'http://localhost').pathname
+
+  console.log(`Got ${pathname}`)
+
+  if (pathname === '/') {
+    res.write('Root page')
+    res.end()
+  } else if (pathname === '/test/exists') {
+    res.writeHead(302, undefined, { location: '/test/exists/' })
```
VitaliyR, line 12: examples of the site which makes redirect, so then in netlify toml we want to land on this page, which then wants to jump to the next

tests/integration/__fixtures__/hugo-site/netlify.toml
```diff
@@ -2,6 +2,6 @@
 command = "npm run serve"
 framework = "#custom"
 functions = "functions/"
-port = 7000
+port = 7001
```
VitaliyR, line 5: 7000 by default on OS X taken by airplay, failed my tests

src/utils/proxy.ts
```diff
@@ -433,9 +437,13 @@ const serveRedirect = async function ({
       match.force ||
       (!staticFile && ((!options.framework && destStaticFile) || isInternal(destURL) || matchingFunction))
     ) {
+      // 3xx redirects parsed above, here are 2xx meaning just override the url of proxying page and use the status
+      // which comes from that url
       req.url = destStaticFile ? destStaticFile + dest.search : destURL
       const { status } = match
-      statusValue = status
+      if (match.force || status !== 200) {
```
VitaliyR, line 444: override status code only if forcing or status is not 200, when 200 is just shadowing the page by the url, so it's status should be propagated (e.g. 404 because it's not there or idk 201 because server decided to return 201 by that url)

khendrikse:
after talking with vitalii, decided to close

PR #7046, opened 2025-02-19
feat!: remove deprecated camelCase option variants

serhalp:
#### Summary

These commands option have been deprecated since 12.8.0 (25 months ago): https://github.com/netlify/cli/releases/tag/v12.8.0.

Removes:
- `--httpProxyCertificateFilename`
- `--prodIfUnlocked`
- `--functionsPort`
- `--targetPort`
- `--edgeInspect`
- `--edgeInspectBrk`
- `--replaceExisting`
- `--gitRemoteName`

The change in 12.8.0 also intended to rename `--httpProxy` to `--http-proxy` but there was a typo resulting in that never happening. While we're cutting a major, this his fixes that by actually renaming it. This was never actually a documented option, so this isn't truly a breaking change.

CPLA-2533


src/commands/base-command.ts
```diff
@@ -192,17 +192,16 @@ export default class BaseCommand extends Command {
       .addOption(new Option('--cwd <cwd>').hideHelp(true))
       .addOption(new Option('-o, --offline').hideHelp(true))
       .addOption(new Option('--auth <token>', 'Netlify auth token').hideHelp(true))
+      // TODO(serhalp) Remove in next major release. This was intended to be deprecated in 2023 but a typo
+      // delayed it to Feb 2025.
```
serhalp, line 196: I could also see an argument for just renaming this one:
- this particular one was hidden from help to begin with, so you could argue it was undocumented
- we're cutting a major here anyway

ndhoule (approved):

PR #7047, opened 2025-02-19
feat!: remove `integration` (SDK v1) commands

serhalp:
#### Summary

These were used for the Netlify SDK v1. See https://developers.netlify.com/sdk/ for information about SDK v2.

PR #7048, opened 2025-02-19
types: mildly improve type safety of build command

serhalp:
#### Summary

I ran into some confusing poorly typed code while working on https://github.com/netlify/cli/pull/7047, so I improved it and pulled it into a separate PR.

We should upstream these fixes to `@netlify/build` sooner rather than later though...

### DIFF
--- a/src/lib/build.ts
+++ b/src/lib/build.ts
@@ -1,54 +1,79 @@
 import fs from 'fs'
 import process from 'process'
 
-import build from '@netlify/build'
+import build, { type OnEnd, type OnPostBuild } from '@netlify/build'
 // @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'toml... Remove this comment to see the full error message
 import tomlify from 'tomlify-j0.4'
+import type { OptionValues } from 'commander'
 
 import { getFeatureFlagsFromSiteInfo } from '../utils/feature-flags.js'
+import type { EnvironmentVariables } from '../utils/types.js'
 
 import { getBootstrapURL } from './edge-functions/bootstrap.js'
 import { featureFlags as edgeFunctionsFeatureFlags } from './edge-functions/consts.js'
 
-/**
- * The buildConfig + a missing cachedConfig
- * @typedef BuildConfig
- * @type {Parameters<import('@netlify/build/src/core/main.js')>[0] & {cachedConfig: any}}
- */
+export interface CachedConfig {
+  env: EnvironmentVariables
+  siteInfo: {
+    id?: string
+    account_id?: string
+    feature_flags?: Record<string, boolean | boolean | number>
+  }
+  // TODO(serhalp) Add remaining properties
+  [k: string]: unknown
+}
+
+interface DefaultConfig {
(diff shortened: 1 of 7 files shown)


src/lib/build.ts
```diff
@@ -1,70 +1,89 @@
...
 
 import { getFeatureFlagsFromSiteInfo } from '../utils/feature-flags.js'
 
 import { getBootstrapURL } from './edge-functions/bootstrap.js'
 import { featureFlags as edgeFunctionsFeatureFlags } from './edge-functions/consts.js'
 
-/**
- * The buildConfig + a missing cachedConfig
- * @typedef BuildConfig
- * @type {Parameters<import('@netlify/build/src/core/main.js')>[0] & {cachedConfig: any}}
```
serhalp, line 16: this wasn't doing anything, because `src/` should be `lib/` and it exports `buildSite`, not a default export... 🤦🏼

src/lib/build.ts
```diff
@@ -112,6 +125,8 @@ export const runBuild = async (options) => {
       scheme: apiUrl.protocol.slice(0, -1),
       host: apiUrl.host,
     }
+    // @ts-expect-error(serhalp) -- I don't know what's going on here and I can't convince myself it even works as
+    // intended. TODO(serhalp) Investigate and fix types.
```
serhalp, line 129: Sorry, I had to cut my losses here eventually. It's sad but I'm still trading 13 for 1...

src/lib/build.ts
```diff
@@ -1,54 +1,79 @@
...
+  accountId?: string
+  edgeFunctionsBootstrapURL: string
+}
+
+interface HandlerResult {
+  newEnvChanges?: Record<string, string>
+  configMutations?: Record<string, string>
+  status?: string
+}
+// The @netlify/build type incorrectly states a `void | Promise<void>` return type.
```
serhalp, line 45: 😓 I guess this ended up growing and growing as I hacked away. I should probably upstream most of this instead of merging this as is, I guess.

src/lib/build.ts
```diff
@@ -92,18 +116,13 @@ export const getBuildOptions = async ({
       ...getFeatureFlagsFromSiteInfo(cachedConfig.siteInfo),
       functionsBundlingManifest: true,
     },
+    // @ts-expect-error(serhalp) -- TODO(serhalp) Upstream the type fixes above into @netlify/build
```
serhalp, line 119: This is because even though we've patched the wrong types, it expects us to pass those wrong types back in here

src/lib/build.ts
ndhoule: Seems fine to merge as is and upstream in another pass

ndhoule (approved):

src/lib/build.ts
serhalp: Yeah, it's also that in the coming days we'll be making several more changes like this. It would be a lot of back-and-forth to validate here + open PR in build + release PR in build + bump here + repeat. I'd rather batch it later.

PR #7049, opened 2025-02-19
fix(deps): add type declarations for 20 of 29 untyped dependencies

serhalp:
#### Summary

This repo needs some work to be fully type safe. To get started on tackling this, we must start from the leaf nodes of the module graph. This means starting by addressing untyped NPM dependencies.

I identified 29 of these and came up with a plan for each.

This PR knocks out the 20 most trivial ones (add missing `@types/` package, remove, upgrade, or replace with a drop-in).

#### Details

- replace untyped `better-opn` with `open`
- upgrade untyped `cookie` to typed v1: https://github.com/jshttp/cookie/releases/tag/v1.0.0
- remove unnecessary `unixify`
- add missing `@types/` packages for the other 17 and fix a few revealed errors


package.json
```diff
@@ -99,7 +98,7 @@
     "comment-json": "4.2.5",
     "configstore": "7.0.0",
     "content-type": "1.0.5",
-    "cookie": "0.7.2",
+    "cookie": "^1.0.2",
```
serhalp, line 101: not to self: I think I'm supposed to pin this and `open` because of something something yarn lockfile

src/recipes/vscode/settings.ts
serhalp, line 1: TODO - test this recipe manually, since it doesn't have test coverage and makes some assumptions about a VS Code plugin

src/utils/get-repo-data.ts
```diff
@@ -51,16 +49,19 @@ const getRepoData = async function ({ remoteName, workingDir }: { remoteName?: s
     }
 
     const { url } = gitConfig.remote[remoteName]
-    const { host, name, owner, repo } = parseGitRemote(url)
+    const parsedUrl = parseGithubUrl(url)
+    if (parsedUrl == null) {
+      throw new Error(`The specified Git remote ${remoteName} does not a valid URL: ${url}`)
```
serhalp, line 54:
```suggestion
      throw new Error(`The specified Git remote ${remoteName} is not a valid URL: ${url}`)
```

src/utils/open-browser.ts
```diff
@@ -1,7 +1,6 @@
 import process from 'process'
 
-// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'bett... Remove this comment to see the full error message
-import open from 'better-opn'
+import open from 'open'
```
serhalp, line 3: TODO worth a quick manual QA for good measure

package.json
ndhoule: Yeah, we should fix these two to a specific version.

src/recipes/vscode/settings.ts
```diff
@@ -1,13 +1,15 @@
...
-import unixify from 'unixify'
+
+const toUnixPath = (path: string): string => path.replace(/\\/g, '/')
 
 // @ts-expect-error TS(7006) FIXME: Parameter 'existingSettings' implicitly has an 'an... Remove this comment to see the full error message
 export const applySettings = (existingSettings, { denoBinary, edgeFunctionsPath, repositoryRoot }) => {
-  const relativeEdgeFunctionsPath = unixify(relative(repositoryRoot, edgeFunctionsPath))
+  // TODO(serhalp) I'm not convinced we want to convert to Unix paths on Windows? Does this even work? Was this a
+  // workaround for something, perhaps https://github.com/denoland/vscode_deno/pull/745?
+  const relativeEdgeFunctionsPath = toUnixPath(posix.normalize(relative(repositoryRoot, edgeFunctionsPath)))
```
ndhoule, line 12: I only did a marginal amount of digging, but my bet is that we actually do want to be unixifying paths. My main thought here is that from what I can tell, unix paths work on Windows, but I wonder if the reverse is true? Given that people often commit vscode settings to a repository, I wonder if Windows-style paths works for unix-like users?

src/utils/get-repo-data.ts
ndhoule: Yes, please

ndhoule (approved):
Other than your notes to self, looks great.

src/recipes/vscode/settings.ts
serhalp: ✅ Tested manually. Works fine. Did not test on Windows.

PR #7050, opened 2025-02-20
feat(utils): replace fastify and express with tinyhttp and friends

talentlessguy:

#### Summary

Replaces express, express-loggin, fastify and fastify-static with tinyhttp, @tinyhttp/logger, sirv and milliparsec

See https://github.com/e18e/ecosystem-issues/issues/156

---

      passes our tests.


tests/integration/utils/mock-api-vitest.ts
```diff
@@ -90,7 +91,7 @@ export const startMockApi = ({ routes, silent }: MockApiOptions): Promise<MockAp
   app.all('*', function onRequest(req, res) {
     addRequest(requests, req)
     if (!silent) {
-      console.warn(`Route not found: (${req.method.toUpperCase()}) ${req.url}`)
+      console.warn(`Route not found: (${req.method!.toUpperCase()}) ${req.url}`)
```
serhalp, line 94: What's the deal here? If `req.method` is optional, let's use optional chaining. If the type is wrong, can we fix it in tinyhttp?

src/utils/static-server.ts
```diff
@@ -1,40 +1,36 @@
...
-    reply.header('age', '0')
-    reply.header('cache-control', 'public, max-age=0, must-revalidate')
+  app.use((req, res, next) => {
+    res.header('age', '0')
+    res.header('cache-control', 'public, max-age=0, must-revalidate')
     const validMethods = ['GET', 'HEAD']
-    if (!validMethods.includes(req.method)) {
-      reply.code(405).send('Method Not Allowed')
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    if (!validMethods.includes(req.method!)) {
```
serhalp, line 25: same here

src/utils/static-server.ts
```diff
@@ -1,40 +1,36 @@
...
+      res.status(405).send('Method Not Allowed')
     }
-    done()
+    return next?.()
   })
-  await server.listen({ port: settings.frameworkPort })
-  const [address] = server.addresses()
+
+  const server = createServer((req, res) => app.handler(req as Request, res as Response))
```
serhalp, line 31: Type assertions are unsafe. Why is there a mismatch here? How can we resolve it?

src/utils/static-server.ts
```diff
@@ -1,40 +1,36 @@
...
+  server.listen(settings.frameworkPort)
+  const address = server.address() as AddressInfo
```
serhalp, line 33: same here — how can we avoid this `as`?

src/lib/functions/server.ts
```diff
@@ -72,24 +72,23 @@ const buildClientContext = function (headers) {
...
 }
 
-// @ts-expect-error TS(7006) FIXME: Parameter 'req' implicitly has an 'any' type.
-const hasBody = (req) =>
+const hasBody = (req: Request) =>
   // copied from is-type package
-  // eslint-disable-next-line unicorn/prefer-number-properties
-  (req.header('transfer-encoding') !== undefined || !isNaN(req.header('content-length'))) &&
+  (req.headers['transfer-encoding'] !== undefined ||
+    !Number.isNaN(Number.parseInt(req.headers['content-length'] as string))) &&
```
serhalp, line 78: same here and a few more people — let's avoid `as` please

src/lib/functions/server.ts
```diff
@@ -269,24 +269,20 @@ interface GetFunctionsServerOptions {
...
+        payloadLimit: 6_291_456,
+      })(req, res, next)
+    }
+    return next?.()
+  })
+  app.use(raw({ payloadLimit: 6_291_456 }))
   app.use(createFormSubmissionHandler({ functionsRegistry, siteUrl }))
-  app.use(
-    expressLogging(console, {
-      blacklist: ['/favicon.ico'],
```
serhalp, line 287: Are we losing this behaviour?

serhalp (commented):
🙌🏼 Thanks for taking this on! I've taken a first pass at reviewing it. Could you please give this a quick rebase and take a look at my comments? I haven't had time to debug your integration tests yet, but I'll get to it soon.

tests/integration/utils/mock-api-vitest.ts
talentlessguy: the type is inherited from `node:http` `IncomingMessage`

src/utils/static-server.ts
talentlessguy: tinyhttp's Request and Response extend `node:http`'s Request and Response. Probably could pass as a generic to `createServer` but if I recall correctly, it throws a type error

src/utils/static-server.ts
talentlessguy: by doing a check for `null`

src/lib/functions/server.ts
talentlessguy: a header value might be a type of string or array of strings if it is sent multiple times

src/lib/functions/server.ts
talentlessguy: the blacklisting of the logging yes - I can add it to `@tinyhttp/logger`

src/utils/static-server.ts
talentlessguy: this is a TypeScript issue. `Request` extends `IncomingMessage`, which itself extends `Stream`. TypeScript thinks that `Request` doesn't extends `Stream` because of that, even though it does

serhalp:
@talentlessguy Thanks for your contributions! I just want to give you a heads up that we likely won't be merging this one, since we're in the process of extracting much of the code you're touching out of this codebase (and incidentally removing express/fastify in the process).

PR #7058, opened 2025-02-24
fix(deps): replace `chalk` with smaller and faster `ansis`

serhalp:
#### Summary

Replace `chalk` with `ansis`: https://github.com/webdiscus/ansis?tab=readme-ov-file#how-to-switch-to-ansis-1.

Honestly, I did this before realizing the savings are absolutely tiny (44 KB → 6 KB), but I fixed a few types along the way so we might as well ship it.


src/lib/functions/registry.ts
```diff
@@ -348,43 +343,42 @@ export class FunctionsRegistry {
 
     if (event === 'loaded') {
       const icon = warningsText ? NETLIFYDEVWARN : NETLIFYDEVLOG
-      const color = warningsText ? chalk.yellow : chalk.green
+      const color = warningsText ? picocolors.yellow : picocolors.green
       const mode =
-        // @ts-expect-error TS(2339) FIXME: Property 'logLambdaCompat' does not exist on type ... Remove this comment to see the full error message
         func?.runtimeAPIVersion === 1 && this.logLambdaCompat
```
serhalp, line 348: This was actually a legit bug. `logLambdaCompat` was a static field so this was always falsy.

src/utils/command-helpers.ts
```diff
@@ -59,30 +45,30 @@ export const USER_AGENT = `${name}/${version} ${platform}-${arch} node-${process
 /** A list of base command flags that needs to be sorted down on documentation and on help pages */
 const BASE_FLAGS = new Set(['--debug', '--http-proxy', '--http-proxy-certificate-filename'])
 
-export const NETLIFY_CYAN = chalk.rgb(40, 180, 170)
+// XXX(serhalp) Restore equivalent of `chalk.rgb(40, 180, 170)`
+export const NETLIFY_CYAN = picocolors.blueBright
```
serhalp, line 49: Just need to decide what to do about this...

src/utils/command-helpers.ts
```diff
@@ -59,30 +45,30 @@ export const USER_AGENT = `${name}/${version} ${platform}-${arch} node-${process
...
- * @param {import('commander').Option} optionB
- * @returns {number}
  * @example
- * options.sort(sortOptions)
+ * options.sort(compareOptions)
  */
-// @ts-expect-error TS(7006) FIXME: Parameter 'optionA' implicitly has an 'any' type.
-export const sortOptions = (optionA, optionB) => {
+export const compareOptions = (optionA: Option, optionB: Option): number => {
+  const longOptionA = optionA.long ?? ''
```
serhalp, line 64: Though we always define one, an option is allowed to have no `long` variant

src/utils/command-helpers.ts
```diff
@@ -59,30 +45,30 @@ export const USER_AGENT = `${name}/${version} ${platform}-${arch} node-${process
...
+  const longOptionB = optionB.long ?? ''
   // base flags should be always at the bottom
-  if (BASE_FLAGS.has(optionA.long) || BASE_FLAGS.has(optionB.long)) {
+  // FIXME(serhalp) This logic can't be right? Look at it... This must only happen to work.
+  if (BASE_FLAGS.has(longOptionA) || BASE_FLAGS.has(longOptionB)) {
     return -1
   }
```
serhalp, line 70: Reviewer, am I tripping? This should be something like this, right?
```suggestion
  if (BASE_FLAGS.has(longOptionA)) {
    return 1
  } else if (BASE_FLAGS.has(longOptionB)) {
    return -1
  }
```

mrstork: Your suggestion looks to be correct. A negative value indicates that `a` should come before `b`, and a positive value indicates that `a` should come after `b`.

src/utils/command-helpers.ts
webdiscus: @serhalp the smaller drop-in replacement for `chalk` is `ansis` (~6 KB). Ansis has both the `ansis.hex('#28B4AA')` and `ansis.rgb(40, 180, 170)` functions.

tests/integration/utils/call-cli.js
```diff
@@ -16,6 +16,10 @@ export const callCli = async function (args, execOptions = {}, parseJson = false
   const { stdout } = await execa.node(cliPath, args, {
     timeout: CLI_TIMEOUT,
     nodeOptions: [],
+    env: {
+      // TODO(serhalp) Why not exercise colorization in integration tests? Remove and update snapshots?
+      NO_COLOR: '1',
```
serhalp, line 21: chalk was doing this magically in our tests, so this is just preserving that behaviour for now

webdiscus: @serhalp

some tests failed because expected `/^VERSION/` but received the string `\u001b[1mVERSION\u001b[22m\n` containing ANSI codes.

Ansis automatically detects CI environments as supported color.

To force disable color should be created separate file with the `NO_COLOR` variable, e.g. `no-color.js`:
```js
// set env variable to force disable color
process.env.NO_COLOR = '1';
```

This file must be imported in the test file `before` other files used `ansis`:

```js
import { expect, describe, test } from 'vitest';

// import env variables to force disable color
import './no-color.js';

// ... import others
```

Alternatively, you can use the `ansis.strip()` function to clean a string from ANSI codes:

_[e2e/install.e2e.ts](https://github.com/netlify/cli/pull/7058/files#diff-bec9cf642fedf82b6fd7503e7dd47c27cc7f8532c2b58e6a92c648d8de436af1):179_
```js
import { expect, describe, test } from 'vitest';
import ansis from 'ansis';
// ...

expect(ansis.strip(stdout.trim()), `Help command does not start with 'VERSION':\n\n${ansis.strip(stdout)}`).toMatch(/^VERSION/)
```

See please the [no-color.test.js](https://github.com/webdiscus/ansis/blob/master/test/no-color.test.js)

src/commands/functions/functions-create.ts
```diff
@@ -304,7 +304,7 @@ const promptFunctionsDirectory = async (command) => {
   ])
 
   try {
-    log(`${NETLIFYDEVLOG} updating site settings with ${chalk.magenta.inverse(functionsDir)}`)
+    log(`${NETLIFYDEVLOG} updating site settings with ${ansis.inverse(ansis.magenta(functionsDir))}`)
```
webdiscus, line 307: Hello @serhalp, I'm curious why nested calls are often used in some cases, while chained syntax is preferred in others.
Ansis supports both, but chaining is shorter, more readable, and performs better:

```diff
- log(`${NETLIFYDEVLOG} updating site settings with ${chalk.magenta.inverse(functionsDir)}`)
+ log(`${NETLIFYDEVLOG} updating site settings with ${ansis.magenta.inverse(functionsDir)}`)
```

package.json
```diff
@@ -74,10 +74,10 @@
     "@pnpm/tabtab": "0.5.4",
     "ansi-escapes": "7.0.0",
     "ansi-to-html": "0.7.2",
+    "ansis": "^3.17.0",
```
webdiscus, line 77: @serhalp,
meanwhile I have released a new version - 4.0.0. 
This release brings a stable and more compact ANSI library.
v4 is ~15.7% smaller than v3.17.

serhalp:
PR is stale and full of conflicts, value is minimal, closing. keeping type fixes in https://github.com/netlify/cli/pull/7383.

PR #7069, opened 2025-02-27
chore: simplify and standardize `package.json` scripts and CI setup

serhalp:
📣 You should probably review this commit by commit!

#### 2593f33dd9060918f04695add8c95f476734dfbd Remove `postinstall` hook that builds the package

- It's nonstandard
- It's surprising
- Since we use `tsc` to build, it causes every CI check to fail if you have type errors
- It causes issues when rewriting git history
- It's a bit magical

I refactored an affected script in a separate commit first (be729a26c5906ead57142b1e8d898011bab245e7), for easier review.

#### b57d815a564752549e9dd97a26f0ac0a0fd272cc Simplify site build scripts

There were two separate scripts for doing the same thing for some reason. I ran into this oddity while making the above change.

#### 5c3f7673b6972fe16979152e3f8c0a92bf902da1 Remove leftover refs to `ava`

ava was removed a couple years ago

#### f4119f86e1370de71708bc1926148b5cc04cb8ec Rename misleading CI jobs

They handle two kinds of benchmarks, only one of which is package size

#### bf2bfa203f5932bd8f38535ce99d089989e17fbf Split formating, linting, and unit testing

- They were all jumbled together 😢
- CI wouldn't run unit tests if you had a prettier or eslint error
- The scripts were nonstandard and surprising
- The scripts were unnecessarily complicated
- CI was running formatting and linting on Windows, macOS, and Linux for no reason

#### 846d1c4f8dcb717cfba1d21723bc5027e55f3263 Simplify CI workflow skipping for release branches

You can just skip the whole job instead of skipping every step!

#### 7169000bd6db1238575230e11953c0b10af95fa2 chore(package.json): sort scripts

:feelsgood: This one is dedicated to @ndhoule. Keep fighting the good fight ✌🏼.

#### ec2e8603 Add missing dependency on `npm-run-all`

We're using `run-s` which comes from this dep but it only happens to work because one of
our dependencies has a transitive dependency on `npm-run-all`.

IMO we should kill this, but for now this just adds the missing dep.

Note: npm-run-all2 is a maintained fork of the unmaintained npm-run-all


package.json
```diff
@@ -62,7 +62,6 @@
     "site:build:install": "cd site && npm install --no-audit",
     "site:build:assets": "cd site && npm run build",
     "postinstall-pack": "node ./scripts/postinstall.js",
```
serhalp, line 64: This `postinstall-pack` hook gets renamed to `postinstall` in a script triggered from the `prepublishOnly` hook, so that when _users_ install this package they see some output we want them to see at install time.

package.json
```diff
@@ -62,7 +62,6 @@
...
-    "postinstall": "npm run build && node ./scripts/postinstall.js",
```
serhalp, line 65: See comment above – we were also triggering this when running `npm i` in this repo when _developing_ in this repo, but that seems unnecessary so I didn't bother preserving it.

serhalp: ... In fact this may be why we had the preinstall build step in the first place... so that this script can run (it uses some npm deps).

CONTRIBUTING.md
```diff
@@ -33,7 +32,7 @@ First, fork and clone the repository. If you’re not sure how to do this, pleas
 Run:
 
 ```bash
-npm install && npm run site:build:install
+npm install
```
serhalp: This being mentioned here didn't make any sense so I removed it (there's a section below about the docs site)

package.json
```diff
@@ -56,11 +56,8 @@
     "test:ci:vitest:unit": "vitest run --coverage tests/unit/",
     "test:ci:vitest:integration": "vitest run --coverage tests/integration/",
     "e2e": "node ./tools/e2e/run.js",
-    "docs": "npm run --prefix=site build",
+    "site:build": "npm run build && npm install --prefix=site --no-audit && npm run --prefix=site build",
```
serhalp, line 59: I don't love the `npm install` inside a script but that's a whole can of worms that didn't seem worth pursuing

package.json
```diff
@@ -31,40 +31,32 @@
...
   "scripts": {
-    "prepare": "is-ci || husky install node_modules/@netlify/eslint-config-node/.husky/",
-    "start": "node ./bin/run.js",
     "build": "tsc",
+    "certs": "openssl req -x509 -out localhost.crt -keyout localhost.key -newkey rsa:2048 -nodes -sha256 -subj \"/CN=localhost\" -extensions EXT -config certconf",
     "dev": "tsc --watch",
-    "test": "run-s format test:dev",
-    "format": "run-s format:check-fix:*",
-    "format:ci": "run-s format:check:*",
-    "format:check-fix:lint": "run-s format:check:lint format:fix:lint",
```
serhalp, line 41: bonus win: this was running eslint twice... fixing also checks, so it was redundant!

ndhoule (approved):
clapping_futurama

PR #7071, opened 2025-02-28
ci: fix new job config

serhalp:
#### Summary

Whoops. I split up a job in https://github.com/netlify/cli/pull/7069, but it turns out if your workflow is invalid, it doesn't even report as failed. And I hadn't marked the new jobs as required yet. 🤦🏼

**Note:** Now that these jobs have run successfully once on this PR, I was able to mark the checks as required.


khendrikse (approved):
it is beautiful

PR #7072, opened 2025-02-28
ci: let other jobs run despite type errors

serhalp:
#### Summary

Because we use `tsc` to build our package, typechecking is coupled with building. Since building is a prerequisite step to many CI jobs, they all fail without actually running what they intend when there is a single type error.

This allows CI to actually report all the other checks when you have type errors, by splitting typechecking into a separate job.

#### To do

- [ ] once it has run successfully once and this is about to be merged, mark the new check as required


khendrikse (approved):
This makes sense!

PR #7074, opened 2025-02-28
fix: better ip version detection when using skip wait port property i…

VitaliyR:
…n dev command

#### Summary

Fixes WRFL-2315

1. Leaving intact the skipWaitPort: for node > 18 default is ipv6, for others ipv4. But also sets the `detectFrameworkHost`
2. If `detectFrameworkHost` is set:
2.1. When received an error `ECONNREFUSED`, don't cancel the request with failure. Instead, proxy it to different ipVer. If originally was 6, proxy to 4 and vice versa. Set `isChangingTarget: true` for the proxy chain.
2.2. If received a response from proxy and `isChangingTarget` is `true` -> "remember" this ipVer by changing it in the settings via `changeSettings` function and also changing in proxy.

---

      passes our tests.


src/utils/proxy.ts
```diff
@@ -557,6 +583,23 @@ const initializeProxy = async function ({
       res.setHeader(NFRequestID, requestID)
     }
 
+    // @ts-expect-error TS(2339) FIXME: Property 'proxyOptions' does not exist on type 'In... Remove this comment to see the full error message
+    const options = req.proxyOptions
```
VitaliyR, line 587: moved it here, so we have only 1 ts-expect-error

src/utils/proxy.ts
```diff
@@ -805,11 +841,16 @@ const onRequest = async (
...
       settings.frameworkHost && isIPv6(settings.frameworkHost) ? `[${settings.frameworkHost}]` : settings.frameworkHost
     }:${settings.frameworkPort}`,
+    detectTarget: settings.detectFrameworkHost,
+    targetHostname: settings.frameworkHost,
     publicFolder: settings.dist,
     functionsServer,
     functionsPort: settings.functionsPort,
     jwtRolePath: settings.jwtRolePath,
     framework: settings.framework,
+    changeSettings(newSettings: Partial<ServerSettings>) {
```
VitaliyR, line 851: Added this to change the settings on the fly

PR #7089, opened 2025-03-06
ci: split lint, format tasks into jobs

ndhoule:
This changeset splits the lint and format tasks--formerly run as part of the unit test job--into their own CI jobs.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,25 @@
+---
+name: Format
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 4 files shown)


serhalp:
@ndhoule isn't this redundant with my (reverted) PR?

ndhoule:
@serhalp Yep. I think we should recommit your changes in smaller hunks; I was forced to revert all three commits because they were very intertwined.

DanielSLew (approved):

PR #7090, opened 2025-03-06
build: relocate site netlify.toml to site/ directory

ndhoule:
This change moves the site `netlify.toml` to the `site/` directory.

Previously, the site-build workflow went like this:

- Netlify Build executed `npm install` at the root of the project (by default)
- After package installation, the `postinstall` hook triggered a CLI build
- Netlify Build would execute `netlify.toml#build.command`, which:
  - Entered the docs directory
  - Ran `npm install` against `docs/package.json`
  - Ran a site build (which implicitly relied on the root directory `npm install` and `npm run build` steps)

To make this change, I modified the logic to:

- Netlify Build starts in the `site/` directory and executes `npm install`
- `site/netlify.toml#build.command`  enters the root and preps the CLI, then returns to the `docs/` directory
- We then build the site

In this way, the site's dependency on the CLI's source is more clear--we assume nothing about the state of the root directory before building the docs site.

Some other noteworthy changes here:

- While making these changes, I realized the (JavaScript) site build scripts assume that the CLI has been built to the `dist` directory. I modified them to be transpiled at runtime and to import from `src`.
  - I'm not particularly attached to this change, but it reduces the amount of setup steps required and will let us convert these scripts to TypeScript.
- I moved the site functions to the standard Netlify functions location, `site/netlify/functions`. Doing so also let me move a dependency that was only used on the site (`@bugsnag/js`) out of the root, which slims the CLI's dependency tree.
- I got rid of most of the `site:` tasks from the top-level `package.json`. (I left the `docs` task, though.)
- These changes require that we change the CLI site's [base directory](https://app.netlify.com/sites/cli/configuration/deploys) to `docs`. The site is refusing to build on `main` right now anyway, so I already made that change.

### DIFF
--- a/.github/workflows/verify-docs.yml
+++ b/.github/workflows/verify-docs.yml
@@ -11,19 +11,22 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - name: Git checkout
-        uses: actions/checkout@v4
-      - name: Use Node.js
-        uses: actions/setup-node@v4
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
         with:
           node-version: lts/*
           cache: npm
-      - name: Install core dependencies
+
+      - name: Install dependencies
         run: npm ci --no-audit
-      - name: Install site dependencies
-        run: npm run site:build:install
+
+      - name: Install ./site dependencies
+        run: npm ci  --prefix=site --no-audit
+
       - name: Generate docs
-        run: npm run docs
+        run: npm run --prefix=site build
+
       - name: Check for changes
         run: |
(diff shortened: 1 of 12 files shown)


jmarlena (approved):
@ndhoule Assuming the docs CLI site still builds fine with these changes, this is fine to merge! If it's hard to test the site build, we can always merge this and prepare to revert if needed.

ndhoule:
@jmarlena Thanks! This is testable (pre-merge) by visiting the deploy preview:

https://deploy-preview-7090--cli.netlify.com/

I did a spot check on a few of the pages by flipping back and forth between the preview and cli.netlify.app and didn't notice any difference.

PR #7099, opened 2025-03-10
fix: ensure we pass build_data in the calls to the deploys api

JakeChampion:
#### Summary

This ensures that the bundle will have the correct metadata applied to the deployment. previously this metadata would only get applied for CLI deployments that included the `--build` flag (unless the `--skip-functions-cache` flag was also applied)


khendrikse:
Closing!

PR #7115, opened 2025-03-20
feat: Add `db init` and `db status` commands

CalebBarnes:
## Summary
This PR introduces a new `db` command to the Netlify CLI, enabling users to provision and manage production-ready Postgres databases with minimal effort.

### Features

- Adds `netlify db` command with two primary subcommands:
  - `db init` - Initialize a new database and connects it to the current site
  - `db status` - Check the status of the database connection

- Seamless integration with Drizzle ORM:
  - Optional initialization of Drizzle configuration and schema boilerplate
  - Adds helpful database scripts to package.json
  - Auto-installs required Drizzle dependencies

- Flexible setup options:
  - Interactive mode with prompts for setup preferences
  - `--assume-no` flag for non-interactive setup (ideal for CI/AI tools)
  - `--boilerplate=drizzle` and `--no-boilerplate` flags to control ORM setup
  - `--overwrite` option to replace existing boilerplate files

## Example Usage

```bash
# Initialize a database with interactive prompts
netlify db init

# Check database connection status
netlify db status

# Non-interactive setup for AI agents or CI environments
netlify db init --assume-no

# Initialize with Drizzle ORM and overwrite existing files
netlify db init --boilerplate=drizzle --overwrite
```
The implementation automatically handles extension installation, creating a database, and connecting it to the site via environment variables.

### Documentation Generation Enhancement

I expanded the CLI documentation generator to support subcommands more robustly. Previously, subcommand docs were only generated if the command name contained a colon (`:`), such as `dev:exec`, which would be treated as a subcommand of `dev`. This approach missed subcommands defined as nested classes or those with explicit parent references.

Now, the docs generator also recognizes subcommands by checking the command's `parent` property, allowing it to correctly generate documentation for subcommands like those under the new `db` command without relying solely on colon-separated naming.

 
---

      passes our tests.


src/commands/database/database.ts
```diff
@@ -0,0 +1,276 @@
...
+        type: 'confirm',
+        name: 'drizzle',
+        message: 'Use Drizzle?',
+      },
+    ].filter((q) => !initialOpts[q.name]),
+  )
+
+  if (!initialOpts.drizzle) {
+    command.setOptionValue('drizzle', answers.drizzle)
+  }
```
DanielSLew, line 37: wouldn't we only want to prompt if they don't provide it as an optional
```suggestion
    if (!initialOpts.drizzle) {
  const answers = await inquirer.prompt(
    [
      {
        type: 'confirm',
        name: 'drizzle',
        message: 'Use Drizzle?',
      },
    ].filter((q) => !initialOpts[q.name]),
  )


    command.setOptionValue('drizzle', answers.drizzle)
  }
```

src/commands/database/utils.ts
```diff
@@ -0,0 +1,104 @@
+import fs from 'fs'
+import inquirer from 'inquirer'
+import path from 'path'
+
+const JIGSAW_URL = 'https://jigsaw.services-prod.nsvcs.net'
```
DanielSLew, line 5: can we make sure this can work with staging too

src/commands/database/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  accountId,
+  slug,
+  hostSiteUrl,
+}: {
+  token: string
+  accountId: string
+  slug: string
+  hostSiteUrl: string
+}) => {
+  const installExtensionResponse = await fetch(`https://app.netlify.com/.netlify/functions/install-extension`, {
```
DanielSLew, line 66: same here about staging

src/commands/database/database.ts
CalebBarnes: Definitely! 

I updated `--drizzle` to be a negatable boolean option as well `--no-drizzle` and am now not prompting here unless `initialOpts.drizzle !== true && initialOpts.drizzle !== false`. 

If both `--drizzle` and `--no-drizzle` are not passed then `initialOpts.drizzle` will be `undefined`.

khendrikse:
We could add logic that if we notice that the site is not linked yet, we ask them if they want to do that, if they say yes we continue into that workflow?

khendrikse (approved):

src/commands/database/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const NEON_DATABASE_EXTENSION_SLUG = process.env.NEON_DATABASE_EXTENSION_SLUG ?? 'neon'
+export const JIGSAW_URL = process.env.JIGSAW_URL ?? 'https://api.netlifysdk.com'
```
ndhoule:
```suggestion
export const JIGSAW_URL = process.env.JIGSAW_URL ?? 'https://jigsaw.services-prod.nsvcs.net'
```

This is the canonical URL for Jigsaw we use elsewhere (e.g. in the SDK)

src/commands/database/database.ts
```diff
@@ -0,0 +1,43 @@
...
+    .command('db')
+    .alias('database')
+    .description(`Provision a production ready Postgres database with a single command`)
+    .addExamples(['netlify db status', 'netlify db init', 'netlify db init --help'])
+
+  dbCommand
+    .command('init')
+    .description(`Initialize a new database for the current site`)
+    .option(`--drizzle`, 'Initialize basic drizzle config and schema boilerplate')
+    .option('--no-drizzle', 'Does not initialize drizzle and skips any related prompts')
```
ndhoule: IIRC, commander automatically handles `--no-` for all boolean flags. Or is this explicitly for adding the `--no-drizzle` to the printed docstring? If so, can you add an explanatory comment inline?

src/commands/database/database.ts
```diff
@@ -0,0 +1,43 @@
...
+    .option(
+      '--minimal',
```
ndhoule: I don't know if it's too late or too much work elsewhere to change this, but most CLI tools:

- call this out as being "interactive" or "non-interactive"
- name this flag after the behavior it models, e.g. `-y --assume-yes` or `--assume-no`.

I would call this `--assume-no` and make the docstring `Non-interactive setup. Does not initialize any boilerplate or install any tools, e.g. Drizzle. Ideal for CI environments or AI tools.`

ndhoule (commented):
submitting a few comments while I review the rest of the PR

src/commands/database/database.ts
```diff
@@ -0,0 +1,43 @@
...
+  const dbCommand = program
+    .command('db')
+    .alias('database')
+    .description(`Provision a production ready Postgres database with a single command`)
+    .addExamples(['netlify db status', 'netlify db init', 'netlify db init --help'])
+
+  dbCommand
+    .command('init')
+    .description(`Initialize a new database for the current site`)
+    .option(`--drizzle`, 'Initialize basic drizzle config and schema boilerplate')
```
ndhoule: I would instead make this something like `--boilerplate=drizzle` and make `drizzle` the default. This will make it so we can easily change the list of tools we support later on without adding tons of flags. Eventually, one of the new flags we want to add is going to conflict with another flag.

src/commands/database/init.ts
```diff
@@ -0,0 +1,181 @@
...
+    const installed = await installExtension({
+      accountId: siteInfo.account_id,
+      netlifyToken: netlifyToken,
+      slug: NEON_DATABASE_EXTENSION_SLUG,
+      hostSiteUrl: extension.hostSiteUrl,
+    })
+    if (!installed) {
+      throw new Error(`Failed to install extension on team "${account.name}": "${extension.name}"`)
+    }
+    log(`Extension "${extension.name}" successfully installed on team "${account.name}"`)
```
ndhoule, line 54: Now now, but in the future: Should we replace this with a call to the extension host site and let the host install the extension, so we're not re-implementing this logic in quite so many places?

src/commands/database/status.ts
```diff
@@ -0,0 +1,79 @@
...
+  const account = await getAccount(command, { accountId: siteInfo.account_id })
+
+  let databaseUrlEnv: Awaited<ReturnType<typeof command.netlify.api.getEnvVar>> | undefined
+  let unpooledDatabaseUrlEnv: Awaited<ReturnType<typeof command.netlify.api.getEnvVar>> | undefined
+
+  try {
+    databaseUrlEnv = await command.netlify.api.getEnvVar({
+      accountId: siteInfo.account_id,
+      siteId: command.siteId,
+      key: 'NETLIFY_DATABASE_URL',
```
ndhoule, line 31: Note for the future: What will happen if we want to support installing two databases to the site? E.g. letting someone set up both Postgres and Redis through Netlify DB?

src/commands/database/utils.ts
```diff
@@ -0,0 +1,261 @@
+import { createRequire } from 'module'
```
ndhoule, line 1: Note for the future: We have a schema for Jigsaw, we can auto-generate a client for this. We have an example of how to do so in the SDK

ndhoule (approved):
Minor comments throughout; the only ones I feel strongly about are changing the `--minimal` flag to be more explanatory, and changing `--drizzle` to `--orm`/`--boilerplate`/something else.

src/commands/database/status.ts
CalebBarnes: Somewhere I have `ManagedEnvs` in the extension repo. We ~~could~~ should probably loop over all of the envs the extension manages. Also this whole status command could be made way better.

CalebBarnes: Also if we want more than one db installed on a single site then we have to change a bunch of other stuff.

CalebBarnes: In that case, `status` would probably list all of the DBs on the project/site.

docs/commands/db.md
```diff
@@ -0,0 +1,86 @@
...
+- `filter` (*string*) - For monorepos, specify the name of the application to run the command in
+- `overwrite` (*boolean*) - Overwrites existing files that would be created when setting up boilerplate
+- `debug` (*boolean*) - Print debugging information
+- `auth` (*string*) - Netlify auth token - can be used to run this command without logging in
+- `no-boilerplate` (*boolean*) - Don't add any boilerplate to your project.
+
+**Examples**
+
+```bash
+netlify db init --minimal
```
CalebBarnes: @ndhoule Assuming you removed/changed `--minimal` flag, we should remove this example too.

src/commands/database/database.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      const options = _options as DatabaseInitOptions
+      if (options.assumeNo) {
+        options.boilerplate = false
+        options.overwrite = false
+      }
+
+      await init(options, command)
+    })
+    .addExamples([`netlify db init --minimal`, `netlify db init --boilerplate=drizzle --overwrite`])
```
CalebBarnes: @ndhoule  Assuming you removed/changed `--minimal` flag, we should remove this example too.

PR #7119, opened 2025-03-20
build: remove build step from post-install hook

ndhoule:
The CLI currently runs `npm run build` in a post-install script. Many of commands in the CLI implicitly depend on this behavior, which is unnecessarily complicated. This post-install script conflicts with the user-facing post-install script, which means we need to do a little pre-publish post-install script juggling to ensure that the end user runs the correct script after installing the CLI.

This change removes the post-install script in favor of explicitly building the CLI before e.g. CI steps that require a built CLI.

Partial un-reversion of https://github.com/netlify/cli/pull/7069, which was reverted in https://github.com/netlify/cli/pull/7085.


.github/workflows/integration-tests.yml
```diff
@@ -138,43 +152,58 @@ jobs:
...
+
       - name: Install core dependencies
         run: npm ci --no-audit
         if: '${{!steps.release-check.outputs.IS_RELEASE}}'
+
+      - name: Build project
+        run: npm run build
+
+      - name: Run E2E tests
+        run: npm run e2e
```
serhalp, line 196: ❓ copypasta?

```suggestion
```

.github/workflows/unit-tests.yml
```diff
@@ -27,19 +28,27 @@ jobs:
...
           check-latest: true
         if: '${{!steps.release-check.outputs.IS_RELEASE}}'
+
       - name: Install core dependencies
         run: npm ci --no-audit
         if: '${{!steps.release-check.outputs.IS_RELEASE}}'
+
+      - name: Build project
+        run: npm run build
+        if: '${{!steps.release-check.outputs.IS_RELEASE}}'
```
serhalp, line 50: IIRC we didn't need build for unit tests? did you check?

scripts/prepare-for-publish.js
```diff
@@ -1,49 +0,0 @@
...
-await writeFile(packageJsonPath, JSON.stringify(pkgJson, null, 2))
-spinner.success()
-
-spinner = createSpinner('Running `npm install --no-audit`').start()
-await execa('npm', ['install', '--no-audit'])
-spinner.success()
-
-spinner = createSpinner('Running `npm shrinkwrap`').start()
-await execa('npm', ['shrinkwrap'])
-spinner.success()
```
serhalp, line 49: Don't we still need this in the release workflow?

serhalp (commented):
LGTM, just two things 👁️👁️

.github/workflows/unit-tests.yml
ndhoule: I copied this out of your changes

https://github.com/netlify/cli/pull/7069/files#diff-cd0effce06f425599e96952f4d5f684641a683f46d9909e7c74a9f8b2f03ccd3R33-R34

scripts/prepare-for-publish.js
ndhoule: Ah yes, good catch! I'll add it back in.

.github/workflows/pre-release.yml
```diff
@@ -33,14 +40,18 @@ jobs:
...
           echo version=${{ steps.extract.outputs.version }}
-      - name: Setup git user
-        run: git config --global user.name github-actions
-      - name: Setup git email
-        run: git config --global user.email github-actions@github.com
+
+      - name: Configure Git user
+        run: |
+          git config --global user.name "Netlify"
+          git config --global user.email "82042599+token-generator-app[bot]@users.noreply.github.com"
```
serhalp, line 47: if you say so

serhalp (approved):

.github/workflows/pre-release.yml
ndhoule: I do

PR #7130, opened 2025-03-25
fix(types): fix 388 type errors and fully type `src/lib/functions/`

serhalp:
### Summary

This started as a bump of https://github.com/netlify/local-functions-proxy/pull/137 to pull in the now-typed version of `@netlify/functions-proxy` but devolved into fixing hundreds of types.

This fully types `src/lib/functions/` and fixes a number of core types used across the codebase.

These types revealed various minor runtime errors / bugs that I also fixed, but in a separate PR that I'll open stacked on this one. I tried to mostly keep behavioural changes for that follow-up PR, but I did include a few here, which I'll call out in inline comments.

### Reviewing

Sorry this is big and sorry I didn't break this down into smaller commits much. I'd recommend reviewing the smaller commits first and finishing with `fix(types): fix functions types and much more` then `fix: fix functions regression and fully type lib/functions/`.

### To do

- [x] Fix the skipped test in `tests/integration/commands/dev/dev-miscellaneous.test.ts`
- [x] Fix the strange `Host machine does not support local functions proxy server` failure on ubuntu-latest node 18.17.0
- [ ] Maybe release a few `netlify/build` fixes and bump here to fix a few more errors (or just do as a follow-up)


src/commands/init/init.ts
```diff
@@ -7,31 +7,25 @@ import getRepoData from '../../utils/get-repo-data.js'
...
-// @ts-expect-error TS(7031) FIXME: Binding element 'siteInfo' implicitly has an 'any'... Remove this comment to see the full error message
-const logExistingAndExit = ({ siteInfo }) => {
+const logExistingAndExit = ({ siteInfo }: { siteInfo: SiteInfo }): void => {
   log()
   log(`This site has been initialized`)
   log()
   log(`Site Name:  ${chalk.cyan(siteInfo.name)}`)
   log(`Site Url:   ${chalk.cyan(siteInfo.ssl_url || siteInfo.url)}`)
-  log(`Site Repo:  ${chalk.cyan(getRepoUrl({ siteInfo }))}`)
+  log(`Site Repo:  ${chalk.cyan(getRepoUrl(siteInfo))}`)
```
serhalp, line 28: this would have always been `undefined`

bin/run.js
```diff
@@ -17,9 +17,8 @@ try {
     pkg,
     updateCheckInterval: UPDATE_CHECK_INTERVAL,
   }).notify()
-} catch (error_) {
-  error('Error checking for updates:')
-  error(error_)
+} catch (error) {
+  logError(`Error checking for updates: ${error?.toString()}`)
```
serhalp, line 21: Maybe should've made this one a separate commit/PR, but FYI this is the only intentional behavioural change in the `error()` refactor commit — This was previously exiting when an error is thrown from the logic that notifies users of an available netlify-cli update. I change this to log and carry on.

src/commands/api/api.ts
```diff
@@ -28,12 +28,13 @@ export const apiCommand = async (apiMethodName: string, options: OptionValues, c
   }
 
   if (!apiMethodName) {
-    error(`You must provide an API method. Run "netlify api --list" to see available methods`)
+    return logAndThrowError(`You must provide an API method. Run "netlify api --list" to see available methods`)
```
serhalp, line 31: I didn't remember this about `never`, but TS only seemed to carry over the "unreachable" refinement from the function call's `never` when I `return`.

src/commands/sites/sites-delete.ts
```diff
@@ -18,21 +18,17 @@ export const sitesDelete = async (siteId: string, options: OptionValues, command
...
+      return logAndThrowError(`No site with id ${siteId} found. Please verify the siteId & try again.`)
     } else {
-      error(error_)
+      return logAndThrowError(error_)
     }
   }
 
-  if (!siteData) {
-    error(`Unable to process site`)
-  }
```
serhalp, line 29: This was just compensating for incorrectly refined types

src/commands/status/status.ts
```diff
@@ -29,16 +29,18 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
...
+      return logAndThrowError(error_)
     }
   }
 
   const ghuser = command.netlify.globalConfig.get(`users.${current}.auth.github.user`)
   const accountData = {
-    Name: user?.full_name,
-    Email: user?.email,
+    Name: user.full_name,
+    Email: user.email,
```
serhalp, line 43: This was compensating for incorrectly refined types

src/commands/functions/functions-build.ts
```diff
@@ -35,7 +34,6 @@ export const functionsBuild = async (options: OptionValues, command: BaseCommand
 
   log(`${NETLIFYDEVLOG} Building functions`)
 
-  // @ts-expect-error TS(2345) FIXME: Argument of type '{ skipGo: boolean; }' is not ass... Remove this comment to see the full error message
-  zipFunctions(src, dst, { skipGo: true })
+  zipFunctions(src, dst)
```
serhalp, line 37: https://github.com/netlify/zip-it-and-ship-it/pull/373

src/commands/init/init.ts
```diff
@@ -166,47 +156,42 @@ const createOrLinkSiteToRepo = async (command) => {
...
   if (repoUrl && !options.force) {
-    logExistingAndExit({ siteInfo })
+    logExistingAndExit({ siteInfo: existingSiteInfo })
   }
 
   // Look for local repo
   const repoData = await getRepoData({ workingDir: command.workingDir, remoteName: options.gitRemoteName })
-  if (repoData.error) {
-    await handleNoGitRemoteAndExit({ command, error: repoData.error, state })
+  if ('error' in repoData) {
```
serhalp, line 190: trust me, this makes sense — see the `getRepoData` type

src/commands/types.d.ts
```diff
@@ -1,14 +1,13 @@
-import type { NetlifyConfig } from '@netlify/build'
-import type { NetlifyTOML } from '@netlify/build-info'
```
serhalp, line 2: These are so, so incomplete and wrong that it isn't worth dealing with them at the moment. This PR tries to get CLI to have its own reliable build types and we'll someday fix them upstream hopefully.

src/commands/types.d.ts
```diff
@@ -18,57 +17,30 @@ export type NetlifySite = {
   set id(id: string): void
 }
 
-type PatchedConfig = NetlifyTOML &
```
serhalp, line 21: bye bye

src/commands/types.d.ts
```diff
@@ -18,57 +17,30 @@ export type NetlifySite = {
...
-}
-
 /**
  * The netlify object inside each command with the state
+ * TODO(serhalp) Rename. These aren't options. They're more like context.
  */
 export type NetlifyOptions = {
   accounts: Account[]
-  // poorly duck type the missing api functions
-  api: NetlifyAPI & Record<string, (...args: $TSFixMe) => Promise<$TSFixMe>>
```
serhalp, line 59: 😀 bye bye

src/commands/types.d.ts
```diff
@@ -18,57 +17,30 @@ export type NetlifySite = {
...
+  },
   repositoryRoot: string
   /** Absolute path of the netlify configuration file */
   configFilePath: string
   /** Relative path of the netlify configuration file */
   relConfigFilePath: string
   site: NetlifySite
-  siteInfo: $TSFixMe
-  config: PatchedConfig
+  siteInfo: SiteInfo
```
serhalp, line 39: 🕺🏼

src/commands/types.d.ts
```diff
@@ -18,57 +17,30 @@ export type NetlifySite = {
...
+  config: NormalizedCachedConfigConfig
   cachedConfig: CachedConfig
-  globalConfig: $TSFixMe
-  state: StateConfig
```
serhalp, line 71: bye bye

src/lib/api.ts
```diff
@@ -1,27 +1,42 @@
...
 const MAX_PER_PAGE = 100
 
-// @ts-expect-error TS(7023) FIXME: 'listSites' implicitly has return type 'any' becau... Remove this comment to see the full error message
-export const listSites = async ({ api, options }): Promise<SiteInfo[]> => {
+export const listSites = async ({
+  api,
+  options,
+}: {
+  api: NetlifyAPI
+  // FIXME(serhalp) `page` and `maxPages` are missing from `netlify` package types
```
serhalp, line 27: As discussed with @ndhoule, this is fine for me and pretty limited in scope, but we'll have to pause on adding types to Netlify API stuff in CLI until we solve this properly. Nathan has a bunch of knowledge in this space from doing this in the SDK.

src/lib/build.ts
```diff
@@ -1,36 +1,112 @@
...
 
 import { getFeatureFlagsFromSiteInfo } from '../utils/feature-flags.js'
-import type { EnvironmentVariables } from '../utils/types.js'
+import type { Account, EnvironmentVariables, Plugin, SiteInfo } from '../utils/types.js'
 
 import { getBootstrapURL } from './edge-functions/bootstrap.js'
 import { featureFlags as edgeFunctionsFeatureFlags } from './edge-functions/consts.js'
+import type { EdgeFunctionDeclaration } from './edge-functions/proxy.js'
 
 export interface CachedConfig {
```
serhalp, line 16: 😰 the crown jewel of this PR

src/lib/build.ts
```diff
@@ -1,36 +1,112 @@
...
+    command?: string | undefined
+    commandOrigin?: 'default' | undefined
+    publish?: string | undefined
+    publishOrigin?: 'default' | undefined
+  }
+  plugins?: { package: unknown; origin: 'default' }[]
+}
 
 // TODO(serhalp) This is patching weak or missing properties from @netlify/build. Fix there instead.
-export type BuildConfig = Parameters<typeof build>[0] & {
```
serhalp, line 29: bye bye

src/lib/edge-functions/registry.ts
```diff
@@ -15,19 +15,14 @@ import {
...
+  type NormalizedCachedConfigConfig,
 } from '../../utils/command-helpers.js'
 import type { FeatureFlags } from '../../utils/feature-flags.js'
 import { MultiMap } from '../../utils/multimap.js'
 import { getPathInProject } from '../settings.js'
 
 import { INTERNAL_EDGE_FUNCTIONS_FOLDER } from './consts.js'
 
-//  TODO: Replace with a proper type for the entire config object.
-export interface Config {
```
serhalp, line 26: 👯‍♂️ byyye

src/lib/functions/runtimes/index.ts
```diff
@@ -1,30 +1,70 @@
...
+
+export type BaseBuildResult = {
+  includedFiles?: undefined | string[]
+  mainFile?: undefined | string
+  outputModuleFormat?: undefined | string
+  schedule?: undefined | string
+  srcFiles: string[]
+
+  // TODO(serhalp) This module and type shouldn't know about these zisi types. Refactor to allow the JS runtime's zisi
+  // builder to define this on its extended base build result type.
```
serhalp, line 22: The way this runtime stuff is typed could really be improved, but I had to just cut my losses here and ship something that is strongly typed and works. We can always follow up.

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
```diff
@@ -65,13 +69,10 @@ export const detectNetlifyLambda = async function ({ packageJson } = {}) {
...
-  const exists = await fileExistsAsync('package.json')
-  if (!exists) {
-    return false
-  }
-
-  const content = await readFile('package.json', 'utf-8')
-  const packageJson = JSON.parse(content)
-  return detectNetlifyLambda({ packageJson })
+export default async function detectNetlifyLambdaBuilder() {
+  const result = await readPackageUp({ normalize: true })
```
serhalp, line 73: ⚠️ To be extra safe, I should change this to `read-pkg` instead of `read-package-up`. I'll push that tiny fix before merging.

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
```diff
@@ -1,16 +1,18 @@
...
 import { memoizedBuild } from '../../../memoized-build.js'
+import type { BaseBuildResult } from '../../index.js'
 
-// @ts-expect-error TS(2525) FIXME: Initializer provides no value for this binding ele... Remove this comment to see the full error message
-export const detectNetlifyLambda = async function ({ packageJson } = {}) {
-  const { dependencies, devDependencies, scripts } = packageJson || {}
-  if (!((dependencies && dependencies['netlify-lambda']) || (devDependencies && devDependencies['netlify-lambda']))) {
+export type NetlifyLambdaBuildResult = BaseBuildResult
+
+export const detectNetlifyLambda = async ({ packageJson }: { packageJson: NormalizedPackageJson }) => {
```
serhalp, line 13: fun fact: I have zero clue what this whole file is about. what in the world is https://www.npmjs.com/package/netlify-lambda? is it super old and deprecated? did I waste my time? I should've looked into that earlier, but I'll get this merge for now and follow up.

src/lib/functions/runtimes/js/builders/zisi.ts
```diff
@@ -76,18 +87,22 @@ const buildFunction = async ({
   } = await memoizedBuild({
     cache,
     cacheKey: `zisi-${entryPath}`,
-    // @ts-expect-error TS(2345) FIXME: Argument of type '{ archiveFormat: string; basePat... Remove this comment to see the full error message
-    command: () => zipFunction(entryPath, targetDirectory, zipOptions),
+    command: async () => {
+      const result = await zipFunction(entryPath, targetDirectory, zipOptions)
+      if (result == null) {
+        throw new Error('Failed to build function')
+      }
```
serhalp, line 94: A lot of code would have been a (worse) nightmare to type without this, so this seems worth including in this PR

src/lib/functions/runtimes/js/index.ts
```diff
@@ -3,45 +3,57 @@ import { dirname } from 'path'
...
 
 // The netlify-lambda builder can't be enabled or disabled on a per-function
 // basis and its detection mechanism is also quite expensive, so we detect
 // it once and cache the result.
-const detectNetlifyLambdaWithCache = () => {
-  // @ts-expect-error TS(7005) FIXME: Variable 'netlifyLambdaDetectorCache' implicitly h... Remove this comment to see the full error message
+const detectNetlifyLambdaWithCache = async () => {
   if (netlifyLambdaDetectorCache === undefined) {
-    netlifyLambdaDetectorCache = detectNetlifyLambdaBuilder()
+    netlifyLambdaDetectorCache = await detectNetlifyLambdaBuilder()
```
serhalp, line 37: this wasn't a sort of debounce mechanism or anything, it was just a plain result cache, so we don't need to hold on to the promise

src/lib/functions/runtimes/js/index.ts
```diff
@@ -54,15 +66,23 @@ export const getBuildFunction = async ({ config, directory, errorExit, func, pro
   const functionDirectory = dirname(func.mainFile)
   const srcFiles = functionDirectory === directory ? [func.mainFile] : [functionDirectory]
 
-  // @ts-expect-error TS(2532) FIXME: Object is possibly 'undefined'.
-  return () => ({ schedule: metadata.schedule, srcFiles })
+  // eslint-disable-next-line @typescript-eslint/require-await -- Must be async to match the interface
+  const build: BuildFunction<JsBuildResult> = async () => ({ schedule: metadata?.schedule, srcFiles })
```
serhalp, line 70: @ndhoule I gotchu:
```suggestion
  const build: BuildFunction<JsBuildResult> = () => Promise.resolve({ schedule: metadata?.schedule, srcFiles })
```

(I don't like it, but whatever)

src/utils/get-package-json.ts
```diff
@@ -1,16 +1,20 @@
...
+
+// TODO(serhalp) Consider renaming. We read all sorts of different `package.json` files in this repo for all sorts of
+// different reasons, and this one is only for reading our OWN package.json.
+const getPackageJson = async (): Promise<NormalizedPackageJson> => {
   if (!packageJson) {
-    const packageJsonPath = join(dirname(fileURLToPath(import.meta.url)), '../../package.json')
-    // @ts-expect-error TS(2345) FIXME: Argument of type 'Buffer' is not assignable to par... Remove this comment to see the full error message
-    packageJson = JSON.parse(await readFile(packageJsonPath))
+    const cliProjectRoot = join(dirname(fileURLToPath(import.meta.url)), '..', '..')
+    const result = await readPackageUp({ cwd: cliProjectRoot, normalize: true })
```
serhalp, line 13: ⚠️ here too I should technically use read-pkg instead of read-package-up

tests/integration/commands/functions-with-args/functions-with-args.test.js
```diff
@@ -866,7 +866,7 @@ describe.concurrent('serving functions', () => {
           path: 'hello.js',
           handler: async () => ({
             statusCode: null,
-            body: 42,
+            body: 'hello',
```
serhalp, line 869: this was a test verifying the behaviour when `statusCode` is invalid, but it was also passing an invalid `body` (covered in another test)

tests/integration/commands/dev/dev-miscellaneous.test.ts
```diff
@@ -152,7 +152,8 @@ describe.concurrent('commands/dev-miscellaneous', () => {
     })
   })
 
-  test('background function clientContext,identity should be null', async (t) => {
+  // XXX(serhalp) - Unskip and fix before merging.
+  test.skip('given a background function, context should have empty `clientContext` and null `identity`', async (t) => {
```
serhalp, line 156: reviewer: ignore this for now

src/utils/live-tunnel.ts
```diff
@@ -92,16 +102,34 @@ const installTunnelClient = async function () {
...
+  siteId?: string | undefined
+  slug: string
+}) => {
+  if (!siteId) {
+    console.error(
+      `${NETLIFYDEVERR} Error: no siteId defined, did you forget to run ${chalk.yellow(
+        'netlify init',
+      )} or ${chalk.yellow('netlify link')}?`,
+    )
+    process.exit(1)
```
serhalp, line 122: This is not a behavioural change. I just moved this from one method to another, but the result is the same

src/utils/detect-server-settings.ts
```diff
@@ -253,8 +259,8 @@ const detectServerSettings = async (
 ): Promise<ServerSettings> => {
   validateProperty(devConfig, 'framework', 'string')
 
-  /** @type {Partial<import('./types.js').BaseServerSettings>} */
-  let settings = {}
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- TODO(serhalp) Set to `BaseServerSettings`. Good luck!
+  let settings: any = {}
```
serhalp, line 263: had to back out of this rabbit hole unfortunately

src/lib/functions/server.ts
```diff
@@ -399,15 +406,12 @@ const startWebServer = async ({
...
+  settings: FunctionsSettings
 }) => {
   await new Promise<void>((resolve) => {
-    // @ts-expect-error TS(7006) FIXME: Parameter 'err' implicitly has an 'any' type.
-    server.listen(settings.functionsPort, (err) => {
-      if (err) {
-        errorExit(`${NETLIFYDEVERR} Unable to start functions server: ${err}`)
-      } else if (debug) {
-        log(`${NETLIFYDEVLOG} Functions server is listening on ${settings.functionsPort}`)
+    server.listen(settings.functionsPort, () => {
```
serhalp, line 412: Feel free to triple-check this, but there didn't seem to be such a thing as an `err` param here

src/utils/command-helpers.ts
```diff
@@ -225,12 +218,15 @@ export const exit = (code = 0) => {
  * several ways. It detects it by checking if `build.publish` is `undefined`.
  * However, `@netlify/config` adds a default value to `build.publish`.
  * This removes 'publish' and 'publishOrigin' in this case.
- * @param {*} config
+ * TODO(serhalp) Come up with a better name (or kill it?). This sucks but it's descriptive.
```
serhalp, line 221: Really hate this, but I think a weird descriptive name is better than a normal misleading one. I'm still not _completely_ sure I understand the intent of `cachedConfig.config` — I assume the path forward here is to rename either `cachedConfig` or `.config` or both, but I don't understand this well enough to be sure.

src/lib/functions/registry.ts
```diff
@@ -350,7 +348,6 @@ export class FunctionsRegistry {
       const icon = warningsText ? NETLIFYDEVWARN : NETLIFYDEVLOG
       const color = warningsText ? chalk.yellow : chalk.green
       const mode =
-        // @ts-expect-error TS(2339) FIXME: Property 'logLambdaCompat' does not exist on type ... Remove this comment to see the full error message
         func?.runtimeAPIVersion === 1 && this.logLambdaCompat
```
serhalp, line 351: This was actually a legit bug: `logLambdaCompat` was a static field so this was always falsy.

src/lib/functions/registry.ts
```diff
@@ -318,9 +316,9 @@ export class FunctionsRegistry {
   /**
    * Logs an event associated with functions.
    */
-  static logEvent(
+  private logEvent(
```
serhalp, line 319: fixing `logLambdaCompat` to not be static forced this change

ndhoule (changes requested):
Mostly looks good, thanks for these changes! I noted some real and potential breaking changes/bugs, and also noted some changes I think we should defer to another PR. (It would be a huge time sink to have to revert these changes, so I think we should be as conservative as possible here.)

You made a bunch of changes while I was reviewing, so please be sure to review any comments that might now be outdated.

src/lib/functions/runtimes/js/builders/netlify-lambda.ts
ndhoule: I made a comment to this effect in the other place we're doing this, but:

If we can assume that we don't need to locate the closest `package.json`, can you explain why it's necessary to use a package here at all? E.g. Do we want `read-pkg`'s parsing/normalization behavior, and if so, can we just use `normalize-package-data` (which will almost certainly already be in the cache) rather than dragging in `read-pkg`'s extra deps?

serhalp: We were finding and parsing `package.json` files in four different parts of this codebase, four different ways. One of these was using `read-package-up`, which is typed, and the other three were using ad hoc methods, which are not.

Since `read-package-up` wraps `read-pkg`, pulling in `read-pkg` (at the same version) won't result in any additional dependencies. It's also [very tiny](https://node-modules.dev/grid#install=read-pkg) anyway. (Funny enough, most of the size is `normalize-package-data` actually!)

As for `normalize-package-data`:
```
$ npm prune --omit=dev
$ npm why normalize-package-data
npm ERR! No dependencies found matching normalize-package-data
```

src/lib/functions/server.ts
serhalp: Where do you see the err argument? I've followed the docs as far as they go, and there's no indication of such a thing. It seems to me the callback gets registered on the `"listening"` event, and errors can only be emitted viathe `'error'` event, which needs to be handled separately.

Wow, this is a confusing, under-documented space: https://github.com/nodejs/node/issues/21482

src/commands/api/api.ts
```diff
@@ -1,11 +1,16 @@
...
-import { chalk, error, exit, log, logJson } from '../../utils/command-helpers.js'
-import BaseCommand from '../base-command.js'
+import { chalk, logAndThrowError, exit, log, logJson } from '../../utils/command-helpers.js'
+import type BaseCommand from '../base-command.js'
 
-export const apiCommand = async (apiMethod: string, options: OptionValues, command: BaseCommand) => {
+type ApiMethodName = keyof NetlifyAPI
+
+const isValidApiMethod = (api: NetlifyAPI, apiMethod: string): apiMethod is ApiMethodName =>
+  Object.hasOwn(api, apiMethod)
```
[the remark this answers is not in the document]
serhalp: I only did it this way because I couldn't come up with a pattern that works to refines both. `Object.hasOwn(api, apiMethod)` doesn't refine a `string`, so you can't do the `typeof` check in the same function. I'm not sure if I'm explaining that clearly.


I'll just push something that does it with two functions. Let me know if you know of a better way.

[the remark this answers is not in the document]
serhalp: This works:

```typescript
const isKeyOf = <T extends object>(obj: T, key: string | number | symbol): key is $Keys<T> => Object.hasOwn(obj, key)

const isValidApiMethod = (api: NetlifyAPI, apiMethod: string): apiMethod is NetlifyAPIMethod =>
  isKeyOf(api, apiMethod) && typeof api[apiMethod] === 'function'
```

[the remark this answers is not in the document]
serhalp: oh, right, that doesn't work either 😓
```
typescript: Expression produces a union type that is too complex to represent. [2590]
```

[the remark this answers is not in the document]
serhalp: I'm gonna punt on this, but we can of course do a follow-up PR.

src/commands/base-command.ts
```diff
@@ -548,14 +546,14 @@ export default class BaseCommand extends Command {
       certificateFile: flags.httpProxyCertificateFilename,
     })
     const apiOpts = { ...apiUrlOpts, agent }
-    // TODO: remove typecast once we have proper types for the API
-    const api = new NetlifyAPI(token || '', apiOpts) as NetlifyOptions['api']
+    const api = new NetlifyAPI(token || '', apiOpts)
```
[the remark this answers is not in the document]
serhalp: I tried to only do this when it was certain it wouldn't change behaviour. I'd be scared of implicit expected behaviours here with an empty string.

src/commands/base-command.ts
```diff
@@ -548,14 +546,14 @@ export default class BaseCommand extends Command {
...
 
     actionCommand.siteId = flags.siteId || (typeof flags.site === 'string' && flags.site) || state.get('siteId')
 
     const needsFeatureFlagsToResolveConfig = COMMANDS_WITH_FEATURE_FLAGS.has(actionCommand.name())
     if (api.accessToken && !flags.offline && needsFeatureFlagsToResolveConfig && actionCommand.siteId) {
       try {
+        // FIXME(serhalp) Remove `any` and fix errors. API types exist now.
```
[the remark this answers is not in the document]
serhalp: Can you clarify this comment? Are you saying I should replace the `any`s with the existing `$TSFixMe` helper? I thought we'd discussed unifying them all in a separate PR.

src/lib/functions/config.ts
```diff
@@ -1,28 +1,50 @@
...
-        nodeBundler: config.node_bundler === 'esbuild' ? 'esbuild_zisi' : config.node_bundler,
-        // @ts-expect-error TS(2339) FIXME: Property 'AWS_LAMBDA_JS_RUNTIME' does not exist on... Remove this comment to see the full error message
+        ignoredNodeModules: 'ignored_node_modules' in value ? value.ignored_node_modules : undefined,
+        nodeBundler: value.node_bundler === 'esbuild' ? 'esbuild_zisi' : value.node_bundler,
         nodeVersion: siteEnv.AWS_LAMBDA_JS_RUNTIME,
         processDynamicNodeImports: true,
-        // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
-        schedule: config.schedule,
         zipGo: true,
+        // XXX remove after bumping netlify/build
```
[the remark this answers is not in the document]
serhalp: Ah yep, good catch, this one wasn't supposed to still look like this. I meant to push a few type fixes to netlify/build and bump here, but I've punted that to the stacked fix PR. I'll update this comment.

src/lib/functions/runtimes/go/index.ts
```diff
@@ -40,16 +58,20 @@ const checkGoInstallation = async ({ cwd }) => {
...
-export const getBuildFunction = ({ func }) => {
+export const getBuildFunction = async ({
+  func,
+}: // eslint-disable-next-line @typescript-eslint/require-await -- Must be async to match the interface
+GetBuildFunctionOpts<GoBuildResult>): Promise<BuildFunction<GoBuildResult>> => {
   const functionDirectory = dirname(func.mainFile)
   const binaryPath = temporaryFile(isWindows ? { extension: 'exe' } : undefined)
 
-  return () => build({ binaryPath, functionDirectory })
+  return async () => build({ binaryPath, functionDirectory })
```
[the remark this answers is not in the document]
serhalp: `:thanks-i-hate-it:`

src/utils/get-package-json.ts
```diff
@@ -1,16 +1,20 @@
...
+    if (result?.packageJson == null) {
+      throw new Error('Could not find package.json')
+    }
+    packageJson = result.packageJson
```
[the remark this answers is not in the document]
serhalp: Addressed on the other similar comment

src/utils/get-repo-data.ts
```diff
@@ -66,8 +75,7 @@ const getRepoData = async function ({ remoteName, workingDir }: { remoteName?: s
     }
   } catch (error) {
     return {
-      // @ts-expect-error TS(2571) FIXME: Object is of type 'unknown'.
-      error: error.message,
+      error: error instanceof Error ? error.message : error?.toString() ?? 'Failed to get repo data',
```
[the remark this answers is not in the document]
serhalp: ? Every value except `null` and `undefined` has a `toString()` method since it's defined on all primitives and inherited on all objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString.

```
> (3).toString()
'3'
> (true).toString()
'true'
> (NaN).toString()
'NaN'
> ({}).toString()
'[object Object]'
```

tests/integration/commands/functions-serve/functions-serve.test.ts
```diff
@@ -42,10 +59,10 @@ describe.concurrent('functions:serve command', () => {
         .withNetlifyToml({ config: { functions: { directory: 'functions' } } })
         .withFunction({
           path: 'ping.js',
-          handler: async () => ({
+          handler: dedent/* js */ `async () => ({
```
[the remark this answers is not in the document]
serhalp: Thanks!

src/lib/functions/netlify-function.ts
```diff
@@ -226,25 +239,28 @@ export default class NetlifyFunction {
...
     await this.buildQueue
 
     if (this.buildError) {
-      return { result: null, error: { errorMessage: this.buildError.message } }
+      // TODO(serhalp) I don't think this error handling works as expected. Investigate.
+      return { result: null, error: { errorType: '', stackTrace: [], errorMessage: this.buildError.message } }
     }
 
     const timeout = this.isBackground ? this.timeoutBackground : this.timeoutSynchronous
+    if (timeout == null) {
```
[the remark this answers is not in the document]
serhalp: ... in fact, this is actually dead code.

package.json
```diff
@@ -66,7 +66,7 @@
     "@netlify/edge-bundler": "12.4.0",
     "@netlify/edge-functions": "2.11.1",
     "@netlify/headers-parser": "8.0.0",
-    "@netlify/local-functions-proxy": "^2.0.2",
+    "@netlify/local-functions-proxy": "^2.0.3",
```
serhalp, line 69: `npm i --save-exact @netlify-local-functions-proxy`

src/utils/get-global-config-store.ts
```diff
@@ -41,6 +41,7 @@ export class GlobalConfigStore<
   }
 
   public get(key: string): T[typeof key] {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
serhalp, line 44: This is due to an indirect merge conflict from the rebase on https://github.com/netlify/cli/pull/7165. I had removed a bunch of file-wide eslint suppressions for this file, but that PR reintroduced some. I didn't want to bother digging too much here (`T[type of key]` is `any` by design...).

src/commands/base-command.ts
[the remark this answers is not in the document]
ndhoule: So should we change the type on `token` to reflect that you're not actually sure if it's a `string | null`?

src/commands/base-command.ts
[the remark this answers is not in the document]
ndhoule: You're adding an `any` that you know should be a `$TSFixMe`. Why add it (and potentially lose track of it) when we know it's a `$TSFixMe` right now?

src/utils/get-repo-data.ts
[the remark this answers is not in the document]
ndhoule: `Object.create(null)`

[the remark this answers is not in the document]
ndhoule: (Ultimately, if you annotate `error` as an `unknown` as TypeScript suggests, rather than letting it default to `any`, TypeScript won't be OK with this.)

ndhoule (approved):
:ship:

src/lib/functions/server.ts
serhalp: (We touched base elsewhere. The docs are bad but all signs point toward there not being an `error` arg here.)

src/commands/base-command.ts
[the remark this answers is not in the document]
serhalp: Oh jeez, never mind. My concern was about changing the behaviour when token is `""`, but we're literally falling back to an empty string, so it wouldn't change the behaviour. 👍🏼

src/commands/base-command.ts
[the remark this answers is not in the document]
serhalp: I did not add this `any`; I only added the FIXME.

It wasn't clear at all from the codebase and from discussions with you that `$TSFixMe` was a desirable pattern. We aren't "tracking" them any more than we are `any`s. 🤷🏼

Let's tackle this separately later?

src/utils/get-repo-data.ts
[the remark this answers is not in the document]
serhalp: TypeScript is already annotating caught errors as `unknown` and it's perfectly happy with it!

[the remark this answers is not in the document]
serhalp: Your `Object.create(null)` example is a... radical counterexample, and totally unique. And arguing for this mattering is equivalent to arguing for `try-catch`ing mathematical and logical operations:

```
> let x = Object.create(null)
undefined
> x
[Object: null prototype] {}
> x == 3
Uncaught TypeError: Cannot convert object to primitive value
```

PR #7133, opened 2025-03-25
test: remove certs script

ndhoule:
We currently have a `certs` npm script that generates self-signed
certificates for use in a single zisi test. This script must be run
before running tests, otherwise the zisi test in question will fail.

This changeset moves this certificate generation into a test helper
function, ensuring that a certificate will always be available when
needed by a test.

It would probably be simpler to instead generate a certificate that
never expires and inline it into the fixture (eliminating the need for
an available `openssl` binary), but this works fine.


tests/integration/commands/dev/dev.zisi.test.ts
```diff
@@ -19,6 +21,51 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url))
 
 const testMatrix = [{ args: [] }]
 
+const generateSSLCertificate = async () => {
+  const tmpdir = await fs.mkdtemp(path.join(os.tmpdir(), 'certs-'))
```
serhalp, line 25: 👁️ you look like someone who would enjoy removing the `tempy` dev dep `:grindog:`

serhalp (approved):

tests/integration/commands/dev/dev.zisi.test.ts
ndhoule: dev deps are free!

Issue #7124, opened 2025-03-22
Never ending "Waiting for framework" in 19.0.3


Simbul:
### Describe the bug

Since upgrading to 19.0.3, `netlify dev` is constantly showing the following message, with a spinner on the left:
```
⠦ Waiting for framework port 5173. This can be configured using the 'targetPort' property in the netlify.toml
```

Running the same command with 19.0.2 does not result in the same behaviour. In fact, the wait seems to be over pretty quickly:
```
✔ Waiting for framework port 5173. This can be configured using the 'targetPort' property in the netlify.toml
```
This is printed out before the message saying `Server now ready on http://localhost:8888`

### Steps to reproduce

1. Start in the root of my repo (apologies, this is a private project)
2. Upgrade netlify-cli to 19.0.3
3. Run `netlify dev`
4. Everything seems to be working fine, but the CLI keeps "waiting for framework port 5173"

### Configuration

```toml
[build]
command = "react-router build"
publish = "build/client"

[dev]
command = "react-router dev"
```

### Environment

```
  System:
    OS: macOS 15.3.2
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 3.13 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.10.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.9.2 - /usr/local/bin/npm
  npmGlobalPackages:
    netlify-cli: 19.0.3
```

ffaubert:
I'm having the same issue, had to downgrade back to 19.0.2

ndhoule:
@Simbul @ffaubert I'm unable to reproduce this myself, but I can definitely investigate more if either of you can put together a minimal reproducible example! Doesn't need to be your full app, just enough to consistently trigger this behavior.

ffaubert:
@ndhoule It seems to be an issue with netlify and vite. I put together a very simplistic test case at https://github.com/ffaubert/netlify-cli-test. 

To reproduce after npm install, run netlify dev and then make a change to public/index.html and you'll see 
⠋ Waiting for framework port 5173. This can be configured using the 'targetPort' property in the netlify.toml
forever.

Also new in v19.0.3 at startup I see:
[...]

PR #7135, opened 2025-03-25
fix: fix incorrect dev config types and `netlify dev` regression

serhalp:
#### Summary

Fixes https://github.com/netlify/cli/issues/7124.

The regression was introduced here: https://github.com/netlify/cli/pull/7112/files#r2005549184. The detected framework config's `pollingStrategies` property was no longer being threaded through correctly.

This was introduced because the `DevConfig` type was incorrect and the weak type assertion (and general type weakness in this codebase) allowed it.

This also fixes a few nearby type isues.


src/commands/dev/types.d.ts
```diff
@@ -8,23 +8,14 @@ export type BuildConfig = NonNullable<NetlifyTOML['build']>
 export type DevConfig = NonNullable<NetlifyTOML['dev']> & {
```
serhalp, line 8: You can see here we're extending [this type](https://github.com/netlify/build/blob/1b9f2f40de7ded6285e9b7c08a1adb43550a0ccf/packages/build-info/src/settings/netlify-toml.ts#L378-L389), but we were duplicating most of it here. Most of this diff is cleaning up this unnecessary duplication + adding `undefined` to optional properties.

src/commands/dev/types.d.ts
```diff
@@ -8,23 +8,14 @@ export type BuildConfig = NonNullable<NetlifyTOML['build']>
...
-    certFile: string
-  }
-  envFiles?: string[]
+  base?: string | undefined
+  staticServerPort?: number | undefined
+  envFiles?: string[] | undefined
 
-  jwtSecret: string
-  jwtRolePath: string
-  pollingStrategies?: PollingStrategy[]
```
serhalp, line 29: This is normalized to a string before it gets here: https://github.com/netlify/build/blob/1b9f2f40de7ded6285e9b7c08a1adb43550a0ccf/packages/build-info/src/settings/get-build-settings.ts#L102

src/utils/types.ts
```diff
@@ -12,7 +13,7 @@ export type FrameworkInfo = {
   dev: {
     commands: string[]
     port: number
-    pollingStrategies: { name: string }[]
+    pollingStrategies: PollingStrategy[]
```
serhalp, line 16: This one was correct. I made it reference the canonical type while I was here.

JakeChampion (approved):

ndhoule (approved):

PR #7140, opened 2025-03-26
style: lint all files

ndhoule:
I modernized the eslint configuration Back in https://github.com/netlify/cli/commit/25fcdd8e3a059879c7b0049af64af61aca8b4eb8. At the time, I chose not to make any changes to the list of filepaths we historically passed to eslint, thinking that maybe this would include a bunch of files (like fixtures) that we intentionally ignore.

This turned out to be not the best idea. I didn't notice at the time that the list of filepaths was not great to begin with: We were passing a bunch of file extensions (mjs, cjs) that we no longer have in the repository, other extensions (md, html) that ESLint doesn't lint, and we weren't linting TypeScript files.

Since ESLint 9, you no longer ever really have reason to pass a path or paths to `eslint` (e.g. you often used to pass `.`, but this is now the implied default). So, this change just changes our use of ESLint to match what ESLint now recommends.

This change exposed a lot of unlinted files. I autofixed as many as possible and removed some outdated lint rule suppression--I did not fix any lint errors by hand, so this should include zero change to logic.  We can fix the rest of the suppressions as we go.


eslint_temporary_suppressions.js
```diff
@@ -0,0 +1,1882 @@
...
+          ignores: ['FormData', 'ReadableStream', 'Response', 'Request', 'fetch', 'fs/promises.cp'],
+        },
+      ],
+    },
+  },
+
+  /* Per-file rule suppressions */
+
+  { ignores: ['functions-templates/**'] },
+  { ignores: ['site/**'] },
```
serhalp, line 35: 🤔 are these really temporary?

serhalp (approved):
Seems better than the alternative!

I'm kind of permanently scarred from past experiences with bulk "temporary suppressions"... but it might work for us.

image

eslint_temporary_suppressions.js
ndhoule: They are until we decide they're permanent!

PR #7152, opened 2025-03-27
build: slim down npm tasks and make them more consistent

ndhoule:
Notable changes here:

- Make testing targets more consistent. E.g. no more `npm run e2e`,
- Slim down how many tasks we have, including by getting rid of `test:ci` tasks. We already pass additional flags to Vitest in several CI workflows, so I don't see the benefit in maintaining these tasks.
- Don't automatically run `test:init`. We already manually run it in the integration CI workflow, so I'm following that workflow's lead. (Hopefully) soon we'll incorporate these setup steps into the tests themselves so you don't have to run any pre-test setup steps at all.
- Remove the explicit Vitest reporter. (This will make Vitest automatically use the GitHub Actions reporter in CI.)


.github/workflows/integration-tests.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           timeout_minutes: 15
           max_attempts: 3
           retry_on: error
-          command: npm run test:ci:vitest:integration -- --shard=${{ matrix.shard }}
+          command: npm run test:integration -- --coverage --shard=${{ matrix.shard }}
```
serhalp, line 79: huh, why do we enable coverage only for integration and not unit? 🤔 I guess were aren't reporting it anywhere anyway so...

ndhoule: We did enable coverage for both:

https://github.com/netlify/cli/pull/7152/files#diff-4761c5069b46ce96f49d74476da61f43ced3abc73d22254cd95a2be0e250a593R79

and still do enable coverage for unit tests:

https://github.com/netlify/cli/pull/7152/files#diff-cd0effce06f425599e96952f4d5f684641a683f46d9909e7c74a9f8b2f03ccd3R48

But yeah, we don't do anything meaningful with the reports.

PR #7161, opened 2025-03-31
fix(npm-shrinkwrap.json): re-remove dev deps

serhalp:
#### Summary

We accidentally reintroduced dev dependencies in the `npm-shrinkwrap.json` file here: https://github.com/netlify/cli/pull/7119/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.yml&show-viewed-files=true&show-deleted-files=false#diff-3d80e6f9eb8af7a28cbe2209d086150e29b68f860376ee618a0041ca9b5c786dL18-L49.

The deleted script was deleting `package.json#devDependencies` before running `npm i && npm shrinkwrap`, which resulted in the shrinkwrap being built from a modified `package-lock.json` with dev deps removed 😓.

cc @ndhoule

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "test:integration": "vitest run tests/integration/",
     "test:unit": "vitest run tests/unit/",
     "postinstall": "node ./scripts/postinstall.js",
-    "prepublishOnly": "npm shrinkwrap",
+    "prepublishOnly": "node ./scripts/prepublishOnly.js",
     "typecheck": "tsc",
     "typecheck:watch": "tsc --watch"
   },
--- a/scripts/prepublishOnly.js
+++ b/scripts/prepublishOnly.js
@@ -1,0 +1,35 @@
+import * as cp from 'node:child_process'
+import * as fs from 'node:fs/promises'
+import * as path from 'node:path'
+
+const main = async () => {
+  // It's best practice to include a shrinkwrap when shipping a CLI. npm has a bug that makes it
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "test:integration": "vitest run tests/integration/",
     "test:unit": "vitest run tests/unit/",
     "postinstall": "node ./scripts/postinstall.js",
-    "prepublishOnly": "npm shrinkwrap",
+    "prepublishOnly": "npm install --omit=dev && npm shrinkwrap",
```
serhalp, line 56: hrm, turns out this won't work... https://github.com/npm/cli/issues/4323. We should maybe just bring in https://www.npmjs.com/package/clean-publish 😢.

ndhoule: Don't we need to just add an `npm prune --production` step to the release workflow?

package.json
```diff
@@ -53,7 +53,7 @@
     "test:integration": "vitest run tests/integration/",
     "test:unit": "vitest run tests/unit/",
     "postinstall": "node ./scripts/postinstall.js",
-    "prepublishOnly": "npm shrinkwrap",
+    "prepublishOnly": "node ./scripts/prepublishOnly.js",
```
serhalp, line 56: `:skinner-children-wrong:`

serhalp (commented):
LGTM (can't review because I originally opened it)

ndhoule (approved):

PR #7171, opened 2025-04-07
perf!: perform minimal accounts query when loading config

ndhoule:
:warning: Depends on https://github.com/netlify/build/pull/6184. Merge and release that change first, and update the `@netlify/config` dependency in this PR to a non-RC release.

---

We've done this performance improvement many times elsewhere, but for posterity:

When we load a user's configuration, we perform a query to the Netlify API requesting information about which accounts a user is a member of. To do this, we query the `/accounts` endpoint, which is an expensive operation: It frequently takes 4+ seconds to get a response from this endpoint. This query happens as part of every CLI startup cycle.

By requesting a more minimal set of `accounts` data, we reduce almost every command's startup time.

Non-scientific testing shows this reduces `netlify status` runtime (on a Netlify dev's MacBook Pro M-something) from 3.907s to 2.001s, wall time. For longer-running commands the impact won't be so dramatic, proportionally speaking, but this is a nice perf bump anyway.

### DIFF
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -581,7 +581,7 @@ export default class BaseCommand extends Command {
       token,
       ...apiUrlOpts,
     })
-    const { accounts, buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
+    const { accounts = [], buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
     let { env } = cachedConfig
     if (flags.offlineEnv) {
       env = {}
--- a/src/commands/status/status.ts
+++ b/src/commands/status/status.ts
@@ -42,16 +42,8 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
     Name: user.full_name,
     Email: user.email,
     GitHub: ghuser,
+    Teams: accounts.map(({ name }) => name),
   }
-  const teamsData = {}
-
-  accounts.forEach((team) => {
-    // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
-    teamsData[team.name] = team.roles_allowed.join(' ')
-  })
-
-  // @ts-expect-error TS(2339) FIXME: Property 'Teams' does not exist on type '{ Name: a... Remove this comment to see the full error message
-  accountData.Teams = teamsData
 
   // @ts-expect-error
   const cleanAccountData = clean(accountData)
--- a/src/utils/types.ts
+++ b/src/utils/types.ts
@@ -184,23 +184,17 @@ type EnvVarValue = {
   context: string
 }
 
-export interface Account {
+export type MinimalAccount = {
   id: string
   name: string
   slug: string
-  type: string
-  capabilities: Record<string, { included: string; used: string }>
-  billing_name: string
-  billing_email: string
-  billing_details: string
-  billing_period: string
-  payment_method_id: string
+  default: boolean
+  team_logo_url: string | null
+  on_pro_trial: boolean
+  organization_id: string | null
   type_name: string
-  type_id: string
-  owner_ids: string[]
-  roles_allowed: string[]
-  created_at: string
-  updated_at: string
+  type_slug: string
+  members_count: number
 }
 
 export interface GitHubRepo {
--- a/tests/integration/commands/status/__snapshots__/status.test.ts.snap
+++ b/tests/integration/commands/status/__snapshots__/status.test.ts.snap
@@ -13,5 +13,8 @@ exports[`fixture: empty-project > should print status for a linked site 2`] = `
 {
   "Email": "test@netlify.com",
   "Name": "Test User",
+  "Teams": [
+    "Test User",
+  ],
 }
 `;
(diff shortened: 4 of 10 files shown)


src/utils/types.ts
```diff
@@ -184,6 +184,19 @@ type EnvVarValue = {
   context: string
 }
 
+export type MinimalAccount = {
```
serhalp, line 187: Why not import the type from `@netlify/config`?

src/commands/status/status.ts
```diff
@@ -42,16 +42,8 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
...
   }
-  const teamsData = {}
-
-  accounts.forEach((team) => {
-    // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
-    teamsData[team.name] = team.roles_allowed.join(' ')
-  })
-
-  // @ts-expect-error TS(2339) FIXME: Property 'Teams' does not exist on type '{ Name: a... Remove this comment to see the full error message
-  accountData.Teams = teamsData
```
serhalp, line 54: Technically this makes this PR breaking, since we're changing the output format and contents (including the structured json output). Was that intentional? If so, we should update the PR title to cut a major, and mention the output change in the PR description.

tests/integration/commands/status/__snapshots__/status.test.ts.snap
```diff
@@ -13,5 +13,8 @@ exports[`fixture: empty-project > should print status for a linked site 2`] = `
 {
   "Email": "test@netlify.com",
   "Name": "Test User",
+  "Teams": [
+    "Test User",
+  ],
```
serhalp, line 18: This accounts API response fixture is now wrong: https://github.com/netlify/cli/blob/f0fdfef9821fbfb806c9648811b5ee9ac0e5d173/tests/integration/commands/status/status.test.ts#L16. Can we update it? Maybe also pull in the `MinimalAccount` type there.

src/utils/types.ts
```diff
@@ -184,6 +184,19 @@ type EnvVarValue = {
...
+  default: boolean
+  team_logo_url: string | null
+  on_pro_trial: boolean
+  organization_id: string | null
+  type_name: string
+  type_slug: string
+  members_count: number
+}
+
 export interface Account {
```
serhalp, line 200: You can delete this! I checked.

src/commands/status/status.ts
```diff
@@ -42,16 +42,8 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
     Name: user.full_name,
     Email: user.email,
     GitHub: ghuser,
+    Teams: accounts.map(({ name }) => name),
```
serhalp, line 45: 🤔 so if it's true that this is the only use of `netlify.{cachedConfig\.,}accounts`, should we do a fast-follow PR to remove it and query here? 😈 We should take any oppportunity to remove stuff from BaseCommand, I think.

serhalp (approved):
LGTM. Maybe worth a quick sentence in the PR description about why this is safe? It took me pulling down the branch and poking around all the ways we use account data to understand that every other case queries for the account data some other way, i.e. that this is only used by `status` despite being on `cmd.netlify.accounts` and `cmd.netlify.cachedConfig.accounts`.

src/utils/types.ts
ndhoule: We don't export it from that module and this is not a public API, so we don't want to make this part of any module's public API.

src/commands/base-command.ts
```diff
@@ -575,7 +575,7 @@ export default class BaseCommand extends Command {
       token,
       ...apiUrlOpts,
     })
-    const { accounts, buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
+    const { accounts = [], buildDir, config, configPath, repositoryRoot, siteInfo } = cachedConfig
```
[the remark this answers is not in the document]
ndhoule: I can go double check/find the relevant LoCs, but IIRC it returns `undefined` if any of the HTTP calls in `netlify/config` fail. This can happen e.g. if you delete a site via the Netlify UI but your local repo is still linked.

src/commands/status/status.ts
ndhoule: It's a porcelain interface, so this is not a breaking change. (We noted that this would be a breaking change in the issue for this, and that the `--json` output for this command isn't working--for me, anyway.)

src/commands/status/status.ts
ndhoule: Probably. I'd think a little bigger: I'd like to figure out what the impact of lifting _all_ of the API calls out of `@netlify/config` would be so we have better control over startup sequence timing/reduce unnecessary calls/investigate a short-lived persistent cache.

src/commands/status/status.ts
serhalp: This changes both the porcelain output and the `--json` output, both of which work for me at the moment. 🤔

ndhoule: Noted! I'll make this a breaking change, then.

PR #7173, opened 2025-04-08
perf: speed up all commands by 75+ ms by enabling V8 code cache when available (`node>=22.8.0`)

serhalp:
#### Summary

See https://nodejs.org/api/module.html#module-compile-cache.

In local benchmarking on various machines, this led to a **75-400 ms improvement** to command run time across the board, at the expense of a moderate increase in first-time run time.

See inline - to account for the common CI single-run use case, this is disabled when we detect a CI environment.

See individual commits for context on the other changes here.

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -1,17 +1,48 @@
 #!/usr/bin/env node
+import * as module from 'module'
+import { tmpdir } from 'os'
+import * as path from 'path'
 import { argv } from 'process'
 
+import { isCI } from 'ci-info'
 import updateNotifier from 'update-notifier'
 
-import { createMainCommand } from '../dist/commands/main.js'
-import { logError } from '../dist/utils/command-helpers.js'
 import getPackageJson from '../dist/utils/get-cli-package-json.js'
-import { runProgram } from '../dist/utils/run-program.js'
 
 // 12 hours
 const UPDATE_CHECK_INTERVAL = 432e5
 
+/* eslint-disable n/no-unsupported-features/node-builtins */
+const maybeEnableCompileCache = () => {
+  // The Netlify CLI is often used in CI workflows. In this context, we wouldn't want the overhead of the first run
+  // because we almost certainly would not get any benefits on "subsequent runs". Even if the user has configured
+  // caching of the CLI itself, there's no chance they've configured the V8 compile cache directory to be cached.
+  if (isCI) return false
+
+  // This was added in node 22.8.0, but we currently support >=18.14.0. As the CLI is generally run hundreds of times,
+  // it is worth a small overhead on the very first run post-install to shave tens to hundreds of milliseconds on every
+  // subsequent run.
+  // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+  if ('enableCompileCache' in module && typeof module.enableCompileCache === 'function') {
+    // By design, this cannot throw.
+    // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+    module.enableCompileCache()
+    // This has no impact on this process. It enables the compile cache for spawned subprocesses.
+    process.env.NODE_COMPILE_CACHE ??= path.join(tmpdir(), 'node-compile-cache')
+    return true
+  }
+  return false
+}
+/* eslint-enable n/no-unsupported-features/node-builtins */
+
 const main = async () => {
+  // Must come first, including before any imports
+  const didEnableCompileCache = maybeEnableCompileCache()
+
+  const { createMainCommand } = await import('../dist/commands/main.js')
+  const { logError } = await import('../dist/utils/command-helpers.js')
+  const { runProgram } = await import('../dist/utils/run-program.js')
+
   const pkg = await getPackageJson()
 
   try {
@@ -23,7 +54,7 @@ const main = async () => {
     logError(`Error checking for updates: ${error?.toString()}`)
   }
 
-  const program = createMainCommand()
+  const program = createMainCommand(didEnableCompileCache)
 
   try {
     await runProgram(program, argv)
--- a/bin/run.js
+++ b/bin/run.js
@@ -1,44 +1,15 @@
 #!/usr/bin/env node
-import * as module from 'module'
-import { tmpdir } from 'os'
-import * as path from 'path'
 import { argv } from 'process'
 
-import { isCI } from 'ci-info'
 import updateNotifier from 'update-notifier'
 
 import getPackageJson from '../dist/utils/get-cli-package-json.js'
+import { maybeEnableCompileCache } from '../dist/utils/nodejs-compile-cache.js'
 
 // 12 hours
 const UPDATE_CHECK_INTERVAL = 432e5
 
-/* eslint-disable n/no-unsupported-features/node-builtins */
-const maybeEnableCompileCache = () => {
-  // The Netlify CLI is often used in CI workflows. In this context, we wouldn't want the overhead of the first run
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -203,7 +204,6 @@
     "eslint-config-prettier": "^10.1.1",
     "eslint-plugin-n": "^17.16.1",
     "form-data": "4.0.2",
-    "is-ci": "4.1.0",
```
serhalp, line 206: was unused

src/commands/main.ts
```diff
@@ -239,6 +238,8 @@ export const createMainCommand = () => {
   createWatchCommand(program)
   createLogsCommand(program)
 
+  program.setAnalyticsPayload({ didEnableCompileCache })
```
serhalp, line 241: This will allow us to see the real-world perf impact

src/utils/live-tunnel.ts
```diff
@@ -74,7 +74,7 @@ const connectTunnel = function ({
   }
 
   const ps = execa(execPath, args, { stdio: 'inherit' })
-  ps.on('close', (code) => process.exit(code))
+  ps.on('close', (code) => process.exit(code ?? undefined))
```
serhalp, line 77: only change needed after downgrading `@types/node`

bin/run.js
```diff
@@ -1,32 +1,69 @@
...
 }
+/* eslint-enable n/no-unsupported-features/node-builtins */
+
+const main = async () => {
+  // Must come first, including before any imports
+  const didEnableCompileCache = maybeEnableCompileCache()
+
+  const { createMainCommand } = await import('../dist/commands/main.js')
+  const { logError } = await import('../dist/utils/command-helpers.js')
+  const { runProgram } = await import('../dist/utils/run-program.js')
```
serhalp, line 45: I could also have just split this `bin/run.js` entrypoint into two files, the first of which just calls `maybeEnableCompileCache` and then `await import('./inner_entrypoint.js')`. If the reviewer cares, I can refactor to that instead of this.

src/commands/base-command.ts
```diff
@@ -378,7 +380,11 @@ export default class BaseCommand extends Command {
         duration,
         status,
       })
-    } catch {}
+    } catch (err) {
+      debug(`${this.name()}:onEnd`)(
+        `Command: ${command}. Telemetry tracking failed: ${err instanceof Error ? err.message : err?.toString()}`,
+      )
+    }
```
serhalp, line 387: Not very related but I ran into this time sink while developing locally and it seems worth fixing

tests/integration/telemetry.test.ts
```diff
@@ -110,20 +113,15 @@ await withMockApi(routes, () => {
...
-      })
+      expect(request!.body).toHaveProperty(
+        'properties',
+        expect.objectContaining({
+          frameworks: ['next'],
+          buildSystem: [],
+          // TODO: this should be NPM however some CI tests are using pnpm which changes the value
+          packageManager: expect.stringMatching(/npm|pnpm/),
+        }),
+      )
```
serhalp, line 124: this test is `should add frameworks, buildSystem, and packageManager` — there's no need to re-test the same behaviours already tested elsewhere

src/commands/main.ts
ndhoule: Ideally, `module` would offer an API for checking if the cache is enabled, but oh well.

Given that this is diagnostic information about the environment, rather than threading it, I would break `maybeEnableCompileCache` into a module that captures the result of calling `module.enableCompileCache` and exports a function `isCompileCacheEnabled` that you can use here. That would let you avoid threading this information around and avoid this mysterious `boolean` argument.

If you continue to thread this information, I would move this into an options bag so it's more self-documenting (and to make room for other options later).

ndhoule (changes requested):
Minor suggestions around how to make this a little cleaner and better documented--address those as you see fit. I marked the only potential blocking issue with a :warning:.

bin/run.js
```diff
@@ -1,32 +1,68 @@
...
+  // This was added in node 22.8.0, but we currently support >=18.14.0. As the CLI is generally run hundreds of times,
+  // it is worth a small overhead on the very first run post-install to shave tens to hundreds of milliseconds on every
+  // subsequent run.
+  // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+  if ('enableCompileCache' in module && typeof module.enableCompileCache === 'function') {
+    // By design, this cannot throw.
+    // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+    module.enableCompileCache()
+    // This has no impact on this process. It enables the compile cache for spawned subprocesses.
+    process.env.NODE_COMPILE_CACHE ??= path.join(tmpdir(), 'node-compile-cache')
```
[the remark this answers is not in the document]
serhalp: Every spawned child process that I could find was already inheriting the parent process's `env`, so this should just work, although it's rather difficult to verify.

What was the ⚠️ for here? I might be misunderstanding. Is there a functional risk here?

bin/run.js
```diff
@@ -1,32 +1,68 @@
...
-
-try {
-  updateNotifier({
-    pkg,
-    updateCheckInterval: UPDATE_CHECK_INTERVAL,
-  }).notify()
-} catch (error) {
-  logError(`Error checking for updates: ${error?.toString()}`)
+
+/* eslint-disable n/no-unsupported-features/node-builtins */
```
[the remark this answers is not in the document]
serhalp: Yeah, I guess. I just didn't like needing to disable it twice.

bin/run.js
```diff
@@ -1,32 +1,68 @@
...
+  // caching of the CLI itself, there's no chance they've configured the V8 compile cache directory to be cached.
+  if (isCI) return false
+
+  // This was added in node 22.8.0, but we currently support >=18.14.0. As the CLI is generally run hundreds of times,
+  // it is worth a small overhead on the very first run post-install to shave tens to hundreds of milliseconds on every
+  // subsequent run.
+  // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
+  if ('enableCompileCache' in module && typeof module.enableCompileCache === 'function') {
+    // By design, this cannot throw.
+    // @see https://nodejs.org/api/module.html#moduleenablecompilecachecachedir
```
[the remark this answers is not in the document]
serhalp: 😅 Yeah, I'll give it a try. I moved all these comments around so many times and nothing felt right.

src/commands/main.ts
serhalp: It does provide that information. Good call. It wasn't even accurate to mark it as enabled in all cases. Done.

bin/run.js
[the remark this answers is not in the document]
serhalp: (We discussed elsewhere. For now, we'll remove this and keep the scope of this PR limited to just enabling the compile cache in the main process.)

bin/run.js
```diff
@@ -1,32 +1,68 @@
...
+    // This has no impact on this process. It enables the compile cache for spawned subprocesses.
+    process.env.NODE_COMPILE_CACHE ??= path.join(tmpdir(), 'node-compile-cache')
+    return true
+  }
+  return false
 }
+/* eslint-enable n/no-unsupported-features/node-builtins */
+
+const main = async () => {
+  // Must come first, including before any imports
```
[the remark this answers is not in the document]
serhalp: Yeah, I hadn't had enough coffee yet when I did this and I couldn't remember how to get default exports to work with async imports 😶. Done now!

ndhoule (approved):

ndhoule:
:racehorse:

PR #7180, opened 2025-04-09
fix: await `zipFunctions()` in functions build

serhalp:
#### Summary

The types fixes in https://github.com/netlify/cli/pull/7130 identified a number of bugs. This fixes one of those.


src/utils/functions/functions.ts
```diff
@@ -22,7 +22,11 @@ export const getFunctionsDir = (
     options: OptionValues
   },
   defaultValue?: string,
-) => options.functions || config.dev?.functions || config.functionsDirectory || defaultValue
+): string | undefined =>
+  ('functions' in options && typeof options.functions === 'string' ? options.functions : null) ??
+  config.dev?.functions ??
+  config.functionsDirectory ??
```
ndhoule, line 28: Isn't this a behavioral change? Won't this now allow `config.dev.functions` and `config.functionsDirectory` to be an empty string? Or do we sanitize empty strings out somewhere?

src/commands/functions/functions-build.ts
```diff
@@ -10,7 +10,7 @@ import type BaseCommand from '../base-command.js'
 export const functionsBuild = async (options: OptionValues, command: BaseCommand) => {
   const { config } = command.netlify
 
-  const src = options.src || config.build.functionsSource
+  const src = ('src' in options && typeof options.src === 'string' ? options.src : null) ?? config.build.functionsSource
```
ndhoule, line 13: Same here--this will now allow an empty string, won't it?

ndhoule (changes requested):

src/utils/functions/functions.ts
serhalp: I'm not sure how you could even end up with an empty string, other than explicitly using `""` in your `netlify.toml` or something... 🤔

I feel like it isn't worth the added complexity to this line to support ignoring hypothetical empty strings here, but in the interest of limiting behavioural changes I'll go ahead and add the conditions 🙈.

ndhoule: I don't think it's worth the added complexity here or the change risk, however low that risk is (you're still adding in new behavior by checking for strings with only whitespace in them). I don't think that bundling it in with a clear and totally unrelated bug is really the way to go about this change, either.

ndhoule (approved):

src/utils/functions/functions.ts
serhalp: I thought that `trim()` wouldn't be controversial, but alright, I've reverted that too.

PR #7181, opened 2025-04-09
fix: actually populate auth `user` field

serhalp:
#### Summary

The types fixes in https://github.com/netlify/cli/pull/7130 identified a number of bugs. This fixes one of those. We were reading from the wrong property on the GitHub API response.


src/commands/base-command.ts
```diff
@@ -401,7 +401,7 @@ export default class BaseCommand extends Command {
   }
 
   async expensivelyAuthenticate() {
-    const webUI = process.env.NETLIFY_WEB_UI || 'https://app.netlify.com'
+    const webUI = process.env.NETLIFY_WEB_UI ?? 'https://app.netlify.com'
```
ndhoule, line 404: This is a behavior change, is this intentional?

ndhoule (changes requested):

src/commands/base-command.ts
serhalp: Yes. It seemed totally reasonable to me to treat `NETLIFY_WEB_UI=''` — an env var intended to be a URL override — as absent, and to not consider this a breaking change.

Especially given that it's an undocumented env var really only used by Netlify employees 😄.

ndhoule: What is the benefit to changing the behavior here?

Everywhere else in the CLI that I can find, we treat empty strings as though they're unset. This is how every other program I've ever used treats empty environment variables, too.

serhalp: Sounds like you feel strongly. Reverted.

src/utils/gh-auth.ts
```diff
@@ -32,7 +32,7 @@ const promptForAuthMethod = async () => {
         'What would you like to do?',
       choices: authChoices,
     },
-  ])
+  ])) as { authMethod: typeof authChoices[number] }
```
ndhoule, line 35:
```suggestion
  ]))
```

ndhoule (approved):

PR #7182, opened 2025-04-09
refactor: remove unused args in calls to `getBuildSettings`

serhalp:
#### Summary

The types fixes in https://github.com/netlify/cli/pull/7130 identified a number of issues. This fixes one of those — just some unused args passed to a function.

### DIFF
--- a/src/utils/init/config-github.ts
+++ b/src/utils/init/config-github.ts
@@ -19,11 +19,11 @@ const PAGE_SIZE = 100
  * Get a valid GitHub token
  */
 export const getGitHubToken = async ({ globalConfig }: { globalConfig: GlobalConfigStore }): Promise<string> => {
-  const userId = globalConfig.get('userId')
+  const userId: string = globalConfig.get('userId')
 
   const githubToken: Token = globalConfig.get(`users.${userId}.auth.github`)
 
-  if (githubToken && githubToken.user && githubToken.token) {
+  if (githubToken.user && githubToken.token) {
     try {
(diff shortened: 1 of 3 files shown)


src/utils/init/config-github.ts
```diff
@@ -19,11 +19,11 @@ const PAGE_SIZE = 100
  * Get a valid GitHub token
  */
 export const getGitHubToken = async ({ globalConfig }: { globalConfig: GlobalConfigStore }): Promise<string> => {
-  const userId = globalConfig.get('userId')
+  const userId: string = globalConfig.get('userId')
 
   const githubToken: Token = globalConfig.get(`users.${userId}.auth.github`)
 
-  if (githubToken && githubToken.user && githubToken.token) {
+  if (githubToken.user && githubToken.token) {
```
ndhoule, line 26: Is this right? Where do we validate that `githubToken` is not `undefined`?

ndhoule (changes requested):

src/utils/init/config-github.ts
serhalp: Bleh, good catch. What happened here is `eslint --fix` changed this automatically due to the incorrect `Token` type.

The `GlobalConfigStore` refactor was great but we should fix that `any`. Dunno what the best approach will be there though... maybe use zod?

ndhoule: Yep, we should parse the config file when we load it (the challenge is being very careful about the shape of the global config file).

ndhoule (approved):

PR #7185, opened 2025-04-09
fix: mark `siteInfo.published_deploy` as optional

serhalp:
#### Summary

The types fixes in https://github.com/netlify/cli/pull/7130 identified a number of issues. This fixes a couple of issues.

- mark `siteInfo.published_deploy` as optional
- remove unnecessary handling of nullish `siteInfo`
- fix some nearby eslint suppressions

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -841,7 +841,7 @@ export const deploy = async (options: OptionValues, command: BaseCommand) => {
     return triggerDeploy({ api, options, siteData, siteId })
   }
 
-  const deployToProduction = options.prod || (options.prodIfUnlocked && !siteData.published_deploy.locked)
+  const deployToProduction = options.prod || (options.prodIfUnlocked && siteData.published_deploy?.locked !== true)
 
   let results = {} as Awaited<ReturnType<typeof prepAndRunDeploy>>
 
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -841,7 +841,7 @@ export const deploy = async (options: OptionValues, command: BaseCommand) => {
     return triggerDeploy({ api, options, siteData, siteId })
   }
 
-  const deployToProduction = options.prod || (options.prodIfUnlocked && siteData.published_deploy?.locked !== true)
+  const deployToProduction = options.prod || (options.prodIfUnlocked && !(siteData.published_deploy?.locked ?? false))
 
   let results = {} as Awaited<ReturnType<typeof prepAndRunDeploy>>
 
--- a/src/commands/status/status.ts
+++ b/src/commands/status/status.ts
@@ -16,14 +16,14 @@ import type BaseCommand from '../base-command.js'
 
 export const status = async (options: OptionValues, command: BaseCommand) => {
   const { accounts, api, globalConfig, site, siteInfo } = command.netlify
-  const current = globalConfig.get('userId') as string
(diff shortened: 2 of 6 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -841,7 +841,7 @@ export const deploy = async (options: OptionValues, command: BaseCommand) => {
     return triggerDeploy({ api, options, siteData, siteId })
   }
 
-  const deployToProduction = options.prod || (options.prodIfUnlocked && !siteData.published_deploy.locked)
+  const deployToProduction = options.prod || (options.prodIfUnlocked && siteData.published_deploy?.locked !== true)
```
ndhoule, line 844: Minor point, but I think it's way clearer to default here rather than compare `undefined | boolean` to `boolean`.

src/commands/status/status.ts
```diff
@@ -37,16 +46,18 @@ export const status = async (options: OptionValues, command: BaseCommand) => {
     }
   }
 
-  const ghuser = command.netlify.globalConfig.get(`users.${current}.auth.github.user`)
+  const ghuser = command.netlify.globalConfig.get(`users.${current}.auth.github.user`) as string
```
ndhoule, line 49: What guarantees that this is a string and not `undefined`?

src/commands/status/status.ts
```diff
@@ -1,13 +1,22 @@
...
+  logJson,
+  warn,
+  type APIError,
+} from '../../utils/command-helpers.js'
+import type BaseCommand from '../base-command.js'
 
 export const status = async (options: OptionValues, command: BaseCommand) => {
   const { accounts, api, globalConfig, site, siteInfo } = command.netlify
-  const current = globalConfig.get('userId')
+  const current = globalConfig.get('userId') as string
```
ndhoule, line 19: What guarantees that this is a string and not `undefined`?

src/utils/functions/functions.ts
```diff
@@ -22,7 +22,11 @@ export const getFunctionsDir = (
     options: OptionValues
   },
   defaultValue?: string,
-) => options.functions || config.dev?.functions || config.functionsDirectory || defaultValue
+): string | undefined =>
```
ndhoule, line 25: https://github.com/netlify/cli/pull/7180

I think you've duplicated the fix from 7130 here; same comments apply.

ndhoule (changes requested):

src/commands/status/status.ts
serhalp: This function explodes if it isn't, but you're right — I'll fix the types and deal with the fallout

src/commands/status/status.ts
serhalp: I believe there's some convoluted implicit guarantees (something like: if it failed to get a token/user, it wouldn't reach this code), but as above... I'll fix the type and deal with the fallout. 👍🏼

src/commands/deploy/deploy.ts
serhalp: Hmm, can you share a concrete diff suggestion? Are you suggesting this:

```suggestion
  const deployToProduction = options.prod || (options.prodIfUnlocked && (siteData.published_deploy?.locked ?? false))
```

ndhoule:
@serhalp Ping me once CI is running again and you've resolved conflicts and I'll have a look.

ndhoule (approved):

PR #7190, opened 2025-04-10
fix: print site information in `link` output

ndhoule:
This changeset updates `@netlify/config` to fix an error in `netlify link` where we sometimes print `undefined` in place of site information.

This addresses the root issue, but I'll create a follow-up PR that addresses the root issue (bad types).

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,10 +60,10 @@
   "dependencies": {
     "@fastify/static": "7.0.4",
     "@netlify/blobs": "8.1.2",
-    "@netlify/build": "31.0.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
     "@netlify/blobs": "8.1.2",
     "@netlify/build": "30.1.1",
```
serhalp, line 63: we can't bump build here too at this point? this pulls in two versions of it now

serhalp (approved):

PR #7194, opened 2025-04-10
refactor: type deploy command input args

ndhoule:
This change adds basic TypeScript types to the `deploy` command's input arguments (i.e. the command-line flags that a user passes to the `netlify deploy` command).

I also lifted some input argument validation out of the `deploy` command's business logic into the `createDeployCommand` function as a minor optimization so we don't pull in the full deploy dependency tree when we know we can't service a user's request.

### DIFF
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -148,6 +148,15 @@ async function getRepositoryRoot(cwd?: string): Promise<string | undefined> {
   }
 }
 
+export type BaseOptionValues = {
+  auth?: string
+  cwd?: string
+  debug?: boolean
+  filter?: string
+  httpProxy?: string
+  silent?: string
+}
+
 /** Base command class that provides tracking and config initialization */
 export default class BaseCommand extends Command {
   /** The netlify object inside each command with the state */
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -2,17 +2,23 @@ import { type Stats } from 'fs'
 import { stat } from 'fs/promises'
 import { basename, resolve } from 'path'
 
-import { type NetlifyConfig, runCoreSteps } from '@netlify/build'
-import { type OptionValues } from 'commander'
+import { type NetlifyConfig, type OnPostBuild, runCoreSteps } from '@netlify/build'
 import inquirer from 'inquirer'
 import isEmpty from 'lodash/isEmpty.js'
 import isObject from 'lodash/isObject.js'
 import { parseAllHeaders } from '@netlify/headers-parser'
 import { parseAllRedirects } from '@netlify/redirect-parser'
+import type { NetlifyAPI } from 'netlify'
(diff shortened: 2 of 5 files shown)


src/lib/build.ts
```diff
@@ -115,7 +115,7 @@ interface HandlerResult {
   status?: string
 }
 // The @netlify/build type incorrectly states a `void | Promise<void>` return type.
-type PatchedHandlerType<T extends (opts: any) => void | Promise<void>> = (
```
serhalp, line 118: 🤔 Wait, what is this existing comment even saying?

src/commands/base-command.ts
```diff
@@ -148,6 +148,15 @@ async function getRepositoryRoot(cwd?: string): Promise<string | undefined> {
...
 }
 
+export type BaseOptionValues = {
+  auth?: string
+  cwd?: string
+  debug?: boolean
+  filter?: string
+  httpProxy?: string
+  silent?: string
+}
```
serhalp, line 158: 😍

src/commands/base-command.ts
```diff
@@ -148,6 +148,15 @@ async function getRepositoryRoot(cwd?: string): Promise<string | undefined> {
   }
 }
 
+export type BaseOptionValues = {
```
serhalp, line 151: `json` is a weird one IIRC. I moved it out of BaseCommand but it's still referenced in a generic context in a couple places. I guess we'll gross that bridge when we get there.

src/commands/deploy/deploy.ts
```diff
@@ -834,6 +853,12 @@ export const deploy = async (options: OptionValues, command: BaseCommand) => {
     }
   }
 
+  if (!siteId) {
+    return logAndThrowError(
+      "Unable to determine which site to deploy to. Make sure you've run 'netlify link' or that you're specifying your desired site using the '--site' option.",
+    )
+  }
+
```
serhalp, line 853: I think you only need this because `initChoice` is untyped above, so TS doesn't know that the checks above are exhaustive?

I believe you can just add `as const` to line 834 and `as {initChoice: typepof initializeOpts[number]}` on line 842. Come to think of it, I think I have this on a branch somewhere...

serhalp (approved):
Woo, thanks for establishing a pattern for this!

src/commands/deploy/deploy.ts
ndhoule: This doesn't have anything to do with `initChoice`--need this because the `if`/`else if` conditions on L837-845 are not exhaustive and do not unconditionally set `siteId`.

serhalp: No, the conditions are exhaustive. You need this because `initChoice` is `any` and `siteId` is `any`. I'll open a PR.

serhalp: https://github.com/netlify/cli/pull/7220

PR #7195, opened 2025-04-10
feat!: run build by default before deploy

ndhoule:
## Summary

This updates the `deploy` command to run the configured `build` command before deployment by default. Previously, this was achieved by passing the `--build` flag.

To opt into the previous behavior, users can now specify `netlify deploy --no-build`.

## Why are we changing this?

This has long been one of the most common sources of confusion with the Netlify CLI. Developers expect a "deploy" command to deploy their project. We're making this change to follow the Principle of Least Surprise and streamline the experience for new users.

### (most common) Build and deploy

Remove the `--build` flag:

```sh
netlify deploy --build [...] # before
```
→
```sh
netlify deploy [...] # after
```

### Separate build and deploy

> [!WARNING]
> Make sure your deploy command runs in your expected environment, e.g. if you were running `FOO=bar netlify build && netlify deploy`, make sure `FOO` is visible to the deploy command, which is now running the build: `FOO=bar netlify deploy`.

Collapse into one command, if you can:

```sh
netlify build && netlify deploy # before
```
→
```sh
netlify deploy # after
```

Otherwise, add the `--no-build` flag:

```sh
make spaghetti && netlify deploy # before
```
→
```sh
make spaghetti && netlify deploy --no-build # after
```

### (rare) Deploy without build step

Add the `--no-build` flag:

```sh
netlify deploy [...] # before (there was no `--build` here)
```
→
```sh
netlify deploy --no-build [...]
```

### DIFF
--- a/src/commands/deploy/index.ts
+++ b/src/commands/deploy/index.ts
@@ -3,7 +3,7 @@ import { env } from 'process'
 import { Option } from 'commander'
 
 import BaseCommand from '../base-command.js'
-import { logAndThrowError, warn } from '../../utils/command-helpers.js'
+import { chalk, logAndThrowError, warn } from '../../utils/command-helpers.js'
 import type { DeployOptionValues } from './option_values.js'
 
 export const createDeployCommand = (program: BaseCommand) =>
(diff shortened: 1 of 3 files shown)


src/commands/deploy/index.ts
```diff
@@ -109,7 +109,19 @@ Support for package.json's main field, and intrinsic index.js entrypoints are co
         'build',
       ),
     )
-    .option('--build', 'Run build command before deploying', false)
+    .addOption(
+      new Option('--build', 'Do not use - this is now the default. Will be removed in future versions.')
```
serhalp, line 113: FWIW I don't think this string can ever be seen

sarahetter: Yeah these get added to generated cli.netlify.com docs

tests/integration/commands/deploy/deploy.test.ts
```diff
@@ -293,6 +295,111 @@ describe.skipIf(process.env.NETLIFY_TEST_DISABLE_LIVE === 'true').concurrent('co
...
+          path: 'public/index.html',
+          content,
+        })
+        .withNetlifyToml({
+          config: {
+            build: { publish: 'public' },
+            plugins: [{ package: './plugins/log-env' }],
+          },
+        })
+        .withBuildPlugin({
```
sarahetter, line 312: we're not asserting on having a build plugin, this just adds mess?

src/commands/deploy/index.ts
serhalp: Well, because we have `hideHelp()` it isn't even included there

PR #7203, opened 2025-04-14
test: fix flakey snapshot normalization logic

serhalp:
#### Summary

This logic was normalizing integration test snapshots by replacing any sequence of five digits with `88888`. The intent is to handle some randomness in test output.

It wasn't working some percentage of the time because randomly selected ports are sometimes _four_ digits. You can see an example in this failed run: https://github.com/netlify/cli/actions/runs/14434953424/job/40474516002?pr=7202#step:10:1143.

While I was at it, I made it less much less generic and moved it into the only test suite that relies on it.

### DIFF
--- a/tests/integration/framework-detection.test.js
+++ b/tests/integration/framework-detection.test.js
@@ -13,8 +13,8 @@ const content = 'Hello World!'
 // Normalize random ports
 const normalizeSnapshot = (output, opts) =>
   normalize(output, opts)
-    .replaceAll(/localhost:\d+/g, 'localhost:88888')
-    .replaceAll(/listening to \d+/g, 'listening to 88888')
+    .replaceAll(/localhost:\d+/g, 'localhost:<SNAPSHOT_PORT_NORMALIZED>')
+    .replaceAll(/listening to \d+/g, 'listening to <SNAPSHOT_PORT_NORMALIZED>')
 
 describe.concurrent('frameworks/framework-detection', () => {
   test('should default to process.cwd() and static server', async (t) => {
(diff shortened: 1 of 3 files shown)


tests/integration/framework-detection.test.js
```diff
@@ -10,6 +10,12 @@ import { normalize } from './utils/snapshots.js'
 
 const content = 'Hello World!'
 
+// Normalize random ports
+const normalizeSnapshot = (output, opts) =>
+  normalize(output, opts)
+    .replaceAll(/localhost:\d+/g, 'localhost:88888')
+    .replaceAll(/listening to \d+/g, 'listening to 88888')
```
ndhoule, line 17: Not critical, but: I would change this to `{PORT_NORMALIZED}` to make it completely clear to anyone viewing the output/snapshot that we normalize this. Using an out-of-range value is a hint but I think it'd be kinder to be more explicit.

tests/integration/framework-detection.test.js
```diff
@@ -10,6 +10,12 @@ import { normalize } from './utils/snapshots.js'
 
 const content = 'Hello World!'
 
+// Normalize random ports
+const normalizeSnapshot = (output, opts) =>
```
ndhoule, line 14: :clap:

Love to see this vs. one big world-normalizing function.

ndhoule (approved):

tests/integration/framework-detection.test.js
serhalp: Good idea! Pushing.

PR #7207, opened 2025-04-15
ci(e2e): add same Windows perf config as for integration tests

serhalp:
#### Summary

We have this incantation in the integration test setup, and e2e tests are _excruciatingly_ slow on Windows (4-7 mins PER TEST, vs. 400-900ms for Ubuntu/macOS), so let's give this a try.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -24,6 +24,12 @@ jobs:
         node-version: ['18.14.0', '20.12.2', '22']
       fail-fast: false
     steps:
+      # This improves Windows network performance. We need this since we open many ports in our tests.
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -17,15 +17,19 @@ jobs:
     name: E2E Windows tests
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
-    env:
-      DEBUG: true
```
serhalp, line 21: who knows, this could have been hurting perf too. It looks like this has been here from the very start. 🤷🏼

ndhoule (approved):

PR #7208, opened 2025-04-15
fix(deps): upgrade misc. dependencies

serhalp:
#### Summary

We've let a bit too many accumulate to manage effectively with Renovate, especially with the flakiness (now resolved!?). This upgrades all the easy ones.

As a bonus, this pulls in some recent `esbuild` deduping from netlify/build which significantly shrinks the dependency tree size (though about half of this is a mirage, as it had recently introduced _more_ esbuild versions).


package.json
```diff
@@ -147,23 +147,23 @@
     "source-map-support": "0.5.21",
     "strip-ansi-control-characters": "2.0.0",
     "tempy": "3.1.0",
-    "terminal-link": "3.0.0",
+    "terminal-link": "4.0.0",
```
serhalp, line 150: just requires node 18: https://github.com/sindresorhus/terminal-link/releases/tag/v4.0.0

package.json
```diff
@@ -186,7 +186,7 @@
     "@types/parallel-transform": "1.1.4",
     "@types/parse-github-url": "1.0.3",
     "@types/parse-gitignore": "1.0.2",
-    "@types/picomatch": "^3.0.2",
+    "@types/picomatch": "4.0.0",
```
serhalp, line 189: we're on picomatch@4, so this was wrong anyway

ndhoule (approved):
clapping_futurama

PR #7211, opened 2025-04-15
fix(deps): remove or type last remaining untyped deps

serhalp:
### 563e0b4b add type declaration for untyped `express-logging`

There's also https://github.com/netlify/cli/pull/7050 that removes it entirely, but I don't want to be blocked on this to continue improving type safety.

### fe8e4a05 remove untyped `strip-ansi-control-characters` dep

`stripVTControlCharacters` was added in 16.11.0 (we support >+18.14.0), so we don't need this tiny transform stream wrapper anyway

### 1049c9a1 fix types and eslint errors in `utils/shell.ts`

### DIFF
--- a/src/utils/copy-template-dir/copy-template-dir.ts
+++ b/src/utils/copy-template-dir/copy-template-dir.ts
@@ -24,13 +24,9 @@ import path from 'path'
 import { pipeline } from 'stream'
 import { promisify } from 'util'
 
-// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'maxstache... Remove this comment to see the full error message
 import maxstache from 'maxstache'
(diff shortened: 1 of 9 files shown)


src/utils/copy-template-dir/copy-template-dir.ts
```diff
@@ -56,9 +52,7 @@ async function writeFile(outDir: string, vars: Record<string, string>, file: Ent
 }
 
 // High throughput template dir writes
-export async function copyTemplateDir(srcDir: string, outDir: string, vars: any): Promise<string[]> {
-  if (!vars) vars = noop
```
serhalp, line 60: there's only one caller and it always passes this in

ndhoule (approved):

PR #7222, opened 2025-04-17
test: remove skipped monorepo integration tests

serhalp:
#### Summary

I'm not sure why we [added skipped tests](https://github.com/netlify/cli/commit/05aed176ecf23eaa6047d555262f177c26ea9741), but these aren't providing any value and they are creating eslint/typescript noise.


serhalp:
hoping to fix instead

PR #7234, opened 2025-04-23
refactor(utils/env): rename variables, add comments, and clarify intent

serhalp:
#### Summary

In preparation for a bug fix, this refactors code paths related to fetching and filtering environment variables for the optionally user-specified context and/or scope.

This should be pretty self-explanatory. I just renamed variables for clarity, fixed some missing types, added some inline doc comments, and did some minor refactoring.

Feel free to review commit by commit.

### DIFF
--- a/src/utils/env/index.ts
+++ b/src/utils/env/index.ts
@@ -100,15 +100,18 @@ export const getValueForContext = (
   /**
    * The deploy context or branch of the environment variable value
    */
-  context: string,
-): EnvelopeEnvVarValue | undefined =>
-  values.find((val) => {
-    if (!(SUPPORTED_CONTEXTS as readonly string[]).includes(context)) {
+  contextOrBranch: ContextOrBranch,
+): EnvelopeEnvVarValue | undefined => {
+  const valueForContext = values.find((val) => {
+    const isSupportedContext = (SUPPORTED_CONTEXTS as readonly string[]).includes(contextOrBranch)
+    if (!isSupportedContext) {
       // the "context" option passed in is actually the name of a branch
-      return val.context === 'all' || val.context_parameter === context
+      return val.context === 'all' || val.context_parameter === contextOrBranch
     }
-    return val.context === 'all' || val.context === context
+    return val.context === 'all' || val.context === contextOrBranch
   })
+  return valueForContext ?? undefined
+}
 
 /**
  * Finds environment variables that match a given source
--- a/src/utils/env/index.ts
+++ b/src/utils/env/index.ts
@@ -1,6 +1,5 @@
 import type { NetlifyAPI } from 'netlify'
 
-import { $TSFixMe } from '../../commands/types.js'
 import { logAndThrowError } from '../command-helpers.js'
 import type { SiteInfo, EnvironmentVariableSource } from '../../utils/types.js'
 
@@ -22,12 +21,16 @@ const SUPPORTED_CONTEXT_ALIASES = {
  * These exactly match possible `scope` values returned by the Envelope API.
  * Note that `any` is also supported.
  */
-export const SUPPORTED_SCOPES = ['builds', 'functions', 'runtime', 'post_processing']
+export const ALL_ENVELOPE_SCOPES = ['builds', 'functions', 'runtime', 'post_processing'] as const
 
-type EnvelopeEnvVarContext = 'all' | 'production' | 'deploy-preview' | 'branch-deploy' | 'dev'
-
-type EnvelopeEnvVarScope = 'builds' | 'functions' | 'runtime' | 'post_processing'
-
+// TODO(serhalp) Netlify API is incorrect - the returned scope is `post_processing`, not `post-processing`
+type EnvelopeEnvVarScope =
+  | Exclude<NonNullable<Awaited<ReturnType<NetlifyAPI['getEnvVars']>>[number]['scopes']>[number], 'post-processing'>
+  | 'post_processing'
+type EnvelopeEnvVar = Awaited<ReturnType<NetlifyAPI['getEnvVars']>>[number] & {
+  scopes: EnvelopeEnvVarScope[]
+}
+type EnvelopeEnvVarContext = NonNullable<EnvelopeEnvVar['values']>[number]['context']
 type EnvelopeEnvVarValue = {
   /**
    * The deploy context of the this env var value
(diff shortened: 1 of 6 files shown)


src/utils/env/index.ts
```diff
@@ -4,57 +4,89 @@ import { $TSFixMe } from '../../commands/types.js'
...
+}
+/**
+ * Supported values for the user-provided env `scope` option.
+ * These exactly match possible `scope` values returned by the Envelope API.
+ * Note that `any` is also supported.
+ */
+export const SUPPORTED_SCOPES = ['builds', 'functions', 'runtime', 'post_processing']
 
-type EnvironmentVariableContext = 'all' | 'production' | 'deploy-preview' | 'branch-deploy' | 'dev'
+type EnvelopeEnvVarContext = 'all' | 'production' | 'deploy-preview' | 'branch-deploy' | 'dev'
```
serhalp: since the `context` enum on Envelope env vars isn't exactly the same as the allowed values for the `context` user option, this seems worth naming specifically

src/utils/env/index.ts
```diff
@@ -1,65 +1,117 @@
...
-type EnvironmentVariableValue = {
-  context: EnvironmentVariableContext
+// TODO(serhalp) Netlify API is incorrect - the returned scope is `post_processing`, not `post-processing`
+type EnvelopeEnvVarScope =
+  | Exclude<NonNullable<Awaited<ReturnType<NetlifyAPI['getEnvVars']>>[number]['scopes']>[number], 'post-processing'>
+  | 'post_processing'
+type EnvelopeEnvVar = Awaited<ReturnType<NetlifyAPI['getEnvVars']>>[number] & {
+  scopes: EnvelopeEnvVarScope[]
+}
+type EnvelopeEnvVarContext = NonNullable<EnvelopeEnvVar['values']>[number]['context']
```
ndhoule, line 33: Do you mean for `undefined` to be part of this union? The intent here is _really_ unclear to me because of all the type magic.

src/utils/env/index.ts
```diff
@@ -1,65 +1,117 @@
...
+type EnvelopeEnvVarValue = {
+  /**
+   * The deploy context of the this env var value
+   */
+  context: EnvelopeEnvVarContext
```
ndhoule, line 38: Possibly related to my point above: In the client's OpenAPI types, this is an optional`?` but here is a required-but-could-be-`undefined` prop. (I'm also a little confused by the inconsistency here--are you inlining the `EnvelopeEnvVarValue` type out of convenience, because it's not possible to pull it out of a return value, or because it needs patching?)

src/utils/env/index.ts
```diff
@@ -272,10 +327,10 @@ export const getHumanReadableScopes = (scopes?: (EnvironmentVariableScope | 'pos
 export const translateFromMongoToEnvelope = (env: Record<string, string> = {}) => {
```
ndhoule, line 327:
```suggestion
export const translateFromMongoToEnvelope = (env: Record<string, string> = {}): EnvelopeItems[] => {
```

This reduces the need to explicitly assert strings as `const`s.

tests/unit/utils/env/index.test.ts
```diff
@@ -65,24 +65,24 @@ test("should filter, sort, and format Envelope's response correctly", () => {
   const envelopeItems = [
```
ndhoule, line 65: If you export the `EnvelopeItem` type, you can:

```suggestion
  const envelopeItems: EnvelopeItem[] = [
```

Which eliminates the need to assert strings as `const`s.

ndhoule (approved):
Looks great, thanks for making these improvements! Some minor comments on typing that I found confusing, take them or leave them.

src/utils/env/index.ts
```diff
@@ -1,65 +1,117 @@
...
+  scopes: EnvelopeEnvVarScope[]
+  values: EnvelopeEnvVarValue[]
 }
 
+// AFAICT, Envelope uses only `post_processing` on returned env vars; the CLI documents and expects
+// only `post-processing` as a valid user-provided scope; the code handles both everywhere. Consider
+// explicitly normalizing and dropping undocumented support for user-provided `post_processing`.
+type SupportedScope = EnvelopeEnvVarScope | 'post_processing' | 'any'
+
+type ContextOrBranch = string
```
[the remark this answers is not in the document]
serhalp: I was personally really happy with this change, so unless you feel strongly I'm going to keep it 😁

src/utils/env/index.ts
```diff
@@ -1,65 +1,117 @@
...
+ * Note that `any` is also supported.
+ */
+export const ALL_ENVELOPE_SCOPES = ['builds', 'functions', 'runtime', 'post_processing'] as const
 
-type EnvironmentVariableValue = {
-  context: EnvironmentVariableContext
+// TODO(serhalp) Netlify API is incorrect - the returned scope is `post_processing`, not `post-processing`
+type EnvelopeEnvVarScope =
+  | Exclude<NonNullable<Awaited<ReturnType<NetlifyAPI['getEnvVars']>>[number]['scopes']>[number], 'post-processing'>
+  | 'post_processing'
```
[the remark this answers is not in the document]
serhalp: I kind of agree, but as previously discussed I'm just doing the bare minimum to get `netlify` API client types working when necessary, and I'll let you eventually apply your learnings from the SDK to establish solid patterns around API client types. `:homer-bush:`

src/utils/env/index.ts
serhalp: Yeah, I had tried that but it opened up a can of worms that I decided to back out of 😞

src/utils/env/index.ts
serhalp: Good catch, thanks. No, that was not intentional.

src/utils/env/index.ts
serhalp: Good catch, thanks. I assume that's incorrect in the OpenAPI types, but I'll update it to match.

serhalp: > I'm also a little confused by the inconsistency here--are you inlining the EnvelopeEnvVarValue type out of convenience, because it's not possible to pull it out of a return value, or because it needs patching?

Good question. I guess I'm just redefining the top-level `EnvelopeEnvVarValue` so that I can add describe doc comments on its fields?

PR #7235, opened 2025-04-23
fix: match the `branch-deploy` context when using `--context branch:foo`

serhalp:
### Context (pun intended)

The `--context` parameter can be passed to `dev`, `serve`, `functions:serve`,`deploy`, `env:*`.

As (poorly) documented, its value can be a valid context (`'all', 'production', 'deploy-preview', 'branch-deploy', 'dev'`) or `branch:foo` where `foo` is the name of a branch.

In the former case, we match env vars with a matching `context`.

In the latter case, the intent is to match env vars with `context: "branch"` and a `context_parameter` matching the branch name (`foo`).

All these behaviours are working as intended.

Here's the UI to illustrate how this works from a user's perspective:

Screenshot 2025-04-23 at 18 15 25

And to better understand the code in this PR, here's what happens currently on main:

```
$ ntl env:list --context branch:feat/make-it-pop
1 environment variable for site [...redacted...] in the feat/make-it-pop branch
.---------------------------------------------------------.
|                  Environment variables                  |
|---------------------------------------------------------|
| Key |        Value         |           Scope            |
|-----|----------------------|----------------------------|
| FOO | feat/make-it-pop-val | Builds, Functions, Runtime |
```

### Bug

However, in addition to these, it's expected that using `--context branch:foo` implicitly inherits the `branch-deploy` context as well (although a matching `branch` context value should take precedence). This was not implemented.

You might think this is unnecessary: if you don't need the override, just use `--context branch-deploy`. However, consider multiple environment variables, some of which have branch-specific overrides, some of which do not. There is currently no mechanism available to make this work as expected.

For example, if you also have this `BAR` env var defined:

Screenshot 2025-04-24 at 09 37 14

and you run the command above, you still only get `FOO`, no `BAR`.

### Fix

This commit fixes that, by simply adding the missing logic. It just... didn't exist.

Note that it is our responsibility here to override `branch-deploy` context values for a given env var with any matching `branch` context values.

Here's the output from above again against this branch:

```
$ ./bin/run.js env:list --context branch:feat/make-it-pop
2 environment variables for site [...redacted...] in the feat/make-it-pop branch
.---------------------------------------------------------.
|                  Environment variables                  |
|---------------------------------------------------------|
| Key |        Value         |           Scope            |
|-----|----------------------|----------------------------|
| BAR | bd-val               | Builds, Functions, Runtime |
| FOO | feat/make-it-pop-val | Builds, Functions, Runtime |
```

This also overhauls the `--context` documentation to be more precise and actionable.

Note that this is stacked on https://github.com/netlify/cli/pull/7234.


src/utils/env/index.ts
```diff
@@ -102,15 +102,18 @@ export const getValueForContext = (
...
   contextOrBranch: ContextOrBranch,
 ): EnvelopeEnvVarValue | undefined => {
-  const valueForContext = values.find((val) => {
-    const isSupportedContext = (SUPPORTED_CONTEXTS as readonly string[]).includes(contextOrBranch)
-    if (!isSupportedContext) {
-      // the "context" option passed in is actually the name of a branch
-      return val.context === 'all' || val.context_parameter === contextOrBranch
+  const isSupportedContext = (SUPPORTED_CONTEXTS as readonly string[]).includes(contextOrBranch)
+  if (!isSupportedContext) {
+    const valueMatchingAsBranch = values.find((val) => val.context_parameter === contextOrBranch)
```
serhalp, line 107: It didn't seem worth optimizing perf here since `values` will contain at most a few elements

src/utils/env/index.ts
```diff
@@ -102,15 +102,18 @@ export const getValueForContext = (
...
+    // This is a `branch` context, which is an override, so it takes precedence
+    if (valueMatchingAsBranch != null) {
+      return valueMatchingAsBranch
     }
-    return val.context === 'all' || val.context === contextOrBranch
-  })
-  return valueForContext ?? undefined
+    const valueMatchingContext = values.find((val) => val.context === 'all' || val.context === 'branch-deploy')
```
serhalp, line 112: I'm fairly certain there's another bug here (which I haven't changed the behaviour of at all in this PR), where whether the `all` context value matches or the one that should be overriding does depends on the order of the elements returned from the API. I assume it's returned in order of precedence and so just happens to not exhibit the bug? In any case, I'll open a follow-up PR later.

tests/unit/utils/env/index.test.ts
```diff
@@ -20,27 +21,78 @@ test('should find a value from a given context', () => {
...
   expect(result).toHaveProperty('value', 'bar')
 })
 
-test('should find a value from a given branch', () => {
-  const values = [
+test('should return a value from the `branch` context with a matching `context_parameter` given a branch', () => {
+  const values = shuffle([
     {
-      context: 'branch-deploy' as const,
+      context: 'branch' as const,
```
serhalp, line 32: This test data was invalid, but happened to work because it technically always checks for a matching `context_parameter`

khendrikse (approved):

PR #7236, opened 2025-04-24
feat!: remove support for deprecated `netlify-lambda` package

serhalp:
#### Summary

https://github.com/netlify/netlify-lambda was deprecated in Feb 2023

This removes support for it from Netlify CLI.

To migrate away from netlify-lambda, see https://github.com/netlify/netlify-lambda/blob/main/MIGRATE.md.

(internal link) See [thread](https://netlify.slack.com/archives/C03ETTLQ9BP/p1744834944707249) where removal was approved


src/lib/functions/runtimes/js/index.ts
```diff
@@ -5,54 +5,36 @@ import { Worker } from 'worker_threads'
...
 
 export const name = 'js'
 
-export type JsBuildResult = ZisiBuildResult | NetlifyLambdaBuildResult
+type SimpleJsBuildResult = {
+  schedule?: string
+  srcFiles: string[]
+}
+
+export type JsBuildResult = ZisiBuildResult | SimpleJsBuildResult
```
serhalp, line 23: There were actually _three_ different ways to run a JS function. When I removed one, it revealed that this union of two types just happened to work. This adds the missing one and removes the deprecated one... Hope that makes esnse.

src/lib/functions/registry.ts
```diff
@@ -168,25 +168,11 @@ export class FunctionsRegistry {
   }
 
   /**
-   * Runs before `scan` and calls any `onDirectoryScan` hooks defined by the
-   * runtime before the directory is read. This gives runtime the opportunity
-   * to run additional logic when a directory is scanned.
+   * TODO(serhalp): Is this really the right place to do this mutation? Should
+   * we do this at all?
```
ndhoule, line 172: I can't suggest the diff but we should convert this from a jsdoc comment to an inline comment

ndhoule (approved):

PR #7242, opened 2025-04-28
fix: fix framework server loading spinner ("Waiting for framework port")

### LINKED ISSUE #7124: Never ending "Waiting for framework" in 19.0.3

serhalp:
### Summary

This is a regression in the loading spinner shown before "Waiting for framework port", introduced in 19.0.3.

See:
1. https://github.com/netlify/cli/pull/7026
2. https://github.com/netlify/cli/pull/7131
3. https://github.com/netlify/cli/issues/7124

Oops, I was really overthinking this. All this needs to do is, every time a chunk comes in from the underlying framework dev server, clear the spinner line, write the chunk, then resume the spinner.

This wasn't set up right at all, which was obfuscated by the erroneous use of `isSpinning`.

Fixes #7124.

Thank you @Florian-Mt for spotting the (embarrassing!) fix.

#### Before

Monosnap screencast 2025-05-01 08-05-27

#### After

Monosnap screencast 2025-05-01 08-03-20

### DIFF
--- a/src/utils/framework-server.ts
+++ b/src/utils/framework-server.ts
@@ -68,7 +68,7 @@ export const startFrameworkServer = async function ({
         throw new Error(`Timed out waiting for port '${settings.frameworkPort}' to be open`)
       }
     }
-    stopSpinner({ error: false, spinner })
+    spinner.success()
   } catch (error_) {
     stopSpinner({ error: true, spinner })
     log(NETLIFYDEVERR, `Netlify Dev could not start or connect to localhost:${settings.frameworkPort}.`)
--- a/src/utils/shell.ts
+++ b/src/utils/shell.ts
@@ -4,7 +4,7 @@ import { stripVTControlCharacters } from 'util'
 
 import execa from 'execa'
 
-import { stopSpinner, type Spinner } from '../lib/spinner.js'
+import { type Spinner } from '../lib/spinner.js'
 
 import { chalk, log, NETLIFYDEVERR, NETLIFYDEVWARN } from './command-helpers.js'
 import { processOnExit } from './dev.js'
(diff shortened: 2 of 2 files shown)


src/utils/framework-server.ts
```diff
@@ -68,7 +68,7 @@ export const startFrameworkServer = async function ({
         throw new Error(`Timed out waiting for port '${settings.frameworkPort}' to be open`)
       }
     }
-    stopSpinner({ error: false, spinner })
+    spinner.success()
```
serhalp, line 71: We want to draw the ✔️, erase it

src/utils/shell.ts
```diff
@@ -61,15 +61,9 @@ export const runCommand = (
...
-    stopSpinner({ error: false, spinner })
-  }
   const pipeDataWithSpinner = (writeStream: NodeJS.WriteStream, chunk: string | Uint8Array) => {
-    if (spinner?.isSpinning) {
-      spinner.clear()
-    }
+    // Clear the spinner, write the framework command line, then resume spinning
+    spinner?.clear()
     writeStream.write(chunk, () => {
       spinner?.spin()
```
serhalp, line 68: I believe this still isn't quite right — spin() renders a single frame, but here we want to keep looping even when we aren't receiving chunks from the framework server.

But this fix is worth getting in now and I'll follow up when I get the full fix working.

serhalp: (In that follow-up I'll also look into how we can add test coverage for this)

sarahetter (approved):

PR #7251, opened 2025-05-02
feat(telemetry): send command opts and args

serhalp:
#### Summary

It's impossible for us currently to identify the potential impact of option deprecation.

This won't address that retroactively for past CLI versions, but at least it will give us data going forward.

Attaching the `source` is important, otherwise this would include applied default values and we wouldn't be able to distinguish this from actual user-provided values.

Unfortunately there doesn't appear to be a commander API to get named positional arguments, so `args` is just an array. Seems ok.


tests/integration/utils/call-cli.ts
```diff
@@ -17,11 +17,16 @@ export const callCli = async function (
   parseJson = false,
 ): // eslint-disable-next-line @typescript-eslint/no-explicit-any
 Promise<any> {
-  const { stdout } = await execa.node(cliPath, args, {
+  const { stdout, stderr } = await execa.node(cliPath, args, {
     timeout: CLI_TIMEOUT,
     nodeOptions: [],
     ...execOptions,
   })
+  if (process.env.DEBUG_TESTS) {
```
serhalp, line 25: Urelated but this is so annoying. We have this `DEBUG_TESTS` mechanism that surfaces spawned process logs but it wasn't being leveraged in the `callCli()` integration test helper case for some reason.

serhalp:
@khendrikse that was unrelated! fixed by https://github.com/netlify/cli/pull/7256 😁

khendrikse (approved):

PR #7260, opened 2025-05-05
feat: add new `netlify clone` command, to clone an existing repo + connected Netlify site in one step

serhalp:
### Summary

When a user already has an existing Netlify site connected to a hosted git repo, pulling this down and preparing it for use with Netlify CLI currently requires numerous steps:
1. find and copy the git remote URL from somewhere (in some cases this is two steps: find git link in Netlify UI → copy git remote URL from there)
2. run `git clone <git url>`
3. run `cd <created dir>`
4. run `netlify link`
5. follow prompt to select "Use current git remote origin" (even though this is obviously what is wanted)

This is a very common flow, e.g. when using a "Deploy to Netlify" button, creating a site from a Netlify template, creating a site from an AI creator... [any workflow where the site is created outside Netlify CLI](https://docs.netlify.com/welcome/add-new-site/#import-from-an-existing-repository) requires several steps to get it ready for use with Netlify CLI—and to make matters worse, this isn't particularly well documented.

This introduces a new command which vastly simplifies these steps, bridging the gap between Netlify CLI and other parts of the platform:
1. find and copy the git remote URL
2. run `netlify clone <git url>` 🚀

This will clone the git repo, find the corresponding Netlify site, and link the created directory automatically. It will only prompt the user if necessary - for example if the git repo URL matches multiple sites.

### Screenshots

`netlify help clone` or `netlify clone --help`:
Screenshot 2025-05-05 at 17 43 07

`netlify clone clone git@github.com:serhalp/netlify-cache-inspector.git`:
ezgif com-animated-gif-maker


src/commands/link/link.ts
```diff
@@ -214,8 +223,6 @@ or run ${chalk.cyanBright('netlify sites:create')} to create a site.`)
       }
       break
     }
-    default:
-      return
```
serhalp, line 218: This is not a breaking change, doubly so:
- this was an exhaustive switch-case
- the condition right below this `(if !site)` results in the same behaviour as before

docs/commands/link.md
```diff
@@ -20,6 +20,7 @@ netlify link
 
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `git-remote-name` (*string*) - Name of Git remote to use. e.g. "origin"
+- `git-remote-url` (*string*) - URL of the repository (or Github `owner/repo`) to link to
```
serhalp, line 23: It was one extra line of code to make this an option for `netlify link` as well, so I went ahead and added it

src/commands/link/link.ts
```diff
@@ -1,18 +1,77 @@
...
 import { chalk, logAndThrowError, exit, log, APIError } from '../../utils/command-helpers.js'
 import getRepoData from '../../utils/get-repo-data.js'
 import { ensureNetlifyIgnore } from '../../utils/gitignore.js'
 import { track } from '../../utils/telemetry/index.js'
 import type { SiteInfo } from '../../utils/types.js'
 import BaseCommand from '../base-command.js'
+import type { LinkOptionValues } from './option_values.js'
 
-const linkPrompt = async (command: BaseCommand, options: OptionValues) => {
+const findSiteByRepoUrl = async (api: NetlifyAPI, repoUrl: string): Promise<SiteInfo> => {
```
serhalp, line 17: This is extracted from below (see first commit). The behaviour is unchanged, though I did polish the output (added a spinner and changed some copy)

tests/integration/commands/clone/clone.test.ts
```diff
@@ -0,0 +1,421 @@
...
+import { createMock } from '../../utils/mock-execa.js'
+import execa from 'execa'
+import { cliPath } from '../../utils/cli-path.js'
+
+describe('clone command', () => {
+  let execaMock: Awaited<ReturnType<typeof createMock>>[0]
+  let removeExecaMock: Awaited<ReturnType<typeof createMock>>[1]
+
+  beforeEach(async () => {
+    ;[execaMock, removeExecaMock] = await createMock(js`
```
serhalp, line 16: oof, this took me a while to figure out. Because we run CLI commands via a spawned child process, we can't actually mock anything. This seemingly hasn't come up much before, but here we're interacting with a third-party server (to clone from git remotes), so it's strongly desirable to mock that out. It turns out there was already this helper and pattern for this. Hope this is clear enough...

tests/integration/commands/link/link.test.ts
```diff
@@ -9,6 +9,53 @@ import { getCLIOptions, withMockApi } from '../../utils/mock-api.js'
 import { withSiteBuilder } from '../../utils/site-builder.js'
 
 describe('link command', () => {
+  test.todo('should link to matching site given `--site-id`')
+  test.todo('should print an error and exit when no site with given `--site-id` is found')
+
+  test.todo('should link to matching site given `--site-name`')
+  test.todo('should print an error and exit when no site with given `--site-name` is found')
```
serhalp, line 16: I noticed missing coverage for `link`. It felt pretty out of scope but I wanted to at least leave these breadcrumbs.

tests/integration/commands/clone/clone.test.ts
```diff
@@ -0,0 +1,530 @@
...
+        // For any other command, use the real execa
+        // Normalize paths for Windows
+        const realExeca = require('${require.resolve('execa').split(sep).join('/')}')
+        // execa's APi is... really weird
+        const result = Promise.resolve(realExeca(command, args))
+        result.unref = () => {}
+        return result
+      }
+    `)
+  })
```
serhalp, line 39: oof, this took me a while to figure out 😓. Because we run CLI commands via a spawned child process, we can't actually mock anything. This seemingly hasn't come up much before, but here we're interacting with a third-party server (to clone from git remotes), so it's strongly desirable to mock that out. It turns out there was already this helper and pattern for this, but then of course every possible complication turned up here. Hope this is clear enough...

src/utils/normalize-repo-url.ts
```diff
@@ -0,0 +1,24 @@
+import parseGitHubUrl from 'parse-github-url'
```
serhalp, line 1: To start with this only supports GitHub, but I believe I've implemented in such a way that only this util would need to be updated to add support for GitLab, etc.

PR #7280, opened 2025-05-08
feat: automatically detects IDE - windsurf or cursor - for AI context files.

smnh:
#### Summary

Instead of asking the user for which IDE they are using when running `netlify recipes ai-context`:
CleanShot 2025-05-08 at 22 48 50@2x

It now automatically detects from which IDE the command was run and adds the context files to the correctfolder:
CleanShot 2025-05-09 at 00 01 10@2x
CleanShot 2025-05-09 at 00 02 26@2x

If it didn't detect anything, it will continue to the original step where it presents a list to chose from.
User can also pass `--skip-detection` flag to skip the IDE detection logic:
CleanShot 2025-05-09 at 00 04 38@2x

---

      passes our tests.

### DIFF
--- a/src/commands/recipes/index.ts
+++ b/src/commands/recipes/index.ts
@@ -17,10 +17,6 @@ export const createRecipesCommand = (program: BaseCommand) => {
     .argument('[name]', 'name of the recipe')
     .description(`Create and modify files in a project using pre-defined recipes`)
     .option('-n, --name <name>', 'recipe name to use')
-    .option(
-      '--skip-detection',
-      'Skips automatic IDE detection. Use this with the ai-context recipe to specify paths manually.',
-    )
     .addExamples(['netlify recipes my-recipe', 'netlify recipes --name my-recipe'])
     .action(async (recipeName: string, options: OptionValues, command: BaseCommand) => {
       const { recipesCommand } = await import('./recipes.js')
--- a/src/commands/recipes/index.ts
+++ b/src/commands/recipes/index.ts
@@ -17,6 +17,10 @@ export const createRecipesCommand = (program: BaseCommand) => {
     .argument('[name]', 'name of the recipe')
     .description(`Create and modify files in a project using pre-defined recipes`)
     .option('-n, --name <name>', 'recipe name to use')
+    .option(
+      '--skip-detection',
+      'Skips automatic IDE detection. Use this with the ai-context recipe to specify paths manually.',
+    )
     .addExamples(['netlify recipes my-recipe', 'netlify recipes --name my-recipe'])
     .action(async (recipeName: string, options: OptionValues, command: BaseCommand) => {
       const { recipesCommand } = await import('./recipes.js')
--- a/src/commands/recipes/recipes.ts
+++ b/src/commands/recipes/recipes.ts
@@ -14,16 +14,15 @@ const SUGGESTION_TIMEOUT = 1e4
 export interface RunRecipeOptions {
   args: string[]
   command?: BaseCommand
-  options?: OptionValues
   config: unknown
   recipeName: string
   repositoryRoot: string
 }
 
-export const runRecipe = async ({ args, command, options, config, recipeName, repositoryRoot }: RunRecipeOptions) => {
+export const runRecipe = async ({ args, command, config, recipeName, repositoryRoot }: RunRecipeOptions) => {
   const recipe = await getRecipe(recipeName)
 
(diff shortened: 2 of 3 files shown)


src/commands/recipes/index.ts
```diff
@@ -17,6 +17,10 @@ export const createRecipesCommand = (program: BaseCommand) => {
     .argument('[name]', 'name of the recipe')
     .description(`Create and modify files in a project using pre-defined recipes`)
     .option('-n, --name <name>', 'recipe name to use')
+    .option(
```
eduardoboucas, line 20: This will add a flag to the entire `recipes` command, not just this recipe. Right now the command itself is fully decoupled from the individual recipes, by design, and is not aware of any specific recipe. The fact that you're adding a parameter that is specific to a recipe, accompanied by help text that says it is relevant only to a specific recipe, breaks that.

If we keep using this pattern, we'll end up with a myriad of flags, each associated with a given recipe. I think that will get messy.

Do we really need this flag? Can we not just do the selection automatically when we detect thr IDE? If we really want an opt-out mechanism, I would recommend an environment variable as a stopgap solution until we build support for flags into the recipes themselves.

src/recipes/ai-context/index.ts
```diff
@@ -56,11 +62,69 @@ const promptForPath = async (): Promise<string> => {
...
+  },
+  {
+    name: 'Cursor',
+    command: 'cursor',
+    rulesPath: IDE_RULES_PATH_MAP.cursor,
+  },
+]
+
+const getPathByDetectingIDE = async (): Promise<string | null> => {
+  const getIDEFromCommand = (command: string): IDE | null => {
```
eduardoboucas, line 84: nit: Any reason for this function to be declared inside another function?

src/recipes/ai-context/index.ts
```diff
@@ -56,11 +62,69 @@ const promptForPath = async (): Promise<string> => {
...
+    const match = IDE.find((ide) => command.includes(ide.command))
+    return match ?? null
+  }
+
+  async function getCommandAndParentPID(pid: number): Promise<{
```
eduardoboucas, line 89: nit: Any reason for this function to be declared inside another function? Also we have function declarations and function expressions in the same scope, might be good to standardise.

src/recipes/ai-context/index.ts
```diff
@@ -56,11 +62,69 @@ const promptForPath = async (): Promise<string> => {
...
+
+  // Go up the chain of ancestor process IDs and find if one of their commands matches an IDE.
+  const ppid = process.ppid
+  let result: Awaited<ReturnType<typeof getCommandAndParentPID>>
+  try {
+    result = await getCommandAndParentPID(ppid)
+    while (result.parentPID !== 1 && !result.ide) {
+      result = await getCommandAndParentPID(result.parentPID)
+    }
+  } catch (_) {
```
eduardoboucas, line 114: nit

```suggestion
  } catch {
```

src/recipes/ai-context/index.ts
smnh: nope, I can move it outside

smnh:
@eduardoboucas removed `--skip-detection` flag and fixed other things you mentioned.

src/recipes/ai-context/index.ts
```diff
@@ -56,11 +62,79 @@ const promptForPath = async (): Promise<string> => {
...
+  return result.ide ? result.ide.rulesPath : null
+}
+
+export const run = async ({ args, command, options }: RunRecipeOptions) => {
   // Start the download in the background while we wait for the prompts.
   const download = downloadFile(version).catch(() => null)
 
-  const filePath = args[0] || (await promptForPath())
+  const filePath =
+    args[0] || ((options?.skipDetection ? null : await getPathByDetectingIDE()) ?? (await promptForPath()))
```
eduardoboucas, line 137: We can now remove references to `options.skipDetection`, right?

src/commands/recipes/recipes.ts
```diff
@@ -14,15 +14,16 @@ const SUGGESTION_TIMEOUT = 1e4
 export interface RunRecipeOptions {
   args: string[]
   command?: BaseCommand
+  options?: OptionValues
```
eduardoboucas, line 17: I don't think we need this anymore?

sean-roberts (approved):

PR #7284, opened 2025-05-10
feat: support granular context file setting

sean-roberts:

#### Summary

Docs has been refactored to provide granular context data: https://github.com/netlify/docs/pull/4352 - this needs to be deployed before this CLI version goes out.

This update to the recipe for ai context introduces a few changes:
- dynamic lookup of the available consumers/clients, docs will tell us what clients we allow explicit path installation for.
- context files will be created in a granular fashion instead of a massive netlify-development.mdc file, it will put all of the context files assigned to the given tool into specific files. So if the consumer (e.g. windsurf) is assigned 5 different context files on the docs, we will add all five
- it moves the file naming, file truncation, and file download URLs to the configuration from docs
- we prefix all local files now with netlify-... given this, we've removed the explicit confirmation about overriding existing, non-netlify files. 

All that to say, going forward, if we want to add a new consumer tool (like Sourcegraph's Amp) we just update the docs with the new consumer, assign the context files it will need, the path to put it in, etc. and this recipe will automatically pick it up without needing an update. The goal is that as we find more nuances to the consumer tools, their needs for files, etc. we can update those on the context hosting (docs) and we can move more quickly.

---

      passes our tests.


src/recipes/ai-context/context.ts
```diff
@@ -2,21 +2,90 @@ import { promises as fs } from 'node:fs'
 import { dirname } from 'node:path'
 
 const ATTRIBUTES_REGEX = /(\S*)="([^\s"]*)"/gim
-const BASE_URL = 'https://docs.netlify.com/ai-context'
-export const FILE_NAME = 'netlify-development.mdc'
+const BASE_URL = process.env.AI_CONTEXT_BASE_URL ?? 'https://docs.netlify.com/ai-context'
```
eduardoboucas, line 5: How do we envision this environment being used? Should we do some validation to make sure we fail nicely if it isn't a valid URL?

src/recipes/ai-context/context.ts
```diff
@@ -2,21 +2,90 @@ import { promises as fs } from 'node:fs'
...
+  if (!res.ok) {
+    return []
+  }
+
   try {
-    const res = await fetch(`${BASE_URL}/${FILE_NAME}`, {
+    const data = (await res.json()) as { consumers: ConsumerConfig[] } | undefined
+    contextConsumers = data?.consumers ?? []
+  } catch {
+    return []
```
eduardoboucas, line 57: nit: No need for an another exit point.

```suggestion
    // no-op
```

src/recipes/ai-context/context.ts
```diff
@@ -2,21 +2,90 @@ import { promises as fs } from 'node:fs'
...
+    url.searchParams.set('consumer', consumer.key)
+
+    if (process.env.AI_CONTEXT_BASE_URL) {
+      const overridingUrl = new URL(process.env.AI_CONTEXT_BASE_URL)
+      url.host = overridingUrl.host
+      url.port = overridingUrl.port
+      url.protocol = overridingUrl.protocol
+    }
+
+    const res = await fetch(url.toString(), {
```
eduardoboucas, line 79: nit: No need to convert to string.

```suggestion
    const res = await fetch(url, {
```

src/recipes/ai-context/context.ts
```diff
@@ -137,3 +208,12 @@ export const writeFile = async (path: string, contents: string) => {
   await fs.mkdir(directory, { recursive: true })
   await fs.writeFile(path, contents)
 }
+
+export const deleteFile = async (path: string) => {
+  try {
+    // delete file from file system - not just unlinking it
+    await fs.rm(path)
+  } catch (_) {
```
eduardoboucas, line 216: nit: No use for variable.

```suggestion
  } catch {
```

src/recipes/ai-context/index.ts
```diff
@@ -12,30 +13,51 @@ import {
...
 
-const presets = [
-  { name: 'Windsurf rules (.windsurf/rules/)', value: '.windsurf/rules' },
-  { name: 'Cursor rules (.cursor/rules/)', value: '.cursor/rules' },
-  { name: 'Custom location', value: '' },
-]
+// context consumers endpoints returns all supported IDEs and other consumers
+// that can be used to pull context files. It also includes a catchall consumer
+// for outlining all context that an unspecified consumer would handle.
+const allContextConsumers = (await getContextConsumers(version)).filter((consumer) => !consumer.hideFromCLI)
```
eduardoboucas, line 28: I think `getContextConsumers` can throw, since we're not wrapping the `fetch` call in a `try/catch`, so this could lead to an uncaught rejection?

src/recipes/ai-context/index.ts
```diff
@@ -48,78 +70,171 @@ const promptForPath = async (): Promise<string> => {
...
   }
 
   log('You must select a path.')
 
-  return promptForPath()
+  return promptForContextConsumerSelection()
+}
+
+/**
+ * Checks if a command belongs to a known IDEs by checking if it includes a specific string.
```
eduardoboucas, line 82:
```suggestion
 * Checks if a command belongs to a known IDE by checking if it includes a specific string.
```

src/recipes/ai-context/index.ts
```diff
@@ -48,78 +70,171 @@ const promptForPath = async (): Promise<string> => {
...
+ * Receives a process ID (pid) and returns both the command that the process was run with and its parent process ID. If the process is a known IDE, also returns information about that IDE.
+ */
+const getCommandAndParentPID = async (
+  pid: number,
+): Promise<{
+  parentPID: number
+  command: string
+  consumerKey: string | null
+}> => {
+  const { stdout } = await execa('ps', ['-p', String(pid), '-o', 'ppid=,comm='])
```
eduardoboucas, line 103: What happens on Windows?

src/recipes/ai-context/index.ts
```diff
@@ -48,78 +70,171 @@ const promptForPath = async (): Promise<string> => {
...
+  const output = stdout.trim()
+  const spaceIndex = output.indexOf(' ')
+  const parentPID = output.substring(0, spaceIndex)
+  const command = output.substring(spaceIndex + 1).toLowerCase()
+  return {
+    parentPID: parseInt(parentPID, 10),
```
eduardoboucas, line 109: nit: This should be faster and a bit nicer.

```suggestion
    parentPID: Number(parentPID),
```

src/recipes/ai-context/index.ts
```diff
@@ -48,78 +70,171 @@ const promptForPath = async (): Promise<string> => {
...
+    command: command,
```
eduardoboucas, line 110: nit

```suggestion
    command,
```

src/recipes/ai-context/index.ts
```diff
@@ -48,78 +70,171 @@ const promptForPath = async (): Promise<string> => {
...
   }
 
-  if (minimumCLIVersion && semver.lt(version, minimumCLIVersion)) {
-    return logAndThrowError(
-      `This command requires version ${minimumCLIVersion} or above of the Netlify CLI. Refer to ${chalk.underline(
-        'https://ntl.fyi/update-cli',
-      )} for information on how to update.`,
-    )
+  if (!consumer?.contextScopes) {
+    log('No context files found for this consumer. Try again or let us know if this happens again.')
```
eduardoboucas, line 157: Should we add some instructions for _how_ they should let us know if this happens again?

eduardoboucas (commented):
Do we want to add some tests here?

sean-roberts:
> Do we want to add some tests here?

yes, working on those right now

src/recipes/ai-context/index.ts
smnh: Is this async call to the `getContextConsumers` made in the main scope to try and fetch the data as soon as possible, so by the time user gets to the promp, we already have the needed data?

Even so, I suggest moving it to the beginning of the `run` function. I don't think there is a huge gain between calling it from the main scope and calling it from within the `run` function which should be the first thing that executes when user runs the command. There is something about executing `fetch` in the main scope feels wrong to me but I can't explain what it is exactly :)

src/recipes/ai-context/index.ts
sean-roberts: I don't know, @smnh will need to verify. Currently, this work is all wrapped in try/catch and will skip if any errors are thrown and, if it simply doesn't work, then it will also continue as it does without this logic.

smnh: @eduardoboucas on Windows and on systems that do not support `ps` it will fallback to IDE selection.
We can add later detection for OS and run appropriate command that lists processes with their IDs and commands.

eduardoboucas (approved):

PR #7293, opened 2025-05-14
feat: publish `netlify` package

eduardoboucas:
#### Summary

Adds a script that publishes the same code as a second package (`netlify`). We have different options to make this happen, like making `netlify-cli` a dependency of `netlify`. But to minimise the changes to `netlify-cli` (which doesn't even have a `main` field at the moment), I went for a script that is meant to be run in the CI and patches `package.json` with the new name and `main` before publishing it.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -54,7 +54,18 @@ jobs:
       - name: Run E2E tests
         run: npm run test:e2e
 
-      - name: Publish package
+      - name: Publish netlify-cli package
         run: npm publish --provenance
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+
+      - name: Prepare netlify package
+        run: node scripts/netlifyPackage.js prepare
+
+      - name: Publish netlify package
+        run: npm publish --provenance
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+
+      - name: Verify netlify package
+        run: node scripts/netlifyPackage.js verify
--- a/scripts/netlifyPackage.js
+++ b/scripts/netlifyPackage.js
@@ -1,0 +1,54 @@
+// @ts-check
+import assert from 'node:assert'
+import { basename, dirname, resolve } from 'node:path'
+import { argv } from 'node:process'
+import { fileURLToPath } from 'node:url'
+import { readFile, writeFile } from 'node:fs/promises'
+
+import execa from 'execa'
+
+const packageJSON = await getPackageJSON()
+
+async function getPackageJSON() {
+  const packageJSONPath = resolve(fileURLToPath(import.meta.url), '../../package.json')
+  const contents = JSON.parse(await readFile(packageJSONPath, 'utf8'))
+
+  return {
+    contents,
+    path: packageJSONPath,
+  }
+}
+
+/**
+ * @type {Record<string, function>}
+ */
+const commands = {
+  prepare: async () => {
+    const newPackageJSON = {
+      ...packageJSON.contents,
+      main: './dist/index.js',
+      name: 'netlify',
+    }
+
+    console.log(`Writing updated package.json to ${packageJSON.path}...`)
+    await writeFile(packageJSON.path, `${JSON.stringify(newPackageJSON, null, 2)}\n`)
+
+    console.log('Re-installing dependencies to update lockfile...')
+    await execa('npm', ['install'], {
+      cwd: dirname(packageJSON.path),
+    })
+  },
+
+  verify: async () => {
+    const { stdout } = await execa('npx', ['-y', 'netlify', '--version'])
+    const version = stdout.match(/netlify-cli\/(\d+\.\d+\.\d+)/)
+
+    assert.equal(version, packageJSON.contents.version, 'Version installed via npx matches latest version')
+  },
+}
+
+if (typeof commands[argv[2]] === 'function') {
+  commands[argv[2]]()
+} else {
+  console.error(`Usage: node ${basename(argv[1])} <command> (available commands: ${Object.keys(commands).join(', ')})`)
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,4 @@
+// This is an entrypoint that mirrors the interface that the `netlify` package
+// used to have, before it was renamed to `@netlify/api`. We keep it for
+// backwards-compatibility.
+export { NetlifyAPI, methods } from '@netlify/api'


scripts/netlifyPackage.js
```diff
@@ -0,0 +1,54 @@
...
+  verify: async () => {
+    const { stdout } = await execa('npx', ['-y', 'netlify', '--version'])
+    const version = stdout.match(/netlify-cli\/(\d+\.\d+\.\d+)/)
+
+    assert.equal(version, packageJSON.contents.version, 'Version installed via npx matches latest version')
+  },
+}
+
+if (typeof commands[argv[2]] === 'function') {
+  commands[argv[2]]()
```
JakeChampion, line 51: question: do we need to use `await` here?

eduardoboucas: Done in https://github.com/netlify/cli/pull/7293/commits/a56363a27a3f0fbcb4e10204d0edd23423cd99a5.

scripts/netlifyPackage.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @type {Record<string, function>}
+ */
+const commands = {
+  prepare: async () => {
+    const newPackageJSON = {
+      ...packageJSON.contents,
+      main: './dist/index.js',
+      name: 'netlify',
```
eduardoboucas, line 30: Now I'm wondering if we should also change the name of the binary so that `npx netlify` and any globally-installed `netlify-cli` do not collide?

scripts/netlifyPackage.js
serhalp, line 1: nit: I was going to comment that this codebase does not follow this naming convention, but I see that `scripts/prepublishOnly` is the one exception, which you probably saw right next to this one!

🤷🏼

scripts/netlifyPackage.js
```diff
@@ -0,0 +1,54 @@
...
+import { fileURLToPath } from 'node:url'
+import { readFile, writeFile } from 'node:fs/promises'
+
+import execa from 'execa'
+
+const packageJSON = await getPackageJSON()
+
+async function getPackageJSON() {
+  const packageJSONPath = resolve(fileURLToPath(import.meta.url), '../../package.json')
+  const contents = JSON.parse(await readFile(packageJSONPath, 'utf8'))
```
serhalp, line 14: nit: We already have a dep on `normalize-package-data` you can use here to avoid `contents` being `any`

There's even a util `src/utils/get-cli-package-json.ts` you should be able to use

src/index.ts
```diff
@@ -0,0 +1,4 @@
+// This is an entrypoint that mirrors the interface that the `netlify` package
```
serhalp, line 1: thought: Should we log a warning here that the package has been renamed?

.github/workflows/release-please.yml
```diff
@@ -54,7 +54,18 @@ jobs:
...
+      - name: Prepare netlify package
+        run: node scripts/netlifyPackage.js prepare
+
+      - name: Publish netlify package
+        run: npm publish --provenance
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
+
+      - name: Verify netlify package
+        run: node scripts/netlifyPackage.js verify
```
serhalp, line 71: suggestion: It seems to me that this is an e2e test, but it isn't quite framed that way, it can't be run locally against local changes nor in CI, and if it catches an issue someday it would be pretty late in the process (after it's public). Since we already have a nice e2e setup in this repo that does _precisely_ what we want here (using verdaccio), I'd suggest adding a test there and removing this!

scripts/netlifyPackage.js
```diff
@@ -0,0 +1,54 @@
...
+      cwd: dirname(packageJSON.path),
+    })
+  },
+
+  verify: async () => {
+    const { stdout } = await execa('npx', ['-y', 'netlify', '--version'])
+    const version = stdout.match(/netlify-cli\/(\d+\.\d+\.\d+)/)
+
+    assert.equal(version, packageJSON.contents.version, 'Version installed via npx matches latest version')
+  },
```
serhalp, line 47: nit: this is moot given my e2e comment, but I believe this would be flakey, since publication to an npm registry under a tag (`latest`) is eventually consistent

scripts/netlifyPackage.js
```diff
@@ -0,0 +1,54 @@
...
+      name: 'netlify',
+    }
+
+    console.log(`Writing updated package.json to ${packageJSON.path}...`)
+    await writeFile(packageJSON.path, `${JSON.stringify(newPackageJSON, null, 2)}\n`)
+
+    console.log('Re-installing dependencies to update lockfile...')
+    await execa('npm', ['install'], {
+      cwd: dirname(packageJSON.path),
+    })
```
serhalp, line 39: I'm not following what this step is for. Could we document it in a comment? Is this just so it updates the `name` which is also in the lockfile?

serhalp, line 39: Hrm, his is getting difficult to reason about... So, before this script runs in CI, we'll have run this as part of the `netlify-ci` publish step: https://github.com/netlify/cli/blob/47242ce76c7aa57978e14e9e075ca6f0f76af348/scripts/prepublishOnly.js.

This means we'll have mutated `package.json` to prune dev deps, converted the `package-lock.json` to an `npm-shrinkwrap.json`, then restored the original `package.json`... now here we'll mutate `package.json` again to update `name` and add `main`, then run `npm i` again, which will restore the dev deps into the `npm-shrinkwrap.json`, making `netlify` twice as big to install as `netlify-cli`. 😰 Did I get all that right?

@ndhoule and I just recently decomplicated all of this (with many false starts due to how tangled and brittle it was), so I think it would be worth avoiding getting back into this state if we can help it. I don't have concrete suggestions that fit in a comment box, but I'm happy to pair on it if you'd like.

scripts/netlifyPackage.js
serhalp: As far as I understand the wonky npm heuristics here:
- if a user has both `netlify` and `netlify-cli` installed (both locally or both globally), `npx netlify` will run `netlify`
- if a user has `netlify-cli` and does not have `netlify`, `npx netlify` will always run `netlify-cli` rather than try to install `netlify`
- if a user has neither installed, `npx netlify` will install and run `netlify`

(It's likely worth verifying these)

Does this seem ok or are any of these scenarios problematic?

scripts/netlifyPackage.js
eduardoboucas: I have removed this command.

.github/workflows/release-please.yml
eduardoboucas: Agreed. Moved to the E2E setup in https://github.com/netlify/cli/pull/7293/commits/0bbc22dd1b3d398d66303a65003ed23970ce5c34.

src/index.ts
eduardoboucas: That feels a bit extreme? We might be logging something that is not actionable to the user, since this might be a transitive dependency.

scripts/netlifyPackage.js
eduardoboucas: This use case is a bit different. We're reading `package.json` and then writing it back, so I don't think we should be normalising it.

I have used the `Package` type now.

scripts/netlifyPackage.js
eduardoboucas: My question is whether `npx netlify` will also install the binary into the PATH, in the same way that globally-installing a package does. If that's the case, then the `netlify` binary will overwrite the binary with the same name that had been injected by a globally-installed `netlify-cli`, which is not intended.

But if `npx netlify` simply runs the binary without adding it to the PATH (which I feel is the expected behaviour), this should be fine.

.github/workflows/release-please.yml
eduardoboucas: Ugh. It's failing on Windows somehow. I'll come back to this when I have some time.

scripts/netlifyPackage.js
ndhoule: I did name `scripts/prepublishOnly` as it is (despite not matching other file naming conventions) specifically so it would match up exactly with the npm script name.

JakeChampion (approved):

PR #7300, opened 2025-05-15
fix: avoid double encoding when notifying dev server of activity

VitaliyR:
…oxying

#### Summary

Fixes double encoding and marking preview server activity.

---

      passes our tests.

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -365,7 +365,7 @@ const serveRedirect = async function ({
     (await isEndpointExists(decodeURIComponent(reqUrl.pathname), options.target))
   if (staticFile || endpointExists) {
     const pathname = staticFile || reqUrl.pathname
-    req.url = encodeURI(pathname) + reqUrl.search
+    req.url = encodeURI(decodeURI(pathname)) + reqUrl.search
     // if there is an existing static file and it is not a forced redirect, return the file
     if (!match.force) {
       return proxy.web(req, res, { ...options, staticFile })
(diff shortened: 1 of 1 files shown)


src/utils/proxy.ts
```diff
@@ -365,7 +365,7 @@ const serveRedirect = async function ({
     (await isEndpointExists(decodeURIComponent(reqUrl.pathname), options.target))
   if (staticFile || endpointExists) {
     const pathname = staticFile || reqUrl.pathname
-    req.url = encodeURI(pathname) + reqUrl.search
+    req.url = encodeURI(decodeURI(pathname)) + reqUrl.search
```
serhalp, line 368: why `decodeURIComponent` on line 365 and `decodeURI` here? can we use the same one and share the result?

src/utils/proxy.ts
```diff
@@ -858,7 +858,17 @@ const onRequest = async (
     },
   }
 
+  const checkNotifyActivity = () => {
```
serhalp, line 861: nit:
```suggestion
  const maybeNotifyActivity = () => {
```

src/utils/proxy.ts
VitaliyR: Let's iterate on this, I need this released faster, and this micro optimization not worth the additional testing effort

serhalp (approved):
SGTM if you need this hotfixed and you've tested it manually, but afterward could you please add more context to the PR description and consider adding test coverage? 🙏🏼

ndhoule:
Request: Once we merge this, can you come back through and add comments that explain the problem and why we're doing what we're doing in this code? And a test, if possible?

The problem is non-obvious from this PR and from the code (I still don't know what the problem this solves is).

ndhoule:
Oh. Well...what Phil said. :laughing:

VitaliyR:
I've added it in the task btw, but I'm planning to put some refactor there, so I will add the comments into the code as well

PR #7319, opened 2025-06-02
feat: get user to initialize site first if they have neon installed for dev

sarahetter:

#### Summary

Modifies dev so a user must initialize a site first (so we can install the extension, create a db, and add env vars).

---

      passes our tests.

### DIFF
--- a/src/lib/extensions.ts
+++ b/src/lib/extensions.ts
@@ -9,10 +9,11 @@ export const packagesThatNeedSites = ['@netlify/neon']
 export async function handleExtensionRequirements(options: OptionValues, command: BaseCommand) {
     const { project } = command
     const { site, siteInfo } = command.netlify
-    // Go through all the packages that need sites, see if they exist in the package.json
-    // and if they do then we have to do a site init or whatever
+
     const packageJson = await project.getPackageJSON()
     const dependencies = packageJson.dependencies ?? {}
+
+    // If we don't have a site, these extensions need one initialized
     const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
     if (!hasSiteData) {
         for (const packageName of packagesThatNeedSites) {
--- a/src/lib/extensions.ts
+++ b/src/lib/extensions.ts
@@ -1,27 +1,27 @@
-import { OptionValues } from "commander";
+import { OptionValues } from 'commander'
 import isEmpty from 'lodash/isEmpty.js'
-import BaseCommand from "../commands/base-command.js";
-import { init } from "../commands/init/init.js";
+import BaseCommand from '../commands/base-command.js'
+import { init } from '../commands/init/init.js'
 import { log } from '../utils/command-helpers.js'
 
 export const packagesThatNeedSites = ['@netlify/neon']
 
 export async function handleExtensionRequirements(options: OptionValues, command: BaseCommand) {
-    const { project } = command
-    const { site, siteInfo } = command.netlify
+  const { project } = command
+  const { site, siteInfo } = command.netlify
 
-    const packageJson = await project.getPackageJSON()
-    const dependencies = packageJson.dependencies ?? {}
+  const packageJson = await project.getPackageJSON()
+  const dependencies = packageJson.dependencies ?? {}
 
-    // If we don't have a site, these extensions need one initialized
-    const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
-    if (!hasSiteData) {
-        for (const packageName of packagesThatNeedSites) {
-            if (dependencies[packageName]) {
-              log(`Found ${packageName} in package.json, initializing a site`)
-              await init(options, command)
-              return
-            }
-          }
+  // If we don't have a site, these extensions need one initialized
+  const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
+  if (!hasSiteData) {
+    for (const packageName of packagesThatNeedSites) {
+      if (dependencies[packageName]) {
+        log(`Found ${packageName} in package.json, initializing a site`)
+        await init(options, command)
+        return
+      }
     }
+  }
 }
--- a/src/lib/extensions.ts
+++ b/src/lib/extensions.ts
@@ -1,17 +1,26 @@
-import { Project } from "@netlify/build-info";
+import { OptionValues } from "commander";
+import isEmpty from 'lodash/isEmpty.js'
+import BaseCommand from "../commands/base-command.js";
+import { init } from "../commands/init/init.js";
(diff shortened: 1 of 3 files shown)


src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
...
+  const packageJson = await project.getPackageJSON()
+  const dependencies = packageJson.dependencies ?? {}
+
+  // If we don't have a site, these extensions need one initialized
+  const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
+  if (!hasSiteData) {
+    for (const packageName of packagesThatNeedSites) {
+      if (dependencies[packageName]) {
+        log(`Found ${packageName} in package.json, initializing a site`)
+        await init(options, command)
```
serhalp: btw, passing the dev command's `options` as the `init` command's `options` only happens to work because we mostly use this very weak `BaseCommand` type instead of command-specific types.

I don't have a great suggestion here, honestly. It would be nice if we had a util to extract just the "global" shared options from an options bag, so you could pass those down but not the `dev`-specific ones.

🤷🏼 I guess just food for thought for later.

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { site, siteInfo } = command.netlify
+
+  const packageJson = await project.getPackageJSON()
+  const dependencies = packageJson.dependencies ?? {}
+
+  // If we don't have a site, these extensions need one initialized
+  const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
+  if (!hasSiteData) {
+    for (const packageName of packagesThatNeedSites) {
+      if (dependencies[packageName]) {
```
serhalp: what if it's a dev dependency? 🤔

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function handleExtensionRequirements(options: OptionValues, command: BaseCommand) {
+  const { project } = command
+  const { site, siteInfo } = command.netlify
+
+  const packageJson = await project.getPackageJSON()
+  const dependencies = packageJson.dependencies ?? {}
+
+  // If we don't have a site, these extensions need one initialized
+  const hasSiteData = Boolean(site.id ?? options.site) && !isEmpty(siteInfo)
```
serhalp: I don't have full context on this, but is this condition the intended behaviour? it seems like this is capturing only cases where we have a site id but no site data for that id.

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
...
+import { log } from '../utils/command-helpers.js'
+
+export const packagesThatNeedSites = ['@netlify/neon']
+
+export async function handleExtensionRequirements(options: OptionValues, command: BaseCommand) {
+  const { project } = command
+  const { site, siteInfo } = command.netlify
+
+  const packageJson = await project.getPackageJSON()
+  const dependencies = packageJson.dependencies ?? {}
```
serhalp: perf nit: move this below into `if (!hasSiteData)` to avoid this I/O in 99.9% of cases

src/lib/extensions.ts
sarahetter: We don't care! :D Only regular deps

src/lib/extensions.ts
sarahetter: Moved back to the logical or that exists in deploy, but this is what we do in deploy :)

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
+import { OptionValues } from 'commander'
+import isEmpty from 'lodash/isEmpty.js'
+import BaseCommand from '../commands/base-command.js'
```
serhalp: huh, I could've sworn there was an eslint rule that automatically adds an empty line between groups here, weird

src/lib/extensions.ts
serhalp: Actually, on second thought I do have a suggestion for a temporary solution to at least de-risk this:
```suggestion
        await init(getBaseOptions(options), command)
```

with this in `src/commands/base-command.ts`:

```typescript
import pick from 'lodash/pick.js'

export const getBaseOptionValues = (options: OptionValues): BaseOptionValues =>
  pick(options, ['auth', 'cwd', 'debug', 'filter', 'httpProxy', 'silent'])
```

This way there's no way there can be weird bugs due to conflicting options that happen to have the same name, plus we establish a decent pattern going forward.

src/lib/extensions.ts
serhalp: Yeah, my brain totally didn't brain on this one, ignore me!

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  // If we don't have a site, these extensions need one initialized
+  // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+  const hasSiteData = Boolean(site.id || options.site) && !isEmpty(siteInfo)
+  if (!hasSiteData) {
+    const packageJson = await project.getPackageJSON()
+    const dependencies = packageJson.dependencies ?? {}
+    for (const packageName of packagesThatNeedSites) {
+      if (dependencies[packageName]) {
+        log(`Found ${packageName} in package.json, initializing a site`)
```
ndhoule: I don't think this log line adequately explains what's going on to an end user. How about something like:

```suggestion
        log(`Dependency ${packageName} requires a linked project, but you don't have one linked yet. Let's do that first.`)
```

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
+import { OptionValues } from 'commander'
+import isEmpty from 'lodash/isEmpty.js'
+import BaseCommand from '../commands/base-command.js'
+import { init } from '../commands/init/init.js'
+import { log } from '../utils/command-helpers.js'
+
+export const packagesThatNeedSites = ['@netlify/neon']
+
+export async function handleExtensionRequirements(options: OptionValues, command: BaseCommand) {
```
ndhoule: I would instead refactor this function to `doesProjectRequireLinkedSite` that returns a boolean. It'd accept a `project`, a `site`, and a `siteInfo`. You'd then move the `await init(...)` call out to `dev.ts`.

src/lib/extensions.ts
```diff
@@ -0,0 +1,27 @@
+import { OptionValues } from 'commander'
+import isEmpty from 'lodash/isEmpty.js'
+import BaseCommand from '../commands/base-command.js'
+import { init } from '../commands/init/init.js'
```
ndhoule: FYI, this will be loaded eagerly, which gets rid of some lazy loading perf improvements

PR #7332, opened 2025-06-05
feat!: drop support for node 18.x and 19.x

khendrikse:

#### Summary

This removes support for node 18.x and 19.x

---

      passes our tests.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [windows-latest]
         # Pinning 20.x version as a temporary workaround due to this https://github.com/nodejs/node/issues/52884
-        node-version: ['18.14.0', '20.12.2', '22']
+        node-version: ['20.0.0', '20.12.2', '22']
       fail-fast: false
     steps:
       # This improves Windows network performance. We need this since we open many ports in our tests.
@@ -79,7 +79,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        node-version: ['18.14.0', '22']
+        node-version: ['20.0.0', '22']
       fail-fast: false
     steps:
       - name: Git checkout
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [windows-latest]
         # Pinning 20.x version as a temporary workaround due to this https://github.com/nodejs/node/issues/52884
-        node-version: ['20.0.0', '20.12.2', '22']
+        node-version: ['20.12.2', '22']
       fail-fast: false
     steps:
       # This improves Windows network performance. We need this since we open many ports in our tests.
@@ -79,7 +79,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        node-version: ['20.0.0', '22']
+        node-version: ['20.12.2', '22']
       fail-fast: false
     steps:
       - name: Git checkout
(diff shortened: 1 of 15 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         os: [windows-latest]
         # Pinning 20.x version as a temporary workaround due to this https://github.com/nodejs/node/issues/52884
-        node-version: ['18.14.0', '20.12.2', '22']
+        node-version: ['20.0.0', '20.12.2', '22']
```
eduardoboucas: I think we had pinned 20.12.2 because there was an issue with 20.0.0?

khendrikse: ohhh oversight. on it!

src/utils/nodejs-compile-cache.ts
```diff
@@ -17,7 +17,7 @@ export let didEnableCompileCache = false
  * user's disk in a temp dir. If any of these changes or the temp dir is cleared, the next run is a cold run.
  *
  * The programmatic API to enable this (`enableCompileCache()`) was added in node 22.8.0, but we currently support
- * >=18.14.0, hence the conditional below. (For completeness, note that support via the env var was added in 22.1.0.)
+ * >=20.12.2, hence the conditional below. (For completeness, note that support via the env var was added in 22.1.0.)
```
eduardoboucas, line 20: I don't think this makes sense anymore?

khendrikse: wait, I think it still does?

eduardoboucas: Ignore me, sorry. I read this wrong.

khendrikse: If it was added in node 22.8.0 then 20.12.2 is not high enough right

tests/integration/commands/dev/dev.config.test.ts
```diff
@@ -459,7 +459,7 @@ describe.concurrent('commands/dev/config', () => {
     })
   })
 
-  test.runIf(gte(version, '18.14.0'))('should support functions with streaming responses', async (t) => {
+  test.runIf(gte(version, '20.12.2'))('should support functions with streaming responses', async (t) => {
```
eduardoboucas, line 462: I don't think we need this anymore because now the lowest version we test with has support for `Request`/`Response`, which is what this condition was about.

```suggestion
  test('should support functions with streaming responses', async (t) => {
```

tests/integration/commands/dev/v2-api.test.ts
```diff
@@ -25,7 +25,7 @@ const setup = async ({ fixture }: { fixture: { directory: string } }) => {
   await execa('npm', ['install'], { cwd: fixture.directory })
 }
 
-describe.runIf(gte(version, '18.13.0')).concurrent('v2 api', async () => {
+describe.runIf(gte(version, '20.12.2')).concurrent('v2 api', async () => {
```
eduardoboucas, line 28: I don't think we need this anymore because now the lowest version we test with has support for `Request`/`Response`, which is what this condition was about.

```suggestion
describe.concurrent('v2 api', async () => {
```

src/lib/functions/netlify-function.ts
```diff
@@ -25,7 +25,7 @@ export type InvokeFunctionResultWithSuccess = { error: null; result: InvokeFunct
 export type InvokeResult = InvokeFunctionResultWithError | InvokeFunctionResultWithSuccess
 
 const TYPESCRIPT_EXTENSIONS = new Set(['.cts', '.mts', '.ts'])
-const V2_MIN_NODE_VERSION = '18.14.0'
+const V2_MIN_NODE_VERSION = '20.12.2'
```
eduardoboucas, line 28: This is technically not needed, since the minimum version needed to have V2 functions (with `Request`/`Response`) is still the same. It just happens that this version is now lower than the minimum version supported by the CLI.

This means that this constant and the `isSupported()` method that uses it can go away. But we can do that in a follow-up.

PR #7352, opened 2025-06-16
feat: add ai command for cli to start the project

denar90:

#### Summary

---

      passes our tests.

### DIFF
--- a/src/commands/init/ai-rules.ts
+++ b/src/commands/init/ai-rules.ts
@@ -2,7 +2,17 @@ import { resolve } from 'node:path'
 import { promises as fs } from 'node:fs'
 import type { NetlifyAPI } from '@netlify/api'
 
-import { chalk, log, logAndThrowError, type APIError } from '../../utils/command-helpers.js'
+import {
+  chalk,
+  log,
+  logPadded,
+  logAndThrowError,
+  type APIError,
+  NETLIFY_CYAN,
+  NETLIFYDEVLOG,
+  NETLIFYDEVWARN,
+  NETLIFYDEVERR,
+} from '../../utils/command-helpers.js'
 import { normalizeRepoUrl } from '../../utils/normalize-repo-url.js'
 import { runGit } from '../../utils/run-git.js'
 import { startSpinner } from '../../lib/spinner.js'
@@ -19,6 +29,10 @@ import type BaseCommand from '../base-command.js'
 import type { SiteInfo } from '../../utils/types.js'
 import inquirer from 'inquirer'
 
+const SPARK_URL = process.env.SPARK_URL ?? 'https://spark.netlify.app'
+const AI_SITE_PROMPT_GEN_URL = `${SPARK_URL}/site-prompt-gen`
+const DOCS_URL = process.env.DOCS_URL ?? 'https://docs.netlify.com'
+
 /**
  * Project information interface for AI projects
  */
@@ -30,7 +44,7 @@ interface ProjectInfo {
 
 // Trigger IDE-specific MCP configuration
 const triggerMcpConfiguration = async (ide: ConsumerConfig, projectPath: string): Promise<boolean> => {
-  log(`\n${chalk.blue('🔧 MCP Configuration for')} ${chalk.cyan(ide.presentedName)}`)
+  log(`\n${chalk.blue('🔧 MCP Configuration for')} ${NETLIFY_CYAN(ide.presentedName)}`)
 
   const { shouldConfigure } = await inquirer.prompt<{ shouldConfigure: boolean }>([
     {
@@ -42,9 +56,9 @@ const triggerMcpConfiguration = async (ide: ConsumerConfig, projectPath: string)
   ])
 
   if (!shouldConfigure) {
-    log(`   ${chalk.gray('You can configure MCP manually later for enhanced AI capabilities:')}`)
+    log(`   ${chalk.dim('You can configure MCP manually later for enhanced AI capabilities:')}`)
     log(
-      `   ${chalk.gray('Documentation:')} ${chalk.cyan(
+      `   ${chalk.dim('Documentation:')} ${NETLIFY_CYAN(
         'https://docs.netlify.com/welcome/build-with-ai/netlify-mcp-server/',
       )}`,
     )
@@ -70,20 +84,11 @@ const triggerMcpConfiguration = async (ide: ConsumerConfig, projectPath: string)
         showGenericMcpConfig(config, ide.presentedName)
     }
 
-    log(`${chalk.green('✅')} MCP configuration completed for ${chalk.cyan(ide.presentedName)}`)
+    log(`${NETLIFYDEVLOG} MCP configuration completed for ${NETLIFY_CYAN(ide.presentedName)}`)
     return true
   } catch (error) {
-    log(`${chalk.red('❌')} Failed to configure MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
+    log(`${NETLIFYDEVERR} Failed to configure MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
     return false
-  }
-}
-
-// Helper functions reused from ai-start.ts
-const decodeHash = (hash: string): string => {
-  try {
-    return atob(hash)
-  } catch (error) {
-    throw new Error(`Failed to decode hash: ${error instanceof Error ? error.message : 'Invalid base64 or URL'}`)
   }
 }
 
@@ -125,10 +130,10 @@ const savePrompt = async (instructions: string, ntlContext: string | null, targe
   try {
     const filePath = resolve(targetDir, 'AI-instructions.md')
     await fs.writeFile(filePath, `Context: ${ntlContext ?? ''}\n\n${instructions}`, 'utf-8')
-    log(`${chalk.green('✅')} AI instructions saved to ${chalk.cyan('AI-instructions.md')}`)
+    log(`${NETLIFYDEVLOG} AI instructions saved to ${NETLIFY_CYAN('AI-instructions.md')}`)
   } catch (error) {
     const errorMessage = error instanceof Error ? error.message : 'Unknown error'
-    log(`${chalk.yellow('⚠️')} Warning: Failed to save AI instructions: ${errorMessage}`)
+    log(`${NETLIFYDEVWARN} Warning: Failed to save AI instructions: ${errorMessage}`)
   }
 }
 
@@ -149,45 +154,34 @@ export const initWithAiRules = async (hash: string, command: BaseCommand): Promi
   await command.authenticate()
   const { api } = command.netlify
 
-  log(`${chalk.blue('🤖 Initializing AI project')} with rules...`)
-  log(`${chalk.gray('User:')} ${api.accessToken ? 'Authenticated ✅' : 'Not authenticated ❌'}`)
+  log(`${NETLIFY_CYAN('🤖 Initializing AI project')} with rules...`)
+  log(`${NETLIFY_CYAN('User:')} ${api.accessToken ? 'Authenticated ✅' : 'Not authenticated ❌'}`)
 
   try {
     // Step 1: Decode hash and fetch project information
     log('\n📋 Extracting project details...')
-    const decodedUrl = decodeHash(hash)
-    log(`${chalk.cyan('Decoded URL:')} ${decodedUrl}`)
+    const decodedUrl = `${AI_SITE_PROMPT_GEN_URL}/${hash}`
+    log(`${NETLIFY_CYAN('Decoded URL:')} ${decodedUrl}`)
 
     log('\n🔍 Fetching project information...')
     const projectInfo = await fetchProjectInfo(decodedUrl)
 
-    log(`${chalk.cyan('Project ID:')} ${projectInfo.projectId}`)
-
     // Step 2: Get repository URL from project ID via Netlify site API
     log('\n🔗 Linking to Netlify project and fetching repository...')
     const repositoryUrl = await getRepoUrlFromProjectId(api, projectInfo.projectId)
-    log(`${chalk.cyan('Repository:')} ${repositoryUrl}`)
 
     // Step 3: Clone repository
     const { repoUrl, repoName } = normalizeRepoUrl(repositoryUrl)
     const targetDir = `ai-project-${repoName}-${hash.substring(0, 8)}`
 
-    const cloneSpinner = startSpinner({ text: `Cloning repository to ${chalk.cyan(targetDir)}` })
+    const cloneSpinner = startSpinner({ text: `Cloning repository to ${NETLIFY_CYAN(targetDir)}` })
 
     await cloneRepo(repoUrl, targetDir, false)
-    cloneSpinner.success({ text: `Cloned repository to ${chalk.cyan(targetDir)}` })
+    cloneSpinner.success({ text: `Cloned repository to ${NETLIFY_CYAN(targetDir)}` })
 
     // Step 4: Save AI instructions to file
     if (projectInfo.prompt) {
-      const ntlContext = await fetch(
-        'https://docs.netlify.com/ai-context/scoped-context?scopes=serverless,blobs,forms',
-        {
-          method: 'GET',
-          headers: {
-            'Content-Type': 'text/plain',
-          },
-        },
-      )
+      const ntlContext = await fetch(`${DOCS_URL}/ai-context/scoped-context?scopes=serverless,blobs,forms`)
         .then((res) => res.text())
         .catch(() => {
           return null
@@ -202,7 +196,7 @@ export const initWithAiRules = async (hash: string, command: BaseCommand): Promi
     let mcpConfigured = false
 
     if (detectedIDE) {
-      log(`${chalk.green('✅')} Detected development environment: ${chalk.cyan(detectedIDE.presentedName)}`)
+      log(`${NETLIFYDEVLOG} Detected development environment: ${NETLIFY_CYAN(detectedIDE.presentedName)}`)
       mcpConfigured = await triggerMcpConfiguration(detectedIDE, targetDir)
     }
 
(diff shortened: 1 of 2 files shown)


src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
+import { resolve } from 'node:path'
```
denar90, line 1: part of code here is duplication, but we still not setelled the command and want to see how it feels so then we can extract stuff into common utils after we work on making command fully public

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    const targetDir = `ai-project-${repoName}-${hash.substring(0, 8)}`
+
+    const cloneSpinner = startSpinner({ text: `Cloning repository to ${chalk.cyan(targetDir)}` })
+
+    await cloneRepo(repoUrl, targetDir, false)
+    cloneSpinner.success({ text: `Cloned repository to ${chalk.cyan(targetDir)}` })
+
+    // Step 4: Save AI instructions to file
+    if (projectInfo.prompt) {
+      const ntlContext = await fetch(
```
denar90: in future we don't wanna fetch context if we installing MCP

src/commands/init/ai-rules.ts
eduardoboucas: I'm not sure why trialing the new command should mean more code duplication? If you move the generic methods you're reusing to a separate file (which I would argue should already be the case), you're making it easier for this PR to be reviewed, you're going to have less code to maintain, and you reduce the number of places where you might need to adjust things if an issue or an improvement come up.

If you then decide that you want to keep the new command in its current shape, great — you already have the code in the structure you want and no further changes are needed; if you decide you want to bin it, you can just delete it and the code you abstracted away into a separate file still makes sense on its own.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    if (contextConsumer) {
+      return contextConsumer
+    }
+  }
+
+  return null
+}
+
+// Generate MCP configuration for the detected IDE
+const generateMcpConfig = (ide: ConsumerConfig): string => {
```
eduardoboucas: Why are you returning the config as a string here only to then re-parse it into an object in `configureMcpForVSCode`?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+      // File doesn't exist or is invalid JSON
+    }
+
+    const mcpConfig = JSON.parse(config) as Record<string, unknown>
+    const updatedConfig = { ...existingConfig, ...mcpConfig }
+
+    await fs.writeFile(configPath, JSON.stringify(updatedConfig, null, 2), 'utf-8')
+    log(`${chalk.green('✅')} VS Code MCP configuration saved to ${chalk.cyan('.vscode/mcp.json')}`)
+  } catch (error) {
+    throw new Error(`Failed to configure VS Code MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
```
eduardoboucas: When will `error` not be an instance of `Error`?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+      },
+      null,
+      2,
+    )
+  )
+}
+
+// VS Code specific MCP configuration
+const configureMcpForVSCode = async (config: string, projectPath: string): Promise<void> => {
+  const configPath = resolve(projectPath, '.vscode', 'mcp.json')
```
eduardoboucas: nit: Save the path to the `.vscode` directory in a variable to avoid resolving the path multiple times.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    await fs.writeFile(configPath, config, 'utf-8')
+    log(`${chalk.green('✅')} Cursor MCP configuration saved to ${chalk.cyan('.cursor/mcp.json')}`)
+  } catch (error) {
+    throw new Error(`Failed to configure Cursor MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
+  }
+}
+
+// Windsurf specific MCP configuration
+const configureMcpForWindsurf = async (config: string, _projectPath: string): Promise<void> => {
+  const { homedir } = await import('node:os')
```
eduardoboucas: Why are you using a dynamic import here?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+  const configPath = resolve(homedir(), '.codeium', 'windsurf', 'mcp_config.json')
```
eduardoboucas: nit: Save the path to the `windsurf` directory in a variable to avoid resolving the path multiple times.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+// Cursor specific MCP configuration
+const configureMcpForCursor = async (config: string, projectPath: string): Promise<void> => {
+  const configPath = resolve(projectPath, '.cursor', 'mcp.json')
+
+  try {
+    await fs.mkdir(resolve(projectPath, '.cursor'), { recursive: true })
+    await fs.writeFile(configPath, config, 'utf-8')
+    log(`${chalk.green('✅')} Cursor MCP configuration saved to ${chalk.cyan('.cursor/mcp.json')}`)
+  } catch (error) {
+    throw new Error(`Failed to configure Cursor MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
```
eduardoboucas: Same question as before about the type-check on `error`.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    const updatedConfig = {
+      ...existingConfig,
+      mcpServers: {
+        ...existingServers,
+        ...newServers,
+      },
+    }
+
+    await fs.writeFile(configPath, JSON.stringify(updatedConfig, null, 2), 'utf-8')
+    log(`${chalk.green('✅')} Windsurf MCP configuration saved to global config`)
```
eduardoboucas: What does "global config" refer to here?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    log(`${chalk.gray('💡')} Restart Windsurf to activate the MCP server`)
+  } catch (error) {
+    throw new Error(`Failed to configure Windsurf MCP: ${error instanceof Error ? error.message : 'Unknown error'}`)
+  }
+}
+
+// Generic MCP configuration display
+const showGenericMcpConfig = (config: string, ideName: string): void => {
+  log(`\n${chalk.yellow('📋 Manual Configuration Required')}`)
```
eduardoboucas: nit

```suggestion
  log(`\n${chalk.yellow('📋 Manual configuration required')}`)
```

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+
+// Trigger IDE-specific MCP configuration
+const triggerMcpConfiguration = async (ide: ConsumerConfig, projectPath: string): Promise<boolean> => {
+  log(`\n${chalk.blue('🔧 MCP Configuration for')} ${chalk.cyan(ide.presentedName)}`)
+
+  const { shouldConfigure } = await inquirer.prompt<{ shouldConfigure: boolean }>([
+    {
+      type: 'confirm',
+      name: 'shouldConfigure',
+      message: `Would you like to automatically configure MCP server for ${ide.presentedName}?`,
```
eduardoboucas:
```suggestion
      message: `Would you like to automatically configure the MCP server for ${ide.presentedName}?`,
```

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+      default: true,
+    },
+  ])
+
+  if (!shouldConfigure) {
+    log(chalk.gray('Skipped MCP configuration. You can set it up manually later.'))
```
eduardoboucas: Should we tell people _how_ they can set it up later?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+const fetchProjectInfo = async (url: string): Promise<ProjectInfo> => {
+  try {
+    const response = await fetch(url, {
+      headers: {
+        'content-type': 'text/plain',
+      },
+    })
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${String(response.status)}`)
```
eduardoboucas: Do we want to provide a more user-friendly (and hopefully actionable) error message here?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    const parsedData = JSON.parse(data) as unknown as ProjectInfo
+    return parsedData
+  } catch (error) {
+    throw new Error(`Failed to fetch project information: ${error instanceof Error ? error.message : 'Unknown error'}`)
+  }
+}
+
+const getRepoUrlFromProjectId = async (api: NetlifyAPI, projectId: string): Promise<string> => {
+  try {
+    const SiteInfo = (await api.getSite({ siteId: projectId })) as SiteInfo
```
eduardoboucas: This is not a class?

```suggestion
    const siteInfo = (await api.getSite({ siteId: projectId })) as SiteInfo
```

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+ * Handles AI rules initialization workflow
+ * This is the experimental --ai-rules functionality for the init command
+ */
+export const initWithAiRules = async (hash: string, command: BaseCommand): Promise<void> => {
+  // Authenticate first before any API operations
+  await command.authenticate()
+  const { api } = command.netlify
+
+  log(`${chalk.blue('🤖 Initializing AI project')} with rules...`)
+  log(`${chalk.gray('Hash:')} ${hash}`)
```
eduardoboucas: Do we have to print the hash? Why do people care?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+  log(`${chalk.gray('User:')} ${api.accessToken ? 'Authenticated ✅' : 'Not authenticated ❌'}`)
+
+  try {
+    // Step 1: Decode hash and fetch project information
+    log('\n📋 Decoding project hash...')
```
eduardoboucas: I think "decoding project hash" doesn't mean a lot to people. I suggest something more user-friendly.

```suggestion
    log('\n📋 Extracting project details...')
```

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+    const decodedUrl = decodeHash(hash)
+    log(`${chalk.cyan('Decoded URL:')} ${decodedUrl}`)
+
+    log('\n🔍 Fetching project information...')
+    const projectInfo = await fetchProjectInfo(decodedUrl)
+
+    log(`${chalk.cyan('Project ID:')} ${projectInfo.projectId}`)
+
+    // Step 2: Get repository URL from project ID via Netlify site API
+    log('\n🔗 Linking to Netlify site and fetching repository...')
```
eduardoboucas:
```suggestion
    log('\n🔗 Linking to Netlify project and fetching repository...')
```

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+      await savePrompt(projectInfo.prompt, ntlContext, targetDir)
+    }
+
+    // Step 5: Detect IDE and configure MCP server
+    log('\n🔍 Detecting development environment...')
+    const detectedIDE = await detectIDE()
+    let mcpConfigured = false
+
+    if (detectedIDE) {
+      log(`${chalk.green('✅')} Detected IDE: ${chalk.cyan(detectedIDE.presentedName)}`)
```
eduardoboucas: Can we settle on either "development environment" or "IDE"? Using both next to each other feels unnecessarily confusing.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,447 @@
...
+
+    if (detectedIDE) {
+      if (mcpConfigured) {
+        log(chalk.yellowBright(`🔧 Step 2: MCP Server Configured`))
+        log(`   ${chalk.green('✅')} ${chalk.cyan(detectedIDE.key)} is ready with Netlify MCP server`)
+        log(`   ${chalk.gray('💡 MCP will activate when you reload/restart your IDE window')}`)
+      } else {
+        log(chalk.yellowBright(`🔧 Step 2: Manual MCP Configuration`))
+        log(`   ${chalk.cyan(detectedIDE.key)} detected - MCP setup was skipped`)
+        log(`   ${chalk.gray('You can configure MCP manually later for enhanced AI capabilities')}`)
```
eduardoboucas: How?

eduardoboucas (changes requested):

src/commands/init/ai-rules.ts
denar90: it's always unknown, but here yeah it will be Error, it's more following code pattern across repo

src/commands/init/ai-rules.ts
denar90: means windsurf mcp config file

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+const savePrompt = async (instructions: string, ntlContext: string | null, targetDir: string): Promise<void> => {
+  try {
+    const filePath = resolve(targetDir, 'AI-instructions.md')
+    await fs.writeFile(filePath, `Context: ${ntlContext ?? ''}\n\n${instructions}`, 'utf-8')
+    log(`${chalk.green('✅')} AI instructions saved to ${chalk.cyan('AI-instructions.md')}`)
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : 'Unknown error'
+    log(`${chalk.yellow('⚠️')} Warning: Failed to save AI instructions: ${errorMessage}`)
```
eduardoboucas: The CLI has a `warn` method precisely for this purpose: https://github.com/netlify/cli/blob/f449ad8c00285ee4a2e5ae582c4cfdcaae9beb72/src/utils/command-helpers.ts#L185-L188

src/utils/mcp-utils.ts
```diff
@@ -0,0 +1,144 @@
...
+import { promises as fs } from 'node:fs'
+import { homedir } from 'node:os'
+import { chalk, log } from './command-helpers.js'
+import type { ConsumerConfig } from '../recipes/ai-context/context.js'
+
+/**
+ * Generate MCP configuration for the detected IDE or development environment
+ */
+export const generateMcpConfig = (ide: ConsumerConfig): Record<string, unknown> => {
+  const configs: Record<string, Record<string, unknown>> = {
```
eduardoboucas, line 11: It would be great to properly type this.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,255 @@
...
+// Helper functions reused from ai-start.ts
+const decodeHash = (hash: string): string => {
+  try {
+    return atob(hash)
+  } catch (error) {
+    throw new Error(`Failed to decode hash: ${error instanceof Error ? error.message : 'Invalid base64 or URL'}`)
+  }
+}
+
+const fetchProjectInfo = async (url: string): Promise<ProjectInfo> => {
```
eduardoboucas, line 95: It would be great to have a comment explaining what is the URL we're hitting — is it the Netlify API, something else?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,255 @@
...
+    const parsedData = JSON.parse(data) as unknown as ProjectInfo
+    return parsedData
+  } catch (error) {
+    throw new Error(`Failed to fetch project information: ${error instanceof Error ? error.message : 'Unknown error'}`)
+  }
+}
+
+const getRepoUrlFromProjectId = async (api: NetlifyAPI, projectId: string): Promise<string> => {
+  try {
+    const siteInfo = (await api.getSite({ siteId: projectId })) as SiteInfo
```
eduardoboucas, line 113: Do we not already have the site information populated by the base command? Or is this a different site? If we could use that, we'd avoid making an extra API call.

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,255 @@
...
+    cloneSpinner.success({ text: `Cloned repository to ${chalk.cyan(targetDir)}` })
+
+    // Step 4: Save AI instructions to file
+    if (projectInfo.prompt) {
+      const ntlContext = await fetch(
+        'https://docs.netlify.com/ai-context/scoped-context?scopes=serverless,blobs,forms',
+        {
+          method: 'GET',
+          headers: {
+            'Content-Type': 'text/plain',
```
eduardoboucas: Why are you setting `Content-Type` on a GET request? Did you mean to use `Accept`?

src/commands/init/ai-rules.ts
```diff
@@ -0,0 +1,255 @@
...
+    const detectedIDE = await detectIDE()
+    let mcpConfigured = false
+
+    if (detectedIDE) {
+      log(`${chalk.green('✅')} Detected development environment: ${chalk.cyan(detectedIDE.presentedName)}`)
+      mcpConfigured = await triggerMcpConfiguration(detectedIDE, targetDir)
+    }
+
+    // Update working directory to cloned repo
+    process.chdir(targetDir)
```
eduardoboucas, line 204: What is the purpose of this?

eduardoboucas (approved):
Approving so you can get unblocked, but I left some notes and questions that I think we should be addressed.

src/commands/init/ai-rules.ts
eduardoboucas: I see a couple of instances of that, which doesn't feel like enough of a reason to perpetuate this and just duplicate this code every time we handle an error, especially when we know for a fact that the type will be `Error`.

denar90:
Awesome 🙌 thank you

src/commands/init/ai-rules.ts
ndhoule: This is a normal idiom--TypeScript can't guarantee caught values are errors.

PR #7366, opened 2025-06-25
fix: providing cwd matching the root dir and filter option conflicted…

VitaliyR:
… preventing functions to be discovered

#### Summary

Fixes EX_184
Preview Server always provides a `cwd` matching the root dir. But since it was provided - the `filter` option started to conflict, so edge function dir never set hence functions are never "discovered".

---

      passes our tests.

### DIFF
--- a/src/commands/base-command.ts
+++ b/src/commands/base-command.ts
@@ -500,9 +500,21 @@ export default class BaseCommand extends Command {
   private async init(actionCommand: BaseCommand) {
     debug(`${actionCommand.name()}:init`)('start')
     const flags = actionCommand.opts()
+
     // here we actually want to use the process.cwd as we are setting the workingDir
     // eslint-disable-next-line no-restricted-properties
-    this.workingDir = flags.cwd ? resolve(flags.cwd) : process.cwd()
+    const processCwd = process.cwd()
+
+    if (flags.cwd) {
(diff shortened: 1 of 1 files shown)


src/commands/base-command.ts
```diff
@@ -500,9 +500,21 @@ export default class BaseCommand extends Command {
...
-    this.workingDir = flags.cwd ? resolve(flags.cwd) : process.cwd()
+    const processCwd = process.cwd()
+
+    if (flags.cwd) {
+      const resolvedCwd = resolve(flags.cwd)
+      this.workingDir = resolvedCwd
+
+      // if cwd matches process.cwd, act like cwd wasn't provided
+      if (resolvedCwd === processCwd) {
+        delete flags.cwd
```
cmparsons, line 514: Is this the main change? I'm kinda confused on what behavior has changed because it seems like we're still resolving to the `cwd` from the flag or `process.cwd()`

cmparsons (approved):

src/commands/base-command.ts
VitaliyR: Down the process it is being check https://github.com/netlify/cli/blob/28252a8af5b83d2062fdb46b0a7b12233b649cbc/src/commands/base-command.ts#L544

VitaliyR:
Confirmed fix in RC

PR #7367, opened 2025-06-25
feat: throw error on deploy command if build fails and user has --json option

sarahetter:
#### Summary

When a build fails when a user runs `netlify deploy --json` it currently fails silently. Behaviour can be seen by running:
```
export DEPLOY_JSON=$(netlify deploy --json)
echo $DEPLOY_JSON
```
    
   on a project that has `exit 1` as its build command. 
   
This PR changes that so we now throw an error like elsewhere in the deploy command, regardless of the `--json` option


tests/integration/commands/deploy/deploy.test.ts
```diff
@@ -449,6 +449,33 @@ describe.skipIf(process.env.NETLIFY_TEST_DISABLE_LIVE === 'true').concurrent('co
...
+        })
+
+      await builder.build()
+
+      await expect(
+        callCli(['deploy', '--json'], {
+          cwd: builder.directory,
+          env: { NETLIFY_SITE_ID: context.siteId },
+        }),
+      ).rejects.toThrow('Error while running build')
```
serhalp, line 475: do we expect the output to be json on failures or no?

sarahetter: nope don't do that anywhere else

PR #7408, opened 2025-07-10
feat: load generated functions from Netlify Build

eduardoboucas:
#### Summary

Loads functions generated by the new `functions.generate` util added in https://github.com/netlify/build/pull/6487 and https://github.com/netlify/build/pull/6525.

It depends on https://github.com/netlify/build/pull/6539, so tests will fail until that lands.


src/lib/functions/registry.ts
```diff
@@ -524,10 +539,11 @@ export class FunctionsRegistry {
           return
         }
 
+        const directory = directories.find((directory) => mainFile.startsWith(directory)) ?? srcDir
```
eduardoboucas: I'm not super happy with this. For some context, `directory` contains the top-level functions directory where this specific function is found (not the sub-directory where a function may live). For example, for both `netlify/functions/foo/index.js` and `netlify/functions/bar.js` this value would be `netlify/functions`.

This value was `undefined` for any functions in `generatedFunctions`, because those functions don't typically live inside any of the regular function directories. So for those cases the right value to use is `srcDir`.

I think this is in need of some refactoring though, because I'm not sure why we need to keep track of the parent functions directory and not just the directory where the function lives.

I'm keeping this as is for now, where we're just adding `srcDir` as a fallback, to minimise the impact of this change, and then we'll revisit when possible.

src/utils/run-build.ts
```diff
@@ -47,7 +47,22 @@ const cleanInternalDirectory = async (basePath?: string) => {
...
+type RunNetlifyBuildOptions = {
+  command: BaseCommand
+  // The flags of the command
+  options: $TSFixMe
+  settings: ServerSettings
+  env: NodeJS.ProcessEnv
+  timeline: 'dev' | 'build'
+}
+
+export async function runNetlifyBuild(
```
eduardoboucas, line 59: The main change here is to add two overloads to this function such that when we call it with `timeline: "dev"` we get the return value correctly typed to the result of running the dev timeline, and the same for `build`.

src/lib/functions/registry.ts
```diff
@@ -527,7 +550,6 @@ export class FunctionsRegistry {
         const func = new NetlifyFunction({
           blobsContext: this.blobsContext,
           config: this.config,
-          directory: directories.find((directory) => mainFile.startsWith(directory)),
```
eduardoboucas, line 529: This was weird. We were computing `directory` by looking at the function and ascertaining which one of the main directories it belonged to. We then passed this property along and [further downstream](https://github.com/netlify/cli/pull/7408/files#diff-3177d45a0f449e1b37a03d73a574731300452cc8ddf10c52b2bbae400371d7fdL75) we'd use it to decide what path to send to `zipFunction`, which should always be the path to the function boundary (i.e. the function sub-directory if it lives in a sub-directory, or the function main file if not).

But that's exactly what the `srcPath` that we get from zip-it-and-ship-it is. So I've removed this property altogether, passed `srcPath` to `NetlifyFunction` and then read it from there when calling `zipFunction`.

src/lib/functions/registry.ts
```diff
@@ -464,14 +469,32 @@ export class FunctionsRegistry {
...
+        },
+        user: {
+          // In reality, `directories` contains both directories with user and
+          // generated functions. The registry currently lacks knowledge about
+          // the contents of each directory, so we put them in the same bag and
+          // rely on the order of the directories to get the priority right.
+          // But now that zip-it-and-ship-it accepts an object with mixed paths
+          // that lets us specify exactly which paths contain user functions or
+          // generated functions, we should refactor this call so it encodes
+          // that distiction.
```
ndhoule:
```suggestion
          // that distinction.
```

PR #7426, opened 2025-07-15
refactor: use globalConfig from @netlify/dev-utils

keanu-a:

#### Summary

Fixes #EX-331

Replacing duplicated globalConfig from `cli` to use globalConfig from `@netlify/dev-utils`

---

      passes our tests.

### DIFF
--- a/src/utils/telemetry/telemetry.ts
+++ b/src/utils/telemetry/telemetry.ts
@@ -2,10 +2,10 @@ import { dirname, join } from 'path'
 import process, { version as nodejsVersion } from 'process'
 import { fileURLToPath } from 'url'
 
+import { getGlobalConfigStore } from '@netlify/dev-utils'
 import { isCI } from 'ci-info'
 
 import execa from '../execa.js'
-import { getGlobalConfigStore } from '@netlify/dev-utils'
 
 import { isTelemetryDisabled, cliVersion } from './utils.js'
 import isValidEventName from './validation.js'


src/commands/base-command.ts
```diff
@@ -643,7 +643,7 @@ export default class BaseCommand extends Command {
       siteData = result
     }
 
-    const globalConfig = await getGlobalConfigStore()
+    const globalConfig = await globalConfigStore.default()
```
serhalp: hrm, we should never need to do any esm/commonjs acrobatics like this these days, especially with our own code. could you instead [open a PR over here](https://github.com/netlify/primitives/blob/9da35b3f9e52971cbfe162190b6568fc9c98d2db/packages/dev-utils/src/main.ts#L8) to replace that `export * as` with explicit re-exports? something like `export {default as getGlobalConfigStore, GlobalConfigStore, resetConfigCache} from ...`

src/utils/types.ts
```diff
@@ -4,7 +4,11 @@ import type { IncomingMessage } from 'http'
 import type { PollingStrategy } from '@netlify/build-info'
 import type { Match } from 'netlify-redirector'
 
-export type { GlobalConfigStore } from './get-global-config-store.js'
+import { globalConfig } from '@netlify/dev-utils'
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type GlobalConfigStore<T extends Record<string, any> = Record<string, any>> = globalConfig.GlobalConfigStore<T>
```
serhalp: 🤔 There should already be a type like this exported from `@netlify/dev-utils`

src/commands/base-command.ts
keanu-a: Yup makes sense, heres the [primitives PR](https://github.com/netlify/primitives/pull/359)

src/utils/types.ts
keanu-a: Yup got rid of that 😅 and replaced with a simple
`export type { GlobalConfigStore } from '@netlify/dev-utils'`

src/commands/base-command.ts
```diff
@@ -35,7 +35,7 @@ import {
 } from '../utils/command-helpers.js'
 import type { FeatureFlags } from '../utils/feature-flags.js'
 import { getFrameworksAPIPaths } from '../utils/frameworks-api.js'
-import getGlobalConfigStore from '../utils/get-global-config-store.js'
+import { getGlobalConfigStore } from '@netlify/dev-utils'
```
serhalp: huh, I thought we had a lint rule enforcing this, but fyi we do import groups: first node builtins, empty line, then npm packages, empty line, then relative files

keanu-a: Noted! I'll go over all imports I changed then

keanu-a: Moved all new imports from `@netlify/dev-utils` to correct import grouping: [b38d980](https://github.com/netlify/cli/pull/7426/commits/b38d980307b18f7206002880df00d007ee2ffdc5)

PR #7439, opened 2025-07-18
feat: match edge functions by header

eduardoboucas:
#### Summary

Adds support for matching edge functions by header.

### DIFF
--- a/src/lib/edge-functions/registry.ts
+++ b/src/lib/edge-functions/registry.ts
@@ -21,7 +21,7 @@ import type { FeatureFlags } from '../../utils/feature-flags.js'
 import { MultiMap } from '../../utils/multimap.js'
 import { getPathInProject } from '../settings.js'
 
-import { INTERNAL_EDGE_FUNCTIONS_FOLDER } from './consts.js'
+import { DIST_IMPORT_MAP_PATH, INTERNAL_EDGE_FUNCTIONS_FOLDER } from './consts.js'
 
 type DependencyCache = Record<string, string[]>
 type EdgeFunctionEvent = 'buildError' | 'loaded' | 'reloaded' | 'reloading' | 'removed'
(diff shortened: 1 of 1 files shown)


src/lib/edge-functions/registry.ts
```diff
@@ -421,6 +421,34 @@ export class EdgeFunctionsRegistry {
         return
       }
 
+      if (route.headers && headers) {
```
serhalp: 🤔 what happens if `route.headers` but `!headers`? we might intend to match specific headers but we wouldn't run that logic? is `!headers` every expected? could we handle this more gracefully and explicitly?

src/lib/edge-functions/registry.ts
```diff
@@ -421,6 +421,34 @@ export class EdgeFunctionsRegistry {
...
+            return headers[headerName] === undefined
+          }
+
+          if (headerValueString && headerMatch?.matcher === 'regex') {
+            const pattern = new RegExp(headerMatch.pattern)
+
+            return pattern.test(headerValueString)
+          }
+
+          return false
```
serhalp, line 444: am I reading correctly that this is an invalid user configuration? since we're in dev, it might be helpful DX to print a user warning in this case.

eduardoboucas: Not really. This is an internal structure that has been generated from parsing the edge function files and the TOML config. If there's an error in syntax, it will be thrown by edge-bundler.

From my tests, we might need some additional logic to surface those errors better (unrelated to this PR), which I'll do in a follow-up.

src/lib/edge-functions/registry.ts
eduardoboucas: I don't think we should ever hit `!headers`, though. Removed in https://github.com/netlify/cli/pull/7439/commits/3cc29d043e757922b08cdc9ce112d06f77d3a74e.

src/lib/edge-functions/registry.ts
eduardoboucas: Actually, I've done it in https://github.com/netlify/cli/pull/7439/commits/7d6f789a7639a6a38ebd47ca806b0cf700e68444 because it's a small fix (and you now you're obliged to review!).

PR #7448, opened 2025-07-21
refactor: use ensureNetlifyIgnore from @netlify/dev-utils

keanu-a:

#### Summary

Fixes #EX-350

Refactored `ensureNetlifyIgnore` to use from `@netlify/dev-utils`

---

Ref: https://github.com/netlify/primitives/blob/main/packages/dev-utils/src/lib/gitignore.ts#L20


src/utils/gitignore.ts
```diff
@@ -1,41 +0,0 @@
-import { readFile, writeFile } from 'fs/promises'
-import path from 'path'
-
-import parseIgnore from 'parse-gitignore'
```
serhalp, line 4: you can uninstall this and and `@types/parse-gitignore`!

src/commands/dev/dev.ts
```diff
@@ -2,6 +2,7 @@ import process from 'process'
 
 import type { NetlifyAPI } from '@netlify/api'
 import { applyMutations } from '@netlify/config'
+import { ensureNetlifyIgnore } from '@netlify/dev-utils'
```
serhalp: I know this is a draft! but some pointers: we don't want to lose [these logs](https://github.com/netlify/primitives/blob/578e5f27148c3283aadf29b5b99147d297b43b06/packages/dev-utils/src/lib/gitignore.ts#L40-L41), so we'll need to pass in a logger. Note that we don't currently have/use a logger that conforms to the shape it expects.

it might also be nice to do this in a util (maybe keep `src/utils/gitignore.ts`) so we don't need to think about this everywhere we use it, especially because I think it will need to be a little awkward, something like:

```js
import { ensureNetlifyIgnore as originalEnsureNetlifyIgnore } from '@netlify/dev-utils'
import {log, warn, logError} from './command-helpers.js'

export const ensureNetlifyIgnore = async (rootDir: string) => originalEnsureNetlifyIgnore(
  rootDir,
  {log, warn, error: logError},
)
```

keanu-a: Ah I completely missed that logger parameter...

I added back `gitignore.ts` as a util and reverted all usages of `ensureNetlifyIgnore` back to using this updated util function: [e605e83](https://github.com/netlify/cli/pull/7448/commits/e605e83f584958deb940e130971ebaccae7f7e58#diff-dc24708bd95f9482ee7d38a7836175af3a849aecebaed95790c00b7ad5617f6d)

src/utils/gitignore.ts
keanu-a: Uninstalled! Will look out for more unused packages

PR #7454, opened 2025-07-22
refactor: use LocalState from @netlify/dev-utils

keanu-a:

#### Summary

Fixes #EX-367

Replaced all usages of `CLIState` with `LocalState` from `@netlify/dev-utils`. Deleted `cli-state.js` as no longer used.

Ref: https://github.com/netlify/primitives/blob/main/packages/dev-utils/src/lib/local-state.ts#L27

---

      passes our tests.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.12.2
+          node-version: 20.19.4
           cache: npm
 
       - name: Install dependencies
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20.12.2
+          node-version: 20.19.4
           cache: npm
 
       - name: Install dependencies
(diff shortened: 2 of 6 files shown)


src/utils/cli-state.ts
serhalp, line 1: 👍🏼 appears to be identical, so no need for any other changes

src/utils/cli-state.ts
```diff
@@ -1,125 +0,0 @@
-import fs from 'fs'
-import path from 'path'
-import process from 'process'
-
-import { deleteProperty, getProperty, hasProperty, setProperty } from 'dot-prop'
-import { findUpSync } from 'find-up'
```
serhalp, line 6: looks like this is the only use of `find-up`, so we can remove it!

keanu-a: Ah it's the only use of `findUpSync` but other files do use `findUp` from the `find-up` package still 😅
Ex: https://github.com/netlify/cli/blob/keanualoua/ex-367-use-localstate-in-netlifydev-utils/src/commands/base-command.ts

serhalp: wtf, github search lied to me earlier, but now it's showing all the results. my bad!

serhalp (approved):

src/utils/cli-state.ts
keanu-a: All good! It does lie sometimes 🤣

PR #7467, opened 2025-07-23
refactor: use watchDebounced from @netlify/dev-utils

keanu-a:
#### Summary

Fixes #EX-324

Replaced all usages from cli `watchDebounced` with function from `@netlify/dev-utils` package. Removed function from `command-helpers.ts` and unused imports.

Ref: https://github.com/netlify/primitives/blob/main/packages/dev-utils/src/lib/watch-debounced.ts

---

      passes our tests.


src/utils/command-helpers.ts
```diff
@@ -7,10 +6,7 @@ import { format, inspect } from 'util'
 import type { NetlifyAPI } from '@netlify/api'
 import { getGlobalConfigStore } from '@netlify/dev-utils'
 import { Chalk } from 'chalk'
-import chokidar from 'chokidar'
-import decache from 'decache'
 import WSL from 'is-wsl'
-import debounce from 'lodash/debounce.js'
```
serhalp, line 13: this time I don't think github is fooling me 😁 — we can remove this dependency!

serhalp (approved):
LGTM, just one follow-up

src/utils/command-helpers.ts
keanu-a: This time github isn't fooling you! But we do use other submodules from the `lodash` package 😅 
Ex: https://github.com/netlify/cli/blob/main/src/commands/base-command.ts
```
import merge from 'lodash/merge.js'
import pick from 'lodash/pick.js'
```

PR #7513, opened 2025-07-30
chore: fail test on missing route mock

sarahetter:

#### Summary

---

      passes our tests.

### DIFF
--- a/MOCK_API_IMPROVEMENTS.md
+++ b/MOCK_API_IMPROVEMENTS.md
@@ -1,0 +1,61 @@
+# Mock API Improvements
+
+## Fail on Unmocked Routes (NEW)
+
+By default, tests will now **fail** when they try to access API routes that haven't been mocked. This prevents tests from accidentally hitting real APIs and improves test isolation.
+
+### Example: Before
+```javascript
+// OLD BEHAVIOR: This would silently return 404 and might cause confusing test failures
+await withMockApi([
+  { path: 'sites/site_id', response: siteInfo }
+], async ({ apiUrl }) => {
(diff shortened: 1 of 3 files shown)


MOCK_API_IMPROVEMENTS.md
```diff
@@ -0,0 +1,61 @@
+# Mock API Improvements
```
aitchiss, line 1: I'm curious why to have this in its own file? It feels more like what should be in the PR description for traceability, and then the 'new' info would live in `CONTRIBUTING` for folks checking how to write tests

tests/integration/utils/mock-api-vitest.ts
```diff
@@ -48,7 +49,7 @@ const clearRequests = (requests: unknown[]) => {
   requests.length = 0
 }
 
-export const startMockApi = ({ routes, silent }: MockApiOptions): Promise<MockApi> => {
+export const startMockApi = ({ routes, silent = false, failOnMissingRoutes = true }: MockApiOptions): Promise<MockApi> => {
```
aitchiss, line 52: is `failOnMissingRoutes` ever `false`? I'm wondering if we need to maintain the logic for it or if we can just move to it being the only behaviour

sarahetter: Ahhh sorry Suzanne this was meant to be a draft and just testing if CI acts like local 😅

MOCK_API_IMPROVEMENTS.md
sarahetter: Claude doing Claude shit lol. This won't be merged

PR #7550, opened 2025-08-15
feat: add deploy config files when deploying

eduardoboucas:
#### Summary

Adds files from `.netlify/deploy-config` when deploying.


src/lib/edge-functions/deploy.ts
```diff
@@ -1,41 +0,0 @@
-import { stat } from 'fs/promises'
```
eduardoboucas, line 1: This whole file was moved to `src/utils/deploy/process-files.ts`, since it's not specific to edge functions anymore.

src/utils/deploy/process-files.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const getEdgeFunctionsDistPathIfExists = async (workingDir: string) =>
+  getDirectoryIfExists(join(workingDir, edgeFunctionsDistPath))
+
+const getDirectoryIfExists = async (directoryPath: string) => {
+  try {
+    const stats = await stat(directoryPath)
+
+    if (!stats.isDirectory()) {
+      throw new Error(`Path ${directoryPath} must be a directory.`)
```
JakeChampion, line 45: this gets silently dropped by the empty catch block on line 49

pieh: True, but it was also just moved as-is from other module ( as Eduardo mentioned in https://github.com/netlify/cli/pull/7550#discussion_r2278632498 )

eduardoboucas: Yeah, I didn't want to change the existing behaviour too much. I'm torn on whether we should start failing the deploy if these internal paths exists as non-directories, but we can do that as a follow-up.

PR #7573, opened 2025-08-21
feat: upload zip after deploy

denar90:

#### Summary

---

      passes our tests.


src/commands/deploy/deploy.ts
```diff
@@ -542,8 +543,20 @@ const runDeploy = async ({
     }
 
     const draft = !deployToProduction && !alias
-    results = await api.createSiteDeploy({ siteId, title, body: { draft, branch: alias } })
+    const createDeployBody = { draft, branch: alias, include_upload_url: options.uploadSourceZip }
```
khendrikse, line 546: @denar90 I think it would be good to add something here that shows a warning and then quits for windows users if they are using this new flag, since I think they are not able to use the zip functionality

denar90: good idea, I moved on command level and default falg to false

PR #7585, opened 2025-08-26
fix: load correct environment variables in Preview Server context

ndhoule:
This changeset fixes environment-variable loading a Preview Server context.

Previously, we were missing `dev-server` from the CLI's list of
supported contexts. Because of a variable that (unnecessarily) mixes the
concept of branches and contexts, when we run into an unsupported
context, we assume we must be running a branch preview, and try to find
variables for the branch named `dev-server`. For the majority of users,
variables for that branch won't exist, and they'll end up with an
unexpectedly empty variable.

Why do we assume it's a branch preview when we have enough information
available to us to know it's not one? Why do we throw away the type
information that would have caught this? Why do we load environment
variables from the API in the CLI even though `@netlify/config` already
does this?

These are the mysteries of the universe.

### DIFF
--- a/src/commands/dev/dev.ts
+++ b/src/commands/dev/dev.ts
@@ -137,7 +137,7 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
 
   env[BLOBS_CONTEXT_VARIABLE] = { sources: ['internal'], value: encodeBlobsContext(blobsContext) }
 
-  if (!options.offline && !options.offlineEnv) {
+  if (!(options.offline || options.offlineEnv)) {
     env = await getEnvelopeEnv({ api, context: options.context, env, siteInfo })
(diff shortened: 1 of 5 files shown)


src/commands/dev/dev.ts
```diff
@@ -137,7 +137,7 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
 
   env[BLOBS_CONTEXT_VARIABLE] = { sources: ['internal'], value: encodeBlobsContext(blobsContext) }
 
-  if (!options.offline && !options.offlineEnv) {
+  if (!(options.offline || options.offlineEnv)) {
```
VitaliyR, line 140: well, I find first to easier to read, no?

ndhoule: A matter of taste, but I find "if neither one is true" to be easier to reason about than "if not x and not y." (Fewer logical operations to hold in your head, too.)

PR #7609, opened 2025-09-04
feat: Support AI gateway

kathmbeck:

#### Summary

Provides initial AI gateway support for local dev. 

---

      passes our tests.

### DIFF
--- a/src/commands/dev/dev.ts
+++ b/src/commands/dev/dev.ts
@@ -4,6 +4,7 @@ import type { NetlifyAPI } from '@netlify/api'
 import { applyMutations } from '@netlify/config'
 import { OptionValues } from 'commander'
 
+import { fetchAIGatewayToken } from '../../lib/api.js'
 import { BLOBS_CONTEXT_VARIABLE, encodeBlobsContext, getBlobsContextWithEdgeAccess } from '../../lib/blobs/blobs.js'
 import { promptEditorHelper } from '../../lib/edge-functions/editor-helper.js'
 import { startFunctionsServer } from '../../lib/functions/server.js'
@@ -19,7 +20,7 @@ import {
   netlifyCommand,
 } from '../../utils/command-helpers.js'
 import detectServerSettings, { getConfigWithPlugins } from '../../utils/detect-server-settings.js'
-import { UNLINKED_SITE_MOCK_ID, getDotEnvVariables, getSiteInformation, injectEnvVariables } from '../../utils/dev.js'
+import { UNLINKED_SITE_MOCK_ID, getDotEnvVariables, getSiteInformation, injectEnvVariables, parseAIGatewayContext } from '../../utils/dev.js'
 import { getEnvelopeEnv } from '../../utils/env/index.js'
 import { ensureNetlifyIgnore } from '../../utils/gitignore.js'
 import { getLiveTunnelSlug, startLiveTunnel } from '../../utils/live-tunnel.js'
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -62,10 +62,11 @@
     "@netlify/blobs": "10.0.10",
     "@netlify/build": "35.1.6",
     "@netlify/build-info": "10.0.7",
+    "@netlify/ai-gateway": "file:../primitives/packages/ai-gateway",
```
kathmbeck: Tested this locally, will need to update once https://github.com/netlify/primitives/pull/450 is released

src/commands/dev/dev.ts
```diff
@@ -155,6 +157,16 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
     siteInfo,
   })
 
+  if (!options.offline && !options.offlineEnv) {
+    await setupAIGateway({ api, env, siteID: site.id, siteURL: siteUrl })
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (env.AI_GATEWAY?.value) {
+    process.env.AI_GATEWAY = env.AI_GATEWAY.value
```
eduardoboucas: Injecting the environment variables into the process should happen automatically in the `injectEnvVariables` call above (L148). Any chance we could move the AI gateway initialisation above it so that we don't need to do this manually? I think we'd need to move the `getSiteInformation` call up a bit, but that should be okay.

src/commands/dev/dev.ts
```diff
@@ -155,6 +157,16 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
...
+    log(`${NETLIFYDEVLOG} AI Gateway configured for AI provider SDK interception`)
```
eduardoboucas: I wonder if this log message is going to be useful to people. 🤔

src/commands/functions/functions-serve.ts
```diff
@@ -37,6 +40,15 @@ export const functionsServe = async (options: OptionValues, command: BaseCommand
     siteInfo,
   })
 
+  if (!options.offline) {
+    await setupAIGateway({ api, env, siteID: site.id, siteURL: siteUrl })
+  }
+
+  if (env.AI_GATEWAY?.value) {
+    process.env.AI_GATEWAY = env.AI_GATEWAY.value
```
eduardoboucas: Same comment that I left on the `dev` command about manually injecting the variable into the process.

tests/integration/commands/functions-serve/functions-serve.test.ts
```diff
@@ -193,4 +202,127 @@ describe.concurrent('functions:serve command', () => {
...
+  test('should inject AI Gateway when linked site and online', async (t) => {
+    await withSiteBuilder(t, async (builder) => {
+      const { siteInfo, aiGatewayToken, routes } = createAIGatewayTestData()
+      const checkFunction = createAIGatewayCheckFunction()
+
+      await builder
+        .withNetlifyToml({ config: { functions: { directory: 'functions' } } })
+        .withFunction({
+          path: checkFunction.path,
+          handler: () => {
```
eduardoboucas: This is using the legacy Lambda syntax and I don't think the AI gateway is expected to work with the Lambda compatibility mode since we don't have a bootstrap there to populate the environment.

You can change the test to use a v2 function like this: https://github.com/netlify/cli/blob/961abeec8d354cf02699055d99baa15f9b6251a0/tests/integration/commands/dev/functions.test.ts#L141

tests/integration/utils/ai-gateway-helpers.ts
```diff
@@ -0,0 +1,110 @@
+import { TestContext } from 'vitest'
```
eduardoboucas, line 1: Just pointing out that the pattern of having a util file for a specific chunk of tests, away from the test file itself, is not a pattern I've seen in this repo. Same goes for the usage of a function (`createAIGatewayCheckFunction`) to generate the functions to be tested, which feels like yet another abstraction on top of the site builder that may make tests a bit harder to follow.

Not an issue necessarily, and definitely not a blocker.

src/commands/dev/dev.ts
kathmbeck: 🤔 Hmm I would expect some type of log, but open to suggestions on wording

tests/integration/utils/ai-gateway-helpers.ts
kathmbeck: I see what you're saying, I pulled this out because I had about ~200 lines of duplication. I could really go either way though -- happy to inline back if you think the consistency is more beneficial.

eduardoboucas (approved):
🚀

PR #7636, opened 2025-09-23
fix: zip based deploys ignore pattern and subdir case

VaibhavAcharya:
#### Summary

Fixes #EX-835

### Problem Description

The zip deploy flow was failing for projects without git repositories when using the `--upload-source-zip` flag. The failure occurred during zip file creation with the error:

```bash
Error: Command failed: zip -r /tmp/.../workspace-snapshots/source-xxx.zip . -x node_modules/* ...
zip I/O error: No such file or directory
zip error: Could not create output file
```

### Root Cause

The issue was that the zip command was trying to create a file at a path like `/tmp/abc123/workspace-snapshots/source-xxx.zip`, but the `workspace-snapshots/` subdirectory didn't exist within the temporary directory.

The Netlify API provides filenames that include subdirectory paths, but the CLI wasn't creating the necessary directory structure before attempting to write the zip file.

### Reproduction Steps

To reproduce this issue:

1. Create a test directory with some files (no git repo)
2. Run: `netlify deploy --upload-source-zip --json --draft --dir . --no-build`
3. Observe the zip command failure

Example reproduction:

```bash
mkdir test-deploy && cd test-deploy
echo "<h1>Test</h1>" > index.html
netlify deploy --upload-source-zip --json --draft --dir . --no-build
```

### Solution

1. Added directory creation before zip file creation in src/utils/deploy/upload-source-zip.ts:

   ```ts
   // Ensure the directory for the zip file exists
   // The filename from the API includes a subdirectory path (e.g., 'workspace-snapshots/source-xxx.zip')
   // While temporaryDirectory() creates a new empty directory, the subdirectory within it doesn't exist
   // so we need to create it before the zip command can write the file
   await mkdir(dirname(zipPath), { recursive: true });
   ```

2. Also improved exclusion patterns from `node_modules` to `node_modules*` for more complete directory exclusion.

---

      passes our tests.


VaibhavAcharya:
**Ignore the failing tests, they are not failing. They work after a retry. 😭**

Some tests in this repo are indeterministic or the setup isn't right.
Debugging for some other time.

src/utils/deploy/upload-source-zip.ts
```diff
@@ -60,6 +61,12 @@ const createSourceZip = async ({
   const tmpDir = temporaryDirectory()
   const zipPath = join(tmpDir, filename)
 
+  // Ensure the directory for the zip file exists
+  // The filename from the API includes a subdirectory path (e.g., 'workspace-snapshots/source-xxx.zip')
+  // While temporaryDirectory() creates a new empty directory, the subdirectory within it doesn't exist
+  // so we need to create it before the zip command can write the file
+  mkdirSync(dirname(zipPath), { recursive: true })
```
VitaliyR: better to not use syncs, it blocks entire i/o. Let's use 
```
await fs.mkdir(dirPath, { recursive: true }) // from fs/promises
```

VaibhavAcharya: done

VitaliyR:
@VaibhavAcharya please use `expect(mockFs.mkdir).toHaveBeenCalledWith(path.join(...))` to fix tests

PR #7640, opened 2025-09-23
fix: support non interactive tty when deploying

VitaliyR:

#### Summary

Adds support for non interactive TTY when running `netlify deploy`

---

      passes our tests.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -1,6 +1,7 @@
 import { type Stats } from 'fs'
 import { stat } from 'fs/promises'
 import { basename, resolve } from 'path'
+import { stdin, stdout } from 'process'
 
 import type { NetlifyAPI } from '@netlify/api'
 import { type NetlifyConfig, type OnPostBuild, runCoreSteps } from '@netlify/build'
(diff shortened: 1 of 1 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -10,6 +10,7 @@ import isObject from 'lodash/isObject.js'
 import { parseAllHeaders } from '@netlify/headers-parser'
 import { parseAllRedirects } from '@netlify/redirect-parser'
 import prettyjson from 'prettyjson'
+import { stdin, stdout } from 'process'
```
eduardoboucas: nit: Move this to the top, near the other built-in imports.

src/commands/deploy/deploy.ts
```diff
@@ -118,6 +119,12 @@ const getDeployFolder = async ({
   }
 
   if (!deployFolder) {
+    if (!stdin.isTTY || !stdout.isTTY) {
+      // non interactive - can't get the value, resolve to the cwd
+      deployFolder = command.workspacePackage ? command.jsWorkspaceRoot || site.root : command.workingDir
+      return deployFolder || command.workingDir
```
eduardoboucas: nit: I personally find the nesting of conditionals quite confusing and would prefer a more explicit version, even if more verbose. Up to you, though.

```suggestion
      if (command.workspacePackage) {
        if (command.jsWorkspaceRoot) {
          return command.jsWorkspaceRoot
        }

        if (site.root) {
          return site.root
        }
      }

      // Either not a workspace package, or empty site root.
      return command.workingDir
```

src/commands/deploy/deploy.ts
```diff
@@ -118,6 +119,12 @@ const getDeployFolder = async ({
   }
 
   if (!deployFolder) {
+    if (!stdin.isTTY || !stdout.isTTY) {
```
eduardoboucas, line 122: nit: Might be nice to abstract this to a `isInteractive` method that we use in the different places we're checking this to ensure consistent behaviour. I could also see us introducing an explicit flag to control this behaviour (similar to npm's `--yes`) and that would make it easier to integrate. Not a blocker, though.

eduardoboucas (approved):

PR #7735, opened 2025-10-22
fix: set same Node.js options locally as AWS Lambda does in production

serhalp:
#### Summary

Despite these Node.js features being enabled by default on some subset of these Node.js 20 and 22 versions, AWS Lambda always disables them explicitly via flags: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html#w292aac41c19.

This leads to a mismatch between the local dev Netlify Functions environment and the production (AWS Lambda) environment, leading to confusing bugs like this: https://github.com/netlify/react-router-template/issues/10#issuecomment-3411177684.

This PR replicates the production Netlify Functions behaviour in the local execution environment.

Following the repro steps from the React Router 7 bug above, I've confirmed this fixes it ✅. ~~It would be nice to add test coverage but I failed to write a failing test 😓.~~ Woo, I did it.

### DIFF
--- a/src/lib/functions/runtimes/js/index.ts
+++ b/src/lib/functions/runtimes/js/index.ts
@@ -89,16 +89,22 @@ export const invokeFunction = async ({
       ...process.env,
       // AWS Lambda disables these Node.js experimental features, even in Node.js versions where they are enabled by
       // default: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html#w292aac41c19.
-      // Replace this here to match production as closely as possible.
+      // They also allow users to re-enable (i.e. not disable) these by co-opting the positive flag (which in reality
+      // may or may not exist depending on the exact node.js version). We replicate all this behavior here.
       NODE_OPTIONS: [
         ...(process.env.NODE_OPTIONS?.split(' ') ?? []),
-        '--no-experimental-require-module',
-        '--no-experimental-detect-module',
-      ]
-        // Unfortunately Node.js throws if `NODE_OPTIONS` contains any unsupported flags and these flags have been added
-        // and removed in various specific versions in each major line. Luckily Node.js has an API just for this!
-        .filter((flag) => process.allowedNodeEnvironmentFlags.has(flag))
-        .join(' '),
+        ...[
+          ...(process.env.NODE_OPTIONS?.includes('--experimental-require-module')
+            ? []
+            : ['--no-experimental-require-module']),
+          ...(process.env.NODE_OPTIONS?.includes('--experimental-detect-module')
+            ? []
+            : ['--no-experimental-detect-module']),
+        ]
+          // Unfortunately Node.js throws if `NODE_OPTIONS` contains any unsupported flags and these flags have been
+          // added and removed in various specific versions in each major line. Luckily Node.js has an API just for this!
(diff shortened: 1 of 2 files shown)


src/lib/functions/runtimes/js/index.ts
```diff
@@ -84,7 +84,20 @@ export const invokeFunction = async ({
...
+  const worker = new Worker(workerURL, {
+    env: {
+      ...process.env,
+      // AWS Lambda disables these Node.js experimental features, even in Node.js versions where they are enabled by
+      // default: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html#w292aac41c19.
+      // Replace this here to match production as closely as possible.
+      NODE_OPTIONS: ['--no-experimental-require-module', '--no-experimental-detect-module']
+        // Unfortunately Node.js throws if `NODE_OPTIONS` contains any unsupported flags and these flags have been added
+        // and removed in various specific versions in each major line. Luckily Node.js has an API just for this!
+        .filter((flag) => process.allowedNodeEnvironmentFlags.has(flag))
```
serhalp: TIL: https://nodejs.org/docs/latest-v22.x/api/process.html#processallowednodeenvironmentflags

src/lib/functions/runtimes/js/index.ts
```diff
@@ -84,7 +84,24 @@ export const invokeFunction = async ({
...
+  const worker = new Worker(workerURL, {
+    env: {
+      ...process.env,
+      // AWS Lambda disables these Node.js experimental features, even in Node.js versions where they are enabled by
+      // default: https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html#w292aac41c19.
+      // Replace this here to match production as closely as possible.
+      NODE_OPTIONS: [
+        ...(process.env.NODE_OPTIONS?.split(' ') ?? []),
+        '--no-experimental-require-module',
+        '--no-experimental-detect-module',
```
pieh: To match:
> To enable a disabled experimental feature, set the re-enable flag in the NODE_OPTIONS environment variable. For example, to enable ES module require support, set NODE_OPTIONS to --experimental-require-module. Lambda detects this override and removes the corresponding disable flag.

```suggestion
        process.env.NODE_OPTIONS?.includes('--experimental-require-module') ? '' : '--no-experimental-require-module',
        process.env.NODE_OPTIONS?.includes('--experimental-detect-module') ? '' : '--no-experimental-detect-module',
```

Or something like that?

tests/integration/commands/functions-serve/functions-serve.test.ts
```diff
@@ -203,6 +204,155 @@ describe.concurrent('functions:serve command', () => {
...
+            t.expect(body).toContain('--abort-on-uncaught-exception --trace-exit')
+          },
+        )
+      })
+    },
+  )
+
+  test.runIf(
+    process.allowedNodeEnvironmentFlags.has('--no-experimental-require-module') ||
+      process.allowedNodeEnvironmentFlags.has('--experimental-require-module'),
```
serhalp, line 292: This may seem weird, but it's because some node.js versions support the _positive_ flag and some node.js versions support the _negative_ flag

pieh (approved):

PR #7743, opened 2025-10-24
fix: bundle edge functions if they exist on deploy --no-build

pieh:

#### Summary

Fixes https://linear.app/netlify/issue/FRB-2024/ensure-edge-functions-are-bundled-with-no-build-option

---

      passes our tests.


tests/integration/commands/deploy/deploy.test.ts
```diff
@@ -178,11 +178,13 @@ describe.skipIf(process.env.NETLIFY_TEST_DISABLE_LIVE === 'true').concurrent('co
         .withNetlifyToml({
           config: {
             build: { publish: 'public', command: 'echo "no op"' },
-            edge_functions: [{ function: 'edge', path: '/*' }],
```
pieh: this was making this test pass, but also it meant that
```
'should deploy Edge Functions when directory exists'
```
test description was not correct, as with no explicit edge functions config, edge functions were NOT deployed

src/commands/deploy/deploy.ts
```diff
@@ -727,8 +728,11 @@ const bundleEdgeFunctions = async (options: DeployOptionValues, command: BaseCom
     // We log our own progress so we don't want this as well. Plus, this logs much of the same
     // information as the build that (likely) came before this as part of the deploy build.
     quiet: options.debug ?? true,
-    // @ts-expect-error FIXME(serhalp): This is missing from the `runCoreSteps` type in @netlify/build
+    // (cachedConfig type error hides this one, but it still is valid) @ts-expect-error FIXME(serhalp): This is missing from the `runCoreSteps` type in @netlify/build
     edgeFunctionsBootstrapURL: await getBootstrapURL(),
+    // @ts-expect-error 'CachedConfig' is not assignable to type 'Record<string, unknown>'.
+    // Index signature for type 'string' is missing in type 'CachedConfig'.
+    cachedConfig: command.netlify.cachedConfig,
```
pieh, line 735: this is a bit frustrating, once I do add `cachedConfig` here, it no longer complain about `edgeFunctionsBootstrapURL` and now it actually complain about `@ts-expect-error` for it ... but this is still valid, just somehow hidden because of `cachedConfig` type error ... 

Passing `cachedConfig` seems needed to properly support `--cwd other/dir` case, as otherwise bundling was trying to look in current directory and there doesn't seem to be other arguments we can pass to tell it to look elsewhere (despite passing `cwd` explitly above).

src/commands/deploy/deploy.ts
```diff
@@ -867,10 +871,11 @@ const prepAndRunDeploy = async ({
...
-
-  // build flag wasn't used and edge functions exist
-  if (!options.build && edgeFunctionsConfig && edgeFunctionsConfig.length !== 0) {
+  // build flag wasn't used and edge functions directories exist
+  if (!options.build && (await anyEdgeFunctionsDirectoryExists(command))) {
+    // for the case of directories existing but not containing any edge functions,
+    // there is early bail in edge functions bundling after scanning for edge functions
+    // for this case and to avoid replicating scanning logic here, we defer to the bundling step
     await bundleEdgeFunctions(options, command)
   }
```
pieh, line 880: the previous condition doesn't seem to be met unless you specify some edge functions configuration in toml (?) so instead we check if there are edge functions dir present and trigger on-demand edge bundling if they are

if directories exist but contain on edge functions - there will be early bail - that's just edge bundling function step in `@netlify/build` works, so I opted not trying to repeat same logic here

PR #7768, opened 2025-11-10
feat: create deploy before building

eduardoboucas:
Currently, the `deploy` command (without `--no-build`) first builds the site and then does a deploy. This means that the build process won't be able to reference the deploy, because it doesn't exist at that point.

This is inconsistent with what happens in our build system and creates compatibility issues with certain features. For example, skew protection only works if the `DEPLOY_ID` environment variable is populated with the deploy ID, so it currently doesn't work with CLI deploys.

This PR changes the order of operations such that we first create a deploy, then we run the build referencing that deploy, and finally update the deploy depending on the result of the build and deploy processes (whether successful or not).

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -373,41 +373,46 @@ const prepareProductionDeploy = async ({ api, siteData, options, command }) => {
   }
 }
 
-// @ts-expect-error TS(7006) FIXME: Parameter 'actual' implicitly has an 'any' type.
-const hasErrorMessage = (actual, expected) => {
+const hasErrorMessage = (actual: unknown, expected: string): boolean => {
   if (typeof actual === 'string') {
     return actual.includes(expected)
   }
   return false
 }
 
-// @ts-expect-error TS(7031) FIXME: Binding element 'error_' implicitly has an 'any' t... Remove this comment to see the full error message
-const reportDeployError = ({ error_, failAndExit }) => {
+interface DeployError extends Error {
+  json?: { message?: string }
+  status?: unknown
+}
+const reportDeployError = ({
+  error,
+  failAndExit,
+}: {
+  error: DeployError
+  failAndExit: (err: unknown) => never
+}): never => {
   switch (true) {
-    case error_.name === 'JSONHTTPError': {
-      const message = error_?.json?.message ?? ''
+    case error.name === 'JSONHTTPError': {
+      const message = error.json?.message ?? ''
       if (hasErrorMessage(message, 'Background Functions not allowed by team plan')) {
         return failAndExit(`\n${BACKGROUND_FUNCTIONS_WARNING}`)
       }
-      warn(`JSONHTTPError: ${message} ${error_.status}`)
-      warn(`\n${JSON.stringify(error_, null, '  ')}\n`)
-      failAndExit(error_)
-      return
+      warn(`JSONHTTPError: ${message} ${error.status}`)
+      warn(`\n${JSON.stringify(error, null, '  ')}\n`)
+      return failAndExit(error)
     }
-    case error_.name === 'TextHTTPError': {
-      warn(`TextHTTPError: ${error_.status}`)
-      warn(`\n${error_}\n`)
-      failAndExit(error_)
-      return
+    case error.name === 'TextHTTPError': {
+      warn(`TextHTTPError: ${error.status}`)
+      warn(`\n${error}\n`)
+      return failAndExit(error)
     }
-    case hasErrorMessage(error_.message, 'Invalid filename'): {
-      warn(error_.message)
-      failAndExit(error_)
-      return
+    case hasErrorMessage(error.message, 'Invalid filename'): {
+      warn(error.message)
+      return failAndExit(error)
     }
     default: {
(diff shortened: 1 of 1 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -632,7 +641,12 @@ const runDeploy = async ({
     if (deployId) {
       await cancelDeploy({ api, deployId })
     }
+
     reportDeployError({ error_, failAndExit: logAndThrowError })
+
+    // We'll never get here, as `reportDeployError` calls `logAndThrowError`,
+    // which will throw. But we're throwing anyway just to make types happy.
+    throw error_
```
serhalp, line 649: I pushed a type fix to avoid needing this. Hope that's cool 😁.

eduardoboucas: Nice, thanks!

src/commands/deploy/deploy.ts
```diff
@@ -1080,29 +1095,75 @@ export const deploy = async (options: DeployOptionValues, command: BaseCommand)
...
+      source_zip_filename?: string
+    }
+    const deployId = deployMetadata.id || ''
+    const deployUrl = deployMetadata.deploy_ssl_url || deployMetadata.deploy_url || ''
+
+    command.netlify.cachedConfig.env.DEPLOY_ID = { sources: ['internal'], value: deployId }
+    command.netlify.cachedConfig.env.DEPLOY_URL = { sources: ['internal'], value: deployUrl }
+
+    process.env.DEPLOY_ID = deployId
+    process.env.DEPLOY_URL = deployUrl
```
serhalp, line 1121: 🤔 What purpose does populating these in the CLI process serve? If it serves a purpose, could we leave a comment here explaining that? and is it possible to add test coverage?

src/commands/deploy/deploy.ts
```diff
@@ -703,9 +718,6 @@ const handleBuild = async ({
 
     logAndThrowError(`Error while running build${message}`)
   }
-  if (exitCode !== 0) {
-    exit(exitCode)
-  }
```
serhalp, line 708: Are these changes related to this PR? I'm not sure I'm following. Could we maybe pull them out into a separate PR for traceability? That would also give us a chance to think through it carefully and add test coverage.

As is, I might be missing something, but it seems like this changes some behaviours. Previously, we called `logAndThrowError()` (which does exactly what it says on the tin, i.e. it does _not_ exit) in the `json` case and quietly called `exit()` instead in the non-`json` case. After these changes, we always call `logAndThrowError()` — this changes the output behaviour — and never call `exit()`. If I'm tracing this right, the thrown error will bubble all the way up to `commander`'s handling. This may all end up resulting in similar behaviour, but it seems risky, especially without thorough coverage.

I assume the reason this change was made is to hit line 1163 in the build failure + non-json case, because we now have an orphaned deploy to clean up?

serhalp (commented):
LGTM other than my two comments

src/commands/deploy/deploy.ts
eduardoboucas: > I assume the reason this change was made is to hit line 1163 in the build failure + non-json case, because we now have an orphaned deploy to clean up?

Yes. Without this change, `handleBuild` won't throw, so we can't take any action if the build fails.

I don't think we can move this to a separate PR because cleaning up an orphaned deploy is a critical part of this.

Rather than explicitly calling `exit`, we now throw the error and let it bubble up. 

Can you expand on what behaviours you think this might change?

src/commands/deploy/deploy.ts
eduardoboucas: This is lets `@netlify/config`, which runs in the same process, grab the deploy details.

src/commands/deploy/deploy.ts
serhalp: > I don't think we can move this to a separate PR because cleaning up an orphaned deploy is a critical part of this.

I meant that PR could land first

> Can you expand on what behaviours you think this might change?

I was wondering about other, existing cases where `await runBuild()` might throw, which we previously handled ourselves and exited but would now lead to bubbling the error all the way up to commander, the details of which I'm not intimately familiar with.

but I've just now [looked up the docs (for the correct old commander version we're using)](https://github.com/tj/commander.js/tree/970ecae402b253de691e6a9066fea22f38fe7431?tab=readme-ov-file#override-exit-and-output-handling) and it seems like it should be fine—it should just log the error and exit(1).

Oh, there _is_ a behaviour change here I believe: in the `json` case, this PR will add new (non-JSON) output (the error message, or rather [specifically our custom error output](https://github.com/netlify/cli/blob/d8e2c596445f76310f5967ba224c84781992988c/src/commands/main.ts#L268-L273)) to `stderr`. But since I believe that's where we landed as our ideal behaviour for `--json` (100% JSON on `stdout`, freeform on `stderr`), I think we're good? 👍🏼

src/commands/deploy/deploy.ts
serhalp: In that case, what is the purpose of the mutations to `command.netlify.cachedConfig.env`? and why can't we use the `newEnvChanges` mechanism? I think the why would be worth documenting in a comment in a follow-up PR!

Issue #7747, opened 2025-10-27
ECONNRESET crash when using Netlify CLI with Node.js 24.x and Nuxt 4.x dev server


ElijahJunaid:
### Describe the bug

Netlify CLI consistently crashes with Error: read ECONNRESET when attempting to proxy to a Nuxt 4.x development server on Node.js 24.6.0. The crash occurs immediately after the Nuxt dev server successfully starts and builds. The error is unhandled and terminates the Netlify CLI process.

Error Output:
```
Error: Netlify CLI has terminated unexpectedly
Error: read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:216:20)
```
This appears to be a socket handling incompatibility between Netlify CLI and Node.js 24.x.

### Steps to reproduce

1. Install Node.js 24.6.0
2. Create a fresh Nuxt 4.x project
```
npx nuxi@latest init test-project --packageManager npm --no-gitInit
cd test-project`
3. Create netlify.toml:
`[dev]
  framework = "nuxt"
  port = 8888
  targetPort = 3000
  autoLaunch = false
```
4. Run Netlify Dev
5. Wait for it to complete then load localhost:8888

### Configuration

netlify.toml:
```
[dev]
  framework = "nuxt"
  port = 8888
  targetPort = 3000
  autoLaunch = false
```

package.json:
```
{
  "name": "nuxt-app",
  "private": true,
  "type": "module",
  "scripts": {
    "build": "nuxt build",
    "dev": "nuxt dev",
    "generate": "nuxt generate",
    "preview": "nuxt preview",
    "postinstall": "nuxt prepare"
  },
  "dependencies": {
    "nuxt": "^4.2.0",
    "vue": "latest"
  }
}
```

nuxt.config.ts:
```
export default defineNuxtConfig({
  compatibilityDate: '2024-11-01',
  devtools: { enabled: true }
})
```

### Environment

System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 AMD Ryzen 5 7600X 6-Core Processor
    Memory: 13.65 GB / 31.14 GB
Binaries:
    Node: 24.6.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.5.1 - C:\Program Files\nodejs\npm.CMD

ElijahJunaid:
fixed issue with 6 edits to the netlify cli repository in a local version;
Implemented the Node 24.x “ECONNRESET” resilience work:

src/utils/framework-server.ts
[...]

PR #7772, opened 2025-11-12
fix: handle ECONNRESET errors in Node.js 24.x

ElijahJunaid:
#7747 - Replace wait-port dependency with custom implementation
- Add retry logic with ECONNRESET error handling
- Add HTTP agent with ECONNRESET handling in proxy
- Enable framework host detection for Node 24+
- Update tests to use new wait-port API

Fixes Node.js 24.x dev server startup failures

### DIFF
--- a/src/lib/wait-port.ts
+++ b/src/lib/wait-port.ts
@@ -4,11 +4,12 @@ export const waitPort = async (
   port: number,
   host: string,
   timeout: number,
-  maxRetries = 10,
+  maxRetries?: number,
 ): Promise<{ open: boolean; ipVersion?: 4 | 6 }> => {
   const startTime = Date.now()
+  const retries = maxRetries ?? Math.ceil(timeout / 2000)
 
-  for (let attempt = 0; attempt < maxRetries; attempt++) {
+  for (let attempt = 0; attempt < retries; attempt++) {
     if (Date.now() - startTime > timeout) {
       return { open: false }
     }


ElijahJunaid:
@VitaliyR did the changes and fixed the tests, they all passed on my end

src/utils/run-build.ts
```diff
@@ -127,8 +126,7 @@ export async function runNetlifyBuild({
     })
 
     settings.frameworkHost = ipVersion === 6 ? '::1' : '127.0.0.1'
-    const nodeVersion = process.versions.node
-    settings.detectFrameworkHost = options.skipWaitPort || semver.gte(nodeVersion, '24.0.0')
+    settings.detectFrameworkHost = options.skipWaitPort
```
VitaliyR, line 129: so this change is not needed anymore?

ElijahJunaid: correct, it was updated with wait-port.ts

ElijahJunaid:
hopefully this one fixes the failed tests

src/lib/wait-port.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        socket.on('connect', () => {
+          isResolved = true
+          socket.end()
+          resolve()
+        })
+
+        socket.on('error', (error) => {
+          isResolved = true
+          cleanup()
```
VitaliyR: cleanup never called neither here nor in other places, since isResolved set to true everywhere, isn't it?
seems that cleanup is being redundant?

ElijahJunaid: you are absolutely correct, its been removed

ElijahJunaid: the tests keep failing but it seems unrelated to my changes. my wait port is only to check if a port is open, but the env var isnt being injected, not something i touched

ElijahJunaid:
removed the proxy.ts handling added by an earlier version, caused unnecessary 500 errors, original problem still solved

ElijahJunaid:
@VitaliyR tests that are failing have nothing to do with my pull request, if you want i can investigate further and try to get a spotless test but itd have more files to be involved

src/utils/framework-server.ts
```diff
@@ -56,14 +55,13 @@ export const startFrameworkServer = async function ({
...
       port = { open: true, ipVersion }
     } else {
-      const waitPortPromise = waitPort({
+      const waitPortPromise = waitPort(
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        port: settings.frameworkPort!,
-        host: 'localhost',
-        output: 'silent',
-        timeout: FRAMEWORK_PORT_TIMEOUT_MS,
-        ...(settings.pollingStrategies?.includes('HTTP') && { protocol: 'http' }),
```
serhalp, line 65: I was worried about this removal but I _think_ it works out:
- Technically we allow this `pollingStrategies` configuration. It's unclear and undocumented what values are even supported... but [every single framework is currently configured to use `tcp`](https://github.com/search?q=repo%3Anetlify%2Fbuild+%2FpollingStrategies%2F&type=code).
- Technically a user could directly configure this in their `netlify.toml`... but even more technically [it isn't a documented field](https://docs.netlify.com/build/configure-builds/file-based-configuration/)
- There is no default here, so we were effectively never passing a `protocol` to the `wait-port` lib
- This isn't really documented, but [if you don't pass a `protocol` to `wait-port` it will check via TCP only](https://github.com/dwmkerr/wait-port/blob/235d8ea1f1989ade0eb849dfb3f6e7c8af92a8f2/lib/wait-port.js#L147-L157)

So... I think it works out. But we should remove `pollingStrategies` from `@netlify/build-info`.

VitaliyR:
@ElijahJunaid let me run the tests on my end, if they would pass I'm inclined to merge this PR

src/utils/framework-server.ts
ElijahJunaid: should be done by a maintainer after the merge as its just info

ElijahJunaid:
@VitaliyR youve gotten it to 3 failed tests (lint title does not count)

ElijahJunaid:
@VitaliyR i readded the prior retry count from the old wait-port to see if that fixes the last 3 tests as they seemed to timeout and i wondered if giving them more time would fix it - if the tests dont pass now, idk what im doing wrong

ElijahJunaid:
i need somebody to approve the checks to run! @VitaliyR @serhalp

ElijahJunaid:
@VitaliyR you need to do ur magic work again so we can see whats left in actual problems if any. currently i dont see any problems relating to this pull request, but it may be obstructed because of all of the unrelated errors (github problem)

VitaliyR (approved):

VitaliyR:
@ElijahJunaid thanks!

PR #7796, opened 2025-11-25
feat: support skew protection token

eduardoboucas:
Grabs the skew protection token from the deploy object and passes it to Netlify Build.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -41,7 +41,7 @@ jobs:
         run: echo ${{ github.event.number }} > ./pr_number
 
       - name: Upload deltas
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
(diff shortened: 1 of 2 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -703,6 +707,15 @@ const handleBuild = async ({
     currentDir,
     deployHandler,
   })
+
+  if (deployId) {
```
eduardoboucas: This is a nicer way to pass the deploy ID to Netlify Build. It's what the `deployId` flag is for.

VitaliyR: Perhaps we can just extend the `getRunBuildOptions` method to set these options there?

eduardoboucas: Done in https://github.com/netlify/cli/pull/7796/commits/980d9113b5e4e15ce71568125f5db3bd1549deba.

PR #7814, opened 2025-12-04
fix: include sourceZipFileName in results for printing results if available

khendrikse:

#### Summary

In [this PR](https://github.com/netlify/cli/pull/7768) and CLI v.23.11.0 a change was introduced that stopped the `source_zip_filename` field from being printed in the json output. 

**Why**
The results object used is no longer the one from the deploy, which would have contained the sourceZipFilename. 

This PR ensures we print that field again.

The agent runner parses that JSON output and looks for `source_zip_filename` to set `result_zip_file_name` on the session, but since it's not there, it stays `null`, and the Publish button doesn't show in our UI.

---

      passes our tests.


khendrikse:
BTW: Also looking into making sure this cannot happen again, as it is such an important part of our workflow.

PR #7880, opened 2026-01-20
test: skip yarn install tests on Node 20.12.2 due to transitive dep c…

jaredm563:
Problem: Yarn v1 enforces engine constraints strictly. A transitive dep (chokidar@5) requires node >=20.19.0
Since Node 20 reaches EOL in April 2026, skip these tests rather than doing an override

The reason this is a problem at all is that despite the fact that the netlify-cli package includes [an npm-shrinkwrap.json file](https://docs.npmjs.com/cli/v11/configuring-npm/npm-shrinkwrap-json), which enforces strict dependency tree resolution for netlify-cli consumers, this is only respected by npm and pnpm, not yarn v1. In other words, npm i -g netlify-cli@1.2.3 and pnpm add -g netlify-cli@1.2.3 are deterministic but yarn global add netlify-cli@1.2.3 is not.

The alternatives would be
-Pinning Chokidar to an older version
-Bumping the node version

But considering these can introduce breaking changes, and the fact the Node 20 is EOL in ~3 months, we can just skip this and leave a comment for maintainers to understand why we have to make this workaround.

#### Summary

---

      passes our tests.


serhalp:
@jaredm563 an additional note for the PR description: the reason this is a problem at all is that despite the fact that the netlify-cli package includes [an `npm-shrinkwrap.json` file](https://docs.npmjs.com/cli/v11/configuring-npm/npm-shrinkwrap-json), which enforces strict dependency tree resolution for netlify-cli consumers, this is only respected by npm and pnpm, not yarn v1. In other words, `npm i -g netlify-cli@1.2.3` and `pnpm add -g netlify-cli@1.2.3` are deterministic but `yarn global add netlify-cli@1.2.3` is not.

e2e/install.e2e.ts
```diff
@@ -205,8 +205,11 @@ const installTests: [packageManager: string, config: InstallTest][] = [
 describe.each(installTests)('%s → installs the cli and runs commands without errors', (packageManager, config) => {
   // TODO: Figure out why this flow is failing on Windows.
   const npxOnWindows = platform() === 'win32' && 'run' in config
```
serhalp: while you're here... am I reading this wrong or is this dead code? There's never `'run' in config` in this `describe`. I think this might be leftover from splitting out the run tests into a separate `describe` below?

jaredm563: Nope, you're totally right.

jaredm563: poor cli needs some love

jaredm563:
Removed the dead `npxOnWindows` check entirely since` 'run' in config` was always `false`, then in another area simplified to `skipOnWindows = platform() === 'win32'` since `'run' in config` was always `true`

e2e/install.e2e.ts
```diff
@@ -306,9 +304,9 @@ const runTests: [packageManager: string, config: RunTest][] = [
 
 describe.each(runTests)('%s → runs cli commands without errors', (packageManager, config) => {
   // TODO: Figure out why this flow is failing on Windows.
-  const npxOnWindows = platform() === 'win32' && 'run' in config
+  const skipOnWindows = platform() === 'win32'
 
-  itWithMockNpmRegistry.skipIf(npxOnWindows)('runs commands without errors', async ({ registry }) => {
+  itWithMockNpmRegistry.skipIf(skipOnWindows)('runs commands without errors', async ({ registry }) => {
```
serhalp, line 309: 🤔 but I think this is the case where the 'run' thing _was_ relevant. Now that we've removed the condition, it—oh I see 🤦🏼. In this case it was _always true_ and in the other case it was _always false_ So they were both dead code in different ways. 👍🏼

serhalp (approved):

PR #7890, opened 2026-01-23
fix: address flaky integration tests

jaredm563:

#### Summary
Fix 1:

Multiple file watcher events (onAdd, onChange) could trigger concurrent build() calls, causing race conditions where the bundler accessed functionConfigs[name] before it was populated .

now only one build runs at a time, and when the current build completes, it checks buildPending and runs another build if needed

Fix 2:
tests/integration/utils/dev-server.ts waited up to 90s with no useful error message, now on timeout it should shows the actual server output so far for debugging

Fix 3:
In tests/integration/commands/init/init.test.ts answerWithValue(CONFIRM) sent ['\n', '\n'] (two newlines), which was causing the second newline to answer the wrong prompt.

Now it should use a single CONFIRM for list/confirm selections, so that prompts are answered correctly and the netlify.toml gets created

---

      passes our tests.


src/lib/edge-functions/registry.ts
```diff
@@ -181,7 +183,41 @@ export class EdgeFunctionsRegistry {
     return [...this.internalFunctions, ...this.userFunctions]
   }
 
-  private async build() {
+  private async build(): Promise<{ warnings: Record<string, string[]> }> {
+    // If a build is already in progress, mark that we need another build
+    // and return the current build's promise. The running build will
+    // trigger a rebuild when it completes if buildPending is true.
+    if (this.buildPromise) {
```
VitaliyR, line 212: can we add tests here for this method which will ensure this tests fine?
```
this.build() // resolved firstly
this.build() // resolved together with the next
this.build() // resolved together with the previous
```

src/lib/edge-functions/registry.ts
```diff
@@ -181,7 +183,41 @@ export class EdgeFunctionsRegistry {
     return [...this.internalFunctions, ...this.userFunctions]
   }
 
-  private async build() {
+  private async build(): Promise<{ warnings: Record<string, string[]> }> {
+    // If a build is already in progress, mark that we need another build
+    // and return the current build's promise. The running build will
+    // trigger a rebuild when it completes if buildPending is true.
```
serhalp, line 211: 👀 Instead of reimplementing this, we could reuse the existing battle-tested helper we already use for this same purpose for (non-Edge) Functions! Check it out: https://github.com/netlify/cli/blob/321170d8588d9b90459784ebed079e416e13f6c6/src/lib/functions/runtimes/js/builders/zisi.ts#L75.

src/lib/edge-functions/registry.ts
serhalp: ^ and the test should verify that subsequent calls trigger an actual new build!

src/lib/edge-functions/registry.ts
jaredm563: Good idea thank you!

jaredm563: So I looked into implementing that and unfortunately if build A fails while build B is pending, the caller of build A gets the error but build B runs without anyone waiting for it. I'll add a comment explaining why there is a custom implementation

tests/unit/lib/edge-functions/registry.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    // Create instance with minimal mocked dependencies
+    const registry = Object.create(EdgeFunctionsRegistry.prototype) as InstanceType<typeof EdgeFunctionsRegistry>
+
+    // Initialize only the properties needed for build()
+    // @ts-expect-error -- accessing private members for testing
+    registry.buildPending = false
+    // @ts-expect-error -- accessing private members for testing
+    registry.buildPromise = null
+    // @ts-expect-error -- accessing private members for testing
```
serhalp: Don't sink your day into this, but could you take a few mins to see if we can avoid adding all these `@ts-expect-error`s? 🙏🏼

khendrikse (approved):

PR #7927, opened 2026-02-04
fix(deps): upgrade express to v5 and http-proxy-middleware to v3

serhalp:
## Description

`express > qs` and `express > body-parser > qs` have `high` security vulnerabilities that cannot be addressed without upgrading from v4 to v5, as the fixes have not been backported: https://github.com/advisories/GHSA-6rw7-vpxm-498p.

This PR upgrades to v5, which resolves all remaining security vulnerabilities:

```
❯ npm audit --omit=dev
found 0 vulnerabilities
```

As a bonus, this removes 20 transitive deps and ~1 MB from the install size.

### Upgrade express from v4 to v5

Release notes: https://github.com/expressjs/express/blob/master/History.md#500--2024-09-10

### Upgrade http-proxy-middleware from v2 to v3

This is required due to the express upgrade.

Release notes: https://github.com/chimurai/http-proxy-middleware/blob/master/MIGRATION.md

| v3 Breaking Change | Applies? | Action |
| --- | --- | --- |
| Removed req.url patching | No — proxy isn't mounted on a sub-path | None |
| pathRewrite behavior change | No — not mounted on a sub-path | None |
| Removed shorthand usage | No — already using object form | None |
| Removed context argument | No — not used | None |
| Removed logProvider and logLevel | Yes — logLevel: 'warn' was used | Removed it |
| Refactored proxy events (onX → on: {}) | No — no event callbacks used | None |

## Notes

Note: ideally we wouldn't be using express at all, as it is overkill for our use cases. At the very least, since we're already using fastify, we should use that. Otherwise, we could be using something like https://npmx.dev/package/@tinyhttp/app. But all this is even more moot, because all the code that uses express, fastify, and http-proxy-middleware has already been extracted into [`@netlify/dev`](https://github.com/netlify/primitives/tree/main/packages/dev), where it has been refactored to not use any of these. But pulling that back in is a significant refactor. We'll get to it eventually.


tests/integration/commands/dev/dev.test.ts
```diff
@@ -233,7 +233,6 @@ describe.concurrent('command/dev', () => {
       await withDevServer({ cwd: builder.directory }, async (server) => {
         const getResponse = await fetch(`${server.url}/api/ping`)
         const jsonPingWithGet = await getResponse.json()
-        t.expect(jsonPingWithGet).toHaveProperty('body', {})
```
serhalp, line 236: `req.body` defaults to `undefined` instead of `{}` when no body parser matches now. I deleted the assertions instead of updating them because there's nothing valuable to test here — `GET`s don't have request bodies.

(If you're confused that this appears to be a _response_ body, it's because the test is setting up a Function that echoes back some request fields.)

serhalp:
I did some local testing of functions emulation for good measure ✅.

PR #7946, opened 2026-02-13
feat!: remove `sites:create-template` command

sarahetter:
#### Summary

**BREAKING**: The `sites:create-template` command has been removed.

It gets very very little usage, there are a lot of quicker, more user friendly ways to create a site these days. 

👋 
---


jaredm563 (approved):

src/commands/sites/sites.ts
```diff
@@ -22,23 +22,21 @@ export const createSitesFromTemplateCommand = (program: BaseCommand) => {
...
-      'netlify sites:create-template',
-      'netlify sites:create-template nextjs-blog-theme',
-      'netlify sites:create-template my-github-profile/my-template',
-    ])
-    .action(async (repository: string, options: OptionValues, command: BaseCommand) => {
-      const { sitesCreateTemplate } = await import('./sites-create-template.js')
-      await sitesCreateTemplate(repository, options, command)
+    .addHelpText(
+      'after',
+      `(Deprecated) This command has been deprecated. Learn more: https://docs.netlify.com/start/quickstarts/deploy-from-template/`,
```
serhalp: 💭 I would vote to just remove it entirely. Since it isn't a command you use as part of your existing workflow on a project, but rather something you might use to start a project, I don't think there's much risk of anything trying to run it and being confused. Thoughts?

docs/index.md
```diff
@@ -165,7 +165,7 @@ Handle various project operations
 | Subcommand | description  |
 |:--------------------------- |:-----|
 | [`sites:create`](/commands/sites#sitescreate) | Create an empty project (advanced)  |
-| [`sites:create-template`](/commands/sites#sitescreate-template) | (Beta) Create a project from a starter template  |
+| [`sites:create-template`](/commands/sites#sitescreate-template) | (Deprecated) Create a project from a starter template  |
```
serhalp: regardless of whether we actually "remove" the command, I would vote to remote it from the docs/help

commitlint.config.js
```diff
@@ -1 +1,10 @@
-export default { extends: ['@commitlint/config-conventional'] }
+export default {
+  extends: ['@commitlint/config-conventional'],
+  parserPreset: {
+    parserOpts: {
+      headerPattern: /^(\w+)(?:\(([^)]*)\))?(!)?:\s(.+)$/,
+      breakingHeaderPattern: /^(\w+)(?:\(([^)]*)\))?(!)?:\s(.+)$/,
+      headerCorrespondence: ['type', 'scope', 'breaking', 'subject'],
+    },
+  },
```
serhalp, line 9: 👁️ what's this change about?

sarahetter: the ding dang conventional commit parser I had to switch to because the old one stopped being a thing doesn't pass if you do a `!` for breaking change. (it's also looking for a footer in the title with BREAKING CHANGE but titles don't have footers, it thinks it's linting a commit message). I'm trying to avoid doing 2 rounds of CLI test flake

PR #7950, opened 2026-02-17
feat: begin integrating `@netlify/dev`

eduardoboucas:
Sets the foundation for integrating `@netlify/dev` in the CLI. We start by disabling all features (except for `db`, which is enabled with an environment variable). Then we can gradually start enabling individual features and remove the corresponding code from the codebase.

As for the term "programmatic", I was looking for a way to distinguish the new `@netlify/dev` package (which offers a _programmatic_ interface for the emulation of the Netlify platform) from the existing Netlify Dev command. Naming things is hard. Happy to go for another term if someone has a better idea.

### DIFF
--- a/tests/integration/__snapshots__/framework-detection.test.ts.snap
+++ b/tests/integration/__snapshots__/framework-detection.test.ts.snap
@@ -152,7 +152,7 @@ exports[`frameworks/framework-detection > should use static server when framewor
 ⬥ Unable to determine public folder to serve files from. Using current working directory
 ⬥ Setup a netlify.toml file with a [dev] section to specify your dev server settings.
 ⬥ See docs at: https://docs.netlify.com/cli/local-development/#project-detection
-⬥ Running static server from \\"should-use-static-server-when-framework-is-set-to-static\\"
+⬥ Running static server from \\"should-use-static-server-when-framework-i-cabde4ea\\"
 ⬥ Setting up local dev server
 
 ⬥ Static server listening to <SNAPSHOT_PORT_NORMALIZED>
@@ -168,7 +168,7 @@ exports[`frameworks/framework-detection > should warn if using static server and
 "⬥ Using simple static server because '--dir' flag was specified
 ⬥ Ignoring 'targetPort' setting since using a simple static server.
 ⬥ Use --staticServerPort or [dev.staticServerPort] to configure the static server port
-⬥ Running static server from \\"should-warn-if-using-static-server-and-target-port-is-configured/public\\"
+⬥ Running static server from \\"should-warn-if-using-static-server-and-ta-45f6af30/public\\"
 ⬥ Setting up local dev server
 
 ⬥ Static server listening to <SNAPSHOT_PORT_NORMALIZED>


package.json
```diff
@@ -158,6 +159,8 @@
     "@bugsnag/js": "8.6.0",
     "@eslint/compat": "1.4.1",
     "@eslint/js": "9.36.0",
+    "@netlify/db": "^0.1.0",
+    "@netlify/db-dev": "^0.2.0",
```
serhalp: Are these used? 🤔

eduardoboucas: In tests, yes.

serhalp: huh, test fixtures with a `package.json` don't get `npm install`ed?

eduardoboucas: Ah, I see what you mean. Moved to the fixture deps in https://github.com/netlify/cli/pull/7950/commits/bd20f1ca43e56e159e0830a5215430a0fb281076.

src/commands/blobs/blobs-set.ts
```diff
@@ -60,6 +60,7 @@ export const blobsSet = async (
   if (force === undefined) {
     const existingValue = await store.get(key)
 
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
serhalp, line 63: Is there a wrong type we should be fixing?

src/commands/dev/programmatic-netlify-dev.ts
```diff
@@ -0,0 +1,52 @@
...
+import { NETLIFYDEVWARN, log } from '../../utils/command-helpers.js'
+import type { EnvironmentVariables } from '../../utils/types.js'
+
+interface StartNetlifyDevOptions {
+  projectRoot: string
+  apiToken: string | undefined
+  env: EnvironmentVariables
+}
+
+export const startNetlifyDev = async ({
```
serhalp, line 24:
```suggestion
/**
 * Much of the core of local dev emulation of the Netlify platform was extracted
 * (duplicated) to https://github.com/netlify/primitives. This is a shim that
 * gradually enables *some* of this extracted functionality while falling back
 * to the legacy copy in this codebase for the rest.
 *
 * TODO: Hook this up to the request chain and fall through to the existing handler.
 * TODO: `@netlify/images` follows a different pattern (it is used directly).
 * Move that here.
 */
export const startNetlifyDev = async ({
```

src/commands/dev/dev.ts
```diff
@@ -174,6 +181,16 @@ export const dev = async (options: OptionValues, command: BaseCommand) => {
 
   injectEnvVariables(env)
 
+  const programmaticNetlifyDev = await startProgrammaticNetlifyDev({
```
serhalp, line 184: You may know this already, but for completeness: this would need to also be hooked up in the `dev:exec`, `serve`, and `functions:serve` commands.

src/commands/blobs/blobs-set.ts
eduardoboucas: Linting started failing for this, randomly. I haven't yet had the time to understand why, since I haven't touched this part of the code.

src/commands/dev/dev.ts
eduardoboucas: Yeah, I'll do that in a follow up. Having this on dev for now is fine.

src/commands/blobs/blobs-set.ts
eduardoboucas: I'll look into this as a follow-up.

PR #7958, opened 2026-02-20
chore(ci): fix flakey integration tests

eduardoboucas:


.github/workflows/conventional-commit.yml
```diff
@@ -7,7 +7,6 @@ jobs:
   lint-title:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
-      - name: Install Dependencies
-        run: npm install @commitlint/config-conventional
-      - uses: JulienKode/pull-request-name-linter-action@v19.0.0
+      - uses: amannn/action-semantic-pull-request@v5
```
eduardoboucas, line 10: No idea why, but the previous action [started failing](https://github.com/netlify/cli/actions/runs/22226035715/job/64292849246). Trying this one to see if it helps.

.github/workflows/integration-tests.yml
```diff
@@ -11,8 +11,43 @@ on:
...
+          SITE_NAME="netlify-test-deploy-$(openssl rand -hex 4)"
+          echo "Creating test site: $SITE_NAME"
+          HTTP_CODE=$(curl -s -o response.json -w '%{http_code}' -X POST \
+            -H "Authorization: Bearer $NETLIFY_AUTH_TOKEN" \
+            -H "Content-Type: application/json" \
+            -d "{\"name\": \"$SITE_NAME\"}" \
+            "https://api.netlify.com/api/v1/netlify-integration-testing/sites")
+          if [ "$HTTP_CODE" -ne 201 ]; then
+            echo "::warning::Failed to create site. HTTP $HTTP_CODE. Live tests will be skipped."
+            exit 0
```
serhalp, line 36: This will just quietly skip all the live tests if it fails? Is that really what we want?

.github/workflows/integration-tests.yml
```diff
@@ -11,8 +11,43 @@ on:
...
+      - name: Create site
+        id: create-site
+        run: |
+          SITE_NAME="netlify-test-deploy-$(openssl rand -hex 4)"
+          echo "Creating test site: $SITE_NAME"
+          HTTP_CODE=$(curl -s -o response.json -w '%{http_code}' -X POST \
+            -H "Authorization: Bearer $NETLIFY_AUTH_TOKEN" \
+            -H "Content-Type: application/json" \
+            -d "{\"name\": \"$SITE_NAME\"}" \
+            "https://api.netlify.com/api/v1/netlify-integration-testing/sites")
```
serhalp, line 33: Since we already have `NETLIFY_AUTH_TOKEN`, is there any particular reason we aren't using `netlify sites:create` for this? Is it to avoid using netlify-cli in the test setup for netlify-cli?

.github/workflows/integration-tests.yml
eduardoboucas: The _Create test site_ step will still fail, so we can block the PR from being merged. The alternative is that we block the entire test suite from running. Do we prefer that?

.github/workflows/integration-tests.yml
eduardoboucas: Yeah it feels overkill to install Node.js and the CLI to do something we can solve with an HTTP call.

.github/workflows/integration-tests.yml
serhalp: Ah, that would be fine, but _would_ it fail? We're exiting with code 0 🤔.

eduardoboucas: That wasn't intended. Fixed in https://github.com/netlify/cli/pull/7961.

PR #7982, opened 2026-03-03
fix: suppress spinner output when --json or --silent flag is passed

DavidWells:
## Summary

- When `--json` or `--silent` is passed, `startSpinner` now returns a no-op spinner that satisfies the `Spinner` interface but writes nothing
- This matches the existing behavior of `log()` in `command-helpers.ts`, which already no-ops on `--json` and `--silent`

## Problem

We hit this in CI where a script captures CLI output with `2>&1` (merging stderr into stdout) and pipes it to `jq`:

```bash
AGENT_OUTPUT=$(netlify agents:create "$TRIGGER_TEXT" --json 2>&1 || true)
AGENT_ID=$(echo "$AGENT_OUTPUT" | jq -r '.id // empty')
```

Even though `nanospinner` writes to stderr by default, the `2>&1` pulls spinner frames (`⠋ Creating agent task...`) into the captured output. `jq` then sees the spinner text before the JSON and fails to parse.

While the script could avoid `2>&1`, when `--json` is passed the CLI should produce machine-friendly output only — no spinners, no decorative text, regardless of which stream they target.

## Approach

Rather than adding `if (!options.json)` guards at every spinner call site, the suppression is handled centrally in `startSpinner`. When `--json` or `--silent` is detected in `process.argv`, it returns a no-op object implementing the full `Spinner` interface. This means:

- Zero changes needed at call sites — existing `.stop()`, `.success()`, `.error()` calls work as before
- Consistent with how `log()` already handles `--json`/`--silent` globally
- No risk of future commands forgetting the check

## Test plan

- [ ] `netlify agents:create "test" --json --agent claude` outputs clean JSON with no spinner text on stderr
- [ ] `netlify agents:create "test" --json --agent claude 2>&1 | jq -r '.id'` parses successfully
- [ ] Interactive mode (no `--json`) still shows spinners as before
- [ ] `npm run typecheck` passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@fastify/static": "9.0.0",
-        "@netlify/ai": "0.3.8",
+        "@netlify/ai": "0.4.0",
         "@netlify/api": "14.0.17",
         "@netlify/blobs": "10.7.0",
         "@netlify/build": "35.8.5",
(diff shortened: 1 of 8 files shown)


eduardoboucas (approved):

jaredm563:
nice!

PR #7989, opened 2026-03-05
chore(renovate): reduce noise from docs site dep updates

serhalp:
- reduce noise by grouping most docs bumps together
  - this also preserves precious Renovate rate limit slots for actual cli dep updates
- don't unnecessarily pin site/package.json deps
  - pinning is just something we need for our publish cli npm package


serhalp:
boop @sarahetter

PR #8002, opened 2026-03-10
feat: execute db commands on running instance

paulo:
#### Summary

If there's a running instance, create a `pg` client and execute the migrations using that instead of the `pglite` client.

### DIFF
--- a/tests/unit/commands/database/migrate.test.ts
+++ b/tests/unit/commands/database/migrate.test.ts
@@ -10,6 +10,7 @@ const { mockApplyMigrations, mockCleanup, mockExecutor, logMessages, jsonMessage
 })
 
 vi.mock('@netlify/dev', () => ({
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
   applyMigrations: (...args: unknown[]) => mockApplyMigrations(...args),
 }))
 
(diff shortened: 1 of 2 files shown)


src/commands/database/db-connection.ts
```diff
@@ -0,0 +1,59 @@
...
+    headers: { enabled: false },
+    images: { enabled: false },
+    redirects: { enabled: false },
+    staticFiles: { enabled: false },
+    serverAddress: null,
+  })
+
+  await netlifyDev.start()
+
+  const devConnectionString = state.get('dbConnectionString')
```
eduardoboucas: Why are we checking the state again here? Don't we expect it to be empty here?

src/commands/database/db-connection.ts
```diff
@@ -0,0 +1,54 @@
+import { Client } from 'pg'
+
+import type { SQLExecutor } from '@netlify/dev'
+import { NetlifyDev } from '@netlify/dev'
```
eduardoboucas, line 4: nit

```suggestion
import { NetlifyDev, type SQLExecutor } from '@netlify/dev'
```

src/commands/database/migrate.ts
```diff
@@ -1,7 +1,8 @@
-import { NetlifyDev } from '@netlify/dev'
+import { applyMigrations } from '@netlify/dev'
```
eduardoboucas, line 1: Just realised that when exporting this from `@netlify/dev` we should've made it more specific to make it clear that it's related to the database. Not a big deal, though.

paulo: I'll try to include both of your suggestions in a follow-up 🙏

PR #8032, opened 2026-03-13
fix: explicitly exit process on successful command completion

tobq:
## Summary

Fixes #8031

On success, `onEnd()` relied on the Node.js event loop draining naturally to exit the process. This fails when stdio pipes are closed by the parent process (e.g. CI runners, editor integrations, or tools that spawn the CLI with piped stdio and then close the connection).

On Windows specifically, the orphaned process never exits and **continuously leaks OS handles**, eventually accumulating 400,000+ handles per process. This exhausts the Windows kernel nonpaged pool (~3.4 GB vs normal ~500 MB), causing **full system freezes** including mouse/keyboard input.

### Root cause

`onEnd()` calls `exit(1)` on error but does nothing on success — it just returns, expecting the event loop to drain. However:

1. Node.js 19+ default HTTP agents use `keepAlive: true`, keeping sockets in the pool
2. When parent processes close stdio pipes (normal behavior for CI, editors, etc.), the CLI process becomes orphaned on Windows
3. The orphaned process can't exit because it has active handles (keepAlive sockets, broken pipe references)
4. Handles accumulate indefinitely until the system runs out of kernel resources

### Changes

- **`src/commands/base-command.ts`**: Add `exit()` call at the end of `onEnd()` for the success path (the error path already had `exit(1)`)
- **`bin/run.js`**: `await` the `onEnd()` call in both success and error paths so telemetry tracking completes before exit

### Reproduction

The issue can be reproduced by spawning the CLI with piped stdio, then closing the pipes:

```javascript
const child = spawn('node', ['bin/run.js', 'api', 'listSiteDeploys', '--data', '...'], { stdio: 'pipe' });
child.stdout.on('data', () => {});
// After API response, close pipes (simulating parent abandonment)
setTimeout(() => {
  child.stdout.destroy();
  child.stderr.destroy();
  child.stdin.destroy();
  child.unref();
  // Without fix: process stays alive indefinitely, leaking handles
  // With fix: process exits cleanly via process.exit(0)
}, 1500);
```

**Before fix:** Process stays alive indefinitely after pipe closure, accumulating handles
**After fix:** Process exits cleanly via `process.exit(0)` before pipes are even abandoned


aitchiss:
@tobq sorry about the delay in response here - it looks like you have some failing integration tests - I'm going to pop this in draft for now - but please re-open the PR when you've had a chance to take a look & fix 🙏

PR #8158, opened 2026-04-15
feat!: add improved `logs` command

eduardoboucas:
This PR replaces the separate `logs:function`, `logs:deploy`, and `logs:edge-functions` commands with a single unified `netlify logs` command that can pull logs from any combination of sources — functions, edge functions, and deploys — in one invocation.

Screenshot 2026-04-17 at 15 42 25
Screenshot 2026-04-17 at 15 42 41

## Quick start

```bash
# Show function and edge function logs from the last 10 minutes (default)
netlify logs

# Last hour
netlify logs --since 1h

# Stream in real time
netlify logs --follow

# All three sources
netlify logs --source functions --source edge-functions --source deploy --since 1h
```

## Source selection with `--source`

The `--source` flag controls which log sources to include. It accepts `functions`, `edge-functions`, and `deploy`, and can be passed multiple times. Defaults to `functions` and `edge-functions`.

```bash
netlify logs --source functions --since 30m
netlify logs --source edge-functions --since 1h
netlify logs --source deploy --since 2h
netlify logs --source deploy --source functions --since 1h
```

## Filtering by function name

Use `--function` and `--edge-function` to narrow down to specific functions. Both are repeatable. When either is passed without an explicit `--source`, the corresponding source is inferred automatically.

```bash
# Specific function
netlify logs --function checkout --since 1h

# Multiple functions
netlify logs --function checkout --function cart --since 1h

# Specific edge function
netlify logs --edge-function auth --since 30m

# Mix
netlify logs --function checkout --edge-function auth --since 1h
```

## Historical vs real-time

Historical mode is the default. The command fetches logs for a time window and prints them sorted by timestamp, with deploy logs appearing first when the deploy source is included.

```bash
# Duration (relative to now)
netlify logs --since 1h
netlify logs --since 30m
netlify logs --since 2 hours

# Absolute timestamps
netlify logs --since 2026-04-14T00:00:00Z --until 2026-04-15T00:00:00Z

# Mix: from 2 hours ago until 30 minutes ago
netlify logs --since 2h --until 30m
```

`--since` and `--until` accept either a **duration** (e.g. `10m`, `1h`, `24h`, `2d`, `1h30m`, `2 hours`) or an **ISO 8601 timestamp**, matching the convention used by `docker logs` and `journalctl`. `--until` defaults to now. When neither `--since` nor `--follow` is given, the default is `--since 10m`.

Real-time streaming requires the `--follow` / `-f` flag, which is mutually exclusive with `--since`/`--until`:

```bash
netlify logs --follow
netlify logs --follow --source functions --source edge-functions
```

## Scoping by deploy with `--url`

Paste a deploy URL to scope logs to that specific deploy:

```bash
# Deploy permalink (24-char hex ID)
netlify logs --function checkout --url https://507f1f77bcf86cd799439011--my-site.netlify.app --since 1h

# Branch subdomain (resolves to most recent ready deploy on that branch)
netlify logs --url https://feature-payments--my-site.netlify.app --since 1h

# Production URL or custom domain — no deploy filter
netlify logs --url https://my-site.netlify.app --since 1h
```

The URL is validated against the linked project. Mismatched URLs produce an error with a suggested fix:

```
The URL https://feature-x--other-site.netlify.app doesn't seem to match the linked project (my-site).
Try running netlify logs --since 1h --url https://my-site.netlify.app
```

## Source indicators and legend

Each log line is prefixed with an indicator and the source name inside square brackets, so interleaved output from different sources is easy to scan:

```
Showing logs from functions and edge functions for the last 1h:

  𝒇   Function
  🌐  Edge function

[𝒇 scheduled] 2026-04-16T08:10:30.880Z INFO Tic tac toe {"next_run":"..."}
[𝒇 scheduled] 2026-04-16T08:10:31.496Z INFO Duration: 54.42 ms
[🌐 auth] 2026-04-16T08:11:02.331Z INFO Token validated
```

Each source gets a deterministic color from a 10-color palette (hash-based, consistent across sessions). The legend is printed at the top when there are results to show. In `--json` mode, no legend is printed.

## JSON Lines output

The `--json` flag outputs one JSON object per line, suitable for piping to `jq`, logging infrastructure, or programmatic consumption:

```bash
netlify logs --json --since 1h
```

```json
{"source":"function","name":"checkout","timestamp":"2026-04-15T17:53:57.004Z","level":"info","message":"Order 4312 placed"}
{"source":"edge-function","name":"auth","timestamp":"2026-04-15T17:53:58.221Z","level":"info","message":"Token validated"}
{"source":"deploy","name":"deploy","timestamp":"2026-04-15T17:53:59.000Z","level":"info","message":"Building site...","section":"building"}
```

The `source` field is always first. Levels are lowercase. Deploy entries include a `section` field when available.

## Debug diagnostics

The `--debug` flag prints HTTP request URLs, response status codes, and timing to stderr — useful for troubleshooting empty results or slow queries:

```bash
netlify logs --debug --since 1h
```

```
[debug] → GET https://analytics.services.netlify.com/v2/sites/.../function_logs/scheduled?from=...&to=...
[debug] ← 200 OK (142ms)
```

## Migration from old commands

The old `logs:function`, `logs:deploy`, `logs:build`, `logs:functions`, and `logs:edge-functions` commands are hidden from help and show a migration error with a tailored suggested command:

```
`netlify logs:function` has been replaced by a more comprehensive `netlify logs` command.

To get logs for this function from the last 10 minutes, run:

  netlify logs --source functions --function my-func --since 10m

Run `netlify logs --help` to see all available options.
```

The suggested command captures any function names the user passed to the old command and translates them to `--function` flags.

## Full flag reference

| Flag | Type | Description |
|---|---|---|
| `-s, --source <type>` | repeatable enum: `functions`, `edge-functions`, `deploy` | Log sources to include. Defaults to `functions` + `edge-functions` |
| `--function <name>` | repeatable string | Filter to specific functions by name |
| `--edge-function <name>` | repeatable string | Filter to specific edge functions by name or path |
| `--since <time>` | duration or ISO 8601 | Start of the historical window. Defaults to `10m` |
| `--until <time>` | duration or ISO 8601 | End of the historical window. Defaults to now |
| `-f, --follow` | boolean | Stream in real time (mutually exclusive with `--since`/`--until`) |
| `-u, --url <url>` | string | Scope to a specific deploy |
| `-l, --level <levels>` | repeatable string | Filter by log level |
| `--json` | boolean | Output as JSON Lines |
| `--debug` | boolean | Print HTTP diagnostics to stderr |


DavidWells:
This looks great!

One other thing we should fix while we are at it: I don't think the default behavior of the logs method should be to tail them tho. This will trip up agents and get them stuck

Instead, I propose an explicit tail flag, like pretty much every other cloud cli, `--tail` or `-t` for short

So the logs would come back with the windows defined and output to standard out as the default, not get stuck in a tail loop unless otherwise specified by the user.

eduardoboucas:
> One other thing we should fix while we are at it: I don't think the default behavior of the logs method should be to tail them tho. This will trip up agents and get them stuck
> 
> 
> 
> Instead, I propose an explicit tail flag, like pretty much every other cloud cli, `--tail` or `-t` for short
> 
> 
> 
> So the logs would come back with the windows defined and output to standard out as the default, not get stuck in a tail loop unless otherwise specified by the user. 

100% agreed, and the only reason I haven't done that as part of this PR is that it would be a breaking change and we'd need to cut a major version. 

Nothing against it, but I'd rather do that in a follow-up and see if there are any other breaking changes we'd like to bundle together.

DavidWells:
1 other note. Showing the multiple functions at once is a fantastic idea and will help tremendously with debugging. But if everything is the same color, It can be really hard to parse through the logs themselves. 

So a way that works really well is something that `debug` https://www.npmjs.com/package/debug does, where it will automatically use the name prefix to generate a color so you can easily visually parse which function the log is coming from. 

image


https://npmx.dev/package-code/debug/v/4.4.3/src%2Fnode.js#L35

DavidWells:
> breaking change and we'd need to cut a major version. Nothing against it, but I'd rather do that in a follow-up and see if there are any other breaking changes we'd like to bundle together.

Lets rip off the bandaid and major version this beauty 🚀

ndhoule:
Agreed that we can make no-follow the default in a follow-up PR. Agreed that the breaking change is totally OK.

> Instead, I propose an explicit tail flag, like pretty much every other cloud cli, --tail or -t for short

Let's go with `-f` / `--follow` in that PR. "tail" has a different meaning.

In this PR: What do you think about making explicit that `netlify logs` is a porcelain command in its docstring? If someone out there is parsing its output (without any other option--this command doesn't have a `--json` flag like a lot of others), we're breaking the format for them. Totally OK with breaking it, but I think it would be kind to users to throw a warning in the help text.

docs/commands/logs.md
```diff
@@ -65,21 +65,29 @@ netlify logs:function
...
 **Arguments**
 
-- functionName - Name of the function to stream logs for
+- functionNames - Names of the functions to stream logs for
 
 **Flags**
 
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `level` (*string*) - Log levels to stream. Choices are: trace, debug, info, warn, error, fatal
+- `since` (*string*) - Start of the historical log window. Accepts a duration (e.g. 10m, 1h, 24h, 2d) or an ISO 8601 timestamp
```
aitchiss: [total nitpick] aren't `from` and `to` more standard option names than `since`/`until`?

aitchiss (dismissed):

eduardoboucas:
If we're okay with breaking changes, let's go even further and use a new top-level `logs` command that combines logs from deploys, functions and edge functions (new).

I've also added `--json`, the colour-coding that @DavidWells mentioned and the `--follow` flag that @ndhoule proposed.

I've updated the PR and the description accordingly.

serhalp:
@eduardoboucas I have a couple breaking changes I've been holding off on. please poke me when this one's close to merging!

ndhoule (dismissed):
Looks great! Frankly, I think this is a better UX for _human_ consumers, much less agents.

A few suggestions that should **not** block this PR:

Should we modify the log retrieval endpoints to accept these filters so we don't have to implement filtration and log coalescing/sorting in the client? This command can and will consume a lot of memory when dealing with busy functions, and time-to-interactivity will grow as the number of loglines to retrieve grows (right?) I don't really know how log data is stored, though, so API changes might not be a reasonable suggestion.

I'd suggest we refactor all the log-yielding functions (e.g. `fetchHistoricalLogs`, `streamFunctions`) to return async iterables. It'd let us consolidate a lot of historical and streaming code paths--these functions would instead stop yielding at the end of a log stream, or yield indefinitely in the streaming case. In the future, if we were to update the log endpoints to return consolidated/sorted data, it'd also let us begin streaming logs immediately as we cursor.

VaibhavAcharya (approved):

PR #8171, opened 2026-04-17
fix(deps): unpin all dependencies

serhalp:
#### Summary

We were pinning deps to partially mimic the published npm-shrinkwrap.json for package managers like yarn and pnpm that don't support it.

Since we stopped shipping the shrinkwrap in #8163, the pinning is no longer relevant.

This will help users' dependency trees get deduped much further.


XhmikosR:
This would need to happen in the other netlify packages too. It's clear that the duplicate packages increase by a lot...

But assuming you guys really want to reduce your deps tree, this is a must change. Next, you'd need to start finding other places that deps need deduplication. v25.0.0 has ~100 duplicate packages. With this PR the number will increase, as it seems. Remember, updating deps blindly to the latest version, is not the right approach all the time, if one cares about the deps tree/count. Unless you can do it across all your dependencies, which isn't realistic...

* https://npmgraph.js.org/?q=netlify-cli@24.11.3
* https://npmgraph.js.org/?q=netlify-cli@25.0.0

Just my 2 cents as a netlify-cli user who has reported [the issue](#3941) for years :)

That is exactly the reason why I started contributing to dependents (precinct dependency, etc) packages in the first place.

* https://npmgraph.js.org/?q=precinct@8.3.1
* https://npmgraph.js.org/?q=precinct
* https://arve0.github.io/npm-download-size/#precinct@8.3.1
* https://arve0.github.io/npm-download-size/#precinct

G-Rath (approved):
❤️

serhalp:
> ## 📊 Benchmark results
> 
> _Comparing with [07fa4d1](https://github.com/netlify/cli/commit/07fa4d1b72ba07aa53e75d82aaffa62b42e70286)_
> 
>     * **Dependency count**: 1,134 ⬆️ **6.44% increase** vs. [07fa4d1](https://github.com/netlify/cli/commit/07fa4d1b72ba07aa53e75d82aaffa62b42e70286)
> 
>     * **Package size**: 378 MB ⬆️ **5.64% increase** vs. [07fa4d1](https://github.com/netlify/cli/commit/07fa4d1b72ba07aa53e75d82aaffa62b42e70286)

For the record, this is somewhat misleading. Actual users will see a decrease in count and size. We're seeing an increase due to our own package-lock and the fact that I tried to keep the scope of change in this PR limited (we can follow up to do a deep upgrade of our dep tree in this repo).

VaibhavAcharya (approved):

PR #8177, opened 2026-04-21
feat: add `db migrations reset` command

eduardoboucas:


src/commands/database/db-migrations-reset.ts
```diff
@@ -0,0 +1,157 @@
...
+  const json = options.json ?? false
+
+  if (branch) {
+    if (branch === PRODUCTION_BRANCH) {
+      throw new Error(
+        `Refusing to target the production branch. ${chalk.bold(
+          'db migrations reset',
+        )} against a remote branch is only for preview branches.`,
+      )
+    }
```
pieh: Just thinking out loud here about this:

In the case when users run into out-of-order migrations on their production branch - I could see this happening with git-based repos that don't enforce PR branches to be up to date with main branch before merging for example.

You'd probably want to create pull request against main with the migration fixes - but this check here would prevent using this command.

The command here is not mutating a database - it is managing not applied migration files - so I'm not sure if fully forbidding running this against production branch is actually desirable?

pieh: I would not say this is a blocker, but just something to think about here?

eduardoboucas: That's a great callout. This was a leftover from a different shape of the command I was toying around with.

Removed in https://github.com/netlify/cli/pull/8177/commits/3f759b1722c04933ec495dab5a41140d6bad6895.

src/commands/database/database.ts
```diff
@@ -175,5 +176,16 @@ export const createDatabaseCommand = (program: BaseCommand) => {
...
+    migrationsCommand
+      .command('reset')
+      .description('Delete local migration files that have not been applied yet')
+      .option('-b, --branch <branch>', 'Target a remote preview branch instead of the local development database')
+      .option('--json', 'Output result as JSON')
+      .action(async (options: MigrationsResetOptions, command: BaseCommand) => {
+        const { migrationsReset } = await import('./db-migrations-reset.js')
+        await migrationsReset(options, command)
+      })
+      .addExamples(['netlify db migrations reset', 'netlify db migrations reset --branch my-feature-branch'])
```
pieh, line 189:
```suggestion
    migrationsCommand
      .command('reset')
      .description('Delete local migration files that have not been applied yet')
      .option('-b, --branch <branch>', 'Target a remote branch instead of the local development database')
      .option('--json', 'Output result as JSON')
      .action(async (options: MigrationsResetOptions, command: BaseCommand) => {
        const { migrationsReset } = await import('./db-migrations-reset.js')
        await migrationsReset(options, command)
      })
      .addExamples(['netlify db migrations reset', 'netlify db migrations reset --branch my-feature-branch', 'netlify db migrations reset --branch production'])
```

I don't exactly suggest this (I'm struggling on exact wording) - but just to sync with the removal of the production branch check removal.

Can be done in follow up

pieh: looking at command like `migrations pull` - it also seem like you can do just `--branch` there (and it would use `production`) and maybe some wording for handling those cases can be borrowed here

pieh (approved):

PR #8189, opened 2026-04-22
fix(deps): replace lodash with native utilities

jherr:
## Summary

Removes `lodash` from the CLI's runtime dependencies in favor of tiny native implementations in `src/utils/object-utilities.ts`.

- Replaces `lodash/isEmpty`, `lodash/pick`, `lodash/merge`, `lodash/throttle`, and `lodash/isObject` across 8 call sites.
- Drops `lodash` and `@types/lodash` from `package.json`.
- Adds JSDoc header blocks to each new helper documenting its intentionally narrow contract (leading-edge-only throttle, merge semantics that overwrite arrays and don't mutate inputs, etc.).
- Adds unit tests at `tests/unit/utils/object-utilities.test.ts` covering nullish handling, falsy-value preservation, non-mutation, recursive merge, array overwrite, and throttle timing behavior (26 tests).

Part of the Milestone 1 dependency-pruning work from the CLI modernization RFC. `node-fetch` and `backoff` are deliberately left for follow-up PRs to keep this one reviewable.

## Test plan

- [x] `npm run build`
- [x] `npx tsc --project tsconfig.build.json --noEmit`
- [x] `npm run test:unit` — 344 passed; the 5 failures (`exec-fetcher`, `hash-fns`) are pre-existing on `main` and unrelated (flaky CDN download in traffic-mesh fetch).
- [x] `npm run test:integration` — 545 passed; 5 flaky failures unrelated to this change (interactive completion prompts, Next.js/Hugo fixture tests, a dev plugin-hook test). None touch the lodash replacement call sites.
- [ ] CI (unit + integration + e2e) green

Made with [Cursor](https://cursor.com)

PR #8195, opened 2026-04-22
refactor: drop backoff dep and convert remaining require() calls to ESM

jherr:
## Summary

Continues the Milestone 1 dependency-pruning work from #8189 by finishing items **B** and **C** from the follow-up list:

- **Drop `backoff` / `@types/backoff`.** Its only runtime use was in `src/utils/deploy/upload-files.ts`; replaced with a small inline Fibonacci retry helper that preserves the existing delay sequence (`5s, 5s, 10s, 15s, 25s, …` capped at `UPLOAD_MAX_DELAY`), the `0.5` jitter factor, and the `failAfter(maxRetry)` "6 attempts when maxRetry=5" semantics. `tests/unit/utils/gh-auth.test.ts` migrated off `backoff` to a plain polling loop.
- **Convert the 3 remaining `require()` calls to ESM.**
  - `src/commands/functions/functions-create.ts` (x2): `require(packageJson)` → `JSON.parse(fs.readFileSync(...))` via a small `readJsonFile` helper.
  - `src/commands/database/legacy/utils.ts`: `createRequire(join(dir, 'package.json'))('./package.json')` → `JSON.parse(fs.readFileSync(join(dir, 'package.json'), 'utf8'))`.
  - `src/commands/functions/functions-invoke.ts`: `processPayloadFromFlag` becomes async; `.json` payloads use `readFile` + `JSON.parse`, other files go through dynamic `import()` (via `pathToFileURL`) with `imported.default ?? imported` to preserve the CJS interop the old `require()` provided.
- **New unit tests** to lock in the changed behavior:
  - `tests/unit/commands/database/legacy/utils.test.ts` (8 cases) — exercises `getPackageJSON` happy path plus each of the shape-validation errors.
  - `tests/unit/commands/functions/functions-invoke.test.ts` (7 cases) — exercises `processPayloadFromFlag` with inline JSON strings, `.json` / `.mjs` / `.cjs` file payloads, missing paths, and load-time errors (`processPayloadFromFlag` is now exported for testability).

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm exec vitest -- run tests/unit/utils/deploy/upload-files.test.ts` (2/2 — retry-count-on-retries and no-retry-on-400 still pass against the inline Fibonacci implementation)
- [x] `npm exec vitest -- run tests/unit/utils/gh-auth.test.ts` (1/1)
- [x] `npm exec vitest -- run tests/unit/commands/database/legacy/utils.test.ts` (8/8, new)
- [x] `npm exec vitest -- run tests/unit/commands/functions/functions-invoke.test.ts` (7/7, new)
- [x] `npm run test:unit` — 337 passed. The 5 failures are pre-existing on `main` (`copy-template-dir.test.ts` × 4 hit a Node 25 `fs.rmdirSync` deprecation; `generate-autocompletion.test.ts` × 1 has a stale snapshot).
- [x] `npm exec vitest -- run tests/integration/commands/functions-invoke/functions-invoke.test.ts` (5/5 — validates the now-async `processPayloadFromFlag` path end-to-end)

Made with [Cursor](https://cursor.com)


jaredm563:
overall LGTM, and overall this PR is an improvement but there are a couple of regressions we should fix 

1. functions:invoke payload handling is different
Old code used require(payloadpath) (Node resolver supports extensionless and directory resolution).
New code gates on fs.existsSync(payloadpath) and imports exact path only.

so existsSync('/tmp/.../payload') => false
require('/tmp/.../payload') => { ok: true } only when payload.js exists.

2. BOM JSON parsing has small regression. We used to accept BOM JSON but now looks like we don't
require(file) succeeds
JSON.parse(readFileSync(file,'utf8')) fails with Unexpected token.

WDYT

jherr:
Thanks for the careful read — both regressions confirmed and fixed in 232ecc2a:

**1. Extensionless / directory path resolution.** `processPayloadFromFlag` now mirrors Node's `require()` resolver: an exact path wins; otherwise we try `.js`, `.cjs`, `.mjs`, `.json` in that order; if the path is a directory we read `package.json#main` and then fall back to `index.*`. New unit tests cover extensionless and directory resolution.

**2. BOM JSON parsing.** Both `processPayloadFromFlag` (`.json` payloads) and the `readJsonFile` helper in `functions-create.ts` now strip a leading U+FEFF before `JSON.parse`. Added a unit test that writes a BOM-prefixed payload and asserts it parses.

Also merged main into the branch (the integration test failures from the prior CI run were unrelated infra noise — `unix-dgram` native build failure on Node 24 and an `EEXIST: mkdir 'new_functions'` in `dev/functions.test.ts`).

jaredm563 (approved):

PR #8200, opened 2026-04-23
feat: various improvements to `database` commands

eduardoboucas:
## https://github.com/netlify/cli/pull/8200/changes/fd77a178ce52866150cb4a1521fe54475b830539

Reworks the human-friendly output of `netlify database status ` into a single _Migrations_ section (replacing the previous separate "Applied migrations" and "Migrations not applied" sections).

Also fixed the apply-guidance: the `netlify database migrations apply` hint only appears when truly targeting the local dev DB, and switches to "Deploy these files to apply the migrations." when using a remote database.

The `--json` output is now affected.

## https://github.com/netlify/cli/pull/8200/changes/60c162d180476afdce4c4bb3cf39a47696d55e7f

Fixed `netlify database migrations pull` to match the server's new split API. The list endpoint (`/database/migrations`) is now metadata-only and a new per-migration endpoint (`/database/migrations/{name}`) returns the SQL content, so the command now makes a list call followed by parallel content fetches. 

## https://github.com/netlify/cli/pull/8200/changes/9bb834f67c9112247b453289f590e98c1c6d182c

Added a shared `readApiErrorMessage` helper used by every CLI call site that hits `/database/*`, which extracts the server's message field from JSON error bodies so users see clean messages instead of raw JSON.

### DIFF
--- a/src/commands/database/db-status.ts
+++ b/src/commands/database/db-status.ts
@@ -1,5 +1,5 @@
 import { readFile, readdir } from 'fs/promises'
-import { join, relative } from 'path'
+import { join, relative, sep } from 'path'
 
 import { chalk, log, logJson, netlifyCommand } from '../../utils/command-helpers.js'
 import BaseCommand from '../base-command.js'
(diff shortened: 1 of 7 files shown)


src/commands/database/db-status.ts
```diff
@@ -357,6 +351,35 @@ const renderPretty = (params: RenderParams) => {
...
+    ),
+  )
+  log(renderList(status.pending, INDENT))
+  if (status.pending.length > 0 && status.outOfOrder.length === 0) {
+    const canApplyLocally = isLocal && !hasUrlOverride
+    log('')
+    if (canApplyLocally) {
+      log(chalk.gray(`${INDENT}Run ${formatCommand('database migrations apply')} to apply them to the local database.`))
+    } else {
+      log(chalk.gray(`${INDENT}Deploy these files to apply the migrations.`))
```
pieh, line 380: NIT: the wording might be confusing and what is meant by "deploy these files"

Not great but maybe:
```suggestion
      log(chalk.gray(`${INDENT}Deploy your site to apply the migrations.`))
```

Also feel free to skip. 

In any case - if any changes were applied, tests that look for this string would need to be updated

pieh (approved):

PR #8207, opened 2026-04-24
fix: prevent path traversal in functions:create command [security]

NiteshCodes7:
## Description
There is a path traversal vulnerability in `functions:create`. The interactive 
prompt correctly validates function names with `/^[\w.-]+$/i`, but this validation 
is completely bypassed when the name is supplied via the `--name` flag or as a 
positional argument.

## Affected File
`src/commands/functions/functions-create.ts`

## Reproduction Steps
Run either of the following:

netlify functions:create --name "../../evil"
netlify functions:create "../../evil"

This causes files to be written outside the intended functions directory.

## Root Cause
In `getNameFromArgs`, validation only runs on the interactive prompt:

const { name } = await inquirer.prompt([
  {
    validate: (val) => Boolean(val) && /^[\w.-]+$/i.test(val), // ✅ validated
  },
])

But when --name or a positional arg is passed, the raw unvalidated value goes 
directly into:

const functionPath = path.join(functionsDir, name) // ❌ no validation

## Impact
An attacker or a malicious script can write files to arbitrary locations on the 
filesystem outside the functions directory.

## Proposed Fix
Add a containment check in `ensureFunctionPathIsOk`:

const ensureFunctionPathIsOk = function (functionsDir, name) {
  const functionPath = path.join(functionsDir, name)

  if (!functionPath.startsWith(path.resolve(functionsDir) + path.sep)) {
    log(`Invalid function name: "${name}" resolves outside the functions directory.`)
    process.exit(1)
  }

  if (fs.existsSync(functionPath)) {
    log(`Function ${functionPath} already exists, cancelling...`)
    process.exit(1)
  }

  return functionPath
}

## Additional Note
A second related issue exists in `downloadFromURL` where filenames from the 
GitHub API response are not sanitized before being passed to `path.join`. 
Fix: wrap `name` with `path.basename(name)` before use.


jaredm563:
Thanks for tackling this @NiteshCodes7 The ensureFunctionPathIsOk guard is a good improvement for template scaffolding!

I think there’s still a path traversal gap in the --url flow:

getNameFromArgs() returns --name / positional arg without validation.
downloadFromURL() uses that value directly in path.join(functionsDir, nameToUse).
The new ensureFunctionPathIsOk() check is not applied in this code path.
So with something like --url <repo-url> --name ../../outside-dir, the command can create/write outside the configured functions directory (and then run npm i there).

We should also add an integration test that verifies traversal is rejected for the URL flow (e.g. functions:create --url ... --name ../../evil exits with an error and does not create files outside the functions dir).

Also can we drop unrelated package-lock.json churn from this PR to keep the security fix focused and easier to review?

NiteshCodes7 (commented):
Updated! Here is what changed:

1. getNameFromArgs() now validates --name flag and positional 
   args with the same regex used by the interactive prompt.

2. downloadFromURL() now has an explicit path traversal check 
   before path.join() is called.

3. Added integration test covering the --url + --name ../../evil 
   flow — verifies non-zero exit and no files written outside 
   functions dir.

4. Removed package-lock.json from the PR.

NiteshCodes7 (commented):
Updated! Here is what changed:

1. getNameFromArgs() now validates --name flag and positional 
   args with the same regex used by the interactive prompt.

2. downloadFromURL() now has an explicit path traversal check 
   before path.join() is called.

3. Added integration test covering the --url + --name ../../evil 
   flow — verifies non-zero exit and no files written outside 
   functions dir.

4. Removed package-lock.json from the PR.

jaredm563 (approved):
@NiteshCodes7 Approved & fantastic work! Thank you for the contribution!

PR #8208, opened 2026-04-24
chore: stabilize integration test suite

jherr:
## Summary

- Add a deterministic `--shell` option for `completion:install` and use it in the zsh completion integration tests so they no longer drive the upstream tabtab shell picker interactively.
- Remove fixed dev-server target ports from flaky concurrent dev/redirects integration tests by allocating free ports per test fixture.
- Respect dynamically-set `process.env.CI` in error telemetry reporting so expected CLI failures in integration tests do not spawn telemetry subprocesses that Vitest reports as unhandled rejections.
- Update the stale `netlify help completion` snapshot to match the current option ordering.

## Test plan

- [x] `npm run test:init`
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm exec vitest -- run --retry=3 --reporter=default tests/integration/frameworks/hugo.test.ts tests/integration/commands/dev/redirects.test.ts tests/integration/commands/build/build.test.ts tests/integration/commands/completion/completion-install.test.ts tests/integration/commands/dev/dev.test.ts tests/integration/commands/help/help.test.ts`
- [x] `npm exec vitest -- run --retry=3 --reporter=default tests/integration/commands/sites/sites.test.ts`
- [x] `npm run test:integration` — 60 files passed, 554 passed / 3 skipped / 7 todo

Made with [Cursor](https://cursor.com)

PR #8209, opened 2026-04-24
chore: stabilize integration test suite

jherr:
## Summary

- Replace hardcoded ports (4567, 9999, 3000, 6123) with dynamic `getPort()` calls across integration tests to prevent port collisions during concurrent execution
- Suppress telemetry error-reporting subprocesses when `process.env.CI` is set (previously only checked the `is-ci` package)
- Use dynamic port substitution for the `next-app` fixture in redirect tests
- Fix test artifact leak in `copy-template-dir` unit tests by moving cleanup to `afterEach` with `force: true`
- Add `tests/unit/utils/tmp` to `.gitignore` as a safety net
- Add diagnostic instrumentation to `tests/integration/utils/dev-server.ts`:
  - `--trace-warnings --trace-uncaught --trace-exit --report-on-signal --report-signal=SIGUSR2` in the spawned CLI subprocess's `NODE_OPTIONS`
  - early rejection when the subprocess exits before the "Local dev server ready" banner (previously only error-exits were caught — clean exits hung until the start-timeout)
  - SIGUSR2 to the (presumed-hung) subprocess from the start-timeout fallback so Node emits a diagnostic report (stack traces + libuv handle table), then splice the report into the timeout error message
  - stderr now included in the timeout error message

This PR does **not** touch `package.json` / `package-lock.json`. The separate Node 24 `extract-zip` hang that this work uncovered is being shipped in its own follow-up PR — these test-stabilization changes are independent of that fix.

## Test plan

- [x] Full integration test suite passes on Node 20/22 (60/60 files, 554 tests, 3 consecutive clean runs)
- [x] `copy-template-dir` unit tests pass with no leaked artifacts
- [x] Snapshot tests unaffected by port changes
- [ ] Node 24 integration tests will be green once the companion `extract-zip` fix lands (currently 3 serve-mode tests fail on Node 24 because `extract-zip@2.0.1` hangs on Node 24's `promisify(util.pipeline)` — see the follow-up PR)


aitchiss:
@jherr does this PR replace https://github.com/netlify/cli/pull/8208, or are both needed? Also I notice there are some failing integration tests - I'm not sure if these are flake unrelated to your changes but can you have a look and let me know if this one is definitely ready for review?

serhalp:
@jherr why did the package-lock get completely rewritten here? could you check that file out from main and `npm i` to re-update?

Screenshot 2026-05-28 at 09 18 19

🙏🏼 could you also please split the zip dependency change into a separate PR? this seems quite unrelated

src/utils/telemetry/report-error.ts
```diff
@@ -22,7 +22,7 @@ const dirPath = dirname(fileURLToPath(import.meta.url))
  */
 // @ts-expect-error TS(7006) FIXME: Parameter 'error' implicitly has an 'any' type.
 export const reportError = async function (error, config = {}) {
-  if (isCI) {
+  if (isCI || process.env.CI) {
```
pieh: IIRC `isCI` already checks `CI` env var as part of its checks, so I am not sure if this actually does anything? Ref https://github.com/watson/ci-info/blob/c4e1d0565552fb20ea3c133db2e056a574e78e6b/test.js#L59-L65

serhalp:
```suggestion
```

src/utils/telemetry/report-error.ts
```diff
@@ -22,7 +22,6 @@ const dirPath = dirname(fileURLToPath(import.meta.url))
  */
 // @ts-expect-error TS(7006) FIXME: Parameter 'error' implicitly has an 'any' type.
 export const reportError = async function (error, config = {}) {
-  if (isCI) {
     return
```
serhalp, line 26:
```suggestion
  if (isCI) {
    return
```

serhalp (approved):

PR #8218, opened 2026-04-28
fix: don't error remote database status command if database was not enabled yet to provide more meaningful output

pieh:

#### Summary

---

      passes our tests.


kathmbeck (dismissed):

src/commands/database/db-status.ts
```diff
@@ -410,9 +410,13 @@ export const statusDb = async (options: DatabaseStatusOptions, command: BaseComm
       throw new Error(`You must be logged in with ${netlifyCommand()} login to target a remote branch.`)
     }
 
-    connectionString = await fetchBranchConnectionString({ siteId, accessToken, basePath }, branch)
-    fetchApplied = remoteAppliedMigrations({ siteId, accessToken, basePath, branch })
     branchLabel = branch
+    if (siteHasDatabase) {
```
pieh, line 414: This explicitly is not using `enabled` check, because when used with `@netlify/database-proxy` and `NETLIFY_DB_URL` (IIRC) it is reported as true even if database is not actually created yet, so `fetchBranchConnectionString` that is hitting an API would return failure.

The `enabled` logic _might_ need some more thinking about desired interaction with `@netlify/database-proxy`, but this seems a bit out of scope for this change.

aitchiss:
@pieh @kathmbeck  is this PR still being worked on?

PR #8232, opened 2026-05-04
fix: don't use inquirer for non-interactive variant of "database migrations new" command

pieh:

#### Summary

Main goal here is to avoid using `inquirer` in non-interactive mode.

I did smuggle in here preference to using `timestamp` numbering schema over sequential as it is less problematic when working on parallel changes. We still will honor sequential and default to it if this is a scheme already used by a project. This part is optimistic and can walk it back if deemed undesirable.

Currently non-interactive works only when both `--scheme` and `--description` is provided:
```bash
> netlify database migrations new --description "test" --scheme timestamp  < /dev/null
Created migration: 20260504084016_test
  /Users/misiek/test/pgs-test-init/netlify/database/migrations/20260504084016_test/migration.sql
```

But if they are not - there is non-helpful error happening:
```bash
> netlify database migrations new --description "test" < /dev/null
? Numbering scheme: (Use arrow keys)
❯ Sequential (0001, 0002, ...) 
  Timestamp (20260312143000) Warning: Detected unsettled top-level await at file:///Users/misiek/.nvm/versions/node/v24.0.0/lib/node_modules/netlify-cli/bin/run.js:87
  await main()
  ^

> netlify database migrations new --scheme timestamp < /dev/null
? What is the purpose of this migration? Warning: Detected unsettled top-level await at file:///Users/misiek/.nvm/versions/node/v24.0.0/lib/node_modules/netlify-cli/bin/run.js:87
  await main()
```

With changes in this PR:
- we get more meaningful error for missing `--description`:
```bash
> ntl_local database migrations new --scheme timestamp < /dev/null
 ›   Error: --description <description> argument is required when not running interactively. Provide a description of the migration (e.g. --description "add users table").
```
- we default to a scheme that is already used in a project (or default to timestamp if no migrations exist yet)
```bash
# if no migrations or timestamps used before
> ntl_local database migrations new --description "test"  < /dev/null
Created migration: 20260504084320_test
  /Users/misiek/test/pgs-test-init/netlify/database/migrations/20260504084320_test/migration.sql

# if sequential migrations used before
> ntl_local database migrations new --description "test2"  < /dev/null
Created migration: 0002_test2
  /Users/misiek/test/pgs-test-init/netlify/database/migrations/0002_test2/migration.sql
```

---

      passes our tests.

### DIFF
--- a/src/commands/database/db-migration-new.ts
+++ b/src/commands/database/db-migration-new.ts
@@ -117,8 +117,8 @@ export const migrationNew = async (options: MigrationNewOptions, command: BaseCo
           name: 'scheme',
           message: 'Numbering scheme:',
           choices: [
-            { name: 'Timestamp (20260312143000)', value: 'timestamp' },
-            { name: 'Sequential (0001, 0002, ...)', value: 'sequential' },
+            { name: 'Timestamp (e.g. 20260312143000) [Recommended]', value: 'timestamp' },
+            { name: 'Sequential (e.g. 0001, 0002, ...)', value: 'sequential' },
           ],
           default: defaultScheme,
         },
--- a/src/commands/database/db-migration-new.ts
+++ b/src/commands/database/db-migration-new.ts
@@ -7,6 +7,7 @@ import { log, logJson } from '../../utils/command-helpers.js'
 import BaseCommand from '../base-command.js'
 import { resolveMigrationsDirectory } from './util/migrations-path.js'
 import { utcTimestampPrefix } from './util/timestamp.js'
+import { isInteractive } from '../../utils/scripted-commands.js'
 
 export type NumberingScheme = 'sequential' | 'timestamp'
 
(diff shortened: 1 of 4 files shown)


src/commands/database/db-migration-new.ts
```diff
@@ -89,31 +90,43 @@ export const migrationNew = async (options: MigrationNewOptions, command: BaseCo
...
+
+    if (isInteractive()) {
+      const answers = await inquirer.prompt<{ scheme: NumberingScheme }>([
+        {
+          type: 'list',
+          name: 'scheme',
+          message: 'Numbering scheme:',
+          choices: [
+            { name: 'Timestamp (20260312143000)', value: 'timestamp' },
+            { name: 'Sequential (0001, 0002, ...)', value: 'sequential' },
```
eduardoboucas: I wonder if something like this would help?

```suggestion
            { name: 'Timestamp (e.g. 20260312143000) [Recommended]', value: 'timestamp' },
            { name: 'Sequential (e.g. 0001, 0002, ...)', value: 'sequential' },
```

eduardoboucas (dismissed):

eduardoboucas (approved):

PR #8234, opened 2026-05-06
fix(dev): execute dev command with shell operators (`&&`, `||`, etc.)

### LINKED ISSUE #8228: `dev` command with `&&` incorrectly parsed
### Describe the bug

In `netlify.toml`, when I write this `dev` command

```
[dev]
  command = "bundle install && bundle exec jekyll serve --port 4002 --incremental --future"
```

the shell command `netlify dev` responds with

```
⬥ Starting framework dev server
Unknown switches "--port", "--incremental", "--future"
Did you mean?  "--path"
⠹ Waiting for framework dev server to be ready on port 4002⬥ Command failed with exit code 15: bundle install && bundle exec jekyll serve --port 4002 --incremental --future. Shutting down Netlify Dev server
```

It appears that `netlify dev` parses the string `&&` as if it were an argument to the `bundle install` command, rather than as the `bash` operator.

When I use `netlify build`, this problem doesn't occur.

### Steps to reproduce

1. In `netlify.toml`, specify a `[dev]` command that connects two `bash` commands with `&&`.
2. Run `netlify dev`.

The command fails.

This behavior is the same when I try to use the `--command` switch.

When the `[build]` command has `&&` in it, `netlify build` behaves as expected.

### Configuration

Sample contents of `netlify.toml`:

```
[build]
  command = "bundle install && bundle exec jekyll build --future"
  publish = "www"

[dev]
  command = "bundle install && bundle exec jekyll serve --port 4002 --incremental --future"
  targetPort = 4002
```

### Environment

- Pop!_OS 24.04, Linux 6.18.7-76061807-generic #202601231045~1769703228~24.04~cb87b5b SMP PREEMPT_DYNAMIC Thu J x86_64 x86_64 x86_64 GNU/Linux
- zsh 5.9
- netlify-cli/26.0.0 linux-x64 node-v20.20.2

puneetdixit200:
## Summary
Fixes #8228 by running `dev.command` through a shell only when the command string needs shell parsing, such as shell operators (`&&`, `||`, pipes, redirects, semicolons) or POSIX-style inline environment assignments.

## Changes
- Added shell-mode detection for compound commands and inline env assignments.
- Kept simple commands on the direct `execa.command()` path so existing command execution behavior stays unchanged.
- Improved missing-command detection from command output and avoided misleading missing-command hints for compound command failures.
- Added integration regression tests for shell operators and generic failure messages.

## Why
`execa.command()` does not interpret shell operators by default. Without shell parsing, a command like:

```toml
[dev]
command = "bundle install && bundle exec jekyll serve --port 4002 --incremental --future"
```

is parsed as a single command with literal `&&` args, causing the observed failure.

## Validation
- `npm run build`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'shell operators|generic command failure'`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should run and serve a production build when using the \`serve\` command'`
- `npm exec vitest -- run tests/integration/commands/dev/functions.test.ts -t 'nodeModuleFormat: esm v1 functions should work'`


src/utils/shell.ts
```diff
@@ -111,8 +138,12 @@ export const runCommand = (
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
   commandProcess.then(async () => {
     const result = await commandProcess
-    const [commandWithoutArgs] = command.split(' ')
-    if (result.failed && isNonExistingCommandError({ command: commandWithoutArgs, error: result })) {
+    const commandWithoutArgs = getCommandName(command)
+    if (
+      result.failed &&
+      canReportMissingCommandName(command) &&
+      isNonExistingCommandError({ command: commandWithoutArgs, error: result })
```
jaredm563, line 162: Thanks for the contribution here! 

In this lineisNonExistingCommandError() now treats exitCode === 127 as a missing-command signal.

IMO, that's a little too broad: an exit code 127 can also come from commands that exist but fail for other reasons (or scripts that intentionally return 127). In those cases, we are going to show the user:

``Please verify '<command>' exists``

even though the command _does_ exist.

So, this effectively reintroduces misleading diagnostics which is what your PR is trying to fix

jaredm563: So, here's what I think you should do:

1. Figure out a way to tighten up the missing command detection
You could keep `ENOENT` and the explicit platform message checks ("not found" slash the Windows "not recognized"), but avoid using exitCode === 127 by itself.

2.Add an integration test for a command that exists but exits 127 (like for example, non Windows: ``bash -c 'exit 127'``) and assert we show the generic command failure path, basically "we stopped because of a 127 vs “Please verify 'bash' exists”.

WDYT?  @puneetdixit200 

Thanks again for the contribution! and please let me know if anything is unclear (:

puneetdixit200:
Rebased this branch onto the latest `main` and rechecked the command-failure coverage.

Verification:
- `npm run build`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'shell operators|compound command|env-assigned command|existing command exits with 127'`
- `npm run lint -- src/utils/shell.ts tests/integration/framework-detection.test.ts`
- `npx prettier --check src/utils/shell.ts tests/integration/framework-detection.test.ts`
- `git diff --check upstream/main...HEAD`

src/utils/shell.ts
puneetdixit200: Thanks, this is covered on the current head `4f17001b`.

What changed:
- `exitCode === 127` is no longer enough by itself to show the missing-command hint.
- missing-command reporting now requires a missing-command message that names the command.
- compound commands and env-assigned commands use the generic command-failure path.
- added the requested integration coverage for an existing command that exits 127: `bash -c "exit 127"`.

Verification:
- `npm ci --cache /tmp/netlify-cli-8234-npm-cache --no-audit`
- `npm run build`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t "shell operators|compound command|env-assigned command|existing command exits with 127"`
- `npm run lint -- src/utils/shell.ts tests/integration/framework-detection.test.ts`
- `npm run format:check -- src/utils/shell.ts tests/integration/framework-detection.test.ts`
- `git diff --check`

AI assistance was used for this recheck; I reviewed the current diff and validation output.

puneetdixit200:
Pushed `024d61d` for the integration failures on the previous head. The shell fallback is now limited to commands that need shell parsing, like compound commands and POSIX env assignments, so ordinary long-running framework commands use the previous direct-spawn path.

Verified locally:
- `npm run build`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should support shell operators in `command`|should use generic command failure when a compound command fails|should use generic command failure when an env-assigned command fails|should use generic command failure when an existing command exits with 127|should print specific error when command doesn.t exist' --testTimeout=90000`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should run `command` when both `command` and `targetPort` are configured|should run and serve a production build when using the `serve` command' --testTimeout=90000`
- `npm exec vitest -- run tests/integration/commands/dev/functions.test.ts -t 'nodeModuleFormat: esm v1 functions should work' --testTimeout=90000`
- `npx eslint src/utils/shell.ts tests/integration/framework-detection.test.ts --no-cache`
- `npx prettier --check src/utils/shell.ts tests/integration/framework-detection.test.ts`
- `git diff --check`

puneetdixit200:
Checked the latest Node 24 integration failures on this branch.

What I verified locally with Node 24.15 after rebuilding:
- `npm run build` passes.
- The PR-specific shell command tests pass: `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'shell operators|generic command failure'`.
- The CI-timeout framework detection test passes: `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should run and serve a production build when using the \`serve\` command'`.
- The CI-timeout functions test passes: `npm exec vitest -- run tests/integration/commands/dev/functions.test.ts -t 'nodeModuleFormat: esm v1 functions should work'`.

The failing `tests/integration/commands/dev/serve.test.ts` case also fails on current upstream `main` after rebuilding, with `/foo.txt` returning `Not Found` instead of `foo`, so that one does not appear to be caused by this branch.

puneetdixit200:
Follow-up pushed in `0f83ddd`.

What changed:
- kept the integration test that verifies shell operators execute through `command`
- moved the generic failure-classification coverage for compound commands, env assignments, package managers, and existing commands exiting 127 into focused unit tests for `src/utils/shell.ts`
- removed the extra dev-server integration cases that were duplicating helper-level behavior and adding more long-running server processes to the shard

Local verification:
- `npm ci --cache /private/tmp/netlify-cli-8234-npm-cache`
- `npm run build`
- `npm exec vitest -- run tests/unit/utils/shell.test.ts`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should support shell operators in \`command\`|should run \`command\` when both \`command\` and \`targetPort\` are configured|should print specific error when command doesn.t exist' --testTimeout=90000`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should run and serve a production build when using the \`serve\` command' --testTimeout=90000`
- `DEBUG_TESTS=true npm exec vitest -- run tests/integration/commands/dev/functions.test.ts -t 'nodeModuleFormat: esm v1 functions should work' --testTimeout=90000`
- `npm exec eslint src/utils/shell.ts tests/unit/utils/shell.test.ts tests/integration/framework-detection.test.ts --no-cache`
- `npx prettier --check src/utils/shell.ts tests/unit/utils/shell.test.ts tests/integration/framework-detection.test.ts`
- `git diff --check`

I also reproduced a CI-shaped shard locally earlier: the framework-detection timeout case passed there, while the shard hit an unrelated edge-functions timeout under load. This push reduces the branch-added integration load while preserving the runtime coverage.

puneetdixit200:
Pushed a small test-only follow-up (`0f83ddd`) to move the shell helper assertions out of integration coverage and into a fast unit test. The remaining integration coverage keeps the actual shell-operator `netlify dev` behavior covered without running the extra negative-path dev-server cases.

Local verification:

```text
npm exec vitest -- run tests/unit/utils/shell.test.ts
# 5 passed

npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should support shell operators in `command`'
# 1 passed

npm run build
npx eslint src/utils/shell.ts tests/integration/framework-detection.test.ts tests/unit/utils/shell.test.ts --no-cache
npx prettier --check src/utils/shell.ts tests/integration/framework-detection.test.ts tests/unit/utils/shell.test.ts
git diff --check
# all passed
```

After the push, GitHub currently shows only neutral/skipped provider rows plus CodeRabbit/deploy-preview for the new head; the old Node 24 integration failures are no longer attached to the latest commit.

puneetdixit200:
Checked the current Node 24 integration failures from run `26477394094` on head `0f83ddd`.

Local Node version: `v24.15.0`.

Verification on this branch:
- `npm ci --cache /private/tmp/netlify-cli-8234-npm-cache --no-audit`
- `npm run build`
- `npm exec vitest -- run tests/unit/utils/shell.test.ts` -> 5 passed
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should support shell operators in `command`|should run and serve a production build when using the `serve` command' --testTimeout=90000` -> 2 passed
- `DEBUG_TESTS=true npm exec vitest -- run tests/integration/commands/dev/functions.test.ts -t 'nodeModuleFormat: esm v1 functions should work' --testTimeout=90000` -> 1 passed
- `git diff --check`

I also reran the failing `tests/integration/commands/dev/serve.test.ts` case on this branch and on current upstream `main` (`8a07280b`). It fails the same way on both: `/foo.txt` returns `Not Found` instead of `foo`, so I did not push a branch change for that failure.

The remaining attached Node 24 failures look like CI/runtime flakiness or an upstream serve-test issue rather than a regression from this PR.

jaredm563:
Yeah, our test suite has some flakiness right now. We are in the middle of refactoring it but I will probably have to run it a couple times.

puneetdixit200:
Rechecked the current Node 24 integration failures on head `0f83ddd`.

What I found locally on Node 24.15 after a clean install and rebuild:
- `tests/integration/commands/dev/functions.test.ts -t "nodeModuleFormat: esm v1 functions should work"` passes.
- `tests/integration/framework-detection.test.ts -t "should run and serve a production build when using the \`serve\` command"` passes.
- `tests/integration/commands/dev/serve.test.ts -t "ntl serve should respect blobs, functions, and edge functions generated by integrations and with the Frameworks API"` fails on this branch with `/foo.txt` returning `Not Found`.
- The same `serve.test.ts` case fails the same way on current `upstream/main` (`8a07280b`), so that failure does not appear to be caused by this PR.

No branch code change from me on this pass.

jaredm563:
We have some issues with extract zip tests -> and need to fix our CI suite before I can merge this, will update you when that happens

Thanks again for the contribution! that fix should be in within the week and will update this PR. have a good week!

jaredm563:
Thanks for the updates! The test suite is fixed up so after we address these couple things we should be good to go!

1. Bug: Potential false missing-command hint
isMissingCommandMessage() can mark existing wrapper commands as missing. We should tighten this up to be a little more stronger than "Line contains a command name + not found". 

Example: bash -c "missing-subcommand" will trigger: "Failed running command ... Please verify 'bash' exists"

2. Test: Test flake risk from fixed port number
Your new test should support shell operators in command hardcodes targetPort: 3000. We actually moved tests to dynamic getPort() to avoid conflicts last weekend, so this one should do the same to match the new pattern before we merge it in.

puneetdixit200:
Updated in `1a217dda` after rebasing on current `main`.

What changed:
- tightened missing-command diagnostics so a wrapper command like `bash` is not reported as missing when only a subcommand fails with `command not found`
- changed the shell-operator integration test to use `getPort()` instead of fixed `3000`

Local verification:
- `npm ci --cache /private/tmp/netlify-cli-8234-npm-cache --no-audit`
- `npm run build`
- `npm exec vitest -- run tests/unit/utils/shell.test.ts`
- `npm exec vitest -- run tests/integration/framework-detection.test.ts -t 'should support shell operators in `command`' --testTimeout=90000`
- `npm exec eslint src/utils/shell.ts tests/unit/utils/shell.test.ts tests/integration/framework-detection.test.ts --no-cache`
- `npx prettier --check src/utils/shell.ts tests/unit/utils/shell.test.ts tests/integration/framework-detection.test.ts`
- `git diff --check`

jaredm563 (approved):
@puneetdixit200 Thank you so much for the fantastic work!!! If you would like to continue working on this codebase, a great follow up for this feature would be to support the standalone & operator. Thanks again and have a good day!

PR #8271, opened 2026-05-21
feat: add watch-ignore flag

### LINKED ISSUE #7763: Infinite repeating error message
### Describe the bug

On OSX, trying to use ntl to run my Eleventy site, specifcally:

```
netlify dev -c "eleventy --serve --quiet"
```

I get this error in console repeated infinitely, and a huge amount of processes fired.

Image

It isn't Eleventy as I can it just fine.

### Steps to reproduce

Not really sure - this worked fine in WSL for years. 

### Configuration

I just updated to the latest ntl CLI package today.

### Environment

```
  System:
    OS: macOS 26.0.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 6.05 GB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.0.0 - /Users/raymondcamden/.nvm/versions/node/v25.0.0/bin/node
    npm: 11.6.2 - /Users/raymondcamden/.nvm/versions/node/v25.0.0/bin/npm
  npmGlobalPackages:
    netlify-cli: 23.10.0
```

hrishikesh-k:
#### Summary

Fixes #7763 (https://answers.netlify.com/t/157384/)

TL;DR: The issue is caused because the user has too many files in their repo, causing the watchers in the CLI cause an EMFILE error on MacOS.

This PR does the following:
- fix ignoring `this.servePath` for Edge Functions. Chokidar [doesn't accept globs in v4](https://github.com/paulmillr/chokidar/releases/tag/4.0.0), so `/**` was incorrect - we now convert it to RegEx
- also ignore users' `publishDir` by default (only if it's not the same as `projectDir`) as it shouldn't be possible to import anything from the `publishDir` into Edge Functions (debatable though)
- add a new flag to the `dev` command: `--watch-ignore` to allow users to configure any additional ignore paths (just excluding `publishDir` wasn't enough in case of the above user as they had over 11k files in their `src` folder.
---

      passes our tests.


jaredm563:
@hrishikesh-k Its a bit weird scenario but I think we duplicated the docs dependency. 

i think you can do

npm install --prefix=site
npm run docs

Then we can drop those lockfile changes. NPM just has a hard time finding the right directory, so having it installed at the root package.json technically works but it should be in site/package.json

src/commands/serve/serve.ts
```diff
@@ -210,6 +210,7 @@ export const serve = async (options: OptionValues, command: BaseCommand) => {
     siteInfo,
     state,
     accountId,
+    watchIgnore: [],
```
jaredm563: Is hardcoding as [] intentional? The toml setting is going to be silently ignored no?

jaredm563: We should add this feature to docs/ in general tbh so it's discoverable for users/agents. WDYT?

hrishikesh-k: Sorry for that, and thanks for catching. This is now fixed. As for adding this to docs, the docs on cli.netlify.com don't have a TOML settings section so not sure if we can add it there. Or maybe you meant the main Netlify docs which include the TOML reference?

jaredm563: Ah that lives in the main netlify docs. Nvm in that case

jaredm563 (approved):
W

cfjedimaster:
As an FYI, I see this is merged and released. I'm running ntl version 26.2.0 and I keep an error:

```
Error: unknown option '--watch-ignore'
```

hrishikesh-k:
Sorry, I believe this got changed into a `netlify.toml` config:

```
[dev]
  watchIgnore = ["dir-1/", ...]
```

The reason being, you'd likely want this config to be committed to your repo.

cfjedimaster:
Hmm. Ok so I'm trying:

```
[dev]
  watchIgnore = ["src/posts","_site"]
```

And I get a crash right away with EMFILE too many open files. That's _better_ than before when it would loop error messages, but do I need to use * or some such to ensure it ignores everything under those 2 dirs?

As an FYI, `ulimit -n` returns 65K.

hrishikesh-k:
It's:

```
[dev]
  watchIgnore = ["_site", "src/_includes", "src/posts"]
```

cfjedimaster:
omg - i forgot about my giant comment folder and you did not - thank you ;) 

And it's working again - thank you SO much!

PR #8275, opened 2026-05-27
fix(deps): remove `uuid` dependency

gameroman:


serhalp (approved):
Woo, thank you friend!

PR #8280, opened 2026-06-01
feat(agents): typed API client and shared infra (1/8)

VaibhavAcharya:
#### Summary

Linear: https://linear.app/netlify/issue/EX-2195

Splits the agents CLI revamp (formerly one large PR) into a reviewable stack. This first PR is the foundation: the typed agents API and the shared types, constants, and utilities the rest of the command work builds on. No user-facing command behavior changes yet.

- The static model whitelist is gone. Models are fetched live from the AI gateway, so adding a model upstream no longer needs a CLI patch.
- API calls live in a typed module (`api.ts`) so each command file can focus on UX.
- Attachment upload handling (`attachments.ts`) with size and per-request limits, plus orphan cleanup on partial failure.

The existing `create`/`list`/`show`/`stop` commands keep working unchanged against the new infra; their behavior changes land in the next PR.

Stack (merge in order): **base** → existing commands → diff/open → pr/commit → sync/redeploy → publish/revert → rename/archive → follow-up.
Base branch: `main`.

---

      passes our tests.


VaibhavAcharya:
Link EX-2195

src/commands/agents/api.ts
```diff
@@ -0,0 +1,250 @@
+// TODO: Migrate to @netlify/api once these endpoints are public.
+// They are marked x-internal in bitballoon (#21736).
+// Method names mirror the bitballoon @operation_id to keep the future swap quick.
```
khendrikse: [nit] a smaller comment with less service specificity

```suggestion
// TODO: Migrate to @netlify/api once these endpoints are public.
// Method names mirror the backend @operation_id to keep the future swap quick.
```

src/commands/agents/utils.ts
```diff
@@ -44,15 +58,156 @@ export const validatePrompt = (input: string): boolean | string => {
...
+  sanitizePromptText(title).replace(CONTROL_CHAR_PATTERN, '').trim()
+
+export const validateRunnerTitle = (title: string): true | string => {
+  const sanitized = sanitizeRunnerTitle(title)
+  if (!sanitized) return 'A non-empty title is required'
+  if (sanitized.length > TITLE_MAX_LENGTH) return `Title must be ${TITLE_MAX_LENGTH.toString()} characters or fewer`
+  return true
+}
+
+export const validateAgent = (agent: string): true | string => {
```
khendrikse, line 78: I think with these utilities it would be good to have some tests that check their most important branches etc.

src/commands/agents/api.ts
```diff
@@ -0,0 +1,250 @@
...
+
+import type { NetlifyOptions } from '../types.js'
+import { parseLinkHeader } from './utils.js'
+import type {
```
khendrikse, line 6: @VaibhavAcharya I would have split this file up as well, start with 1 command and the methods we need for it, and the same for the types. This makes it easier to understand how they are being used in the following PR's that need to be reviewed.

khendrikse: Not necessary to do now btw! But for a next time :)

src/commands/agents/api.ts
```diff
@@ -0,0 +1,250 @@
...
+      await throwForStatus(response)
+    }
+    const body = await response.text()
+    const { total, hasNext } = readPagination(response, page, perPage)
+    return { data: body, total, page, perPage, hasNext }
+  }
+
+  const getSessionDiff = async (id: string, sessionId: string, kind: 'result' | 'cumulative'): Promise<string> => {
+    const response = await fetch(`${baseUrl}/agent_runners/${id}/sessions/${sessionId}/diff/${kind}`, getInit())
+    if (response.status === 404) return ''
```
khendrikse: I'd maybe flip this with the next line for consistency. But ok to do that in a next PR :)

src/commands/agents/utils.ts
```diff
@@ -44,15 +58,156 @@ export const validatePrompt = (input: string): boolean | string => {
...
+  const validAgents = AVAILABLE_AGENTS.map((entry) => entry.value) as string[]
   if (!validAgents.includes(agent)) {
     return `Invalid agent. Available agents: ${validAgents.join(', ')}`
   }
   return true
 }
 
+export const validateListStatusFilter = (status: string): true | string => {
+  if ((LIST_STATUS_FILTERS as readonly string[]).includes(status)) return true
```
khendrikse: this could use `isListStatusFilter`?

src/commands/agents/api.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  const listAgentRunnersForAccount = async (
+    accountSlug: string,
+    filters: ListAgentRunnersFilters = {},
+  ): Promise<PaginatedResult<AgentRunner[]>> => {
+    const page = filters.page ?? 1
+    const perPage = filters.per_page ?? DEFAULT_PER_PAGE
+    const params = buildSearchParams({ ...filters, page, per_page: perPage })
+    const response = await fetch(
+      `${baseUrl}/${encodeURIComponent(accountSlug)}/agent_runners?${params.toString()}`,
```
khendrikse, line 109: We're using `encodeURIComponent` here, is there a reason we do not want to do a similar thing to other id's? like a session or AR id? Maybe this happens in the command itself :)

khendrikse (commented):
@VaibhavAcharya I left a few comments :) some are nitty, feel free to do with it what you want!

PR #8323, opened 2026-06-26
fix: silence zip output when creating source zips

ndhoule:
This code path is the cause of a bunch of SWAR errors. Logs don't
provide enough information for us to debug this right now, so this PR
does two things:

- Adds the `-q` flag to `zip` to quiet its output. (We pass this flag
  everywhere else we call out to `zip`.) I have a weak suspicion the
  root cause here could be a too-much-output error (but that seems
  unlikely).
- Improves the error output when we hit this error. This should let us
  debug the issue if it comes up again in the future.

### DIFF
--- a/src/utils/deploy/upload-source-zip.ts
+++ b/src/utils/deploy/upload-source-zip.ts
@@ -83,7 +83,7 @@ const createSourceZip = async ({
     let message = 'zip command failed'
     if (_baseErr instanceof Error && 'command' in _baseErr) {
       const baseErr = _baseErr as ExecaError
-      message = `${baseErr.shortMessage}\n\n${baseErr.all}`
+      message = `${baseErr.shortMessage}\n\n${baseErr.all ?? ''}`
     }
     throw new Error(message, { cause: _baseErr })
   }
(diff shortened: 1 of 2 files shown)


davbree:
Claude:

> One thing on the error output, since the whole point is making this debuggable: the message appends `baseErr.stdout`, but `zip` writes its errors/warnings to **stderr**, and `-q` empties stdout further — so `stdout` will usually be blank, and execa's `shortMessage` is just `Command failed with exit code N: zip …` (no stderr). Net result: the actual reason (`zip error: Nothing to do!`, an I/O error, the exit code, …) still won't show up.
> 
> Since you already set `all: true`, it's a one-word fix to capture it:
> 
> ```ts
> message = `${baseErr.shortMessage}\n\n${baseErr.all}`   // or baseErr.stderr
> ```
> 
> Also worth noting: the error-case test rejects with a plain `new Error('zip command failed')` (no `.command`), so it skips the `ExecaError` branch entirely — the enrichment isn't actually exercised. A test that rejects with a realistic `ExecaError` shape (`command`, `shortMessage`, `stderr`/`all`) would lock it in.
> 
> Context on the urgency: per Humio this is hitting the agent-runner create flow at ~900+ sessions/day over the last few days, and the swallowed reason is exactly what's been blocking diagnosis — so getting the real `zip` stderr into the message is the high-value part here.

aitchiss (approved):

PR #8325, opened 2026-07-01
fix: surface zip stderr and exit code in source-zip upload errors

davbree:
## Summary

`netlify deploy --upload-source-zip` shells out to `zip` to build the source snapshot (`createSourceZip` in `src/utils/deploy/upload-source-zip.ts`). On a nonzero exit, `execFile`'s `error.message` is only the generic `Command failed: zip -r …` — the actual reason and exit code live on `error.stderr` / `error.code`, which were discarded. So the failure dead-ends at `Command failed: zip -r …` with no recoverable cause.

## Change

Include `zip`'s `stderr` and exit code in the reported error, e.g.:

```
Failed to create source zip: Command failed: zip -r … (zip error: Nothing to do! (…); exit code 12)
```

instead of the dead-end `Failed to create source zip: Command failed: zip -r …`.

Behavior is otherwise unchanged — the original error is still thrown, so callers and retry logic are unaffected; only the logged/`warn`ed message (and the `source-zip-upload` error status) is enriched.

## Test

Adds a unit test that mirrors how `util.promisify(execFile)` rejects (error carrying `stderr` + `code`) and asserts both are surfaced in the message. Existing tests unchanged and green.

## Why

Fleet telemetry shows this `zip` step failing on a large, steady volume of source-zip uploads with no recoverable reason in the logs. This is intentionally scoped to **make the failure visible** so the underlying cause can be pinpointed with data rather than guesswork.

### DIFF
--- a/src/utils/deploy/upload-source-zip.ts
+++ b/src/utils/deploy/upload-source-zip.ts
@@ -121,7 +121,13 @@ export const uploadSourceZip = async ({
     try {
       zipPath = await createSourceZip({ sourceDir, filename, statusCb })
     } catch (error) {
-      const errorMsg = error instanceof Error ? error.message : String(error)
+      const execError = error as { message?: string; stderr?: string | Buffer; code?: number | string }
+      const stderr = execError.stderr ? String(execError.stderr).trim() : ''
+      const detail = [stderr, execError.code != null ? `exit code ${String(execError.code)}` : '']
+        .filter(Boolean)
+        .join('; ')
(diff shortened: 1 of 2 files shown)


davbree:
Closing in favor of #8323 (EX-2434) — that's the owning team's fix for the same root issue and it's a superset: same error-output goal, plus it adds `-q` and refactors to execa. Left a review note there about capturing the zip reason from stderr/`all` rather than `stdout`.

PR #8300, opened 2026-06-11
feat: improve agent/CI ergonomics — capabilities command, non-interactive mode, JSON error envelopes, exit-code contract

DavidWells:
## TL;DR

AI agents are now heavy users of `netlify` — and they drive it through pipes, parse its output, and branch on its exit codes. We ran a structured agent-ergonomics audit of every command, flag, and error path, then fixed the 13 highest-impact issues. Everything is **additive**: no existing flag, output field, or success path changes behavior.

- **Added** `netlify capabilities --json` — machine-readable manifest of every command, flag, exit code, and env var
- **Added** `--non-interactive` global flag — prompts fail fast with a clear message instead of hanging headless sessions
- **Fixed** `netlify sites delete x` (space instead of colon) no longer prints help and **exits 0** — it errors with the correct `sites:delete` suggestion
- **Fixed** `status --json` and `env:* --json` always emit JSON, even on auth/link errors — no more zero-byte stdout
- **Added** documented exit codes: `0` ok · `1` error · `2` usage error · `4` input needed
- **Fixed** piped `--help` is parseable again, `NO_COLOR` is honored, typos get better suggestions, corrupt state files are backed up instead of silently reset

Measured with a fresh-context agent on 8 realistic tasks before/after: **5/8 → 8/8 task success**, and the most common workflow (reading env vars) went from 5 round-trips to 1. Each change ships with unit tests; validation: typecheck ✓ lint ✓ 434 unit tests ✓.

---

### 1. Space-form subcommands no longer fake-succeed

**Why:** `netlify sites delete my-site` used to print the `sites` help text and **exit 0**. A script — or an agent doing cleanup — reads that as "site deleted." Nothing happened. This was the single most dangerous false-success in the CLI.

**Better for agents:** exit 0 now always means the command actually ran. The error names the exact colon-form command (flags preserved), so an agent self-corrects in one step.

<details><summary>Before / after</summary>

```
$ netlify sites delete my-precious-site
```

**Before — exit 0**, parent help on stdout:

```
Handle various project operations

USAGE
  $ netlify sites [options]
...
```

**After — exit 2**, stderr:

```
 ›   Error: 'netlify sites delete' is not a command.
 ›   Did you mean 'netlify sites:delete my-precious-site'?
 ›   Run 'netlify sites --help' to see available subcommands.
```

Also covers `netlify env list`, `netlify functions invoke hello`, etc. Trailing flags survive the suggestion (`netlify env list --json` → `netlify env:list --json`).
</details>

[...]

PR #8331, opened 2026-07-06
feat(cli): honor NO_COLOR

DavidWells:
Split from #8300 (3 of 9).

## What

Honors the [NO_COLOR](https://no-color.org) convention: any non-empty `NO_COLOR` value disables ANSI colors — both in `chalk` output and in every `prettyjson.render` call site (`deploy`, `sites:create`, `status`, `status:hooks`, `watch`), which bypassed chalk's existing `--json` guard.

Measured: `NO_COLOR=1 netlify env --help` goes from 30 ANSI escape sequences to 0.

## Why it helps agents

Agents consume CLI output through pipes and set `NO_COLOR` expecting clean text; today they get ANSI noise they have to strip (or worse, parse around). This is the informal standard nearly every modern CLI honors.

## Testing

- New unit tests for `shouldDisableColors`
- typecheck ✓ lint ✓ 407 unit tests ✓

### DIFF
--- a/src/utils/command-helpers.ts
+++ b/src/utils/command-helpers.ts
@@ -33,7 +33,15 @@ const safeChalk = function (noColors: boolean) {
   return new Chalk()
 }
 
-export const chalk = safeChalk(argv.includes('--json'))
+/** Honors the NO_COLOR convention (https://no-color.org): any non-empty value disables colors */
+export const shouldDisableColors = (env: NodeJS.ProcessEnv = process.env): boolean => Boolean(env.NO_COLOR)
+
+export const chalk = safeChalk(argv.includes('--json') || shouldDisableColors())
+
+/** Options for `prettyjson.render()` so human-formatted output honors NO_COLOR and --json like `chalk` does */
+export const prettyJsonRenderOptions = (): { noColor: boolean } => ({
+  noColor: argv.includes('--json') || shouldDisableColors(),
+})
 
 export type ChalkInstance = ChalkInstancePrimitiveType
 
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -35,6 +35,7 @@ import {
   getToken,
   log,
   logJson,
+  prettyJsonRenderOptions,
   warn,
   type APIError,
 } from '../../utils/command-helpers.js'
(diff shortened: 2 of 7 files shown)


src/utils/command-helpers.ts
```diff
@@ -33,7 +33,15 @@ const safeChalk = function (noColors: boolean) {
   return new Chalk()
 }
 
-export const chalk = safeChalk(argv.includes('--json'))
+/** Honors the NO_COLOR convention (https://no-color.org): any non-empty value disables colors */
+export const shouldDisableColors = (env: NodeJS.ProcessEnv = process.env): boolean => Boolean(env.NO_COLOR)
```
serhalp, line 37: chalk already implements all the various colour detection stuff (much more than just NO_COLOR), the problem is that we're overriding that entirely by passing `argv.includes('--json')`. chalk exports `supportsColor` so we should be able to just import that and change to something like `argv.includes('--json') || supportsColor` (that probably isn't right, but you get the idea).

https://main.npmx.dev/package/chalk#user-content-supportscolor

DavidWells: Good instinct, but I checked and chalk doesn't actually handle `NO_COLOR`. It's a deliberate omission on their side.

The `supports-color` logic chalk 5.6.2 ships (vendored at `node_modules/chalk/source/vendor/supports-color/index.js`) checks `--no-color`/`--color` flags, `FORCE_COLOR`, TTY status, CI env vars, and `TERM`, but there is zero mention of `NO_COLOR` anywhere in the file. The chalk maintainers have declined to support it for years and point people to `FORCE_COLOR=0` instead (e.g. chalk/supports-color#105). So `argv.includes('--json') || !supportsColor` would silently drop the `NO_COLOR` behavior this PR exists to add.

Also a small clarification on the "overriding entirely" part: we only override in the force-off case. When the arg is false, `safeChalk` returns `new Chalk()`, which still runs chalk's full auto-detection, so TTY/CI/FORCE_COLOR handling is already intact for chalk output today.

That said, your suggestion does point at a real gap for prettyjson: it does its own coloring and knows nothing about chalk's detection, so piped `netlify status` output gets ANSI codes today even without `NO_COLOR`. Pulling in `supportsColor` there in addition to (not instead of) the `NO_COLOR` check would fix that:

```ts
noColor: argv.includes('--json') || shouldDisableColors() || !supportsColor.stdout
```

Happy to add that here if you're on board, or split it into a follow-up.

PR #8335, opened 2026-07-06
feat(status): --json always emits a JSON object for every auth/link state

DavidWells:
Split from #8300 (7 of 9).

## What

`netlify status --json` emitted **zero bytes of stdout** when logged out or unlinked (exit 1 with no explanation of which). Now every state emits a JSON object:

```json
{
  "loggedIn": true,
  "linked": false,
  "account": { "Name": "…", "Email": "…" },
  "siteData": null,
  "error": { "code": "NOT_LINKED", "fix": "netlify link" }
}
```

Codes: `NOT_LOGGED_IN` (no token or expired session), `NOT_LINKED`. The success payload gains additive `loggedIn`/`linked`/`error: null` fields. Non-JSON output and exit semantics are unchanged.

## Why it helps agents

`status --json` is the first command an agent runs to orient itself. Zero-byte stdout forces guessing between "not logged in" and "not linked"; a stable `error.code` + `fix` makes recovery a single deterministic step, and `JSON.parse(stdout)` never throws.

## Testing

- New unit tests covering all four states (logged out, expired session, unlinked, linked)
- typecheck ✓ lint ✓ 412 unit tests ✓


serhalp (approved):
💯

PR #8356, opened 2026-07-23
feat!: drop support for Node.js 20, require 22.13.0+

serhalp:
#### Summary

**BREAKING**: Node.js 20 is no longer supported. 22.13.0 is now the minimum supported version.

Node.js 20 reached End-of-Life (EOL) in April 2026, three months ago.

We require 22.13.0+ specifically because 22.12.0 added `require(esm)` support. Many of our own packages that Netlify CLI depends on as well as much of the community has been requiring 22.12.0+ in order to modernize and slim down dependency trees. (We go one further to 22.13.0 because a dependency in our tree already requires `^20.19.0 || ^22.13.0 || >=24` and cannot be removed or downgraded at this time.)

### DIFF
--- a/src/utils/nodejs-compile-cache.ts
+++ b/src/utils/nodejs-compile-cache.ts
@@ -16,8 +16,9 @@ export let didEnableCompileCache = false
  * Keep in mind that the cache is specific to a version of netlify-cli and a version of node.js and it is stored on the
  * user's disk in a temp dir. If any of these changes or the temp dir is cleared, the next run is a cold run.
  *
- * The programmatic API to enable this (`enableCompileCache()`) was added in node 22.8.0, but we currently support
- * >=20.12.2, hence the conditional below. (For completeness, note that support via the env var was added in 22.1.0.)
+ * The programmatic API to enable this (`enableCompileCache()`) was added in node 22.8.0, below our minimum supported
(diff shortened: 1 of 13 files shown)


src/utils/nodejs-compile-cache.ts
```diff
@@ -30,6 +30,7 @@ export const maybeEnableCompileCache = (): void => {
   // The docs recommend turning this off when running tests to generate precise coverage
   if (process.env.NODE_ENV === 'test') return
 
+  // TODO(serhalp): Consider enabling unconditionally now that we require Node.js 22.12.0+.
```
serhalp: This would have been a hard-break for 100% of users that are on an older Node.js version, for almost no benefit (deleting two lines of code), so I deferred.

tests/integration/telemetry.test.ts
```diff
@@ -1,16 +1,17 @@
+import { createRequire } from 'node:module'
 import { env as _env, version as nodejsVersion } from 'process'
 
 import type { Options } from 'execa'
 import execa from 'execa'
 import { expect, test } from 'vitest'
 
-import pkg from '../../package.json'
```
serhalp, line 7: We can switch these to `with {type: 'json'}` now in theory, but we need to upgrade prettier first.

VaibhavAcharya (approved):

PR #8382, opened 2026-07-31
build: switch from prettier to 11x faster oxfmt 🚀

serhalp:
#### Summary

This PR moves this repo's build tooling from prettier to [oxfmt](https://oxc.rs/docs/guide/usage/formatter.html). This results in an ~11x speedup of formatting.

On my Apple M3 Pro:
- `prettier --check`: 3413 ms median over 5 runs
- `oxfmt --check`: 304 ms median over 5 runs

Instead of porting over our exact prettier globs, this also starts formatting all oxfmt default globs, including json, toml, markdown, etc. That accounts for much of the diff in this PR.


src/commands/types.d.ts
```diff
@@ -14,7 +14,7 @@ export type NetlifySite = {
   configPath?: string
   siteId?: string
   get id(): string | undefined
-  set id(id: string): void
```
serhalp, line 17: This was invalid TypeScript, obscured by `skipLibCheck: true` but caught by oxfmt's parser.

types/lambda-local/index.d.ts
```diff
@@ -1,7 +1,7 @@
 // This module is "TypeScript" but contains no actual type annotations, so
 // the resulting `.d.ts` file is just useless `any`s.
 declare module 'lambda-local' {
-  declare interface Options {
```
serhalp, line 4: This was invalid TypeScript, obscured by `skipLibCheck: true` but caught by oxfmt's parser.

.oxfmtrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "semi": false,
+  "singleQuote": true,
+  "printWidth": 120,
+  "proseWrap": "always",
+  "trailingComma": "all",
+  "sortPackageJson": false,
```
serhalp: We could turn this one later 😁. I didn't wanna touch too many things here.

.oxfmtrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "semi": false,
+  "singleQuote": true,
+  "printWidth": 120,
+  "proseWrap": "always",
+  "trailingComma": "all",
```
serhalp: Would be nice to just use defaults, but this matches our Netlify code style

ndhoule: I'd prefer `oxfmt`'s defaults. We don't have a very consistent style across Netlify, and this PR is already changing the style of the repository.

oxfmt's default printwidth is 100, which seems wide enough for me, but we could leave it at 120 to avoid too dramatic a diff.

.oxfmtrc.json
ndhoule: I'd be OK with enabling this one, too.

ndhoule (dismissed):
Looks great, thanks for taking the time to make the improvement! My only note here is that I would prefer to see us use oxfmt's defaults. We could match Prettier's current settings to minimize the diff, but I think we might as well have one big diff. (I don't think splitting formatting changes across multiple commits has much benefit--we're still taking the torch to git history either way.)

.oxfmtrc.json
serhalp: I'll do it right away in a follow-up PR.

sirgagahandino9-design (commented):
Check further

ndhoule (dismissed):

.oxfmtrc.json
serhalp: 👍🏼 I kept just `semi: false, singleQuote: true` and the diff is quite small.

ndhoule (approved):

PR #8413, opened 2026-08-14
feat: add deploy-specific environment variable support to `netlify deploy`

ndhoule:
This changeset adds the ability to inject environment variables at deploy time. These variables take priority over any account- or site-level environment variables, and otherwise behave identically to variables defined via the Netlify UI or API.


src/commands/deploy/deploy.ts
```diff
@@ -319,7 +320,17 @@ const generateDeployCommand = (
 
   if (command?.options) {
     for (const option of command.options) {
-      if (['createSite', 'site', 'siteName', 'team'].includes(option.attributeName())) {
+      // `env` and `secretEnv` are skipped because reprinting a secret value here would leak it.
+        if (
```
jaredm563: I think a couple of your changes are using the old formatter,might need to run some NPM commands to make CI happy

src/commands/deploy/index.ts
```diff
@@ -76,6 +77,16 @@ For detailed configuration options, see the Netlify documentation.`,
       'Specify a deploy context for environment variables read during the build ("production", "deploy-preview", "branch-deploy", "dev") or `branch:your-branch` where `your-branch` is the name of a branch (default: dev)',
       normalizeContext,
     )
+    .option(
+      '--env <KEY=VALUE>',
+      'Set an environment variable for this deploy only. Can be specified multiple times.',
```
jaredm563: AX nit: I feel like this might be confusing if an agent sees it and tries to run like `netlify deploy --build --env VITE_API_URL=https://staging.api.example.com` expecting their Vite to bake that into the bundle

WYT about something like "Set an environment variable for this deploy only. O**nly available to serverless functions at runtime, not at build**. Can be specified multiple times"

src/utils/env/deploy-env-vars.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Returns the first key that appears more than once, or `undefined` if all keys are unique.
+ *
+ * The API rejects duplicate keys, and a single flag's parser cannot see the other flag's values,
+ * so callers must check the merged list.
+ */
+export const findDuplicateKey = (variables: DeployEnvironmentVariable[]): string | undefined => {
+  const seen = new Set<string>();
+
+  for (const { key } of variables) {
+    if (seen.has(key)) {
```
jaredm563, line 118: Should we .toUpperCase() in case there is some case sensitive ops later ? could avoid some troubleshooting for us later

jaredm563 (commented):
LGTM overall,left a couples nits for AX improvement and a possible improvement for string handling.

src/utils/env/deploy-env-vars.ts
ndhoule: Environment variables are case sensitive--they're uppercase by convention, not by requirement--so I think we should leave this as is.

jaredm563 (dismissed):
whoops,forgot to press the green

src/commands/deploy/index.ts
ndhoule: Good suggestion! Updated language.

jaredm563 (approved):

PR #8421, opened 2026-08-16
fix(logs): remove duplicated /api/v1 prefix in deploy logs URL

### LINKED ISSUE #8352: logs --source deploy returns 404 due to duplicated /api/v1 in request URL
## Summary

`netlify logs --source deploy` fails with `404 Not Found` because the CLI sends the historical deploy-log request to a URL containing a duplicated `/api/v1` prefix.

This reproduces on two independently linked Netlify sites.

## Environment

- Netlify CLI: `netlify-cli/26.2.0 darwin-arm64 node-v24.18.0`
- Node.js: `v24.18.0`
- OS: macOS 26.5.2, arm64

## Minimal reproduction

1. Link a local directory to a Netlify site that has at least one completed deploy.
2. Run:

   ```shell
   DEBUG=1 netlify logs --source deploy --since 1h
   ```

3. Observe the request URL and error:

   ```text
   [debug] → GET https://api.netlify.com/api/v1/api/v1/deploys/<deploy-id>/log
   [debug] ← 404 Not Found
   Error: Failed to fetch deploy logs: 404 Not Found
   ```

The same result occurs without `DEBUG=1`; the debug flag only exposes the generated URL.

## Expected behavior

The command should return historical deploy logs for the selected/latest ready deploy, as described by `netlify logs --help` and the CLI command documentation.

## Actual behavior

The command always fails with:

```text
Error: Failed to fetch deploy logs: 404 Not Found
```

## Suspected cause

`logsCommand` passes `client.basePath` as `apiBase`:

```ts
const apiBase = client.basePath
```

In a normal authenticated CLI session, this value is already:

```text
https://api.netlify.com/api/v1
```

`fetchDeployHistoricalLogs` then appends another `/api/v1`:

```ts
debugFetch(`${apiBase}/api/v1/deploys/${encodeURIComponent(deployId)}/log`, ...)
```

This produces:

```text
https://api.netlify.com/api/v1/api/v1/deploys/<deploy-id>/log
```

The likely path-construction fix is to append only `/deploys/<deploy-id>/log` when `apiBase` is `client.basePath`.

## Additional observation

I also tested the presumed corrected URL directly with the same authenticated token and deploy IDs:

```text
GET https://api.netlify.com/api/v1/deploys/<deploy-id>/log
```

That request also returned `404 Not Found` on both sites. In addition, the public Netlify OpenAPI specification does not appear to expose a GET endpoint for historical deploy logs (it documents `POST /builds/{build_id}/log`, but not a deploy-log retrieval endpoint).

So there may be two separate issues to verify:

1. The duplicated `/api/v1` in the CLI URL construction.
2. Whether `/deploys/{deploy_id}/log` is the correct and currently available backend endpoint for historical deploy-log retrieval.

## Regression/testing note

The historical deploy-log path was added as part of the unified `logs` command. The current integration tests appear to cover migration from the deprecated `logs:deploy` command, but not an authenticated historical deploy-log HTTP request, which may explain why the malformed URL was not caught.

`netlify logs --source deploy --follow` takes the WebSocket path instead and does not hit this HTTP 404, although it naturally requires an active build.

HosnainRafi:
## Problem

`netlify logs --source deploy` fails with `404 Not Found` because the CLI sends the historical deploy-log request to a URL containing a duplicated `/api/v1` prefix:

```
GET https://api.netlify.com/api/v1/api/v1/deploys/<deploy-id>/log
404 Not Found
```

The root cause is that `fetchDeployHistoricalLogs` appends `/api/v1/deploys/...` to the `apiBase` parameter, but `client.basePath` already contains the full `/api/v1` prefix (e.g., `https://api.netlify.com/api/v1`).

## Fix

Changed the URL construction in `src/commands/logs/sources/deploy.ts` from:

```ts
`${apiBase}/api/v1/deploys/${encodeURIComponent(deployId)}/log`
```

to:

```ts
`${apiBase}/deploys/${encodeURIComponent(deployId)}/log`
```

This produces the correct URL: `https://api.netlify.com/api/v1/deploys/<deploy-id>/log`

## Tests

Added a new unit test file `tests/unit/commands/logs/deploy-source.test.ts` with 4 test cases:

1. Verifies the URL does not contain duplicated `/api/v1` prefix
2. Verifies correct URL format when apiBase already ends with `/api/v1`
3. Verifies empty array is returned for non-array response
4. Verifies error is thrown on non-ok response

All tests pass, ESLint clean, Prettier formatted.

Fixes #8352

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -37,6 +37,7 @@ import {
   logJson,
   warn,
   type APIError,
+  NETLIFYDEVWARN,
 } from '../../utils/command-helpers.js'
 import { DEFAULT_CONCURRENT_HASH, DEFAULT_DEPLOY_TIMEOUT } from '../../utils/deploy/constants.js'
 import { type DeployEvent, deploySite } from '../../utils/deploy/deploy-site.js'
@@ -944,6 +945,22 @@ const prepAndRunDeploy = async ({
 
   const deployFolder = await getDeployFolder({ command, options, config, site, siteData })
   const functionsFolder = getFunctionsFolder({ workingDir, options, config, site, siteData })
+  // When deploying without running a build, warn if build plugins are configured
+  // because their config mutations are lost without a build run
+  // (see https://github.com/netlify/cli/issues/3792).
+  if (!options.build) {
+    type ConfigPlugin = { package?: unknown; origin?: string }
+    const plugins =
+      (config?.plugins as ConfigPlugin[] | undefined) ??
+      (command.netlify.cachedConfig.config as { plugins?: ConfigPlugin[] } | undefined)?.plugins
+    const configuredPlugins = plugins?.filter((plugin) => plugin.origin !== 'default') ?? []
+    if (configuredPlugins.length > 0) {
+      log(
+        `${NETLIFYDEVWARN} Site uses build plugins (${configuredPlugins.map((p) => p.package).join(', ')}) but no build is being run.\n` +
+          `  Config changes made by these plugins will not be applied. Use ${chalk.cyanBright('netlify deploy --build')} to build and deploy together.`,
+      )
+    }
+  }
   const { configPath } = site
 
   // build flag wasn't used and edge functions directories exist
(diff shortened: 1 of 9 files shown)


serhalp:
Superseded by https://github.com/netlify/cli/pull/8452

PR #8436, opened 2026-08-24
feat: forward agent runner ids on deploy creation

aitchiss:

#### Summary

Second part of [PAW-210](https://linear.app/netlify/issue/PAW-210/better-track-preview-state-for-agent-runners) — the `netlify-cli` slice of the [recommended approach](https://linear.app/netlify/issue/PAW-210/better-track-preview-state-for-agent-runners#comment-2d67b45f). Follows netlify/agent-runner-orchestrator#1125.

Today the runner↔deploy link is made at deploy *completion*, so an agent runner just stays `running` through the whole build and bitballoon can't distinguish "the agent is thinking" from "the agent is done, the preview is building". The fix is to link at deploy *creation* instead, which means the runner ids have to reach the `createSiteDeploy` request.

The orchestrator now sets `NETLIFY_AGENT_RUNNER_ID` and `NETLIFY_AGENT_RUNNER_SESSION_ID` in the CLI's env next to the existing `NETLIFY_DEPLOY_SOURCE`. This PR forwards them through as fields on the create-deploy body. One-field passthrough; no change to the deploy flow.

#### What changed

- **`src/utils/deploy/deploy-source.ts`** (new) — `getDeploySourceFields()` returns `deploy_source` plus `agent_runner_id` / `agent_runner_session_id`. Each id is omitted when unset rather than sent as `undefined`.
- **`src/commands/deploy/deploy.ts`** — both `createSiteDeploy` call sites spread that helper in place of the duplicated `deploy_source` line:
  - `runDeploy` (~L604), which creates the deploy on the `--no-build` path
  - `deploy` (~L1394), which creates it up front when building

#### Reviewer notes

- **Both create paths are covered, and they're the only ones.** Those two call sites were the only places the CLI sent `deploy_source`. `--trigger` goes through `createSiteBuild` (a git-based build trigger) and anonymous/drop deploys never sent a deploy source, so neither is in scope.
- **Deploy retries need nothing here.** The orchestrator's retry path re-invokes the CLI, so the env is re-read on each `createSiteDeploy`.
- **The session id is included as well as the runner id.** The orchestrator sets both, and the bitballoon slice needs the session to point at the deploy (`session.deploy_id`). Happy to trim to just `agent_runner_id` if you'd rather the CLI carried only the one field.
- **Unrecognised ids:** as flagged on the orchestrator PR, `bin-local.ts` generates `local-<hex>` ids for local runs, so bitballoon's create path should ignore an id it doesn't recognise rather than erroring.
- **Test results:** new unit tests 5/5; full unit suite 477/477 across 66 files (the 5 reported "errors" are unhandled telemetry rejections from `geo-location`/`exec-fetcher` hitting a 500 on a binary download — pre-existing and environmental). Full deploy integration suite 55/55, including two new tests covering both create paths. `typecheck`, `lint` and `format:check` all clean.

#### Still to come on PAW-210

- `bitballoon`: accept the params in `Api::V1::DeploysController#create`, link at creation in `Deploy.create_with_site!`, and add the derived `preview_state` to the agent runner snapshot serializer.

---

- [x] Open a [bug/issue](https://github.com/netlify/cli/issues/new/choose) before writing your code 🧑‍💻 — tracked in PAW-210.
- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 📖
- [x] Update or add documentation (if features were changed or added) 📝 — n/a, these are internal env vars set by the orchestrator and aren't part of the documented CLI surface (`NETLIFY_DEPLOY_SOURCE` isn't documented either).

a very serious puffin

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -40,6 +40,7 @@ import {
 } from '../../utils/command-helpers.js'
 import { DEFAULT_CONCURRENT_HASH, DEFAULT_DEPLOY_TIMEOUT } from '../../utils/deploy/constants.js'
 import { type DeployEvent, deploySite } from '../../utils/deploy/deploy-site.js'
+import { getDeploySourceFields } from '../../utils/deploy/deploy-source.js'
 import { uploadSourceZip } from '../../utils/deploy/upload-source-zip.js'
 import { getEnvelopeEnv } from '../../utils/env/index.js'
 import { mergeDeployEnvVars } from '../../utils/env/deploy-env-vars.js'
@@ -600,7 +601,7 @@ const runDeploy = async ({
         draft,
         branch: alias,
         include_upload_url: options.uploadSourceZip,
-        deploy_source: process.env.NETLIFY_DEPLOY_SOURCE || 'cli',
+        ...getDeploySourceFields(),
       }
 
       const createDeployResponse = await api.createSiteDeploy({ siteId, title, body: createDeployBody })
@@ -1390,7 +1391,7 @@ export const deploy = async (options: DeployOptionValues, command: BaseCommand) 
       draft,
       branch: alias,
       include_upload_url: options.uploadSourceZip,
-      deploy_source: process.env.NETLIFY_DEPLOY_SOURCE || 'cli',
+      ...getDeploySourceFields(),
     }
 
     // TODO: Type this properly in `@netlify/api`.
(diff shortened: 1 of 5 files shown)


whitep4nth3r:
Started going through this with @aitchiss - best course of action before this can be tested thoroughly is to get the linked bitballoon PR merged!

VaibhavAcharya (approved):

aitchiss:
I was able to verify this E2E on the pre-release - screenshot of deploy linked to agent runner id passed in env vars:
CleanShot 2026-08-24 at 16 27 30

PR #8452, opened 2026-09-01
feat(logs): retrieve historical logs by deploy id

whitep4nth3r:
Full disclaimer this repo is all very new to me and I had a lot of robot help that I tried to double check and verify along the way.

Fixes historical `netlify logs --source deploy`, which returned `404 Not Found`.

Reported in [Slack](https://netlify.slack.com/archives/C090RJS5TC5/p1788211008007509).

## Root cause

- `fetchDeployHistoricalLogs` requested `${apiBase}/api/v1/deploys/:id/log`.
- `apiBase` already ends in `/api/v1` (double prefix), **and** that REST endpoint does not exist
- Build logs are only served over the socketeer websocket (the transport `--follow` already uses).

## Before / After

| | Before | After |
|---|---|---|
| `logs --source deploy --since 1h` | `404 Not Found` | Streams the deploy's build log |
| Transport | non-existent REST endpoint | socketeer websocket |
| Deploy selection (`--source deploy`) | latest `ready` only | latest of **any** state |
| Failed-build logs | unreachable | `--deploy <id>`, or auto-selected when latest |
| Targeting a specific deploy | — | `-d, --deploy <id>` shows the full build log |

## How to test

```bash
# full failed build log — no --since needed
netlify logs --source deploy --deploy <FAILED_DEPLOY_ID>
# auto-selects the latest deploy, even if it failed
netlify logs --source deploy --since 24h
# unit
npm exec vitest -- run tests/unit/commands/logs/
```

## Tested against real infra

`--deploy <failed-id>` returned the full 1311-line failed build log incl. the failure reason:

```
[🚀 deploy] ... INFO ** ERROR **
[🚀 deploy] ... INFO The following serverless functions failed to deploy: palette-agent.spec
[🚀 deploy] ... INFO Failed during stage 'building site': Command was cancelled
[🚀 deploy] ... INFO Failing build: Failed to build site
```

- Socketeer replay confirmed for finished deploys: failed = 1311 lines, successful = 1126 lines, all timestamped, clean close.
- `--deploy … --since 1m` → window ignored, full log returned.
- `--source deploy --since 24h` → auto-selects latest deploy (1125 lines).
- Guards: invalid `--deploy` and `--deploy` + `--url` both error.
- 11 unit tests, typecheck, lint, build all pass.

⚠️ Behavior change: `--source deploy` now shows the latest deploy even if it failed (previously only the last successful one).

### DIFF
--- a/docs/commands/logs.md
+++ b/docs/commands/logs.md
@@ -43,7 +43,7 @@ netlify logs --follow
 netlify logs --follow --source functions --source edge-functions
 netlify logs --json --since 1h
 netlify logs --url https://my-branch--my-site.netlify.app --since 1h
-netlify logs --source deploy --deploy 5b4e23db82d3f1780abd74f2 --since 7d
+netlify logs --source deploy --deploy 5b4e23db82d3f1780abd74f2
 ```
 
 
--- a/docs/commands/logs.md
+++ b/docs/commands/logs.md
@@ -17,6 +17,7 @@ netlify logs
 
 **Flags**
 
+- `deploy` (*string*) - Show logs for a specific deploy by ID, including failed builds. Cannot be combined with --url
 - `edge-function` (*string*) - Filter to specific edge functions by name or path
 - `filter` (*string*) - For monorepos, specify the name of the application to run the command in
 - `follow` (*boolean*) - Stream logs in real time instead of showing historical logs
@@ -42,6 +43,7 @@ netlify logs --follow
 netlify logs --follow --source functions --source edge-functions
 netlify logs --json --since 1h
 netlify logs --url https://my-branch--my-site.netlify.app --since 1h
+netlify logs --source deploy --deploy 5b4e23db82d3f1780abd74f2 --since 7d
 ```
 
 
--- a/docs/commands/logs.md
+++ b/docs/commands/logs.md
@@ -43,7 +43,7 @@ netlify logs --follow
 netlify logs --follow --source functions --source edge-functions
 netlify logs --json --since 1h
 netlify logs --url https://my-branch--my-site.netlify.app --since 1h
-netlify logs --source deploy --deploy 5b4e23db82d3f1780abd74f2
+netlify logs --source deploy --deploy <deploy-id>
 ```
 
 
--- a/src/commands/logs/index.ts
+++ b/src/commands/logs/index.ts
@@ -130,6 +130,12 @@ export const createLogsCommand = (program: BaseCommand) => {
       ),
     )
     .addOption(
+      new Option(
+        '-d, --deploy <deploy-id>',
+        'Show logs for a specific deploy by ID, including failed builds. Cannot be combined with --url',
+      ),
+    )
+    .addOption(
       new Option('-l, --level <levels...>', `Log levels to include. Choices are:${CLI_LOG_LEVEL_CHOICES_STRING}`),
     )
     .addOption(new Option('--json', 'Output logs as JSON Lines'))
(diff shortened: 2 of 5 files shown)


whitep4nth3r:
This PR also fixes this open PR https://github.com/netlify/cli/pull/8421

serhalp:
Writing this out partly for my own understanding... 😰

So, before this PR, `netlify logs` (as of a few months ago) supports a `--source <source>` which allows logs to be filtered to a given type ("source"):

1. `deploy`: build/deploy logs
2. `functions`: Netlify Functions runtime logs
3. `edge-functions`: Netlify Edge Functions runtime logs

With or without `--source`, you can also filter logs a few other ways:

- `--function <name>`: filter to a named function
  - This _implicity sets and is only compatible with_ `--source functions`
- `--edge-function <name>`: filter to a named edge function
  - This _implicity sets and is only compatible with_ `--source edge-functions`
- `--url <url>`: filter to the _deploy ID_ resolved from this URL
  - This filters logs _across all (possibly filtered) sources_ by deploy ID.
- `--since`, `--until`, `--level`: obvious

So, in summary, `--url` is a mechanism to specify _a deploy_, but it does not only support nor does it imply only the _deploy source_.

Finally, one more complexity: historical vs. live logs.

- By default, historical mode is used. It looks up historical logs across all selected sources and filters.
- `--follow` selects live mode instead. It streams logs in real time across all selected sources and filters (but `--since` and `--until` are not supported).
- You cannot request both historical and live logs in one command.

Hopefully I got all that right. Now, issues this PR attempts to solve:

1. bug: `--source deploy` without `--follow` doesn't work at all!
  - This was using an invalid `api.netlify.com/api/v1/api/v1/deploys/<id>/log` endpoint. But even stripping the duplicated `api/v1/` is still invalid. The endpoint just doesn't exist 🙃. This needs to use the websocket endpoint. It replays and follows. That's what `netlify logs:deploy` always used. We just got it wrong in this new code path. This PR fixes this.
2. feature: it would be nice if you could specify `--deploy <deploy id>` as an alternative to `--url <url>`.
  - This PR implements this.
  - To be clear, despite `--function <name>` implying `--source functions` and `--edge-function <name>` implying `--source edge-functions`, `--deploy <id>` would _not_ imply `--source deploy`. I think this is pretty confusing, both for humans and for agents. I'm not sure it's very avoidable though. We should document this clearly. We could alternatively make this `--deploy-id <id>`.
3. bug: `--source deploy` without `--url` (or now `--deploy`) automatically selected the latest deploy but skipped over failed deploys.
  - This PR fixes this.
4. bug: `--source deploy --follow --url <url>` silently ignores the resolved deploy from the given URL 😓; uses the "find the latest deploy" behaviour instead.

Whew.

I believe there is also one additional change in this PR:

5. `--source deploy` with `--url` or `--deploy` now quietly ignores `--since` and `--until`
  - I can see the idea here, but I'm not sure. Personally I would rather keep this as is.

---

Given all this, here's my take:
- I would much prefer breaking this out into smaller, atomic PRs (could be a stack). We have three separate bugs and a feature and I believe they're all independent. I think this will be a lot easier to review now and to understand in the future. (I would argue the refactoring/hardening of the deploy logs parsing stuff can/should be in the PR for #1, btw.)
- I may have missed the reasoning, but I would propose we skip #5.
- This is amazing, thank you for digging into and fixing all this!

whitep4nth3r:
Thanks @serhalp - will split out into a stacked PR. Stand by.

PR #8453, opened 2026-09-02
fix: stop shipping broken source maps and drop direct node-fetch dep

jherr:
First slice of a `node_modules` footprint reduction effort. Scope here is deliberately the low-risk, measurable end of it.

### Stop shipping source maps (the actual win)

We ship 534 map files — **534 of the package's 1128 files** — and none of them can work. Every map references `../src/*.ts`, `src` isn't in `files`, and none carry `sourcesContent`. So in an installed copy, every map points at a path that doesn't exist. Nobody has ever gotten a resolved stack frame out of them.

`sourceMap`/`declarationMap` stay **on** for local development; a `prepack` script strips maps and their `//# sourceMappingURL=` annotations from `dist/` before the tarball is assembled.

| | before | after | |
|---|---|---|---|
| tarball | 480 KB | **300 KB** | −38% |
| unpacked | 2.06 MB | **1.14 MB** | −45% |
| files | 1128 | **595** | −47% |

### Drop the direct `node-fetch` dependency

`src/` no longer imports it (7 call sites), so it moves to `devDependencies` — the test suite still uses it in ~25 files, which is left for a follow-up. Global `fetch` has been available since long before our `>=22.13.0` engine floor.

**This does not shrink end-user installs yet**, and I want to be upfront about that: `@netlify/api` still depends on `node-fetch`, and `@mapbox/node-pre-gyp` pulls `node-fetch@2` via `@vercel/nft`. I verified this against a real install of the packed tarball. The ~9 MB (`node-fetch` + `fetch-blob` + `web-streams-polyfill`) only comes back once `@netlify/api` drops it. This PR removes our half of that blocker.

## Two bugs found along the way

**`functions-create` never awaited its downloads.** Template files were written with `res.body?.pipe(dest)` — not awaited, so `Promise.all` resolved before any write finished, and `npm i` ran against a possibly-incomplete directory. Global `fetch` returns a web `ReadableStream` with no `.pipe()` at all, so this had to change regardless; `await pipeline(...)` fixes both the incompatibility and the pre-existing race. An explicit `res.ok` check replaces what was previously a silent write of an error-page body.

**`npm run clean` left the build broken.** It removed `dist/` but not `tsconfig.build.tsbuildinfo`, so the next `npm run build` saw an up-to-date incremental cache and emitted *nothing* — no `dist/` at all. Both `clean` and the pack-time strip now clear it.

## What I tried and backed out

An `overrides` entry pinning `@octokit/openapi-types` to one version (3 copies, ~14 MB of pure `.d.ts`). I dropped it: npm `overrides` only apply to the root project, so it's inert in both dependency and global installs, and our dev tree already resolves to a single copy. It bought nothing and forced an unrelated `Number(githubRepo.id)` cast. Getting that win needs an upstream `@octokit/types` bump or bundling.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean
- `npm run test:unit` — 499/500, matching the baseline on `main` exactly. The one failure (`generate-autocompletion` snapshot) is pre-existing and unrelated: option ordering only, same content, reproduces on unmodified `main`.
- Verified `fetch` parity by probe against a local server: `Buffer` body with manual `Content-Length` behaves identically to node-fetch (no chunked encoding, `statusText` preserved) — that's the `upload-source-zip` path.
- Round-tripped `clean` → `build` → `pack` → `build` to confirm maps are stripped from the tarball and restored by a subsequent local build.

## Follow-ups (not in this PR)

The remaining footprint is ~361 MB installed, ~1040 packages. Biggest items, all needing more than a config change:

- `@netlify/images` → `ipx` → `sharp` + libvips — **34.8 MB**, 81 packages, only used to emulate Image CDN in `netlify dev`
- `@netlify/database-dev` → `@electric-sql/pglite` — **22.3 MB** embedded Postgres WASM
- `typescript` — **23.6 MB**, via `ts-node` in `@netlify/build` and `detective-typescript` in `zip-it-and-ship-it`
- `@opentelemetry/semantic-conventions` — **12 MB** of upstream packaging bloat
- Dependency source maps + `.d.ts` are ~89 MB (34%) of the installed tree. Only bundling reaches those.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,6 @@
         "multiparty": "^4.2.3",
         "nanospinner": "^1.2.2",
         "netlify-redirector": "^0.5.0",
-        "node-fetch": "^3.3.2",
         "normalize-package-data": "^7.0.1",
         "open": "^11.0.0",
         "p-filter": "^4.1.0",
@@ -161,6 +160,7 @@
         "lodash.shuffle": "^4.2.0",
         "memfs": "^4.56.10",
         "nock": "^14.0.10",
+        "node-fetch": "^3.3.2",
         "npm-run-all2": "^8.0.4",
         "oxfmt": "0.61.0",
         "p-timeout": "^7.0.0",
@@ -5841,10 +5841,6 @@
         "node": ">= 20"
       }
     },
-    "node_modules/@octokit/openapi-types": {
-      "version": "26.0.0",
-      "license": "MIT"
-    },
     "node_modules/@octokit/plugin-paginate-rest": {
       "version": "13.2.0",
       "license": "MIT",
@@ -5924,6 +5920,12 @@
       "dependencies": {
         "@octokit/openapi-types": "^26.0.0"
       }
+    },
+    "node_modules/@octokit/types/node_modules/@octokit/openapi-types": {
+      "version": "26.0.0",
+      "resolved": "https://registry.npmjs.org/@octokit/openapi-types/-/openapi-types-26.0.0.tgz",
+      "integrity": "sha512-7AtcfKtpo77j7Ts73b4OWhOZHTKo/gGY8bB3bNBQz4H+GRSWqx2yvj8TXRsbdTE0eRmYmXOEY66jM7mJ7LzfsA==",
+      "license": "MIT"
     },
     "node_modules/@open-draft/deferred-promise": {
       "version": "2.2.0",
@@ -6581,7 +6583,6 @@
       "cpu": [
         "arm"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6595,7 +6596,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6609,7 +6609,6 @@
       "cpu": [
         "arm64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
@@ -6623,7 +6622,6 @@
       "cpu": [
         "x64"
       ],
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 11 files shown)


jherr:
Closing in favour of three smaller PRs. This one bundled four independent changes behind a single review, and one of them — dropping the direct `node-fetch` dependency — is being cut entirely.

**Split into:**

- #8463 — `chore: clear tsbuildinfo in clean script`
- #8464 — `fix: stop shipping source maps in the published package` (the −45% win)
- #8465 — `fix: await function template downloads in functions-create`

Each is independent, based on current `main`, and can merge in any order.

**Dropped: the `node-fetch` removal.** It doesn't shrink an end user's install yet — `@netlify/api` still depends on `node-fetch`, and `@mapbox/node-pre-gyp` pulls `node-fetch@2` via `@vercel/nft` — so it buys nothing today while carrying 7 `src/` call-site changes and a test rewrite. Worth revisiting once `@netlify/api` drops it.

Cutting it also drops the entire `package-lock.json` diff, which is a win on its own: the lockfile churn here (`@octokit/openapi-types` re-nesting, rollup platform binaries losing `dev: true`) was incidental fallout from the backed-out `overrides` experiment and had no business in the change. None of the three new PRs touches the lockfile.

**Not re-filing the research.** The footprint numbers stay readable here rather than moving to a new issue: the ~361 MB / ~1040-package installed breakdown, the biggest remaining items (`ipx`/`sharp` 34.8 MB, `pglite` 22.3 MB, `typescript` 23.6 MB, `@opentelemetry/semantic-conventions` 12 MB), the `@netlify/api` node-fetch blocker, and the dead-end `@octokit/openapi-types` `overrides` experiment (npm `overrides` only apply to the root project, so it's inert for both dependency and global installs). The new PRs link back here.

Leaving the branch in place for now — it's the only copy of the node-fetch work.

PR #8464, opened 2026-09-03
fix: stop shipping source maps in the published package

jherr:
We ship 534 `.map` files — **534 of the package's 1128 files** — and none of them can work. Every map references `../src/*.ts`, `src` isn't in `files`, and none carry `sourcesContent`. So in an installed copy, every map points at a path that doesn't exist. Nobody has ever gotten a resolved stack frame out of them.

`sourceMap`/`declarationMap` stay **on** for local development; a `prepack` script strips maps and their `//# sourceMappingURL=` annotations from `dist/` before the tarball is assembled.

| | before | after | |
|---|---|---|---|
| tarball | 480 KB | **296 KB** | −38% |
| unpacked | 2.06 MB | **1.14 MB** | −45% |
| files | 1128 | **595** | −47% |

### Why the script also deletes `tsconfig.build.tsbuildinfo`

`tsc --incremental` decides what to emit from that file alone, not from what's on disk. Stripping maps out from under it would otherwise leave a subsequent local build convinced it has nothing to do, silently yielding a `dist/` with no maps. Deleting it forces the next local build to emit in full.

### Idempotency

`.github/workflows/release-please.yml` invokes `npm publish` more than once per release, so `prepack` runs more than once. The second run is a no-op, as is a run with no `dist/` present.

## Verification

- `npm run clean && npm run build` → 1068 files, 534 maps
- `npm pack` → **595 files, 296 KB, 0 `.map` entries in the tarball, 0 `sourceMappingURL` annotations left**
- `npm run prepack` a second time → `removed 0 map file(s)` — clean no-op
- `npm run prepack` with no `dist/` → `no dist/ output found, nothing to do`
- `npm run build` again → 1068 files, 534 maps back. Round trip holds; local debugging is unaffected.
- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean
- `npm run test:unit` — 499/500, matching the baseline on unmodified `main` exactly (the `generate-autocompletion` snapshot failure is pre-existing; I reproduced it on `main` at `97fd77d3c`)

No `package-lock.json` change.

---

Split out of #8453, which bundled this with unrelated work. The wider `node_modules` footprint research (~361 MB installed, ~1040 packages, and where the rest of it lives) is in that PR's description.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "test:integration": "vitest run --retry=3 tests/integration/",
     "test:unit": "vitest run tests/unit/",
     "postinstall": "node ./scripts/postinstall.js",
+    "prepack": "node ./scripts/strip-source-maps.js",
     "typecheck": "node ./node_modules/typescript-native/bin/tsc",
     "typecheck:watch": "node ./node_modules/typescript-native/bin/tsc --watch"
   },
--- a/scripts/strip-source-maps.js
+++ b/scripts/strip-source-maps.js
@@ -1,0 +1,92 @@
+/*
+ * This script runs at pack time (`prepack`), before the tarball is assembled.
+ *
+ * We build with `sourceMap` and `declarationMap` enabled so that local development against `dist/`
+ * has working maps. Those maps are useless to end users, though: they reference `../src/*.ts`, and
+ * `src` is not in the package's `files` list, so every map in a published install points at a file
+ * that isn't there. They accounted for roughly half of the published package, so we strip them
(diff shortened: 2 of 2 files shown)


serhalp:
@jherr Instead of bending over backwards to delete them post hoc, we can just instruct npm not to publish them with 1 line! https://github.com/netlify/cli/pull/8468

serhalp (changes requested):
see comment above

PR #8476, opened 2026-09-04
fix(logs): fetch historical deploy logs over websocket

whitep4nth3r:
## The bug

404 on `netlify logs --source deploy` 

The above command requested `${apiBase}/api/v1/deploys/:id/log`, which both double-prefixed `/api/v1` and hit a REST endpoint that does not exist.

### The fix

Replay stored build logs over the socketeer websocket using the same transport the `--follow` path and the deploy UI use.

#### Command

Run against `netlify-react-ui` on this branch:

`netlify logs --source deploy --since 30d`

#### Output example

image
image

Supersedes #8421.

### DIFF
--- a/src/commands/logs/sources/deploy.ts
+++ b/src/commands/logs/sources/deploy.ts
@@ -1,11 +1,10 @@
 import type { NetlifyAPI } from '@netlify/api'
 
 import { getWebSocket } from '../../../utils/websockets/index.js'
-import { debugFetch } from '../log-api.js'
 import type { LogEntry } from '../log-api.js'
 
-interface DeployLogLine {
-  ts: string
+interface DeployLogMessage {
+  ts?: string | number
   log?: string
   message?: string
   level?: string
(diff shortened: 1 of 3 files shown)


src/commands/logs/sources/deploy.ts
```diff
@@ -1,62 +1,121 @@
...
   from: number
   to: number
 }): Promise<LogEntry[]> => {
-  const response = await debugFetch(`${apiBase}/api/v1/deploys/${encodeURIComponent(deployId)}/log`, {
-    headers: {
-      Authorization: `Bearer ${accessToken ?? ''}`,
-    },
-  })
+  const collected: { entry: LogEntry; hasTimestamp: boolean }[] = []
+  const ws = getWebSocket('wss://socketeer.services.netlify.com/build/logs')
```
serhalp, line 60: I think we have an undocumented mechanism (for internal use) to override the netlify.com domain with `NETLIFY_API_URL`. We should use that here I think (so we can point at staging across the board).

serhalp (approved):
:shipit:

src/commands/logs/sources/deploy.ts
whitep4nth3r: On it.

whitep4nth3r: There's currently only a mechanism for the actual API and not socketeer. Would you like me to invent a mechanism for socketeer? I need to know what he staging URL for socketeer is.

serhalp: ooh that makes sense. nah, punt.

PR #8477, opened 2026-09-04
fix(logs): auto-select the latest finished deploy, including failures

whitep4nth3r:
Stacked on #8476. Part of splitting #8452 into atomic PRs.

`netlify logs --source deploy` auto-selected the latest **`ready`** deploy, so failed builds were silently skipped. 

It now selects the latest **finished** deploy — `ready`, `error`, or `cancelled` — and skips in-progress builds (which have no final log yet).

## Command

`netlify logs --source deploy --since 30d` where the latest build was failed:

image
...
image

### DIFF
--- a/src/commands/logs/logs.ts
+++ b/src/commands/logs/logs.ts
@@ -18,6 +18,7 @@ import { LOG_LEVELS_LIST, CLI_LOG_LEVEL_CHOICES_STRING } from './log-levels.js'
 import {
   fetchDeployHistoricalLogs,
   findCurrentBuildingDeploy,
+  findLatestFinishedDeploy,
   findLatestReadyDeploy,
   streamDeploy,
 } from './sources/deploy.js'
@@ -218,7 +219,9 @@ export const logsCommand = async (options: OptionValues, command: BaseCommand) =
     }
 
     if (!deployId) {
-      const latestId = await findLatestReadyDeploy(client, siteId)
+      const latestId = sources.includes('deploy')
+        ? await findLatestFinishedDeploy(client, siteId)
+        : await findLatestReadyDeploy(client, siteId)
       if (latestId) {
         deployId = latestId
       }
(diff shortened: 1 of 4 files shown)


src/commands/logs/sources/deploy.ts
```diff
@@ -167,3 +167,11 @@ export const findLatestReadyDeploy = async (client: NetlifyAPI, siteId: string):
   const deploys = (await client.listSiteDeploys({ siteId, state: 'ready', per_page: 1 })) as { id: string }[]
   return deploys.length > 0 ? deploys[0].id : undefined
 }
+
+const FINISHED_DEPLOY_STATES = new Set(['ready', 'error', 'cancelled'])
+
+export const findLatestFinishedDeploy = async (client: NetlifyAPI, siteId: string): Promise<string | undefined> => {
+  const deploys = (await client.listSiteDeploys({ siteId, per_page: 10 })) as { id: string; state: string }[]
```
serhalp, line 174: Technically this won't find the latest finished deploy if the last 10 deploys aren't in a finished state. But I think that's an edge case we can live with here.

src/commands/logs/logs.ts
```diff
@@ -218,7 +219,9 @@ export const logsCommand = async (options: OptionValues, command: BaseCommand) =
     }
 
     if (!deployId) {
-      const latestId = await findLatestReadyDeploy(client, siteId)
+      const latestId = sources.includes('deploy')
```
serhalp, line 222: 🙏🏼 can we please add test coverage for this new behaviour?

serhalp (commented):
LGTM, good catch! only blocker is test coverage

src/commands/logs/sources/deploy.ts
whitep4nth3r: yeah I was unsure of what the number should be :salma-crying:

src/commands/logs/logs.ts
whitep4nth3r: Done.

serhalp (dismissed):

serhalp (dismissed):

PR #8479, opened 2026-09-04
feat(logs): add --deploy-id flag to target a specific deploy

whitep4nth3r:
Stacked on #8477.

Adds `-d, --deploy-id <deploy-id>` to `netlify logs` — view logs for a specific deploy (including failed builds) as an alternative to `--url`.

```bash
netlify logs --source deploy --deploy-id <deploy-id> --since 7d
```

Successful command:

image
image

Unsuccessful commands:

image

*Note*: When testing this I felt it was better to *not* need the `--since` for a specific deploy ID, however, I left it in as per @serhalp's comment in the original large PR: https://github.com/netlify/cli/pull/8452

image

PR #8480, opened 2026-09-04
fix(logs): stream the targeted deploy in --follow

whitep4nth3r:
Stacked on #8479.

In `--follow`, stream the explicitly targeted deploy (`--deploy-id` / `--url`) instead of ignoring it and streaming the current building deploy. 

Falls back to the building deploy only when nothing was targeted.

image

*Note:* Currently, if a deploy is targeted and it is a finished deploy, the command doesn't exit at "Site is live". The next PR in the stack will attempt to address this.

### DIFF
--- a/src/commands/logs/logs.ts
+++ b/src/commands/logs/logs.ts
@@ -169,6 +169,7 @@ export const logsCommand = async (options: OptionValues, command: BaseCommand) =
   }
 
   let deployId: string | undefined
+  let deployTargeted = false
   if (options.deployId) {
     const explicitDeployId = (options.deployId as string).trim()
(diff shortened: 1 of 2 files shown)


src/commands/logs/logs.ts
serhalp, line 1: Is it possible to add test coverage for this behaviour? 🙏🏼

whitep4nth3r: incoming

whitep4nth3r: Test coverage added.

serhalp (dismissed):
:shipit:

serhalp (dismissed):

serhalp (approved):

PR #8481, opened 2026-09-04
fix(logs): close the deploy stream when replaying a finished deploy

whitep4nth3r:
Stacked on #8480.

## Problem

`--follow` on an already-finished deploy hangs. 

Socketeer replays the stored log then holds the socket open, and for a **successful** deploy it may never send the `report`/`building` marker that `streamDeploy` closes on. This results in the stream never ending and not printing "Deploy stream closed."

## Fix

When following a **targeted** deploy (`--deploy-id`/`--url`) that **is finished** (`ready`/`error`/`cancelled`) close the stream once the replay goes idle. Live builds are unchanged.

### `scheduleClose` machinery (`streamDeploy`)

`streamDeploy` takes an optional `{ closeWhenIdleMs }`. When it's set (replay mode), a single `closeTimer` drives auto-close through one helper:

```ts
const scheduleClose = (ms) => { clearTimeout(closeTimer); closeTimer = setTimeout(() => ws.close(), ms) }
```

- **on connect** → `scheduleClose(DEPLOY_STREAM_START_TIMEOUT_MS)` (20s) — a no-message safety cap so it never hangs if socketeer stays silent.
- **on the first message** → `scheduleClose(closeWhenIdleMs)` (3s), replacing the 20s cap — so a slow start (up to 20s before the first line) can't cause an empty, premature close.
- **each subsequent message** → re-`scheduleClose(3s)` — quick exit once the replay stops.
- **on close / teardown** → `clearTimeout(closeTimer)`.

When `closeWhenIdleMs` is `undefined` (a live build), no timer is armed and behavior is unchanged (closes only on the terminal build report).

Why: closing too early drops log lines, closing too late only delays the exit. The idle timer is only armed *after* the first message, and a longer cap guards the no-output case.

## Tests

`netlify logs --follow --source deploy --deploy-id 6a9a953eed6f7867a85e8dd3` - where the deploy ID has finished:

image

### DIFF
--- a/src/commands/logs/sources/deploy.ts
+++ b/src/commands/logs/sources/deploy.ts
@@ -13,6 +13,7 @@ interface DeployLogMessage {
 }
 
 const DEPLOY_LOG_REPLAY_TIMEOUT_MS = 30_000
+const DEPLOY_STREAM_START_TIMEOUT_MS = 20_000
 
 const parseDeployLogTimestamp = (message: DeployLogMessage): number | undefined => {
   if (!message.ts) {
@@ -124,8 +125,17 @@ export const streamDeploy = (
   accessToken: string | null | undefined,
   onEntry: (entry: LogEntry) => void,
   onClose: () => void,
+  { closeWhenIdleMs }: { closeWhenIdleMs?: number } = {},
 ): (() => void) => {
   const ws = getWebSocket('wss://socketeer.services.netlify.com/build/logs')
+
+  let closeTimer: ReturnType<typeof setTimeout> | undefined
+  const scheduleClose = (ms: number) => {
+    clearTimeout(closeTimer)
+    closeTimer = setTimeout(() => {
+      ws.close()
+    }, ms)
+  }
 
   ws.on('open', () => {
     ws.send(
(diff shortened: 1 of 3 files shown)


src/commands/logs/sources/deploy.ts
```diff
@@ -175,3 +193,8 @@ export const findLatestFinishedDeploy = async (client: NetlifyAPI, siteId: strin
   const finished = deploys.find((deploy) => FINISHED_DEPLOY_STATES.has(deploy.state))
   return finished?.id
 }
+
+export const isDeployFinished = async (client: NetlifyAPI, siteId: string, deployId: string): Promise<boolean> => {
+  const deploy = (await client.getSiteDeploy({ siteId, deployId })) as { state?: string }
```
serhalp: nit: can we avoid this unsafe type assertion? 🤔

serhalp (dismissed):
:shipit:

src/commands/logs/sources/deploy.ts
whitep4nth3r: I believe I may have fixed it. Please check!

serhalp: 🤔 Doesn't appear to have changed.

serhalp (dismissed):

serhalp (approved):

PR #8482, opened 2026-09-04
refactor(logs): use the typed options pattern

whitep4nth3r:
Stacked on #8481. Addresses @serhalp's note on #8479 that `logs` wasn't using the typed options pattern.

Adds a `LogsOptionValues` type (same convention as `deploy`/`clone`/`link`) used by the command action and `logsCommand`, removing the `as string` / `as string[]` casts on option access. No behavior change.


serhalp (dismissed):
😍😍😍

serhalp (dismissed):

serhalp (approved):

